PR #2, opened 2017-04-20
WIP: Configuring monorepo with Lerna

taras:
I initialized lerna in this project, but this was too easy. Is there more you'd like me to do here?

If so, please provide some direction on what you'd like to accomplish with this beyond the basics.

## TODO

- [x] Get tests to work
- [ ] Make schema resolve through DI
- [ ] Make tests run with lerna
- [ ] Run lerna in travis


ef4:
Thanks. To exercise this, we would need to see if lerna can actually install and link up all the dependencies correctly.

I would suggest testing that locally. Get the test suite working first, then `rm -rf node_modules`, then run `lerna bootstrap`, then see if the tests still work.

Then you can put the right lerna bootstrap command into .travis.yml and take out `yarn install`.

I think lerna has an option to delegate to yarn, which would be good.

taras:
That makes sense, I'm on it.

taras:
I had to use this http://blog.mact.me/2014/10/22/yosemite-upgrade-changes-open-file-limit because I was getting `Error: ENFILE: file table overflow` error.

taras:
I got the tests passing before running `lerna bootstrap`, but tests fail after I run `lerna bootstrap`.

I believe the next step is to declare dependencies on each package. The main package should only have the dependencies necessary to run the tests and build the project. I'm basing this on what I read in ["How bootstrap works" in lerna's README](https://github.com/lerna/lerna/blob/master/README.md#bootstrap)

Does it make sense?

ef4:
Yes, makes sense. The packages already try to declare their dependencies, but you are probably finding places where there are gaps.

taras:
I looked at how different projects that use Lerna run tests and each one has a different way of doing it. Cardstack will need its own way to do it. 

Right now, I'm trying to figure out what dependencies the root project should have. It depends on the role of the `cardstack-server` package. Will 'cardstack-server' package be installed as a dependency or will one of the nested packages be installed instead? 

Babel.js is consumed by installing the nested `babel-core` package. Ember.js is consumed by installing the root `ember-source` package.

Depending on, we'll need different dependencies and devDependencies in the monorepo package.json. 

What are your thoughts about the role of root package in the use of cardstack?

ef4:
I don't think the root package will ever be a thing people use. It doesn't
strictly need a package.json at all once we have all the dependencies
accounted for within each of the packages.

taras:
If we assume that lerna is installed globally on user's machine then we can use `learn run test` to run tests on all packages. Each package will need to be able to run its own tests using it's own dependencies and script in packages.json. 

If you're good with this, then I can start making changes to have each package be able to run it's own tests.

taras:
When I run tests inside of `@cardstack/authentication` module, I get the following error:

```bash
npm run node-tests

> @cardstack/authentication@0.0.9 node-tests /Users/tarasmankovski/Repositories/cardstack/packages/authentication
> mocha -r ../test-support/prepare.js node-tests/*-test.js



  authentication/middleware
    (read)
      1) "before all" hook: setup
      2) "after all" hook: teardown
    (read/write)
      token issuers
        3) "before each" hook: setup for "can update a user"
        4) "after each" hook: teardown for "can update a user"


  0 passing (30ms)
  4 failing

  1) authentication/middleware (read) "before all" hook: setup:
     Error: Cannot find module '@cardstack/hub/writers'
      at require (internal/module.js:20:19)
      at Resolver.retrieve (/Users/tarasmankovski/Repositories/cardstack/packages/di/index.js:69:21)
      at Container.factoryFor (/Users/tarasmankovski/Repositories/cardstack/packages/di/node_modules/@glimmer/di/dist/commonjs/es5/container.js:28:52)
      at Container.lookup (/Users/tarasmankovski/Repositories/cardstack/packages/di/node_modules/@glimmer/di/dist/commonjs/es5/container.js:51:28)
      at Container.lookup (/Users/tarasmankovski/Repositories/cardstack/packages/di/index.js:35:24)
      at exports.createDefaultEnvironment (/Users/tarasmankovski/Repositories/cardstack/packages/hub/node-tests/support.js:47:27)
      at process._tickCallback (internal/process/next_tick.js:109:7)

  2) authentication/middleware (read) "after all" hook: teardown:
     TypeError: Cannot read property 'teardown' of undefined
      at exports.destroyDefaultEnvironment (/Users/tarasmankovski/Repositories/cardstack/packages/hub/node-tests/support.js:80:12)
      at Context.teardown (node-tests/authentication-test.js:126:11)

  3) authentication/middleware (read/write) token issuers "before each" hook: setup for "can update a user":
     Error: Cannot find module '@cardstack/hub/writers'
      at require (internal/module.js:20:19)
      at Resolver.retrieve (/Users/tarasmankovski/Repositories/cardstack/packages/di/index.js:69:21)
      at Container.factoryFor (/Users/tarasmankovski/Repositories/cardstack/packages/di/node_modules/@glimmer/di/dist/commonjs/es5/container.js:28:52)
      at Container.lookup (/Users/tarasmankovski/Repositories/cardstack/packages/di/node_modules/@glimmer/di/dist/commonjs/es5/container.js:51:28)
      at Container.lookup (/Users/tarasmankovski/Repositories/cardstack/packages/di/index.js:35:24)
      at exports.createDefaultEnvironment (/Users/tarasmankovski/Repositories/cardstack/packages/hub/node-tests/support.js:47:27)
      at process._tickCallback (internal/process/next_tick.js:109:7)

  4) authentication/middleware (read/write) token issuers "after each" hook: teardown for "can update a user":
     TypeError: Cannot read property 'teardown' of undefined
      at exports.destroyDefaultEnvironment (/Users/tarasmankovski/Repositories/cardstack/packages/hub/node-tests/support.js:80:12)
      at Context.teardown (node-tests/authentication-test.js:126:11)




npm ERR! Darwin 16.5.0
npm ERR! argv "/Users/tarasmankovski/.nvm/versions/node/v7.9.0/bin/node" "/Users/tarasmankovski/.nvm/versions/node/v7.9.0/bin/npm" "run" "node-tests"
npm ERR! node v7.9.0
npm ERR! npm  v4.2.0
npm ERR! code ELIFECYCLE
npm ERR! errno 4
npm ERR! @cardstack/authentication@0.0.9 node-tests: `mocha -r ../test-support/prepare.js node-tests/*-test.js`
npm ERR! Exit status 4
npm ERR!
npm ERR! Failed at the @cardstack/authentication@0.0.9 node-tests script 'mocha -r ../test-support/prepare.js node-tests/*-test.js'.
npm ERR! Make sure you have the latest version of node.js and npm installed.
npm ERR! If you do, this is most likely a problem with the @cardstack/authentication package,
npm ERR! not with npm itself.
npm ERR! Tell the author that this fails on your system:
npm ERR!     mocha -r ../test-support/prepare.js node-tests/*-test.js
npm ERR! You can get information on how to open an issue for this project with:
npm ERR!     npm bugs @cardstack/authentication
npm ERR! Or if that isn't available, you can get their info via:
npm ERR!     npm owner ls @cardstack/authentication
npm ERR! There is likely additional logging output above.

npm ERR! Please include the following file with any support request:
npm ERR!     /Users/tarasmankovski/.npm/_logs/2017-04-26T21_15_53_313Z-debug.log
```

[2017-04-26T21_15_53_313Z-debug.txt](https://github.com/cardstack/cardstack/files/959960/2017-04-26T21_15_53_313Z-debug.txt)

ef4:
Ah, I think I see what is happening.

The `require` is located inside `@cardstack/di`. But lerna is not linking `@cardstack/hub` into `@cardstack/di`'s node_modules, because it's not listed as a dependency. So `@cardstack/di` can't find `@cardstack/hub`.

And I don't really want to need to list every possible DI-managed module as a dependency of `@cardstack/di`, so we will need to change the implementation a bit.

ember-cli does similar things with addons. It uses https://www.npmjs.com/package/resolve in order to do module resolution relative to places other than the current module. `@cardstack/di` should be told what top-level project it belongs to and do the resolutions relative to that project. In a test suite, the top-level project is the module under test. So in your example above, `@cardstack/di` would do module resolution relative to where `@cardstack/authentication` is located, and since `@cardstack/authentcation` has a dependency on `@cardstack/hub`, it would succeed.

taras:
@ef4 nice. Thank you for direction. I'll work on it from here.

taras:
screen shot 2017-04-27 at 7 36 03 pm

It works if I resolve relative to `process.cwd()` but that doesn't seem like the right solution. I got a bit further and there are other places where similar resolution needs to be done. Is it possible for us centralize it how it's done in ember-cli?

ef4:
Agreed, I think that the `Registry` exported from `@cardstack/di` should require a constructor argument like `projectPath` so that it's set explicitly, instead of `process.cwd`.

taras:
@ef4 I refactored the code to allow passing root path to Registry. The next challenge is resolving plugins. Should the [plugins be resolved](https://github.com/cardstack/cardstack/blob/master/packages/hub/plugins.js#L135) relative to working directory or hub module root?

ef4:
Plugins should be resolved relative to the same place that the Registry is using. 

Ideally, they would just use the registry instead of having any of their own resolution at all. But it gets more complicated, because the set of plugins that are allowed to run is controlled by the set of plugin-config models that exist, and that can be modified at runtime.

For an immediate path forward, perhaps the project path could actually move to being something that gets registered explicitly on the registry, like:

```js
registry.register('config:project', { path: myProjectPath }, { instantiate: false })
```

This would replace the constructor argument which I suggested earlier.

That would allow the hub's plugins module to inject `config:project` and read out the path. `plugins` is not currently loaded via DI, so that would also need to be changed.

taras:
> plugins is not currently loaded via DI, so that would also need to be changed.

IOW, we should be able to retreave plugins via the DI, so it would automatically use the path from `config:project`. I guess first step would be to register each plugin.

ef4:
I was referring to the plugins.js module itself.

Ultimately yes, the plugins themselves also, but that will take more
thinking.

taras:
Should I register Plugins as an injection to have it automatically injected into the `writer` when it's being looked up the first time?

ef4:
You should feel free to charge ahead if you can, but this issue may need to wait until I can address schema-dependent dependency injection.

taras:
I would be happy to pick this up after you `address schema-dependent dependency injection` or if you give me direction then I could try to address that change.

taras:
I spent a bit of time looking at what it would take to get Plugins instantiating via the DI. Here is what I found - saving here for posterity. 

When container.lookup(`hub:writers`) is called, the following happens:

```
Writers
	injects: schema-cache
	create -> 
            schema = this.schemaCache.schemaFromBranch(branch) 
		this._load(branch)
		     this.schemaFrom(models)
	                Schema.loadFrom(models)
                             plugins = Plugins.load(models)
                             new Schema(..., plugins)
```

I now understand that when you say `address schema-dependent dependency injection`, you're referring to instantiating Schema from DI, which intern could instantiate Plugins from DI.

ef4:
Yes. And the Schema you'd instantiate is not a singleton -- it's more like the way @glimmer/di would instantiate a component than the way it would instantiate a service. There's not one long-lived schema (or plugins), there can be multiple schemas that are created with different models as arguments.

This is all motivated by the fact that it should be possible to preview a feature branch directly on your site. The feature branch might contain schema changes. So the hub needs to know to behave differently when doing `POST /some-content` vs `POST /some-content?branch=new-feature`.

The object that *is* a long-lived singleton is SchemaCache. That's what other parts of the hub tend to inject so they can get access to branch-specific schemas.

taras:
Ok, so schemaCache will instantiate a Schema and use it for as long as necessary and then it'll get garbage collected when it's no longer necessary. 

I'm thinking about creating another branch to try doing this refactor. I'll keep this branch open and keep rebasing against master as I push other related PRs. This PR should eventually become mergeable. 

How does this sound?

ef4:
Good idea.

taras:
It looks like this is done in https://github.com/cardstack/cardstack/commit/a9740d5a430c9026b7473955a823b2bdead29e97

ef4:
Yes, thanks for exploring ahead, your work set the direction that let me refactor things to make lerna work.

taras:
👍 Glad to help. Working on getting the latest code to work with the demo app.

PR #14, opened 2017-05-12
Add @glimmer/di to @cardstack/di dependencies

courajs:


ef4:
This one already bit an app, so thanks.

(This class of bug happens because we don't actually depend on the individual packages package.json files in development. There is some initial work in PR #2 to fix that, but it depends on some architectural changes in the way we load plugins.)

PR #18, opened 2017-05-31
Seed directories

ef4:
So that different plugins can each contribute different seeds.

I'm also moving some things out of the current git seeds into the bootstrap schema, since they are directly needed by hub itself.


alexspeller:
@ef4 awesome, this is going to be super useful

PR #20, opened 2017-06-22
Prefix JSONAPI middleware

ef4:
The jsonapi middleware doesn't play nicely with others because it matches every possible path, so no other middleware can ever be sorted after it.

This change gives it a default prefix of /api, so

    /cardstack/articles/1

now becomes

    /cardstack/api/articles/1

While fixing this I also uncovered two places where we were returning relative URLs that should have been absolute (the Location header and the next link used for pagination).


ef4:
@alexspeller this will break the EP when you update cardstack.

PR #21, opened 2017-06-30
General style fixes

knownasilya:
screen shot 2017-06-30 at 8 53 38 am

---

This is what it was before:

screen shot 2017-06-30 at 8 54 46 am


ef4:
Thanks.

PR #25, opened 2017-08-02
Postgres Support

ef4:
I think we should hold off merging this until we land @courajs's container work so we can use that in CI to handle the test suite's PostgreSQL dependency.

 - [x] indexer
 - [x] writer


ef4:
@alexspeller I finished this, and it's merged and released as 0.2.2.

PR #26, opened 2017-08-04
Change di resolver to find the hub specified by the package

courajs:
The resolver was just checking whether it could successfully require the
hub, and fell back to going through test-support.

Due to package hoisting, this might not have the intended effect,
especially if you've locally linked test-support


ef4:
Looks good to me.

(We're getting lots of Travis failures like this one that seem to be due to losing contact with elasticsearch during the test run. I'm hoping we can address that when containerization lands.)

PR #27, opened 2017-08-14
Fix failure if no session is present

alexspeller:


ef4:
:+1:, feel free to merge after resolving the conflict.

PR #31, opened 2017-08-23
Tools update

ef4:
- bring the tools package up to date with recent changes elsewhere in the framework
 - add an acceptance test so it stays that way
 - begin incorporating a new ember-inject-optional addon to deal with optional services

With this PR, the tools dummy app should be fully working, with the exception of the "Create +" button in the top right (that needs to get decomposed so it only appear if cardstack has been told what types are create-able and where their "new" routes are).

cc @balinterdi this is the changes we were looking at earlier today.


packages/tools/tests/acceptance/tools-test.js
```diff
@@ -0,0 +1,22 @@
+import { test } from 'qunit';
+import moduleForAcceptance from '../../tests/helpers/module-for-acceptance';
+
+moduleForAcceptance('Acceptance | tools', {
+  beforeEach() {
+    delete localStorage['cardstack-tools'];
+  },
+  afterEach() {
+    delete localStorage['cardstack-tools'];
```
balinterdi: Do we need to have this both in the `beforeEach` and the `afterEach` hooks?

PR #32, opened 2017-08-25
First version of workflow package

balinterdi:
Implementation for #34


balinterdi:
@ef4 As discussed, we should merge back my branch to master, even if the current version of the workflow package is not widely useful yet.

I merged back the latest master and tests all pass locally. However, the Travis build fails on a weird (to me) error, in the loader.js library, specifically on [this line](https://github.com/ember-cli/loader.js/blob/v4.6.0/lib/loader/loader.js#L172), with the error `mod.module is undefined`. I updated loader.js to the latest version, to no avail.

It's probably not a good idea to merge back this branch to master as it will break the build.

Have you seen this error before or do you have a hunch on what could cause this?

ef4:
I think the problem in Travis was being caused by a failure to load seed models. So I merged in a commit that properly reports those errors. And then the error disappeared. :shrug:

balinterdi:
Great news!

Somewhat relatedly, I haven't npm published the `@cardstack/workflow` package and I'm not sure if I need to.

ef4:
You don't need to, but it does create a hazard since the next time somebody runs `lerna publish`, it will try to publish as a private package and fail. Because npm defaults to private for namespaced packages, and because we're not paying them for private packages.

I will publish it now. Once you do it the first time with `--access=public`, it will continue to default to public after that.

PR #35, opened 2017-09-21
Use a real inflector library for code generation

balinterdi:
I found out about this when `priorities` was singularized as `prioritie` :)

### DIFF
--- a/packages/models/cardstack/code-generator.js
+++ b/packages/models/cardstack/code-generator.js
@@ -1,5 +1,6 @@
 const { declareInjections } = require('@cardstack/di');
 const Handlebars = require('handlebars');
+const inflection = require('inflection');
 
 Handlebars.registerHelper('camelize', function(str) {
   return str.replace(/-(\w)/g, (m, d) => d.toUpperCase());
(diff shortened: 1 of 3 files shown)


ef4:
Thanks, this is a good improvement. 

Eventually I would like to make ember-inflector work within node so we can be more sure of getting the same answers in the browser vs in the hub.

balinterdi:
That would be great. Is there a way to do that via some javascript module re-importing/re-packaging mumbo-jumbo?

I seem to have seen a comment that that's possible but really mind-wrecking in one of the packages...

ef4:
The other place we need to do it (but haven't) is the jsonapi-factory module in test-support. There are currently two copies of it.

The best solution used to be writing in CJS and using browserify to make it work in the Ember app. But now I think we can use [@std/esm](https://www.npmjs.com/package/%40std%2Fesm) to load ES modules in Node and that would be nicer.

If `@std/esm` is mature enough and debuggable enough, I'm in favor of switching all Hub code to use ES modules.

balinterdi:
Do you mean [here](https://github.com/cardstack/cardstack/blob/master/packages/test-support/addon/jsonapi-factory.js)? I haven't found instances of singularization there.

ef4:
Oh, I wasn't referring to other places that need an inflector (although there probably are some). I was referring to places where we need to share a module between node and the browser.

PR #40, opened 2017-09-27
Fix misc upgrade errors

alexspeller:
This has a bunch of yarn.lock files committed but I think that's expected as that's from after running lerna bootstap when changing deps of a package so the yarn.lock should be committed in this case no?

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -3,6 +3,6 @@
   "packages": [
     "packages/*"
   ],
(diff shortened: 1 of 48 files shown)


ef4:
I cherry picked 785324e.

We don't need  51298cf because I already did 0aa3193b18085d30f2313790623a3cd113b2804a.

Released in 0.2.10.

PR #41, opened 2017-09-27
Log output of hub container when it fails to start

courajs:
Collect the output of the hub's container, and log it as an error if it fails to start.

It should be easier to look at the commits separately than the whole PR diff, since I refactored some beforehand.


courajs:
Closed in favor of #42

PR #43, opened 2017-09-28
Allow plugin to specify expiry of auth token

alexspeller:
Allow auth through query params along with header, and allow plugin to specify expiry of auth token


ef4:
This is probably ok, but I want to understand the use case. Putting auth tokens in URLs makes them easier to accidentally leak by accident.

alexspeller:
@ef4 removed the URL param bit as discussed

ef4:
Thanks, looks good.

PR #44, opened 2017-09-29
Better workflow package through dummy music app

balinterdi:


balinterdi:
@ef4 I resolved the merge conflicts ^^^ by updating the workflow package dependencies to 0.2.10, did `yarn` and `lerna update` but after that I get the following error when trying to run tests (or launch `ember server`):

> UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 2): TypeError: model.relationships.fields.data[Symbol.iterator] is not a function

PR #47, opened 2017-10-10
[WIP] Enable adding custom buttons

balinterdi:


balinterdi:
I needed to tweak the css because the component adds an extra div so `.footer > button` css rules did not apply. The other possibility is to define `tagName: ''` but that could interfere with handling the click action of the button.

balinterdi:
One thing to consider is whether the model object (e.g a song) needs to be involved to determine which buttons should be shown (if any). I like the pure-component approach I implemented and we can do the logic in the component itself (see the example below) but if we want to prevent anything to be inserted into the DOM below, we'd have to involve the model.

```hbs
<!-- app/templates/components/song-action-buttons.hbs -->
{{#unless model.isPublished}}
   <button class="cs-version-control-footer--button"
      disabled={{not (eq modificationState "saved")}}
      {{action (perform publish)}}>
     Publish
   </button>
{{/unless}}
```

packages/tools/addon/templates/components/cs-version-control.hbs
```diff
@@ -34,9 +34,14 @@
   {{#liquid-if enabled rules=animationRules growDuration=250 growPixelsPerSecond=1}}
     <footer class="cs-version-control-footer cs-field-editor-section">
       <a class="text-button" {{action (perform delete) }}>Delete</a>
-      <button disabled={{eq modificationState "saved"}} {{action (perform update) }}>
+      <button class="cs-version-control-footer--button" disabled={{eq modificationState "saved"}} {{action (perform update) }}>
         Update
       </button>
+      {{#if (is-component (concat modelType '-action-buttons'))}}
+        {{#with (concat modelType '-action-buttons') as |buttonsComponent|}}
```
ef4: If you move the `with` so it's outside the `if`, you don't need to say `(concat modelType '-action-buttons')` twice.

balinterdi: Good catch, fixed this in https://github.com/cardstack/cardstack/pull/52 .

PR #48, opened 2017-10-11
Soup up building the app docker image

courajs:
This adds the `ember hub:build` command to just build the app image.

The image build now handles any locally linked modules - although it has to `yarn install` for every one, so it may be very slow for large link graphs.

It also does some very interesting stuff with creating a dynamic docker build context so that we don't need to bind any user directories to pull in local code.

`packages/hub/Dockerfile` is for building the `cardstack/hub` base image.


ef4:
Woohoo, nice.

PR #49, opened 2017-10-12
Fix 'Cannot find module 'C' issue

goo32:
Ran across this issue when getting the [stupendous-rentals](https://github.com/cardstack/stupendous-rentals) demo running on my Windows 10 machine. This fix worked for me but see my comment in plugin-loader.js where I believe the problem originates.

### DIFF
--- a/packages/di/index.js
+++ b/packages/di/index.js
@@ -132,11 +132,14 @@ class Resolver {
   }
   retrieve(specifier) {
     let module;
-    let [type, name] = specifier.split(':');
+    let [type, name, rest] = specifier.split(':');
     if (type === 'hub') {
       module = this.hubPath + '/' + name;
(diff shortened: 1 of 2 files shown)


ef4:
Thanks! Good catch. Obviously this is under-tested on windows. 

I think we can split only on the final “:” to address this. 

That said, in order to eliminate the entire class of cross-platform bugs in Hub and to support easy standard deployment, we intend to make the Hub always run inside Docker. There is an open PR right now with the first part of that work.

goo32:
Right, I suppose split(":", 2) would have been sufficient. I can make the change if you're going to merge but if not I can make due until the dockerized hub. Looking forward to it!

ef4:
Thanks, I tweaked this to split with a regex. Please reopen if I got something wrong.

PR #54, opened 2017-10-15
Allow x-requested-with header

venkatd:
I'm not sure if this is the best fix since my knowledge of CORS is limited.

It looks like jQuery adds the [X-Requested-With](https://stackoverflow.com/questions/17478731/whats-the-point-of-the-x-requested-with-header) header.

CORS requests from Electron get rejected because of this.


ef4:
Thanks. X-Requested-With is just a convention (not a standard), but I guess it's prevalent enough for us to bother whitelisting it by default.

@courajs: this kind of bug is a big reason I want to always run cross-origin in dev, instead of proxying through ember-cli.

PR #69, opened 2017-11-01
dotbc-demo improvements

balinterdi:


balinterdi:
@ef4 I propose we merge this branch now to master and probably release a new version of the workflow package (of all cardstack packages)?

I'm suggesting this because:

1) The branch is now almost a month old and have diverged significantly from the released version
2) It has reached a state of maturity where further features/improvements in the workflow package are not as frequent and can more easily be conceptualized which makes it easier to create a new branch for each of those.

PR #70, opened 2017-11-01
Don't pre-emptively check for a running hub using the heartbeat port

courajs:
Instead, just fail the build if a hub isn't running.

It took me a while to figure out decent error handling, but I think I found something decent.

No hub running:
screen shot 2017-11-01 at 3 42 27 pm

Error response from hub:
screen shot 2017-11-01 at 3 43 00 pm


ef4:
Thanks, looks good.

PR #72, opened 2017-11-05
Adding editor and inline-editor options to fields

venkatd:


packages/tools/addon/helpers/cs-field-editor-for.js
```diff
@@ -1,9 +1,18 @@
-import Ember from 'ember';
-
+import { helper } from '@ember/component/helper';
 import { fieldType } from '@cardstack/rendering/helpers/cs-field-type';
+// @TODO: expose public api
+import metaForField from '@cardstack/rendering/-private/meta-for-field';
```
venkatd: Maybe we should move this to `@cardstack/models` and expose it publicly?

packages/tools/addon/helpers/cs-field-editor-for.js
```diff
@@ -16,4 +25,9 @@ export function csFieldEditorFor([content, fieldName], { variant }) {
   return `${prefix}field-editors/${type}-editor`;
 }
 
-export default Ember.Helper.helper(csFieldEditorFor);
+export default helper(csFieldEditorFor);
+
+function fieldOptions(content, fieldName) {
```
venkatd: I might get rid of this in favor of ensuring `meta.options` defaults to an empty has when using `metaForField`

packages/tools/tests/unit/helpers/cs-field-editor-for-test.js
```diff
@@ -20,11 +24,21 @@ test('it maps to a field editor component', function(assert) {
   assert.equal(result, 'field-editors/test-field-type-editor');
 });
 
+test('it can do a custom field editor', function(assert) {
+  let result = csFieldEditorFor([new TestModel(), 'theFieldWithCustomEditor'], {});
```
venkatd: Should we give the default variant a name?

packages/tools/addon/helpers/cs-field-editor-for.js
ef4: Agreed. It doesn't need to be private.

packages/tools/tests/unit/helpers/cs-field-editor-for-test.js
ef4: Sure, it could be "default".

ef4:
This all looks great. I pushed a commit to rename the new fields from `editor` to `editor-component` and `inline-editor` to `inline-editor-component` just because `editor` was already in use elsewhere in the test suite, which implies to me that it's a generally nice name apps might try to use too.

PR #73, opened 2017-11-06
Updated yarn.lock and added .nvmrc to gitgnore

habdelra:
After doing a `lerna clean` and a `lerna bootstrap`, i saw that one of the `yarn.lock` files that was checked in was not reflecting the latest from a `yarn install`.

I use nvm to manage my node versions across different projects as it's a lot to keep track of, I added it to the ignore files. if you'd like me to commit it, I'd be happy to do so.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 /node_modules
 /packages/*/node_modules
 .vagrant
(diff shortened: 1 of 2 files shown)


ef4:
🥂

PR #74, opened 2017-11-06
added mock authenticator

habdelra:


ef4:
This looks good. I just rebased it on top of the default addon blueprint so it's easier to see which parts you wrote vs what was generated. I'll merge as soon as CI goes green again.

PR #77, opened 2017-11-09
Add hub as a dependency in routing

balinterdi:
Routing imports `environment` from the hub package [here](https://github.com/cardstack/cardstack/blob/master/packages/routing/addon/routes/cardstack.js#L2) and yet it doesn't have it listed as a dependency.

### DIFF
--- a/packages/routing/package.json
+++ b/packages/routing/package.json
@@ -18,6 +18,7 @@
     "test": "ember try:each"
   },
   "dependencies": {
+    "@cardstack/hub": "^0.3.1",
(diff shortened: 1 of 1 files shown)


balinterdi:
Some tests fail in test-support, not sure if it's related.

balinterdi:
Closing this as we discussed that hub should be an explicit dependency in all cardstack apps for now.

ef4:
Formally, hub should be a `peerDependency` of all the other packages.

PR #78, opened 2017-11-10
integrate identity into left rail

habdelra:
This logic integrates identity into the left rail. This commit allows an app that wishes to use the cardstack 4-edges layout to be able to load it like this from the app's `application.hbs`template:

```hbs
{{#cardstack-tools}}
  {{outlet}}
{{/cardstack-tools}}
{{head-layout}}
{{cardstack-authentication}}
```

Additionally, I cleaned up the styling a bit for the left rail (we'll still need to do styling cleanup deeper down in the workflow plugin)


ef4:
This looks good.

PR #80, opened 2017-11-22
[WIP] Add activator blueprint for the ephemeral package

balinterdi:
This seems fine (I copied it from the jsonapi one) but the `cardstack-ephemeral-seeds` generator doesn't seem to be added to a project to which I linked this.

Can it be because `ephemeral` is purely a server-side cardstack plugin?


ef4:
Yes, specifically this package doesn't have `ember-addon` in its keywords
in package.json.

balinterdi:
So that's why Ember CLI doesn't detect its blueprint? At any rate, it'd be useful to have an activator for ephemeral, as otherwise it has to be manually activated (which is fine if there is no other way.)

ef4:
Agreed, although I have a branch already now (plugin-control) that changes
plugins to default enabled instead of default disabled. It’s just about
ready to merge.

balinterdi:
Great, then I don't think this is needed at all, closing.

PR #83, opened 2017-11-28
drop a bunch of now-extraneous plugin-configs

ef4:
... and blueprints for making them.

@balinterdi This probably breaks your tutorial instructions because of removed blueprints. Which is mostly a good thing as it lets you simplify.

Probably we should add a blueprint to the ephemeral package for the parts that are still needed.


balinterdi:
Wow, that's a biggie (but great).

Once this is merged in, I'll update the tutorial (and probably start writing the second part after that).

ef4:
Published as 0.4.0.

balinterdi:
Ah, ok, that was quick.

PR #85, opened 2017-11-29
Make @cardstack/search a cardstack plugin

balinterdi:
The `search` packages is not defined as a cardstack plugin and thus, when installed in a client app, throws the following error:

    Error: Unable to check whether plugin @cardstack/search is enabled: got exception Not Found from http://localhost:3000/api/plugins/%40cardstack%2Fsearch

@ef4 Some other packages also don't have the `cardstack-plugin` keyword and the related `api-version` definition (`di` and `bundler` possibly more) but I think that's because those are pure server-side packages that don't need this?

### DIFF
--- a/packages/search/package.json
+++ b/packages/search/package.json
@@ -3,8 +3,12 @@
   "version": "0.4.0",
   "description": "The default blueprint for ember-cli addons.",
   "keywords": [
-    "ember-addon"
+    "ember-addon",
+    "cardstack-plugin"
   ],
+  "cardstack-plugin": {
+    "api-version": 1
+  },
   "license": "MIT",
   "author": "",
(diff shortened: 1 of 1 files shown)


ef4:
Good catch.

Yes, up until recently the only reason to mark a package as a `cardstack-plugin` was if it exposed a server-side feature that the hub would need to discover.

As of yesterday, purely client-side plugins that want to control their build output based on whether the plugin is enabled also need to mark themselves as cardstack-plugins (which caused this bug).

di and bundler can stay as "just libraries" for now, since they don't do anything that would require them to be treated as cardstack plugins.

ef4:
Published `@cardstack/search` 0.4.3.

PR #87, opened 2017-11-30
[WIP] adding user notification cardstack plugin

habdelra:
Still WIP, posting for some early feedback. Note that this PR focuses on the underlying notification framework, and not really any payload considerations. A future PR will work on the actual notification payload.

TODO:
- [ ] Follow a subscribe/unsubscribe pattern that is more general and user agnostic.
- [ ] the decision to connect/disconnect to auth.io should be made depending on clients of the notification plugin that are subscribed to notifications
- [ ] remove the need to use components, and instead use just services with "optional-inject" addon
- [ ] "sessions" hub service that can get sessions from a token and get a token from a session and leverage in authenticator middleware and notification service
- [ ] Cleanup silly test in `mock-auth` plugin
- [ ] add actual tests for server
- [ ] add actual tests for client
- [ ] cleanup authenticator middleware copy/pasted logic in the notification service.js  (@ef4 I'm gonna need a little guidance here on how to share this correctly)

Also I added some eslint module definitions in the root package.json, as my vim is not smart enough to deal with that correctly at the package level.


packages/notification/cardstack/service.js
```diff
@@ -0,0 +1,104 @@
...
+
+    this.hasStarted = true;
+
+    await this.messengers.getMessenger(messengerName).then(({ socketIoUrl }) => {
+      if (!socketIoUrl) {
+        log.error(`Cannot start socket.io server, no 'socketIoUrl' defined for the '${messengerName}' resouce`);
+        return;
+      }
+
+      //TODO how do we know this server is actually the host referenced in the scocket.io URL?
```
habdelra: @ef4 I'm not too sure how to solve this. In the ephemal config you can define the URL for the socket.io server. The client will use the URL to communicate to the server. But we also need to start the socket.io server correctly from this config. I'm not sure how we'll be able to tell that some particular URL (e.g. https://myapp.com:3100/socket) is actually *this* server, or whether the URL is some other server and we should skip the socket.io startup.

courajs:
I'm working on this now, no need to leave this open

PR #88, opened 2017-12-01
add test coverage for our ast transform

ef4:
Because we're going to want to improve it soon and prior to this it was not obvious how to easily test it.


ef4:
@balinterdi in case you have a chance to look into that AST issue, I've added test coverage now so it's easier to extend.

balinterdi:
@ef4 Cool, I'll probably look into that next week, thank you.

PR #91, opened 2017-12-07
Use new logger

courajs:
This is mostly just changing `require('@cardstack/plugin-utils/logger')` -> `require('@cardstack/logger')`, but there are a couple things to look at:
- I tried to standardize on just using a module-scope function, instead of passing it in half the time and using `this.log`.
- Formatters are now registered on a logger instance, not globally on the module. See the [json diff formatter][json-formatter] and [its usage][json-usage], and the [table formatter][table-formatter] and [its usage][table-usage].
- I added configuration of the logging levels for the [full hub server][hub-config] and the [node test suite][test-config].
- I've temporarily removed failing the test suite on unexpected warn/error logs, as the proper hook doesn't exist yet in @cardstack/logger. There's an open issue: https://github.com/cardstack/logger/issues/2

[json-formatter]: https://github.com/cardstack/cardstack/compare/use-new-logger?expand=1#diff-ddd366dd9c585983735f8e8cf9898dd5
[json-usage]: https://github.com/cardstack/cardstack/compare/use-new-logger?expand=1#diff-4a2b8e469adb107ec3b493e2e780f38fR77
[table-formatter]: https://github.com/cardstack/cardstack/compare/use-new-logger?expand=1#diff-ac48e7f8c5627db0349c5cd3e3a89e61
[table-usage]: https://github.com/cardstack/cardstack/compare/use-new-logger?expand=1#diff-75475489e389172a263b367ab4e225aaR15
[hub-config]: https://github.com/cardstack/cardstack/compare/use-new-logger?expand=1#diff-01d0c7c4921de18abe789e1e94502c84
[test-config]: https://github.com/cardstack/cardstack/compare/use-new-logger?expand=1#diff-98d1ffe79c9335665ce07d4374064730R24


packages/hub/package.json
```diff
@@ -21,6 +21,7 @@
     "@cardstack/di": "^0.3.4",
     "@cardstack/elasticsearch": "^0.4.1",
     "@cardstack/handlebars": "^0.3.4",
+    "@cardstack/logger": "^0.1.0",
```
ef4: I think every package that instantiates a logger needs to do this directly.

packages/hub/searchers.js
```diff
@@ -1,5 +1,5 @@
 const { declareInjections } = require('@cardstack/di');
-const logger = require('@cardstack/plugin-utils/logger');
+const log = require('@cardstack/logger')('cardstack/searchers');
```
ef4: :+1: on favoring module-scoped loggers.

packages/hub/package.json
ef4: We have a gap right now in our testing that lets this class of bug slip through -- packages can depend on each other's dependencies without declaring them properly, thanks to yarn workspaces. I think we can plug the gap via eslint.

courajs:
@ef4 I couldn't really get the linting working, so I just manually added the dep for all the modules requiring it. I've opened an issue to add the linting later: #93

PR #92, opened 2017-12-08
Transform content-related dynamic expressions inside attribute assignments

balinterdi:


balinterdi:
@ef4 This is done and I also tested that it works in an actual application (using the "cardstackified" Super Rentals app), so we can write things like: `` and it will do the right thing.

One scenario that I have in mind but doesn't work is when there are multiple attribute assignments within the same tag, like ``, but I don't think we can AST transform it to something that works as `cs-field` can only yield one single value. 

Maybe if we turned it into a hash, but that's probably a longer issue, then.

ef4:
Nice!

Possibly the multiple-attribute case can be rewritten to a `cs-group` component, which yields a hash of multiple fields. We would need some way of naming the group though (the name is user-visible). That will need some more design, it doesn't need to block this PR.

packages/rendering/lib/transform.js
```diff
@@ -28,6 +36,49 @@ BindTransform.prototype.transform = function(ast) {
...
+
+          if (value.type === 'MustacheStatement') {
+            var path = value.path;
+            var parts = path.parts;
+            if (parts.length === 2 && parts[0] === 'content') {
+              foundDynamicContent = true;
+              // contentProperty is the property that is looked up on content
+              // (e.g `imageUrl` in the case of `content.imageUrl`)
+              contentProperty = parts[1];
+              newAttributes.push(b.attr(name, b.mustache(b.path('param1'))));
```
ef4: I think we will get a naming collision if the user happens to be already using something named `param1`.

We can detect this problem statically (by searching through the AST we're wrapping) and pick a free name instead.

balinterdi:
Great point about the potential naming collision. I've now made sure we can't have that and I think this is ready to be merged in. I'm excited to potentially have my first AST-related commit approved for an OS project :)

ef4:
🎉

PR #94, opened 2017-12-09
Lint for extraneous requires

### LINKED ISSUE #93: Add linting for requiring unspecified dependencies
We should pull in eslint-plugin-node, and specifically the no-extraneous-require rule: https://github.com/mysticatea/eslint-plugin-node/blob/master/docs/rules/no-extraneous-require.md

I tried to do this, but ran into trouble getting eslint to load the plugin. I can try again when I have some time to focus on it.

Because we're using a monorepo with yarn workspaces, we're especially susceptible to this - our modules can all require each other, and only break once published individually to the registry.

ef4:
Fixes #93.

This is split into two commits, one adding the linting infrastructure, the other fixing the resulting lint errors.

### DIFF
--- a/packages/eslint-config/index.js
+++ b/packages/eslint-config/index.js
@@ -10,8 +10,12 @@ module.exports = {
     'no-constant-condition': ["error", { checkLoops: false }],
     'require-yield': 0,
     'no-var': "error",
-    semi: ["error", "always"]
+    semi: ["error", "always"],
+    'node/no-extraneous-require': ['error', {
+      'allowModules': []
+    }]
   },
+  plugins: ['node'],
   globals: {
     Promise: false,
     Map: false,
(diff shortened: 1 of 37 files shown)


packages/eslint-config/index.js
```diff
@@ -10,8 +10,12 @@ module.exports = {
     'no-constant-condition': ["error", { checkLoops: false }],
     'require-yield': 0,
     'no-var': "error",
-    semi: ["error", "always"]
+    semi: ["error", "always"],
+    'node/no-extraneous-require': ['error', {
+      'allowModules': []
+    }]
   },
+  plugins: ['node'],
```
courajs: Ah, I think this is what I was missing when I was trying to get this working

PR #98, opened 2017-12-18
Mobiledoc markup

bagby:
Add ability to add markup to Cardstack editor Mobiledoc fields. This pulls in changes that were made in a separate Cardstack project. For now this will include just markup - `h2`, `p`, `b`, `ul`, etc. A future PR will address adding images and videos inline.


ef4:
Woohoo, nice progress.

bagby:
@ef4 I added a couple of acceptance tests. In order to get the dummy app to work I had to move a couple of packages from `devDependencies` to `dependencies` in cardstack-tools. Seems fine...?

ef4:
Yeah, I think that's fine as long as cardstack-tools really needs those packages at runtime.

bagby:
It does, it was getting by on the fact that `cardstack-models` has them as hard dependencies, and they are typically used in concert together. But that's not the case in the dummy app.

ef4:
@bagby I think this is good to merge unless you have more you wanted to do first?

The only thing I'd suggest as a further improvement here is to switch from `cs-icon` to the existing `cs-svg` that's already in the tools package, so we don't have to keep two different icon implementations. But that can be a followup PR.

bagby:
@ef4 nope I think this is good to merge. I'll create an issue for the follow-up thing you mentioned.

PR #99, opened 2018-01-03
Add Live Queries module

courajs:
You may want to review the diff [excluding the default addon blueprint](https://github.com/cardstack/cardstack/compare/cardstack:0741b03...cardstack:dc3d33c).

On the back end, we have a cardstack plugin which starts a socket.io server. Based on subscribe and unsubscribe events from clients, it manages state about which queries each client cares about (using a WeakMap to avoid leaking memory). It also listens to events from the indexer, and sends invalidate events to subscribed clients that may care.

Currently we get really simple events from the indexer - just one after every full indexing, so we always notify all subscribed clients.

On the client, there's a service to create and tear down subscriptions for a given query. The `liveQuery` computed helper uses this service to keep a full RecordArray up-to-date.


courajs:
@ef4 I just master in here to fix a `yarn.lock` merge conflict. I could do a quick walkthrough sometime this week if you want

ef4:
Thanks @courajs, this is great. I'm just switching back to this work after completing some other stuff so I will hopefully get a full review and merge done soon.

packages/hub/indexers.js
```diff
@@ -183,6 +187,7 @@ class Indexers {
     } finally {
       running.destroy();
     }
+    this.events.emit('index_update');
```
habdelra: I think it might be a good idea to include the hints in the event that is emitted so that subscribers to this event can make more nuanced decisions about what to do based on what is being indexed.

ef4: I think we can do one better. Instead of this coarse event, we can emit events whenever an indexer calls the `save` or `delete` operations.

packages/live-queries/cardstack/service.js
```diff
@@ -0,0 +1,104 @@
...
+      let subs = new ClientSubscriptionSet(socket);
+      this._client_subscriptions.set(socket, subs);
+
+      socket.on('query-subscribe', subs.subscribe.bind(subs));
+      socket.on('query-unsubscribe', subs.subscribe.bind(subs));
+    });
+  }
+
+  listenToIndexer(indexEvents) {
+    indexEvents.on('index_update', () => this._invalidateAll());
```
habdelra: maybe not for this PR, but as a future direction, I wonder if we should introduce a new hub feature that plugins can use that will allow a plugin to determine which models to invalidate based on a hint being passed within the index event. maybe it's called "live-indexer" feature?

packages/live-queries/addon/live-query.js
```diff
@@ -0,0 +1,43 @@
...
+  return Ember.computed(...deps, function() {
+    if (query) {
+      // we have already run before, we're recomputing due to a dependent key change
+      query.cleanup();
+    } else {
+      // first time
+      this.on('willDestroyElement', () => query.cleanup());
+    }
+    let inputs = fn(getProperties(this, deps));
+    query = new LiveQuery(inputs, getOwner(this));
```
habdelra: In the readme we mention that we should only use LiveQuery in a component due to the lifecycle operations that a component has. I think we should at the very least emit a warning if this is not within a Component, and probably actually we should throw an error if that is the case.

packages/live-queries/cardstack/service.js
ef4: yeah, Aaron and I discussed this. The plan is to make sure that the client-to-server API provides the full query, giving us a place in the server where we could steadily make the invalidation smarter by understanding more queries, while erring on the side of over-invalidating.

packages/hub/indexers.js
```diff
@@ -182,7 +182,7 @@ class Indexers {
     } finally {
       running.destroy();
     }
-    this.events.emit('index_update');
+    this.events.emit('index_update', hints);
```
habdelra: @ef4 I added the hints to this event for now. from the indexer it doesnt look like I have direct access to the operations. The event listeners need to get access to the event emitter in order to wire up correctly. Right now that happens pretty easily as there is just one `hub:indexers` that can be injected in the live-query service.

packages/hub/indexers.js
```diff
@@ -55,6 +56,7 @@ class Indexers {
     this._realTimeQueue = [];
     this._seedSchemaMemo = null;
     this._schemaCache = null;
+    this.events = new EventEmitter();
```
ef4: Is there a reason we aren't just extending Indexers from EventEmitter, so that it would be `indexers.on('index_update', ...)` instead of `indexers.events.on('index_update', ...)`?

habdelra: good idea. I'll give that a shot

habdelra:
So I've made the update to use more fined grained events, however, i'm seeing awkward timing problems in that the client seems to be reloading the model before it is available. and seems to always get the n - 1 model change. I can leave the more fine grained events in the system, but I think for the time being, we might wanna rely on the more ambiguous `index_updated` event. what do you think?

habdelra:
@ef4 another thought is that the indexer could queue up the add/delete events and then flush them after the `_doUpdate()` method in the indexer has completed

habdelra:
@ef4 actually the queuing up of the fined grained events and then flushing after the index updated was really simple to implement and works great. included in the latest commit.

ef4:
I think even the coarser event is going to suffer the same bug sometimes, it's just less likely because it happens later, and you're racing against when elasticsearch decides to refresh the search index, except in cases where the update had `{realTime: true}`.

I don't think we should queue up all the events because during big indexing operations there may be a huge number of them, and very few would actually be relevant to current live queries. It would be better for the event handler to be able to cheaply rule out ones that don't matter and aggregate dirtyness (like, once you know your query is dirty it doesn't help to track another thousand ids that also changed).

Here's what I think we should do:

  1. Keep the fine-grained events -- the live query module should watch them and use them to build up the minimal state necessary to know which queries have dirtied. (which is going to be much smaller than the whole list of every event that fires)
  2. Keep the coarse-grained event and call it "update_complete". The live query module will wait for this one to take action on the state that it updated based on the fine-grained events.
  3. I think see a way to make _all_ queries realTime: true without hurting perf. That will eliminate the remainder of the race. I will make that PR now.

habdelra:
@ef4 ok, I made the adjustments. the fine grained events now build up in the live-query service, until the coarse `update_completed` event is received, and then they are cleared. I'm not using the fine grained events to make smarter decisions around how to invalidate yet though. I think  we can do that in subsequent PR.

habdelra:
@ef4 I was thinking about how we could be smarter with the invalidation based on the fine grained events. I think not in this PR, but a subsequent PR, we could use the grants associated with the indexed content and the identity (or lack thereof) of the live-query subscribers to determine which subscriptions to send an invalidation signal. we could even be more precise and send the id and type of the models to invalidate...

habdelra:
Published as: ``` - @cardstack/authentication: 0.5.9 => 0.5.10
 - @cardstack/codegen: 0.5.9 => 0.5.10
 - @cardstack/drupal-auth: 0.5.9 => 0.5.10
 - @cardstack/drupal: 0.5.9 => 0.5.10
 - @cardstack/elasticsearch: 0.5.8 => 0.5.10
 - @cardstack/email-auth: 0.5.9 => 0.5.10
 - @cardstack/ephemeral: 0.5.8 => 0.5.10
 - @cardstack/ethereum: 0.5.9 => 0.5.10
 - @cardstack/git: 0.5.9 => 0.5.10
 - @cardstack/github-auth: 0.5.9 => 0.5.10
 - @cardstack/hub: 0.5.9 => 0.5.10
 - @cardstack/jsonapi: 0.5.9 => 0.5.10
 - @cardstack/live-queries: 0.4.1 => 0.5.10
 - @cardstack/mock-auth: 0.5.9 => 0.5.10
 - @cardstack/models: 0.5.9 => 0.5.10
 - @cardstack/postgresql: 0.5.9 => 0.5.10
 - @cardstack/routing: 0.5.9 => 0.5.10
 - @cardstack/test-support: 0.5.9 => 0.5.10
 - @cardstack/tools: 0.5.9 => 0.5.10
 - @cardstack/workflow: 0.5.9 => 0.5.10```

PR #103, opened 2018-02-05
Added CORS headers to the auth config endpoint

habdelra:


ef4:
This definitely solves the issue you saw, but it's broader than I would prefer. This is going to put cors headers on nearly every request to hub, including ones that aren't handled by this plugin.

The `_tokenVerifier` wraps every other middleware in the system (to make the session available to those requests).

I think the cleanest solution is to move `addCorsHeaders` directly into each of the three other endpoints instead (because each of those is terminal -- they don't call `next`, so they don't wrap any other middlwares).

habdelra:
@ef4 sure no prob, I just updated based on your suggestion.

PR #106, opened 2018-02-08
rewrite external user when dealing with authentication result that is a partial session + mock user partial session support

habdelra:
I think partial sessions have enjoyed the fact that the email auth is intrinsically JSON API. I believe we need to convert the auth response using the user template before we start checking for the `type` as this blows up when it deals with an auth response that is not JSON API (like Auth0).

### DIFF
--- a/packages/authentication/cardstack/middleware.js
+++ b/packages/authentication/cardstack/middleware.js
@@ -128,19 +128,7 @@ class Authentication {
       user = await this._processExternalUser(result, source);
     }
 
-    if (result && result.meta && result.meta.partialSession) {
-      if (!user) {
-        // top-level meta is not passed through (it was for
-        // communicating from plugin to us). Plugins could use
-        // resource-level metadata instead if they want to.
-        delete result.meta;
-        user = result;
-      }
-
-      if (user.data.type !== 'partial-sessions') {
-        user.data.type = 'partial-sessions';
-      }
-
+    if (user && user.data && user.data.type === 'partial-sessions') {
       ctxt.body = user;
       ctxt.status = 200;
       return;
@@ -171,6 +159,10 @@ class Authentication {
   async _processExternalUser(externalUser, source) {
     let user = rewriteExternalUser(externalUser, source);
     if (!user.data || !user.data.type) { return; }
+
+    if (user.data.type === 'partial-sessions') {
+      return user;
+    }
 
     let have;
 
--- a/packages/authentication/cardstack/middleware.js
+++ b/packages/authentication/cardstack/middleware.js
@@ -16,6 +16,10 @@ function addCorsHeaders(response) {
   response.set('Access-Control-Allow-Headers', 'Content-Type');
 }
 
+function isPartialSession(doc) {
+  return doc.meta && doc.meta['partial-session'];
+}
+
 module.exports = declareInjections({
   encryptor: 'hub:encryptor',
   searcher: 'hub:searchers',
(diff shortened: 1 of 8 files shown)


ef4:
I was originally assuming that if a plugin wants to establish a partial session, it's required to return a JSON:API document that has `{ meta: { partialSession: true } }`. And we're already checking for `meta`, so it seems like we're already expecting a valid JSON:API structure.

But I think I see where you're going with this, and I think it may be even better than what I have, if we refactor this slightly.

I don't think we need to call `rewriteExternalUser` from two places. I think we can just move the partial session handling to after `processExternalUser` would have already happened. I think [right here](https://github.com/cardstack/cardstack/pull/106/files#diff-1c6d4a972df4b744439f1667c56d22d2R158) is where we could check for `meta.partial-session` and just pass it through.

This is nice because it means apps can control what they consider to be partial by just changing their user template.

habdelra:
sounds good

packages/authentication/cardstack/middleware.js
```diff
@@ -143,6 +128,24 @@ class Authentication {
       user = await this._processExternalUser(result, source);
     }
 
+    if (result && result.meta && result.meta.partialSession) {
```
ef4: If this checks user.meta instead of result.meta, it will give the userTemplate control over whether this is a partial response.

packages/authentication/cardstack/middleware.js
```diff
@@ -143,6 +128,24 @@ class Authentication {
...
+        // top-level meta is not passed through (it was for
+        // communicating from plugin to us). Plugins could use
+        // resource-level metadata instead if they want to.
+        delete result.meta;
+        user = result;
+      }
+
+      if (user.data.type !== 'partial-sessions') {
+        user.data.type = 'partial-sessions';
+      }
```
ef4: This was originally intended as a default, not an override. Users can have their own model for partial sessions, we only care about meta.partialSession.

packages/authentication/cardstack/middleware.js
```diff
@@ -143,6 +128,24 @@ class Authentication {
...
     }
 
+    if (result && result.meta && result.meta.partialSession) {
+      if (!user) {
+        // top-level meta is not passed through (it was for
+        // communicating from plugin to us). Plugins could use
+        // resource-level metadata instead if they want to.
+        delete result.meta;
+        user = result;
+      }
```
ef4: I don't think we really need this whole block anymore.

packages/authentication/cardstack/middleware.js
habdelra: @ef4 Yeah, but email-auth is not returning valid JSONAPI, and hence `_processExternalUser` returns null. So that is why I need to check both `result` and `user`

packages/authentication/cardstack/middleware.js
habdelra: Ok

packages/authentication/cardstack/middleware.js
habdelra: Sure

packages/authentication/cardstack/middleware.js
ef4: I see, the only reason it's not returning valid JSON:API is because it was relying on getting `{type: "partial-sessions" }` as a default, which now happens too late. Let's jettison the idea of having a default type and just make the plugin set it. So email-auth should just add that type explicitly, and we can drop the stuff at line 140 about setting a default type.

habdelra: Sounds good

habdelra:
@ef4 ok, i think we're good now

ef4:
Thanks!

Published as

```
 - @cardstack/authentication: 0.4.8 => 0.4.9
 - @cardstack/drupal-auth: 0.4.8 => 0.4.9
 - @cardstack/email-auth: 0.4.8 => 0.4.9
 - @cardstack/github-auth: 0.4.8 => 0.4.9
 - @cardstack/mock-auth: 0.4.8 => 0.4.9
```

PR #109, opened 2018-02-10
use a service for mock-login for ever friendlier testing

habdelra:
@ef4 It would be great to get a version rev after this PR gets merged,


ef4:
This looks good, I will merge.

One suggestion for next time you're poking at this area: since you're providing a component anyway, the component can have its own input field for the username, instead of (or in addition to) parsing the url.

ef4:
Published in 0.4.11.

PR #111, opened 2018-02-13
added user correlation query

habdelra:


packages/authentication/cardstack/middleware.js
```diff
@@ -164,15 +166,23 @@ class Authentication {
     if (!user.data || !user.data.type) { return; }
 
     let have;
+    let query = source['userCorrelationQuery'];
 
-    if (user.data.id != null) {
-      try {
+    try {
+      if (user.data.id != null) {
         have = await this.userSearcher.get(user.data.type, user.data.id);
```
ef4: I don't think this should take priority over the userCorrelationQuery. It's legit to want to have externally -provided IDs while also doing custom correlation.

I think you can get rid of this step entirely by making the default userCorrelationQuery be one that just finds by ID.

packages/authentication/cardstack/middleware.js
```diff
@@ -164,15 +166,23 @@ class Authentication {
...
+        let compiled = Handlebars.compile(query);
+        have = await this.userSearcher.search(JSON.parse(compiled(user)));
       }
+    } catch (err) {
+      if (err.status !== 404) {
+        throw err;
+      }
+    }
+
+    if (have && have.data && _.isArray(have.data)) {
```
ef4: We have ECMA standard `Array.isArray` now, no need for lodash here.

packages/authentication/cardstack/middleware.js
```diff
@@ -164,15 +166,23 @@ class Authentication {
...
+      have = have.data.length ? _.merge({ data: have.data[0] }, user) : null; // deep merge
```
ef4: Why are we merging? I think the only time we would apply `user` to `have` is later if `mayUpdateUser`, and in that case the writers system does it for us.

In other words, I don't think anything about merging needs to change in this PR.

packages/authentication/cardstack/middleware.js
habdelra: ah--good idea. will do

packages/authentication/cardstack/middleware.js
habdelra: sure, no problem. I was actually think about this on my train ride in, thinking that we actually don't want auth0 to stomp on data here: like what if a user wants to change their name from how it appears in auth0. totally makes sense

PR #114, opened 2018-02-14
Read grants

ef4:
This completes our grant system with may-read-resource and may-read-fields permissions. The best docs at the moment are at the top of packages/hub/schema/grant.js


ef4:
Published as 0.5.0.

PR #119, opened 2018-02-15
add case-insensitive field type

habdelra:


ef4:
Very nice, this looks great.

PR #120, opened 2018-02-21
Allow Hub to start in Windows

bagby:
Windows doesn't understand shebangs (`#!/usr/bin/env node`), so we should explicitly pass the `node` process to `spawn`, and put `bin/cardstack-hub.js` in `args`.


ef4:
Thanks, looks good.

ef4:
Published in 0.5.4.

PR #123, opened 2018-02-28
build hub schema from contract`s ABI

habdelra:
@ef4 this is the very first step in the road to the ethereum plugin, where we are constructing a hub schema based on the ABI of a contract. For the tests I just grabbed a standard ERC-20 token contract ABI and made assertions that the schema generated in the hub reflected the structure of the contract. My intent is that the contract mapping entries (which represent the ledger of an ERC-20 contract), are separate content types for which the contract content-type `has-many` of, and which each mapping entry content-type `belongs-to` the contract. So the idea is that the user's token balance is a document in the hub that we index whenever there is an ethereum transfer event (which will be done in a future PR).


packages/ethereum/node-tests/ethereum-indexer-test.js
```diff
@@ -0,0 +1,205 @@
...
+                address: ""
+              },
+            }
+          }
+        },
+      });
+
+    env = await createDefaultEnvironment(`${__dirname}/..`, factory.getModels());
+
+    await env.lookup('hub:indexers').update({ realTime: true });
```
ef4: createDefaultEnvironment does this for you automatically, it's not needed here.

habdelra:
Yeah, I’m anticipating that I’ll be adding client code in this plugin. If that turns out to not be necessary I’ll remove the ember pieces.

packages/ethereum/node-tests/ethereum-indexer-test.js
habdelra: Cool thanks

PR #124, opened 2018-03-01
Adding private blockchain and integrating truffle testing

habdelra:
@ef4 so for the truffle tests I followed the example the we did for the `ember-tests`, where we discover packages that use truffle based on the presence of a `truffle.js` file (which all truffle projects have). Additionally, I updated the README to describe how to startup the private blockchain (it's an npm script in package.json) for people that want to run the truffle tests. Additionally, I added real solidity contracts into the plugin, and added some sanity checks into the tests that would fail if the contract state was not being reset in between the tests. Also, I augmented the hub schema based test to grab the contract's ABI from an actual compiled solidity contract to prove we can mix hub based tests into truffle's testbed.


habdelra:
Superceded by https://github.com/cardstack/cardstack/pull/127

PR #136, opened 2018-03-09
use hints when they are available for more efficient updateContent

habdelra:
@ef4 it looks like I didn't even need to use `owningDataSource` as `this.updaters` already had everything that I needed.


ef4:
Nice.

Issue #134, opened 2018-03-07
Improved data source management for test suites


ef4:
Planning notes based a call with @habdelra:

The near-term plan:

1. Generic data source seeding
   - lets rename `seeds` to `data-sources`, because that's all that's really supposed to be in there. 
   - then we can use `seeds` for a more generic version of the ephemeral store's `initialModels`. You would manually load seed models into your data sources (_any_ data sources, not just ephemeral) by running a command line tool (`ember hub:seed`). 
   - By convention, the ephemeral store would still load effectively do `hub:seed` automatically at boot, so it behaves the same as today.
   - no more separate dev/test/prod directories for seeds (or data sources). We're finding it better to manage the distinction via env vars.
2. Postgres-specific seeding & testing
    - add command line tools for conveniently creating and destroying databases
    - create includes migrations (though they would still also run automatically if a running indexer sees that they're needed, as they do now).
   - the seed command from above would necessarily let you seed a postgres database, as long as it already exists.
   - `ember test` will require the database to already exist, it will not make it for you.
   - the baseline state between tests is a database that exists and has schema (from the migrations) but it does not include any automatic use of seeds.
   - each test is responsible for its own setup and teardown to get back into the baseline state.
   - the baseline state does not include seed models automatically, but individual tests can choose to load them explicitly (and then clean them up to get back to baseline)
3. Test Support package improvements
    - we need to expose a built-in privileged user for the clientside test-support code to run as, when setting up and tearing down tests
    - the existing test-support/addon/fixtures.js API needs to be extended to support data sources that can't do arbitrary checkpointing.  The `ephemeral-checkpoints` and `ephemeral-restores` data types would go away in favor a generic middleware provided by test-support that can create and restore checkpoints.
    - writers will get new hooks for creating and restoring checkpoints. Create is optional (not every data source can create arbitrary checkpoints). But you can implement restore without implementing create by supporting pre-defined checkpoints (how they are predefined is data-source specific configuration).
    - in the case of postgres, the "empty checkpoint" could be configured as a series of sql statement to run that get you into that state (mostly of the form "delete from some_table", but arbitrary sql is allowed in case your baseline state isn't truly empty).
        ```
        factory.addResource('data-sources', 'postgres').withAttributes({
          sourceType: '@cardstack/postgresql',
          params: {
            branches: {
              master: {
                migrationsDir: 'cardstack/migrations'
              }
            },
            testCheckpoints: {
              empty: [
                `delete from users`,
[...]

PR #141, opened 2018-03-19
Seeds refactor and Checkpoint refactor

habdelra:
Addresses many items in the issue https://github.com/cardstack/cardstack/issues/134 

- uses new `initial-data/` folder layout
- introduces the folder `initial-data/data-sources` folder that holds the data source definitions
- introduces the folder `initial-data/seeds` folder that holds the seed models
- adds a new command `ember hub:seed` to load the seed models that do not come from ephemeral data sources.
- generalizes checkpoint creation and restores to use test-support writer along with writer hooks in the various plugins. along with ephemeral checkpoint refactoring to use this approach
- adding CI session ID support to allow CI to use a privileged session for test setup and teardown via an environment variable.


habdelra:
@ef4 ok ive completed the refactors that we've discussed around both the test-support's writer.js for dispatching checkpoints/restores to the various plugins' writers, as well as using a command for loading seeds instead of an option.

PR #143, opened 2018-03-25
added postgresql checkpoint restore

habdelra:


packages/postgresql/node-tests/writer-test.js
```diff
@@ -197,4 +212,43 @@ describe('postgresql/writer', function() {
...
+      });
+    expect(response).hasStatus(201);
+
+    result = await client.query('select * from articles');
+    expect(result.rows).has.length(0);
+
+    // The DB connection is still being used by the server, such that dropping the table at
+    // this point will result in an error. unsure how to await that, or force the table drop
+    // to not care....
+    await timeout(5000);
```
habdelra: @ef4 I could use some help here. I just couldn't figure out how to await this correctly so that the test teardown doesn't get all upset. I tried reaching directly into the container and awaiting the `teardown()` for the postgresql's writer and indexer, but that didn't seem to make a difference...

ef4: I would have expected `destroyDefaultEnvironment` to already do the right thing. That is what causes the whole container to get torn down, which should cause all the postgres connection pools to get closed.

habdelra: Yeah, same. I’m pretty confused by this.

habdelra: I did put a breakpoint in the `teardown()` in the postgresql writer, and saw  it was never called when the container was torn down in this test.

PR #146, opened 2018-03-28
more consistent case for contract methods

habdelra:


packages/drupal/package.json
```diff
@@ -25,7 +25,7 @@
   "dependencies": {
     "@cardstack/logger": "^0.1.0",
     "@cardstack/plugin-utils": "^0.6.0",
-    "ember-source": "~2.17.0-beta.1",
+    "ember-source": "~2.18.2",
```
ef4: Unintentional change?

habdelra: i figured we should use the same ember version across the various plugins (both use them for Ember.String). seemed wasteful to have different ember versions in the workspace node_modules for such a simple thing

ef4: Oh yeah, good call.

PR #154, opened 2018-03-30
added node eslint rules in cardstack folders

habdelra:
My editor is getting upset that there aren't node eslint rules in the cardstack folders after the recent eslint updates.

### DIFF
--- a/packages/authentication/cardstack/.eslintrc.js
+++ b/packages/authentication/cardstack/.eslintrc.js
@@ -1,0 +1,4 @@
+module.exports = {
+  root: true,
+  "extends": '@cardstack/eslint-config'
+};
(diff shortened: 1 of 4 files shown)


ef4:
This should not be necessary. Please check how your editor is running eslint. If it has its own global copy of eslint, it may be out of date.

Everything is working for me, both when running eslint on the command line and from my editor.

The way it's supposed to work is: `packages/authentication/.eslintrc.js` extends `@cardstack/eslint-config/browser`, that config has `overrides` that apply our node rules to `cardstack/**/*.js`.

I switched to this pattern because it's the same way the current version of ember-cli applies separate rules for node vs browser files.

habdelra:
ok--thanks, will do

PR #156, opened 2018-03-30
More missing remappers

habdelra:
More missing remappers (also includes commit from PR https://github.com/cardstack/cardstack/pull/155)


ef4:
:+1:

Issue #160, opened 2018-04-04
Postgresql should use sql defined in checkpoints for a restore, not sql defined in a restore


habdelra:
Right now the sql restore uses sql that are part of that restore model. Rather, the postgresql restore should look up a checkpoint, and execute the sql associated the checkpoint. The checkpoint with it's sql can be established in the config for the postgresql data source.

habdelra:
![](https://github.trello.services/images/mini-trello-icon.png) [refactor: use sql defined checkpoint for postgresql restore](https://trello.com/c/mMPYCo08/109-refactor-use-sql-defined-checkpoint-for-postgresql-restore)

habdelra:
We addressed this by refactoring Fixtures to make them work for all data sources. Scenarios created with fixtures replace the approach we were previously taking with checkpoints and restores: https://github.com/cardstack/cardstack/pull/169

PR #165, opened 2018-04-04
fixes issue #160 - refactor checkpoint and restores

habdelra:


ef4:
Checkpoints are already a first-class content-type -- they don't belong down inside `params` on the postgres data source. 

Looking up the checkpoint that's being restored is generic -- every data source needs the same thing. So the code for that can go in hug/writers.js:restoreCheckpoint(). The document passed into writer.prepareCheckpointRestore can be the actual checkpoint, all ready to go.

ef4:
Yes, that should be allowed. It does mean you’ll need some place to store them that actually works. 

This is another reason I’m doing my data source routing refactor — you shouldn’t need to worry about this stuff. Merely making Postgres your default data source should take care of creating all the schema that is needed by every plugin, including test-support.

habdelra:
so I shouldn't have to worry about having a migration script that accommodates a checkpoint table is the default data source is postgres?

habdelra:
hmmmm, actually i'm confused why the hub:writer even has checkpoint stuff in it. checkpoints are handled by the `test-support`. and a plain create on the 'checkpoint' or 'restore' content type kicks off the whole thing. I think the hub's checkpoint stuff is no longer used and never got cleaned up when we refactored.

ef4:
Oh yes, I was mistaken. I was confused by the vestigial code in hub:writers. You're right, all the generic parts go in the test support writer.

Let's discuss tomorrow, working backward from the API we really want for tests to use, starting with how test-support/addon/fixtures.js already works.

habdelra:
Closing this to approach the test setup/teardown from a different direction based on our discussion

Issue #159, opened 2018-04-04
Assert ethereum contract model names are plural


habdelra:
Right now the ethereum contract model name inflections has no opinions. this should change to assert that all the content types that are derived from the contract have a plural inflection.

habdelra:
![](https://github.trello.services/images/mini-trello-icon.png) [Github issue https://github.com/cardstack/cardstack/issues/159](https://trello.com/c/qi7tBy37/108-github-issue-https-githubcom-cardstack-cardstack-issues-159)

habdelra:
Fixed in PR https://github.com/cardstack/cardstack/pull/166

PR #166, opened 2018-04-04
fixes issue #159 treat ethereum content types with plural inflection

habdelra:


ef4:
I would rather we could get ember-inflector working in the hub, so that we use the same rules everywhere. But that turns out to be annoyingly hard to do, so this is fine.

PR #169, opened 2018-04-06
Refactoring Fixtures to support all data source types and to provide teardown

habdelra:
Fixture refactoring provides support for all data source types, as well as to provide teardown. This approach will replace the checkpoint/restore approach we were taking previously.


packages/hub/util/bin-utils.js
```diff
@@ -6,6 +6,7 @@ const fs = require('fs');
 
 const seedsFolder = 'seeds';
 const dataSourcesFolder = 'data-sources';
+const defaultTestingToken = 'testing-token';
```
ef4: Please randomize this, I don't want it to be possible for people to accidentally hard code it anywhere.

It can be made available to the browser via a code generator. It can probably just go in plugin-utils/code-generators/environment.js

packages/hub/util/bin-utils.js
```diff
@@ -67,6 +68,8 @@ HUB_ENVIRONMENT                   The environment the hub is running in. Possibl
 
   if (process.env.CI_SESSION_ID) {
     commander.ciSessionId = process.env.CI_SESSION_ID;
+  } else if (process.env.HUB_ENVIRONMENT === 'test') {
+    commander.ciSessionId = defaultTestingToken;
```
ef4: Based on my previous comment, this should just go away.

packages/live-queries/tests/acceptance/live-query-test.js
```diff
@@ -4,17 +4,19 @@ import { visit, fillIn, click } from '@ember/test-helpers';
...
 
-let scenario = new Fixtures('default', factory => {
-  factory.addResource('items').withAttributes({content: 'hello'});
+let scenario = new Fixtures({
+  create(factory) {
+    factory.addResource('items').withAttributes({content: 'hello'});
+  },
+
+  destroy() {
+    return [{ type: 'items' }];
```
ef4: I think this is not needed if we are automatically destroying everything that was made during `create`. Is that implemented?

packages/test-support/code-generator.js
```diff
@@ -0,0 +1,33 @@
...
+    let value = this.ciSessionId && this.ciSessionId.id;
+
+    return value ? template({ properties: [
+      {
+        name: 'ciSessionId',
+        value
+      }
+    ]}) : null;
+  }
+});
```
ef4: Oh, I see here you already have a code generator for the ci session, this is good. So then the only change is randomizing it where it's generated.

packages/hub/util/bin-utils.js
habdelra: it is in a code generator. I added it to the test-support package as that's the closest place to where it is consumed.

but yeah, I'll randomize it

packages/hub/util/bin-utils.js
habdelra: sure.

packages/live-queries/tests/acceptance/live-query-test.js
habdelra: good point. the destroy does automatically remove anything that was in the `create`

packages/hub/util/bin-utils.js
habdelra: @ef4 so one aspect about this i feel a little uneasy about is that we only have a CI session token if the hub is started with that as an env var when in production mode. the CI session token allows unimpeded access. I feel a little uncomfortable allowing the ability to have unimpeded access tot he hub in production mode when you did specifically set that up, that's why we're using an env var to drive this in production mode. In the test env, using a random value for this is totally fine.

packages/hub/util/bin-utils.js
habdelra: @ef4 so one aspect about this i feel a little uneasy about is that we only have a CI session token if the hub is started with that as an env var when in production mode. the CI session token allows unimpeded access. I feel a little uncomfortable allowing the ability to have unimpeded access tot he hub in production mode when you did specifically set that up, that's why we're using an env var to drive this in production mode. In the test env, using a random value for this is totally fine. given that, I'm happy to make this random, but I dont think we should set this value when in production mode and there isn't a env var telling us that we are in CI.

ef4: I see now, the way you have it is good as long as we randomize the generation.

PR #172, opened 2018-04-07
use field type for indexed mappings based on ABI

### LINKED ISSUE #162: Issue indexing content types that have differing fields from ethereum event triggers
some of the content types that we use for ethereum events don't have fields that exactly match the fields that are fired for an Ethereum event. https://github.com/cardstack/cardstack-web/issues/249

In this case, we should just lookup call the method of the contract for the content type instead of using the ethereum event to populate the document for the content type.

habdelra:
Fixes #162
When indexing a mapping entry for a contract, use the field type for the mapping based on the ABI for the contract method instead of hard coded number field type.


habdelra:
This PR is subsumed by https://github.com/cardstack/cardstack/pull/173

PR #174, opened 2018-04-08
Fixes #86 check auth on app boot

### LINKED ISSUE #86: need to explicitly check session tokens on client app start
I'm running into a scenario where it looks like the client is not explicitly checking tokens with the server when the application starts. This is visible when you start the hub without setting `CARDSTACK_SESSIONS_KEY` env var. then login (use the mock-auth plugin's dummy app), then restart the server (this invalidates all the tokens), then reload the client. In this scenario the cardstack-session service's `isAuthenticated` property is `true`, but the token presented to the server does not resolve to a valid session.

habdelra:
Fixes #86 

I performed this in 2 commits. the first commit added additional features to the mock-auth addon's testing dummy app so that it was easy to see this problem. The second commit fixes the problem.

### DIFF
--- a/packages/authentication/addon/cardstack-authenticator.js
+++ b/packages/authentication/addon/cardstack-authenticator.js
@@ -26,11 +26,11 @@ export default Base.extend({
         localStorage.removeItem('cardstack-secret-token'),
         localStorage.removeItem('cardstack-authentication-source');
         this.authenticate( authenticationSource, { secret }).then(resolve, reject);
-      } else if (potentiallyValidSession) {
+      } else if (potentiallyValidSession && authenticationSource) {
         // dont assume the session you have is valid just because session token hasn't yet expired
-        // attempt to fetch the user model that you posses to confirm the session is valid
-        let { type, id, meta: { token } } = rawSession.data;
-        fetch(`${hubURL}/api/${type}/${id}`, {
+        localStorage.removeItem('cardstack-authentication-source');
+        let { meta: { token } } = rawSession.data;
+        fetch(`${hubURL}/auth/${authenticationSource}/status`, {
           method: 'GET',
           headers: { 'authorization': `Bearer ${token}` }
         }).then(response => {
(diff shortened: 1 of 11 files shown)


packages/authentication/addon/cardstack-authenticator.js
```diff
@@ -22,11 +22,25 @@ export default Base.extend({
...
+      if (!potentiallyValidSession && secret && authenticationSource ) {
         localStorage.removeItem('cardstack-secret-token'),
         localStorage.removeItem('cardstack-authentication-source');
         this.authenticate( authenticationSource, { secret }).then(resolve, reject);
-      } else if ( validSession || partialSession ) {
+      } else if (potentiallyValidSession) {
+        // dont assume the session you have is valid just because session token hasn't yet expired
+        // attempt to fetch the user model that you posses to confirm the session is valid
+        let { type, id, meta: { token } } = rawSession.data;
+        fetch(`${hubURL}/api/${type}/${id}`, {
```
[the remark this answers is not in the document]
habdelra: Makes sense, :+1:

packages/authentication/cardstack/middleware.js
```diff
@@ -109,6 +111,40 @@ class Authentication {
...
+              }]
+            };
+            return;
+          }
+          let user = await this.userSearcher.get(session.type, session.id);
+
+          if (!user) { throw new Error(`cant find user type ${session.type} id ${session.id}`); }
+
+          ctxt.status = 200;
+          ctxt.body = user;
```
ef4: I think this leaks any private fields on the user model (one that the user does not have a grant to see). See `applyReadAuthorization` in `_invokeAuthenticationSource` -- I think we need the same thing here. (And hopefully a test to make sure it remains true).

habdelra: whoops! thanks

ef4:
🙌🏻

PR #178, opened 2018-04-11
cleanup abandoned replication slots

habdelra:


packages/postgresql/indexer.js
```diff
@@ -17,6 +17,8 @@ const changePattern = /^table ([^.]+)\.([^:]+): ([^:]+): (.*)/;
 // If your id contains spaces or quotes yer gonna have a bad time. Sorry not sorry.
 const idPattern = /id\[[^\]]+\]:'?([^\s']+)/;
 
+const replicationSlotPrefix = process.env.ELASTICSEARCH_PREFIX || 'cardstack';
```
habdelra: @ef4 i felt conflicted about this env var. I wasn't sure if I should just rename it to be more agnostic of elastic search

ef4: Yeah, the same goes for ELASTICSEARCH_PREFIX itself. I think this is OK.

PR #192, opened 2018-04-26
handled ember-simple-auth deprecation

habdelra:
updated to no longer use simple auth authorizer, per deprecation warning: https://github.com/simplabs/ember-simple-auth#deprecation-of-authorizers


ef4:
Nice. I was about to tackle this, glad you got to it first.

PR #197, opened 2018-04-29
More efficient ethereum indexing

habdelra:
This builds on top of https://github.com/cardstack/cardstack/pull/194. In this PR I'm associating an ethereum data source to a single smart contract (as opposed to many contracts). I've also added logic to keep track of the last indexed block number as the Updater's `meta` property so that when indexing without a hint, we'll only look for events that have occurred after the last indexed block number.


packages/ethereum/cardstack/service.js
```diff
@@ -24,13 +25,13 @@ class EthereumService {
...
     this._hasConnected = false;
-    this._isIndexing = false;
-    this._indexQueue = [];
+    this._isIndexing = {};
+    this._indexQueue = {};
     this._processQueueTimeoutMs = 1000;
-    this._processQueueTimeout = null;
-    this._contractDefinitions = null;
+    this._processQueueTimeout = new WeakMap();
+    this._contractDefinitions = new WeakMap();
```
ef4: I think the keys you're using in both these WeakMaps are strings. If that's so, they're probably not doing what you want them to -- they will act like plain old Maps and the values will be retained forever.

packages/ethereum/cardstack/indexer.js
```diff
@@ -142,17 +140,15 @@ class Updater {
       for (let model of schema) {
         await ops.save(model.type, model.id, model);
       }
-    } else {
-      await ops.beginReplaceAll();
+      await ops.finishReplaceAll();
```
ef4: If you don't have valid meta, or if the schema has changed, you are reindexing everything -- schema and content both.

So `{begin|finish}ReplaceAll` needs to wrap the whole indexing operation, not just the schema part. Otherwise you can end up with stale ethereum records still in the search index.

habdelra: ah, ok. since ethereum contracts never change schema (because the blockchain is immuatable like that), i wonder if it even makes sense to have a `{begin|finish}ReplaceAll`? 

It probably makes sense to reindex everything if there is no meta/or schema, as there should always be both after the first indexing. Not having meta or schema probably means the index has been wiped clean of our ethereum data source.

packages/ethereum/cardstack/service.js
habdelra: whoops--yes. will fix.

packages/ethereum/cardstack/indexer.js
ef4: I think we still need replaceAll. In general there are a bunch of things that might cause us to want to redo indexing work (like adding a computed field to an ethereum-backed record). The only way to convey that to the plugin is to get rid of its meta and force it to start again. Under those conditions, it should get rid of any of its older stuff in the search index.

habdelra: cool, I fixed the placement of the begin/finish replace all in the latest commit.

PR #198, opened 2018-04-30
add login lifecycle actions to support doing things after successful login

habdelra:
this can support our needs around redirecting to different routes in the consuming app after authentication and authorization of the cardstack session.


habdelra:
Apologies for the merge, I'm gonna fix the dashboard redirect after the email verification state using this. I'll definitely fix any issues that arise from a post-merge review.

PR #200, opened 2018-05-01
fixes issue around comparing changed fields deeply when calculating auth restrictions

habdelra:


packages/hub/schema/field.js
```diff
@@ -195,7 +196,7 @@ module.exports = class Field {
 
     if (pendingChange.serverProvidedValues.has(this.id) && pendingChange.serverProvidedValues.get(this.id) === value) {
```
ef4: I think we need to replace this `===` with `isEqual` too.

habdelra: sure

ef4: (The server-provided value comes from a default. Either default-at-create or default-at-update.)

PR #203, opened 2018-05-02
refactored ethereum indexing to buffer through elastic search for large indexing jobs

habdelra:
Because of this approach, we were able to remove a lot of the quirky async from the ethereum service, and all the "special" async needs for the ethereum event handling can be dealt with in the `buffer.loadModels()` that we we reviewing today. The result is that there is one overall promise that is available to the tests that we can `await` for all things ethereum related. Also, I removed all the logic around using hints to decide if ethereum models have been buffered or not, and we rather we use buffer's ES index and the last indexed blockheight vs the buffered model's blockheight to determine if we need to reload the model from ethereum.

I've also tested it against our contract in rinkeby, and the hub starts up and is listening on port 3000 almost instantly 👌 

Also, this introduces `blockheight` as `{ meta }` for ethereum models, which is think is a pretty good  idea, as it give more context in terms of when the data was obtained.

### DIFF
--- a/packages/ethereum/cardstack/abi-utils.js
+++ b/packages/ethereum/cardstack/abi-utils.js
@@ -1,0 +1,56 @@
+const Ember = require('ember-source/dist/ember.debug');
+const { dasherize } = Ember.String;
+
+function fieldTypeFor(contractName, abiItem) {
+  if (!abiItem.outputs || !abiItem.outputs.length) { return; }
+
+  if (!abiItem.inputs.length) {
+    // We are not handling multiple return types for non-mapping functions
+    // unclear what that would actually look like in the schema...
+    switch(abiItem.outputs[0].type) {
+      // Using strings to represent uint256, as the max int
+      // int in js is 2^53, vs 2^256 in solidity
+      case 'uint256':
+      case 'bytes32':
+      case 'string':
+      case 'address':
+        return { fields: [{ type: '@cardstack/core-types::string' }]};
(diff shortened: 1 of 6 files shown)


habdelra:
Apologies for the self-merge of the PR, I want to close out this issue and move on to other things. This includes all the things we talked about ed, so hopefully no surprises here. I'll be happy to respond to any post merge review comments.

ef4:
Does this mean that we can safely delete the normal elasticsearch index and it will reindex ethereum quickly because the special ethereum buffer index is still intact?

That would be very good.

habdelra:
The time it takes to receive all the ethereum events depends on external factors. But from the main content index’s perspective it does all happen at once. In our slack I shared a graph of ES activity for our app in the worst case scenario. The first part of the graph you can see little steps as the buffer’s internal index accumulates events from the contract. Then as soon as the buffer has reached the current block height it flushes all the accumulated events at once  into the main content index, and that’s the vertical line that you see. https://cardstacksyndicate.slack.com/files/U0HBY7LBV/FAK7LGJDD/screen_shot_2018-05-03_at_10.21.16_pm.png

habdelra:
Oh I just realized I misunderstood your question. I was thinking for some reason that all indices where dropped. If you dropped just the content index, then yes, the indexing is immediate. We retain he buffer’s private index. And we should be retaining it thru app restarts, as when the buffer starts up it searches for he last index it was using.

PR #210, opened 2018-05-07
Use SES_REGION env var to configure region for email messenger when using SES

alexspeller:
Also add some debug logging


habdelra:
thanks, I'll add the app deploy starts of this

PR #216, opened 2018-05-10
Add the concept of a session meta, and store the request IP in it in the

alexspeller:
authentication middleware.

A plugin config can be added to allow respecting the X-Forwarded-For header if
the app is hosted behind a reverse proxy.

### DIFF
--- a/packages/authentication/cardstack/middleware.js
+++ b/packages/authentication/cardstack/middleware.js
@@ -54,7 +54,7 @@ class Authentication {
     };
   }
 
-  _tokenToSession(token) {
+  async _tokenToSession(token, ctxt) {
     let ciSessionId = this.ciSession && this.ciSession.id;
     if (ciSessionId && token === ciSessionId) {
       return Session.INTERNAL_PRIVILEGED;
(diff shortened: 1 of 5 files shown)


alexspeller:
@ef4 I think you'll want to look at this as it makes some architectural choices about sessions and plugin configs

packages/authentication/cardstack/middleware.js
```diff
@@ -163,9 +164,28 @@ class Authentication {
     throw new Error(`No such authentication source "${name}"`, { status: 404 });
   }
 
+  async _sessionMeta(ctxt) {
+    let config, ip;
+
+    try {
+       config = await this.searcher.getFromControllingBranch(Session.INTERNAL_PRIVILEGED, 'plugin-configs', '@cardstack/authentication');
```
ef4: Let's not add a round trip to elasticsearch to every request. We can cache this in the middleware. I'm pretty sure we will already rebuild the middleware stack if a plugin-config is changed.

packages/authentication/cardstack/middleware.js
```diff
@@ -163,9 +164,28 @@ class Authentication {
...
+    } catch (e) {
+      // no config for the authentication plugin, assume header should not be respected
+    }
+
+    if (config && config.data.attributes['plugin-config'] && config.data.attributes['plugin-config']['allow-x-forwarded-for']) {
```
ef4: Instead of the field name `plugin-config`, we can use the existing `params` field which serves exactly the same purpose in multiple other places (a bag of plugin-specific configuration that is untyped from the perspective of the JSON:API schema).

ef4:
I started down the path of redoing this, but realized it touches several things that I don't want to touch right this moment, so I'm merging as is, and we can follow up on my configure-koa-trust-proxy branch later.

PR #222, opened 2018-05-24
[wip] Initial queue plugin implementation

alexspeller:
There is one unresolved question here, which is how to configure pg-boss.

I was planning to do it using env variables; so first off, check if `PG_BOSS_DB_URL` is present, then check for `PG_BOSS_DB_HOST`, `PG_BOSS_DB_NAME` etc, then just default to nothing which will try to connect to localhost as is the default behaviour for the pg module.

However, if I do that, I'm not quite sure how I'd override that for testing - I guess I could just do `process.env.PG_BOSS_DB_URL = whatever` in the test setup but that seems a bit unidiomatic?

### DIFF
--- a/packages/hub/plugin-loader.js
+++ b/packages/hub/plugin-loader.js
@@ -24,8 +24,7 @@ const featureTypes = [
   'authenticators',
   'middleware',
   'messengers',
-  'code-generators',
-  'queues'
+  'code-generators'
 ];
 const javascriptPattern = /(.*)\.js$/;
 
--- a/packages/hub/plugin-loader.js
+++ b/packages/hub/plugin-loader.js
@@ -24,7 +24,8 @@ const featureTypes = [
   'authenticators',
   'middleware',
   'messengers',
-  'code-generators'
+  'code-generators',
+  'queues'
 ];
 const javascriptPattern = /(.*)\.js$/;
 
--- a/packages/queue/index.js
+++ b/packages/queue/index.js
@@ -1,0 +1,103 @@
+const PgBoss = require('pg-boss');
+const log = require('@cardstack/logger')('cardstack/queue');
+
+module.exports = class Queue {
+  constructor() {
+    this.boss = null;
+    this.promiseResolveCallbacks = {};
+    this.promiseRejectCallbacks = {};
+    this.callbackedQueues = [];
+    this.jobErrors = {};
+  }
+
+  async _ensureBoss() {
+    if (!this.boss) {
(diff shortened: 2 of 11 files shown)


packages/hub/plugin-loader.js
```diff
@@ -24,7 +24,8 @@ const featureTypes = [
   'authenticators',
   'middleware',
   'messengers',
-  'code-generators'
+  'code-generators',
+  'queues'
```
ef4: I think we're not using this? It would be needed if we were using the plugin-loader to discover queues from arbitrary plugins, but I think the hub is just hard coded to use the one implementation we have, and I think that's good.

packages/queue/node-tests/queue-test.js
```diff
@@ -0,0 +1,202 @@
...
+    env = await createDefaultEnvironment(`${__dirname}/..`);
+
+    queue = env.lookup('hub:queues');
+
+    await env.lookup('hub:indexers').update({ forceRefresh: true });
+  }
+
+  async function teardown() {
+    await destroyDefaultEnvironment(env);
+    await queue.stop();
```
ef4: This should probably happen automatically when the container is torn down. [See here](https://github.com/cardstack/cardstack/blob/aa4454368343218ff19029f5fb1c3f229b62198c/packages/hub/indexers.js#L89) for an example of using the teardown hook.

That will mean that `destroyDefaultEnvironment` automatically stops the queue.

packages/queue/index.js
```diff
@@ -0,0 +1,103 @@
...
+    this.boss = null;
+    this.promiseResolveCallbacks = {};
+    this.promiseRejectCallbacks = {};
+    this.callbackedQueues = [];
+    this.jobErrors = {};
+  }
+
+  async _ensureBoss() {
+    if (!this.boss) {
+      this.boss = new PgBoss({ database: 'postgres', host: 'localhost', user: 'postgres', port: 5444 });
```
ef4: My thoughts on your config question: 

 - this Queue class can take the whole `pgBossConfig` as a constructor argument
 - hub/queues.js can inject `config:pg-boss` and pass that into `new Queue`.
 - in hub/main.js `wireItUp`, would `registry.register('config:pg-boss', opts.pgBossConfig || { database: process.env.PG_BOSS_DB_NAME, ...etc })`

This allows the normal usage to be environment variables, while also allowing someone who is calling `createDefaultEnvironment` to pass the `pgBossConfig` option if they need to override.

ef4:
Thanks, looking good.

PR #232, opened 2018-06-19
Document builder

habdelra:
This is a checkpoint to show my train of thought on the consolidated response builder. My goal is to peel off a `document-builder.js` utility from the `document-context.js` that can be used outside of an indexing/ES context. I made a very small beachhead which includes returning pure JSON API from the new `document-builder`, and accumulating the ES specific stuff in a property that I plan on eventually removing, while maintaining that all the tests are passing.

 My next focus will be to cleanup the property that holds the accumulated ES fields, or figure out a more graceful way of dealing with this. but generally, I'm just whittling away at document-builder until I have something that I can use in the writers as well as jsonapi middleware, or really anywhere that needs to build JSON API documents from models.


packages/hub/indexing/document-builder.js
```diff
@@ -0,0 +1,206 @@
...
+const Model = require('../model');
+const { uniqBy } = require('lodash');
+
+module.exports = class DocumentBuilder {
+  constructor({ schema, type, id, doc, fieldMapping, getResource }) {
+    this.schema = schema;
+    this.type = type;
+    this.id = id;
+    this.doc = doc;
+    this._fieldMapping = fieldMapping ? fieldMapping : fieldName => fieldName; //TODO need a less ES specific way of approaching this
```
ef4: I don't think we need to keep fieldMapping. If I'm remembering right, it's only used to deal with ES schema quirks.

habdelra:
I'm thinking maybe callers to the response builder can pass in a `relationshipObserver` function, that is invoked each time we visit a relationship, which can then deal with the ES specific stuff outside of the document builder....

packages/hub/indexing/document-builder.js
habdelra: yup, my plan is to factor this out of here.

ef4:
What about relationships is ES-specific? Mostly I think it's general (updating references and embedding in-place for deep searchability are both general-purpose, not ES-specific).

habdelra:
it's the adding of the ES specific mappings that hold a related resource. Specifically this https://github.com/cardstack/cardstack/blob/master/packages/hub/indexing/document-context.js#L122-L123 (but also this for attributes too: https://github.com/cardstack/cardstack/blob/master/packages/hub/indexing/document-context.js#L76-L77)

habdelra:
these ES specific embellishments are not JSON API, but we derive them as part of constructing the relationships (and attributes).

habdelra:
maybe a better idea would be to just have some JSONAPI introspection logic to construct these ES embellishments in the document-context.js as opposed to a callback that is invoked for each relationship visited in document-builder.js

ef4:
I see what you mean, but `logicalFieldToES` is already _nearly_ a no-op, and in postgres it will definitely be a no-op. If you just take that function out, the embedding that's happening in the searchDoc is still needed and not ES-specific.

```diff
-let esName = await this._logicalFieldToES(field.id);
-searchDocOut[esName] = related;
+searchDocOut[field.id] = related; // this is still important
```

habdelra:
But is it the case that we want it only in specific circumstances? Like we only want to write those fields in the case where we are creating a doc to put in the index? If we want to use this same module to assert all the includes are present in the json api middle ware, as you were pointing out to me previously, then we would not care for those fields, right? When I say that I want to remove that logic—I’m not being super precise: I really want to move it. Specifically  to a place that make sense for the indexing context, so that I can use this module in scenarios where we would not want to include those extra fields, like the json api middleware that you were pointing out.

Is my reasoning not right? Would you expect that we would not use this module outside of scenarios where we are creating documents to put in the index?

ef4:
I had been assuming that this module would be capable of building both representations (searchDoc and pristineDoc). Mostly because a lot of the work and data involved is shared.

```js
let builder = new DocumentBuilder(type, id, doc, ...etc);
await builder.pristineDoc(); // generates only pristine, but leave a lot of internal cached state in builder
await builder.searchDoc(); // generate only search, but is able to do a lot less work by sharing what was already cached by pristineDoc.
```

I also don't think the DocumentBuilder and the DocumentContent will remain separate types once we're through this refactor. The only reason we need classes here at all (as opposed to pure functions) is that there's a bundle of internal state that we want to keep around, and it's one bundle not two, so I see the need for a single class.

As an aside: In our refactoring work yesterday, @savvymas and I discovered that in pgsearch we will want a _third_ document representation `upstreamDoc`, which is the original thing you got from the data source (it's not needed in today's architecture because we never feed content from ES back into the indexing process, but soon we will).

habdelra:
@ef4 do you wanna take another look at this PR and me know if this looks more inline with your vision?

habdelra:
oh I'll add the logic around not recomputing pristineDoc or searchDoc if you already accessed one of these from teh build like you were mentioning earlier.

packages/hub/indexing/document-builder.js
```diff
@@ -192,3 +220,17 @@ function ensure(obj, section) {
...
+
+  let { data } = record;
+  let { id, type } = data;
+  let flattenedRecord = { id, type };
+
+  for (let attr of Object.keys(data.attributes || {})) {
+    flattenedRecord[attr] = data.attributes[attr];
+  }
+  return flattenedRecord;
+}
```
ef4: I don't think this is right -- we shouldn't be guessing which things to flatten by looking at the value. We know precisely which fields should be flattened (because they are relationships) based on the schema.

It also seems weird to me that it only handles attributes. Flattening needs to happen recursively across all levels of relationships too, but if that's implemented elsewhere, I don't see why attribute flattening would be happening separately here.

habdelra: we know which fields are relationships, but some relationships are strings, some are arrays strings, some are dates, some are arrays of objects. this particular function is used to flatten the fields that ultimately become the search doc embellishments. in that scenario, there is no need to flatten a relationship of a relationship--just the attributes will do. In fact I had inadvertently done that in a initial pass thru this and resulted in failing tests, and my query to you around if we should be able to return results for queries that where indirect relationships match query params, and you said not to do that.

habdelra: right now, the search doc embellishments are not json api and we only capture the top level relationships. this is not to be confused with the json api relationships that are used to feed into the includes. that a totally separate thing (which wis recursive and handled correctly as evidenced by the passing tests).

habdelra: and just as nomenclaure, when I say "searchdoc embellishments" i mean all the fields that appear outside of the pristine part of the search doc.

ef4: All the things that appear inside `included` in the pristine doc also need to appear _inline_ in the search doc, down arbitrarily deep. Top level relationships are not sufficient.

Given pristine:
```
{
  data: {
    type: 'posts',
    id: '1',
    attributes: { title: 'hello world' },
    relationships: {
      author: { 
        data: { type: 'people', id: '1' }
      }
    }
  },
  included: [
    { 
      type: 'people',
      id: '1',
      attributes: { name: 'Hassan' },
      relationships: {
        company: {
          data: { type: 'companies', id: '2' }
        }
      }
   },
   {
      type: 'companies',
      id: '2',
      attributes: { name: 'Cardstack' }
   }
  ]
}
```

We need searchDoc:

```
{
  type: 'posts',
  id: '1',
  title: 'hello world' ,
  author: {
     type: 'people',
     id: '1',
     name: 'Hassan',
     company: {
       type: 'companies',
       id: '2',
       name: 'Cardstack'
     }
  }
}
```

So that a natural search of `author.company.name=Cardstack` will find this document.

habdelra: Ok, we’ll probably have to address some breaking tests for this too, then.

habdelra: This is definitely clear and actionable feedback that I can make headway on.

ef4: I would have hoped we already have test coverage for deeply nested queries like this, but maybe not.

ef4: BTW, the searchDoc flattening I showed above is necessarily safe from namespace collisions because JSON:API requires all attributes, relationships, and "type" and "id" to live in a single namespace. That spec is really smart.

habdelra: I’ll make sure to test for that too

habdelra:
This PR is replaced by this PR https://github.com/cardstack/cardstack/pull/236

PR #233, opened 2018-06-20
Use queueing system for indexing

alexspeller:
Use queueing system for indexing to allow indexing to ensure only one indexing job runs at a time, regardless of how many hub instances are running.

Opens up the possibility of running worker-only hub processes not serving web requests in the future


ef4:
Looking nice.

Issue #237, opened 2018-06-27
pgsearch refactor tracking issue


ef4:
Opening this as a place to collect notes on where we are and where we're going with the pgsearch refactor. The major goals here are

 -  allow us to consistently offer features like server-side computed properties (which can in turn power arbitrarily complex permissions, relationships, etc). By "consistently" I mean you can rely on them being present (1) in the search index, (2) in responses that bypass the search index via custom searchers, (3) in responses echoed back from writers.

 - avoid needing to always go to upstream data sources when recomputing inter-record dependencies (because they can be arbitrarily slow and some may be fragile and easy to overload).

 - make all writes (via hub:writers) also do the correct targeted update in the search index, without relying on triggering a general hub:indexers update (no more "update with hints"). This lets us maintain a clear separation between full updates (which are effectively background processing that is never in a request critical path) and ongoing concurrent writes (which are never blocked by background indexing and yet are always immediately visible in the search index).

In these notes it's useful to keep in mind three different representations for any resource:

 - the upstream format is a single json:api resource as given to us by an indexer or searcher plugin. It never has computed fields, and it can't have includes because it's only a resource, not a complete document.
 - the pristine format is a complete json:api document that is augmented with computed fields and includes.
 - the search format is a flattened representation that nests are included resources inline, flattens away the attributes and relationships POJOs, and allows field-type plugins to customize the storage format of their values for optimal searchability.

The steps here are:

 - [x] replace our built-in elasticsearch-backed index with a postgres-backed index. We need to consistently read our own writes, we don't actually depend internally on any of the things elasticsearch excels at (like deep text searches). This work is already far along on the pgsearch branch.

 - [x] refactor DocumentContext so it can run not just at indexing time, but also when assembling a response from a writer or searcher. Optimize its internals so that it can build both or either of the searchDoc and pristineDoc while caching the common work between them and not preemptively building parts it wasn't asked for.

 - [x] change hub:writers so that it uses DocumentContext to send a complete pristine document format in the response to the client, and also so that it stores the search document format directly in the search index (this allows plugins like jsonapi to drop their explicit calls to `indexers.update()`).

[...]

PR #238, opened 2018-06-28
Pgsearch refactor

ef4:
This swaps elasticsearch for postgres as our internal search index.

There is more cleanup and optimization work to do, but the changes are ready to land on master so that others can continue building off of them.


ef4:
related #237

PR #239, opened 2018-06-28
removed elastic search field-type plugins

savvymas:
This removes leftover field-type plugins from elastic search, and _logicalFieldToES from document context.

### DIFF
--- a/packages/hub/indexing/document-context.js
+++ b/packages/hub/indexing/document-context.js
@@ -70,9 +70,6 @@ module.exports = class DocumentContext {
     return this._read(type, id);
   }
 
-  async _logicalFieldToES(fieldName) {
-    return fieldName;
-  }
 
   // TODO come up with a better way to cache (use Model)
   async _getCachedSearchDoc() {
@@ -105,7 +102,7 @@ module.exports = class DocumentContext {
 
   async _buildAttribute(field, value, pristineDocOut, searchDocOut) {
     // Write our value into the search doc
-    let esName = await this._logicalFieldToES(field.id);
+    let esName = field.id;
     searchDocOut[esName] = value;
 
     // Write our value into the pristine doc
@@ -116,7 +113,7 @@ module.exports = class DocumentContext {
     let derivedFields = field.derivedFields(value);
     if (derivedFields) {
       for (let [derivedName, derivedValue] of Object.entries(derivedFields)) {
-        let esName = await this._logicalFieldToES(derivedName);
+        let esName = derivedName;
         searchDocOut[esName] = derivedValue;
       }
     }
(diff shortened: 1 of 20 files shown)


packages/hub/schema/field.js
```diff
@@ -216,15 +216,10 @@ module.exports = class Field {
...
-    if (typeof this.plugin.customAnalyzer === 'function') {
-      return Object.assign({}, this.plugin.customAnalyzer());
-    }
-  }
 
   mapping(searchTree, allFields) {
     let m = {
-      [this.id]: Object.assign({}, this.plugin.defaultMapping(allFields))
+      [this.id]: Object.assign({}, allFields)
     };
```
ef4: This whole `mapping` function can get removed. And look for whatever is calling it (probably in content-type.js) and prune that back too.

packages/mobiledoc/cardstack/field-type.js
```diff
@@ -6,31 +6,19 @@ module.exports = {
...
-    return `${fieldName}_as_text`;
-  },
 
   // derive a text representation to store in elasticsearch
   derivedFields(fieldName, value) {
     return {
-      [this.queryFieldName(fieldName)]: value ? renderer.render(value).result : null
+      [(fieldName)]: value ? renderer.render(value).result : null
     };
   },
```
ef4: derivedField and derivedMappings can also go

packages/hub/indexing/document-context.js
```diff
@@ -105,7 +102,7 @@ module.exports = class DocumentContext {
 
   async _buildAttribute(field, value, pristineDocOut, searchDocOut) {
     // Write our value into the search doc
-    let esName = await this._logicalFieldToES(field.id);
+    let esName = field.id;
```
ef4: Now `esName` as a variable isn't really needed either. You can replace it everywhere with `field.id`.

packages/hub/indexing/document-context.js
```diff
@@ -179,7 +176,7 @@ module.exports = class DocumentContext {
     //
     // we don't store the type as a regular field in elasticsearch,
     // because we're keeping it in the built in _type field.
-    let esId = await this._logicalFieldToES('id');
+    let esId = 'id';
```
ef4: Same here: "es" is going to just be confusing now.

packages/hub/indexing/document-context.js
```diff
@@ -193,7 +190,7 @@ module.exports = class DocumentContext {
     // we are going inside a parent document's includes, so we need
     // our own type here.
     if (depth > 0) {
-      let esType = await this._logicalFieldToES('type');
+      let esType = 'type';
```
ef4: And same here.

PR #241, opened 2018-06-28
converts mobiledoc to text and corresponding test coverage

savvymas:
This converts mobiledoc documents from json to text so that it can be searched properly.


ef4:
For those playing along at home, this re-implements a general field-type plugin feature we dropped because the API was too tied to elasticsearch. `searchIndexFormat` is the new hook for deciding how your field type looks in the search index.

PR #243, opened 2018-06-28
Hub writers document context

habdelra:
Addresses items discussed in https://github.com/cardstack/cardstack/issues/237

### DIFF
--- a/packages/hub/writers.js
+++ b/packages/hub/writers.js
@@ -2,11 +2,13 @@ const Error = require('@cardstack/plugin-utils/error');
 const log = require('@cardstack/logger')('cardstack/writers');
 const DocumentContext = require('./indexing/document-context');
 const { declareInjections } = require('@cardstack/di');
+const Session = require('@cardstack/plugin-utils/session');
 
 module.exports = declareInjections({
   schema: 'hub:current-schema',
   schemaLoader: 'hub:schema-loader',
   searchers: 'hub:searchers',
+  pgSearchClient: `plugin-client:${require.resolve('@cardstack/pgsearch/client')}`
 },
 
 class Writers {
@@ -33,7 +35,8 @@ class Writers {
         this.schema.invalidateCache();
       }
 
-      // TODO index the search doc
+      await this.pgSearchClient.saveDocument({ context });
+
       return context.pristineDoc();
     } finally {
       if (pending) { await pending.abort();  }
@@ -41,6 +44,7 @@ class Writers {
   }
 
   async update(branch, session, type, id, document) {
+    debugger;
     log.info("updating type=%s id=%s", type, id);
     let schema = await this.schema.forBranch(branch);
     let writer = this._lookupWriter(schema, type);
@@ -59,6 +63,9 @@ class Writers {
       if (newSchema) {
         this.schema.invalidateCache();
       }
+
+      await this.pgSearchClient.saveDocument({ context });
+
       return context.pristineDoc();
     } finally {
       if (pending) { await pending.abort();  }
@@ -74,9 +81,12 @@ class Writers {
     try {
       let newSchema = await schema.validate(pending, { session });
       await pending.finalize();
+
       if (newSchema) {
         this.schema.invalidateCache();
       }
+
+      await this.pgSearchClient.deleteDocument({ branch, type, id });
     } finally {
       if (pending) { await pending.abort();  }
     }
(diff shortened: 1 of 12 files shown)


packages/hub/writers.js
```diff
@@ -26,11 +28,13 @@ class Writers {
...
     try {
       let newSchema = await schema.validate(pending, { type, session });
-      let response = await this._finalizeAndReply(pending);
+      let context = await this._finalizeAndReply(pending, branch, session, type, schema);
       if (newSchema) {
         this.schema.invalidateCache();
       }
-      return response;
+
+      // TODO index the search doc
```
habdelra: @ef4 this is the place where I'm not sure what the correct approach should be to update the index. should I use the `hub:indexers.update()`?

packages/hub/writers.js
```diff
@@ -78,21 +82,21 @@ class Writers {
...
-    if (finalDocument.attributes) {
-      responseDocument.attributes = finalDocument.attributes;
-    }
-    if (finalDocument.relationships) {
-      responseDocument.relationships = finalDocument.relationships;
-    }
-    return responseDocument;
+      upstreamDoc: finalDocument,
+      sourceId: contentType.dataSource.id,
+      read: (type, id) => this.searchers.get(session, branch, type, id)
```
habdelra: @ef4 this is how i'm looking up relationships, but this relies on the fact that the writers update the index as evidenced by the `pgseach/indexer: indexes a belongs-to` test. As this create https://github.com/cardstack/cardstack/blob/master/packages/pgsearch/node-tests/indexer-test.js#L107-L112 would need to be indexed before this create https://github.com/cardstack/cardstack/blob/master/packages/pgsearch/node-tests/indexer-test.js#L114-L122 in order for the searcher to find the related record

packages/hub/writers.js
ef4: You should [inject the pgsearch client](https://github.com/cardstack/cardstack/blob/cc237da819a33c9c50eb5f8d059c6adafaa4f8f3/packages/hub/indexers.js#L50) and use its [`saveDocument`](https://github.com/cardstack/cardstack/blob/cc237da819a33c9c50eb5f8d059c6adafaa4f8f3/packages/pgsearch/client.js#L120)  and `deleteDocument` methods.

When you do that I think you will find that it's worth refactoring `saveDocument` so that it accepts a `DocumentContext` instance as its argument, because most of what it takes now is things that are available there, and it already takes too many arguments.

When you save from `hub:writers`, there is no `generation` to worry about, you can leave that empty.

habdelra: sweet thanks!

packages/hub/writers.js
ef4: Using a searcher here is correct, although we're going to need to think about permissions. When you save the searchDoc into the search index, it should be complete, so it shouldn't depend on the user's current permissions when following relationships. But when we reply to the client with the pristine doc, it should obviously only include things they're allowed to see.

habdelra: cool, make sense. I'll use the `Session.INTERNAL_PRIVILEGED`

habdelra: or wait, i just parsed your last sentence.... yes, i'll figure out a badd way to refactor this.

habdelra: i think when I get to the jsonapi middleware, i'll revisit this again, but with pristineDocs as input instead of the upstreamDoc, and i think that is probably where we need to use the user's session for the reads

habdelra: or maybe there is a `userPristineDoc()` function on document-context which uses user's session? I'm seeing that the jsonapi middleware just returns the response from teh writers directly.

ef4: Something like [applyReadAuthorization](https://github.com/cardstack/cardstack/blob/cc237da819a33c9c50eb5f8d059c6adafaa4f8f3/packages/hub/searchers.js#L57) may be the right thing to use for the writer reply.

packages/hub/writers.js
```diff
@@ -51,48 +62,67 @@ class Writers {
...
-    if (finalDocument.relationships) {
-      responseDocument.relationships = finalDocument.relationships;
-    }
-    return responseDocument;
+      upstreamDoc: finalDocument,
+      sourceId: contentType.dataSource.id,
+      read: async (type, id) => {
+        let result;
+        try {
+          result = await this.searchers.get(Session.INTERNAL_PRIVILEGED, branch, type, id);
```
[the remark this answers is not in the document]
habdelra: sorry, forgot about dealing with that.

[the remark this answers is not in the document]
ef4: No worries, that is what code review is for!

ef4:
Nice. I'm impressed the auth stuff all worked out without a lot of new code.

PR #244, opened 2018-07-10
Invalidation refactor

habdelra:
moving invalidation into the pgsearch client and invoking from both branch-update and the writers.

@ef4 is this what you had in mind?

### DIFF
--- a/packages/hub/indexing/branch-update.js
+++ b/packages/hub/indexing/branch-update.js
@@ -3,7 +3,6 @@ const owningDataSource = new WeakMap();
 const { flatten } = require('lodash');
 const log = require('@cardstack/logger')('cardstack/indexers');
 const DocumentContext = require('./document-context');
-const Session = require('@cardstack/plugin-utils/session');
 
 const FINALIZED = {};
 
@@ -105,12 +104,6 @@ class BranchUpdate {
       schema: await this.schema(),
       branch: this.branch,
       read: (type, id) => this.read(type, id),
-      wasInvalidated: async ({ operation, type, id, doc, sourceId, nonce }) => {
-        this.emitEvent(operation, { type, id, doc });
-        if (operation === 'add' && this.isControllingBranch) {
-          await this._maybeUpdateRealms(type, id, doc, sourceId, nonce);
-        }
-      }
     });
   }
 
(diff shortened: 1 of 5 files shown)


habdelra:
@ef4 one day bug I’m also unsure of is how to write a test to assert invalidation for the writers work correctly, I could use your guidance there too

ef4:
> I think the idea is that if I have say, an author document and a post document that has an author relation, if I use a writer to update the author document, I should see the post document that refers to it also reindexed?

Yes, you would have the author embedded as a default include in the post. Being a default include (and not an on-demand include based on a query parameter) is important, because that's what would test the contents of the search index.

Then edit the author, and then fetch the post, and check that the copy of author inside the post is updated.

packages/hub/indexing/branch-update.js
```diff
@@ -99,7 +99,19 @@ class BranchUpdate {
...
+    await this.client.invalidations({
+      touched: this._touched,
+      touchCounter: this._touchCounter,
+      schema: await this.schema(),
+      branch: this.branch,
+      read: (type, id) => this.read(type, id),
+      wasInvalidated: async ({ operation, type, id, doc, sourceId, nonce }) => {
+        this.emitEvent(operation, { type, id, doc });
+        if (operation === 'add' && this.isControllingBranch) {
+          await this._maybeUpdateRealms(type, id, doc, sourceId, nonce);
```
ef4: I think this whole `wasInvalidated` hook should move into the client and be standard for all invalidations.

For example, the realm update thing is needed if somebody is PATCHing a user.

packages/pgsearch/client.js
```diff
@@ -129,6 +204,14 @@ module.exports = class PgClient {
     let refs = await context.references();
     let realms = await context.realms();
 
+    if (touched) {
+      touched[`${type}/${id}`] = touchCounter++;
```
ef4: This looks like a bug. This isn't updating a shared counter, it's updating a local variable that never goes back out of this method.

More generally, this looks like an opportunity to make the touch-tracking and invalidation part of pgsearch's API. Like:

```
let batch = client.beginBatch();
// saveDocument and deleteDocument would only be available on `batch`, not directly on `client` anymore
await batch.saveDocument({ context });
await batch.deleteDocument(...whatever);

// this would automatically run invalidations. All the `touched` tracking happens inside the `batch`.
await batch.done();
```

This also has the benefit of giving us a logical place to add database transactions in the future if we want that (such that only an entire batch becomes visible at once, never the intermediate states during indexing).

packages/hub/indexing/branch-update.js
habdelra: so one aspect about moving this into the client that might be tricky is that we also use this to emit indexing events for branch-update, as the invalidation of resources currently triggers indexing events from callers of the pgsearch client which you can see a couple lines up on line 109. would that mean we still need to provide a callback for callers to use?

ef4: No, we can move the events to also come from pgclient.

packages/pgsearch/client.js
habdelra: nice. will do

ef4:
Looks good, thanks.

PR #245, opened 2018-07-12
Incorporate document-context into searchers

habdelra:
@ef4 I wanted to do a checkpoint with you to make sure I'm going down the right path here. I overhauled the `handleCollectionGET()` to use pristineDoc's from document-context that already have the included's.

Some questions:
- In the tracking issue, we talk about making sure that we are not doing wasted work when processing the doc in the document-context that we get from the searcher. Does that mean the building of the relationships? Like if the document-context sees that the relationship already exists in the doc it is processing--just use what is already there? (same for attrs)?
- how should we signal to the document-context which additional fields we'd like to have in `included`? I just flattened the the `this.includes` from the `Handler`, but that feels not right as I'm destroying the path information that was specified by the client in the `include` query param.
- I'm not using the `searchTree` in the document-context for the additional included fields. I feel wrong about that, but I don't really understand how to leverage the searchTree mechanism for additional included fields specified by the client. Guidance here would be helpful.

This addresses items in https://github.com/cardstack/cardstack/issues/237

### DIFF
--- a/packages/hub/indexers.js
+++ b/packages/hub/indexers.js
@@ -135,10 +135,10 @@ class Indexers extends EventEmitter {
           // Compare each branch, so we don't invalidate the schemas
           // unnecessarily
           for (let [branch, newSchema] of Object.entries(schemas)) {
-            let oldSchema = priorCache[branch];
+            let oldSchema = (await priorCache)[branch];
             if (!newSchema.equalTo(oldSchema)) {
               log.info('schema for branch %s was changed', branch);
-              priorCache[branch] = newSchema;
+              (await priorCache)[branch] = newSchema;
               if (oldSchema) {
                 await oldSchema.teardown();
               }
--- a/packages/hub/indexers.js
+++ b/packages/hub/indexers.js
@@ -72,7 +72,8 @@ class Indexers extends EventEmitter {
         }
       })();
     }
-    return (await this._schemaCache)[branch];
+    let schema = (await this._schemaCache)[branch];
+    return schema;
   }
 
   async invalidateSchemaCache() {
--- a/packages/hub/indexers.js
+++ b/packages/hub/indexers.js
@@ -72,8 +72,7 @@ class Indexers extends EventEmitter {
         }
       })();
     }
-    let schema = (await this._schemaCache)[branch];
-    return schema;
+    return (await this._schemaCache)[branch];
   }
 
   async invalidateSchemaCache() {
--- a/packages/hub/indexing/document-context.js
+++ b/packages/hub/indexing/document-context.js
@@ -1,11 +1,11 @@
 const authLog = require('@cardstack/logger')('cardstack/auth');
 const log = require('@cardstack/logger')('cardstack/indexers');
 const Model = require('../model');
-const { uniqBy } = require('lodash');
+const { get, uniqBy } = require('lodash');
 
 module.exports = class DocumentContext {
 
-  constructor({ read, schema, type, id, branch, sourceId, generation, upstreamDoc, additionalIncludes=[] }) {
+  constructor({ read, schema, type, id, branch, sourceId, generation, upstreamDoc, includePaths }) {
     this.schema = schema;
     this.type = type;
     this.id = id;
@@ -13,7 +13,7 @@ module.exports = class DocumentContext {
     this.sourceId = sourceId;
     this.generation = generation;
     this.upstreamDoc = upstreamDoc;
-    this.additionalIncludes = additionalIncludes ? additionalIncludes.map(part => part.split('.')) : [];
+    this.includePaths = includePaths ? includePaths.map(part => part ? part.split('.') : null).filter(i => Boolean(i)) : [];
     this._read = read;
     this.cache = {};
     this.isCollection = upstreamDoc && upstreamDoc.data && Array.isArray(upstreamDoc.data);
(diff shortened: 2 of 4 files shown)


packages/jsonapi/middleware.js
```diff
@@ -183,27 +183,17 @@ class Handler {
       sort: this.query.sort,
       page: this.query.page,
       queryString: this.query.q
-    });
+      // TODO: I feel like we can do better here....
+    }, flatten(this.includes || []));
```
habdelra: @ef4 I'm not so happy about this either as I'm collapsing path information by flattening this....

habdelra:
@ef4 also another thought: the jsonapi middleware uses a cache for assembling the collection includes. Would it be a good idea to be able to pass in a document cache to the document-context that can be shared amongst all documents-context instances  for a collection so that we could reuse work like is currently happening in middleware?

packages/hub/indexing/document-context.js
```diff
@@ -125,20 +126,20 @@ module.exports = class DocumentContext {
       return;
     }
     let related;
-    if (value.data && searchTree[field.id]) {
+    if (value.data && (searchTree && searchTree[field.id] || this.additionalIncludes.includes(field.id))) {
```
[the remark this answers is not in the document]
ef4: Yes, the point of searchTree is that it's a representation of additionalIncludes that is already pre-digested for convenient recursion. You shouldn't need this.

packages/hub/searchers.js
```diff
@@ -72,7 +74,7 @@ class Searchers {
     return this.get(session, this.controllingBranch.name, type, id);
   }
 
-  async search(session, branch, query) {
+  async search(session, branch, query, additionalIncludes) {
```
ef4: This is probably the root of your problem. The includes are already represented inside `query`, there's no need to change the signature here.

ef4:
I think if you switch to having a single `DocumentContext`, the rest (like using searchTree to represnt all the includes) will fall out much easier.

habdelra:
awesome--that's super helpful

packages/hub/indexing/document-context.js
```diff
@@ -150,67 +166,112 @@ module.exports = class DocumentContext {
...
+    if (searchTree[segments[0]]) { return; }
+
+    let field = contentType.realAndComputedFields.get(segments[0]);
+    if (!field || !field.relatedTypes) { return contentType; }
+
+    let relatedTypes = Object.keys(field.relatedTypes);
+    if (!relatedTypes.length) { return contentType; }
 
+    //TODO how to decide which related type to use?
+    searchTree[segments[0]] = this._buildSearchTree(Object.assign({}, searchTree), this.schema.types.get(relatedTypes[0]), segments.slice(1));
```
habdelra: @ef4 I have a question. when using the schema to look up the type for a field, i see that for fields they use a `relatedTypes` that is an object whose keys are the related type for the field. This means that it is possible for a field to be related to more than one type. I'm unsure what that semantically means when a field is associated with more than one type. I just arbitrarily chose a related type, but I have a feeling that because I don't understand this structure, I probably did the wrong thing. Can you help me to understand the `Field.relatedTypes` better? Maybe my approach to deriving the type for a field is incorrect?

ef4: Yeah, the semantics here are not fully fleshed out.

In practice right now, practical relationship fields should have one content-type in related-types. The schema is more flexible than that, with the idea that it could eventually support heterogeneous relationships, but that doesn't really get used anywhere and it would take more work to make it do anything. 

So you sense of things being off is correct, but I think OK for the present.

packages/hub/indexing/document-context.js
```diff
@@ -150,67 +178,123 @@ module.exports = class DocumentContext {
...
+      // The next fields in the searchDoc get a "cardstack_" prefix so
+      // they aren't likely to collide with the user's attribute or
+      // relationship.
       searchDoc.cardstack_pristine = pristine;
-      searchDoc.cardstack_references = this._references;
-      searchDoc.cardstack_realms = this.schema.authorizedReadRealms(type, jsonapiDoc);
-      authLog.trace("setting resource_realms for %s %s: %j", type, id, searchDoc.cardstack_realms);
+      if (!isCollection) {
+        searchDoc.cardstack_references = this._references;
+        searchDoc.cardstack_realms = this.schema.authorizedReadRealms(type, jsonapiDoc);
```
ef4: I was about to say these aren't necessary anymore (because we have `realms` as a separate property on the document context). But maybe they are still used on included resources in the search index, as part of the process of filtering a response for only resources the user can see?

habdelra: yeah, those are definitely still used.

habdelra: or well I should amend that--we use that information, but I think it's in a different place after the pgsearch refactor

habdelra: let me delete it and just see what happens....

habdelra: yeah, it's not necessary. i'll clean this up a little more

packages/hub/indexers.js
```diff
@@ -134,10 +134,10 @@ class Indexers extends EventEmitter {
           // Compare each branch, so we don't invalidate the schemas
           // unnecessarily
           for (let [branch, newSchema] of Object.entries(schemas)) {
-            let oldSchema = priorCache[branch];
+            let oldSchema = (await priorCache)[branch];
             if (!newSchema.equalTo(oldSchema)) {
               log.info('schema for branch %s was changed', branch);
-              priorCache[branch] = newSchema;
+              (await priorCache)[branch] = newSchema;
```
ef4: This kind of bug is something TypeScript would actually catch for us. And the build tooling (especially for monorepos) has gotten a lot better now. Perhaps it's time.

habdelra: I don't disagree :-)

PR #246, opened 2018-07-13
[wip] Edges plugin

alexspeller:
Initial implementation of moving the toolbar / modal handling to a seperate edges plugin.

Needs some testing, wanted to get a quick look at the approach before I crystallize with tests


packages/edges/addon/components/cardstack-edge.js
```diff
@@ -0,0 +1,14 @@
+import Component from '@ember/component';
+import { computed } from '@ember/object';
+import layout from '../templates/components/cardstack-edge';
+
+const Edge = Component.extend({
+  layout,
+  componentName: computed('edge', function() {
+    return `in-${this.get('edge')}-toolbar`;
+  })
+});
```
ef4: We can probably reduce this dynamic shenanigans by just extending the base edge component out of ember-toolbars. It's in `ember-toolbars/addon-components/-in-edge`, but that's my addon and we can make it public.

packages/tools/addon/templates/components/cardstack-tools.hbs
```diff
@@ -1,18 +1,16 @@
...
-{{/toolbar-manager}}
+  {{/edges.launcher}}
+{{/cardstack-edges}}
 
 {{#if tools.active}}
-  {{in-top-toolbar show=(component "cardstack-header") }}
-  {{in-right-toolbar show=(component "cardstack-toolbox") }}
+  {{cardstack-edge 'top' (component "cardstack-header")}}
+  {{cardstack-edge 'right' (component "cardstack-toolbox")}}
   {{cs-field-overlays}}
```
ef4: The goal as you keep refactoring will be to delete the cardstack-tools component.

We want any number of optional packages to be able to inject stuff like this:

```hbs
{{#edges.launcher}}
     {{cardstack-tools-launcher}}
{{/edges.launcher}}

{{#if tools.active}}
  {{cardstack-edge 'top' (component "cardstack-header")}}
  {{cardstack-edge 'right' (component "cardstack-toolbox")}}
  {{cs-field-overlays}}
{{/if}}

{{from-elsewhere name="paragraph-tools"}}
```

into the app's top-level template. Probably in the implementation of cardstack-edges it could be something like:

```hbs
{{#each registeredServices as |s|}}
  {{#if s.injectComponent}}
      {{component s.injectComponent edges=edges}}
  {{/if}}
{{/each}}
```

PR #253, opened 2018-07-24
refactor plugins to use a computed to access plugin-config `enabled` property

habdelra:
This addresses issue https://github.com/cardstack/cardstack/issues/237 And removes the need to read from the index as part of indexing plugins via the use of server-side computed properties

Also, I’m seeing that you cannot have a computed property be named the same as a real field, as these share a common namespace in the `ContentType.realAndComputedFields` map. this is problematic if we want a computed called `enabled` that aliases to a real field called `enabled`. so i’m naming the computed `plugin-enabled`, my thinking was that it was better to rename the computed than the real field, as i'td be a bummer for the computed to squat on a really commonly named field like `enabled`.

### DIFF
--- a/packages/core-types/cardstack/computed-field-types/alias.js
+++ b/packages/core-types/cardstack/computed-field-types/alias.js
@@ -1,0 +1,28 @@
+async function resolvePath(model, pathSegments) {
+  if (!model) { return; }
+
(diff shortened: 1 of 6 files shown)


habdelra:
Subsumed by this PR https://github.com/cardstack/cardstack/pull/255

PR #257, opened 2018-07-27
refactored so that upstreamDoc passed to DocumentContext is json:api

habdelra:
Addresses issue https://github.com/cardstack/cardstack/issues/237


ef4:
😅That's a big diff, thanks.

PR #262, opened 2018-08-01
filtering by bool is dicey since filters are coerced to strings, using `exists` instead

habdelra:
I think this relied on elastic search specific semantics, as content types that are not built-in won't even have a `built-in` field. (where ES considered not having the field the same as the field build set to `false`.)

### DIFF
--- a/packages/tools/addon/components/cs-create-menu.js
+++ b/packages/tools/addon/components/cs-create-menu.js
@@ -17,7 +17,7 @@ export default Component.extend({
   availableTypes: computed(function() { return []; }),
 
   loadAvailableTypes: task(function * () {
-    let types = yield this.get('store').query('content-type', { filter: { not: { 'is-built-in' : true } } });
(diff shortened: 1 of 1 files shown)


ef4:
I don’t think clients should need to deal with this kind of distinction. Perhaps filters on Boolean fields should automatically coerce NULL to false.

habdelra:
yeah, there's a TODO right now to use the schema for the field to help to figure out how to filter the field. right now we use solely the filter value to figure out how to assemble the SQL, which is tricky because the request coerces the `false` to a `"false"`. But looking up the field in the schema should help us to deal with this scenario in a better fashion.

ef4:
We already have some field-type-specific filter implementations: https://github.com/cardstack/cardstack/blob/e232a5d5541b60b095612dae6258685310ddcc22/packages/core-types/cardstack/field-types/integer.js#L5-L7

We just need to do the same for boolean.

PR #274, opened 2018-08-05
Handling exceptions that are thrown when migrating DB

habdelra:


habdelra:
That’s not what i’m seeing in the logs:

```
Aug 04 18:23:29 production_f82dfafb2f4d node.log: (node:48) UnhandledPromiseRejectionWarning: Error: Unable to fetch migrations: Error: Another migration is already running
Aug 04 18:23:29 production_f82dfafb2f4d node.log: UnhandledPromiseRejectionWarning: Error: Unable to fetch migrations: Error: Another migration is already running
Aug 04 18:23:29 production_f82dfafb2f4d node.log:     at process._tickDomainCallback (internal/process/next_tick.js:218:9)(node:48) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 1)
```

habdelra:
ok, changed this to log the exceptions we encounter when we cannot migrate the DB

habdelra:
I think actually not handling this might be better, as if the migration is already running then there is no way we can `await` it correctly. blowing up is probably the right thing here...

PR #287, opened 2018-08-27
Implement queryString search

mansona:
I have managed to get the previously skipped tests working with a relatively minor change to the overall system. I tried my best to follow my lead from the other aspects of the code that I was coming across, but please let me know if any of my implementations are overkill and need to be pared back 👍 

One major change that I needed to do as part of this PR is that I upgraded Postgres to v10. The reason for that is because they have built-in `to_tsvector()` support for json and bson columns as of v10 so a lot of the work for a "general search" was done for us. I'm feeling confident that this isn't going to cause us any problems because the entire test suite seems to work fine. Obviously I don't have the full context of the sorts of upstream issues this might cause but it's worth us having a conversation and discussing our options if this upgrade is not possible at this time.

Let me know if you have any questions 😄

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,6 +20,7 @@ cache:
 
 before_install:
   - docker run -d --rm -p 9200:9200 --name elasticsearch cardstack/elasticsearch:dev
+  - pushd packages/postgresql/node-tests/pg-server; docker build . -t cardstack/pg-test; popd;
   - docker run -d --rm -p 5444:5432 --name postgres      cardstack/pg-test
   - export PATH=$HOME/.yarn/bin:$PATH
 
@@ -34,5 +35,3 @@ before_script:
 
 script:
   - yarn test
-
-
--- a/packages/hub/searchers.js
+++ b/packages/hub/searchers.js
@@ -118,7 +118,7 @@ class Searchers {
     let result = await next();
     if (result) {
       let schema = await schemaPromise;
-      let includePaths = (get(query, 'queryString.include') || '').split(',');
+      let includePaths = (get(query, 'include') || '').split(',');
       let pristineResult = await (new DocumentContext({
         branch,
         schema,
--- a/packages/jsonapi/middleware.js
+++ b/packages/jsonapi/middleware.js
@@ -163,6 +163,7 @@ class Handler {
       filter: this.filterExpression(type),
       sort: this.query.sort,
       page: this.query.page,
+      include: this.query.include,
       queryString: this.query.q
     });
     let body = { data: models, meta: { total: page.total } };
--- a/packages/jsonapi/middleware.js
+++ b/packages/jsonapi/middleware.js
@@ -163,7 +163,7 @@ class Handler {
       filter: this.filterExpression(type),
       sort: this.query.sort,
       page: this.query.page,
-      queryString: this.query
+      queryString: this.query.q
     });
     let body = { data: models, meta: { total: page.total } };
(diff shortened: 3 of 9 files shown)


ef4:
Thanks, glad to see you're diving in here. You're on the right track.

I'm fine with upgrading postgres, and we should do that regardless. But I want to suggest a slightly different implementation that will be more extensible going forward. My motivations are:

 - it will need to be possible for users to enable/disable full-text search on a per-field basis. Lots of stuff definitely should not get searched.

 - we may want to help the site owner not unintentionally leak sensitive information via full-text search. For example, you can add a "what-we-really-think-about-them" field to the "users" content type and set it to be readable only by admins. Users are prevented from ever reading the field, but they can do things like `/api/users?q=dumbass` to test what the admins think of them. 😆

I don't think we need to solve either of those right now, but my suggestion is to make `q` be an actual column in the documents table of type `tsvector`. No triggers or anything will be needed to keep it up to date, since we're already doing our own indexing process. But this will allow us to make choices at indexing time about what goes into `q`, and then not worry about it at query time.

The places to look (beyond where you're already working) are:

 - [here is where we write to the documents table](https://github.com/cardstack/cardstack/blob/b55497fb529303f848f53ad3a74ef56dbaff5428/packages/pgsearch/client.js#L202-L214). You will want to do something similar to what `expiresExpression` is doing: build up an expression for the value of the `q` column. It can still be `to_tsvector(search_doc)` for the present, but this is where we will eventually get fancier.
 - [here is where you can add the new column to the postgres schema](https://github.com/cardstack/cardstack/tree/master/packages/pgsearch/migrations). This is using [node-pg-migrate](https://github.com/salsita/node-pg-migrate). It includes a CLI command for generation a new migration.

mansona:
@ef4 thank you for the guidance, I was expecting to have to spend a good part of the day researching this but thanks to my background research yesterday and you pointing me in the right direction I was able to get this working surprisingly easily 😂 

I tried my best to follow the overall style of the `pgsearch/client.js` file but I'm not too sure my addition of a "tsVector" param type is too high level: https://github.com/cardstack/cardstack/pull/287/files#diff-097a43c396f5c2e81c473208b3d1e967R22 I would just make it a kind of `wrapParam` type where you have to define the wrap function where you use it: https://github.com/cardstack/cardstack/pull/287/files#diff-ca6fb7ebcbd4ab7ad3b0e7205c532d9aR207

let me know what you think and I'll clean up this PR so we can hopefully get it merged 🎉

ef4:
Thanks. The mini expression language in here isn't documented anywhere, but this is how it works:

Every expression is an array. Each element of the array is either a string (in which case it is a snippet of raw SQL) or an object with `{ param }`. All user-provided data must go through params to avoid SQL injection. The values passed to `upsert` are expressions, so you can say:

```js
q: ['to_tsvector(', param(searchDoc), ')']
```

If you switch to that you can drop `tsVector` and the `wrap` support from util. I would rather not increase the complexity of the expression language when we can do it this way instead.

Otherwise everything looks good.

mansona:
Thanks again for the info @ef4 👍 

I did come across his expression language in my investigations but it hadn't occurred to me that it might work in this context 🤔 

I have updated the PR now to remove my `wrap` implementation and it still seems to be working. Let me know if you need anything else before we can land this one 👍

mansona:
@ef4 I've simplified queryCondition and it seems to work fine 👍

ef4:
Nice.

PR #288, opened 2018-08-28
Add ability to search all card types at once

mansona:
Adding an endpoint that allows you to query all card types at once.

This is a special case that essentially removes the `type` definition from the filter in the case that the type name is `all`

This allows for the ability to write: http://localhost:3000/api/all?q=testing which will query all types.

### Questions/Comments

- [x] This opens up the ability to GET http://localhost:3000/api/all and essentially list all cards without having to know the type name ahead of time. If this is an issue I can add a restriction so that you can only use this endpoint if you have specified a query string using `?q=`

### DIFF
--- a/packages/jsonapi/middleware.js
+++ b/packages/jsonapi/middleware.js
@@ -100,13 +100,6 @@ class Handler {
       // name itself is taking precedence).
       filter.type = type;
     }
-
-    if (filter.type === 'all') {
-      // this allows for a user to query all document types at once using the
-      // endpoint name 'all'
-      delete filter.type;
-    }
-
     if (id != null) {
       filter.id = id;
     }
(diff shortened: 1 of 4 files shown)


ef4:
I had forgotten that this didn't already exist. 😀

I'm not concerned with hiding the types, that's what the grants system is for. 

Right now the endpoint `/api` doesn't do anything useful, can we move this to there?

mansona:
So I quickly added what you suggested, so now you can either query http://localhost:3000/api/all or http://localhost:3000/api and they will act the same way.

It may seem a little convoluted but I was going for maximum reuse, instead of trying to keep a generic `handleCollectionGET` and `handleCollectionGETAllTypes` in sync.

let me know your thoughts on the implementation and I'll adjust to your liking 👍

ef4:
@mansona notice that I only kept the /api endpoint, not /api/all.

ef4:
Published in 0.10.5.

PR #289, opened 2018-08-28
add ability to pass params to a cardstack-content

mansona:
This opens up the possibility for inter-card communication and will no doubt require some careful discussion.

I am putting together a demo/example now that uses this addition and is the main reason I would like to see it merged 👍


ef4:
This doesn't really fit with the vision for what cards are supposed to be.

`{{cardstack-content}}` is nearly like `{{outlet}}`. You're generally not supposed to know or care what code is inside. It doesn't take customizable parameters for the same reason that outlet doesn't. It's opaque to the caller.

I know you are trying to hack around the lack of some other important features (particularly card routing), but let's come up with an alternative to this that doesn't introduce public API.

PR #290, opened 2018-09-05
Add support for relationship tracking & saving

balinterdi:
Where relationship tracking is provided by the ember-data-relationship-tracker add-on

* [x] The update for a related record fails with a 400, `missing required field "meta.version"`, as the meta data is not sent in the payload
* [x] Related records are saved when they are added or deleted but not when only changing one of their attributes (update the title of a document-chapter)

### DIFF
--- a/packages/models/addon/model.js
+++ b/packages/models/addon/model.js
@@ -12,9 +12,8 @@ export default DS.Model.extend(RelationshipTracker, {
 
   init() {
     this._super();
-    let relationMap = this.constructor.relationshipsByName;
     let dirtyTrackingProperties = [];
-    relationMap.forEach((relation) => {
+    this._relationshipsByName.forEach((relation) => {
       let { kind, meta } = relation;
       let { owned } = meta.options;
       if (owned) {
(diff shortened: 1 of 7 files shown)


packages/models/cardstack/code-generator.js
```diff
@@ -54,6 +54,13 @@ define('@cardstack/models/generated/{{modelName}}', ['exports', '@cardstack/mode
         }),
        {{/if}}
      {{/each}}
+
+     resourceMetadata: Ember.inject.service(),
```
balinterdi: @ef4 Not sure how we can import `inject` and `computed` so that we can use the direct imports and potentially benefit from code splitting down the line. (And I'm also uncertain whether that matters at this point.)

packages/models/addon/model.js
```diff
@@ -1,3 +1,85 @@
...
+  relationshipTrackerVersion: computed(function() {
+    let meta = this.get('resourceMetadata').read(this);
+    return meta && meta.version;
+  }),
+
+  async save() {
+    // this._super is not safe to use asynchronously
+    // see https://github.com/ember-cli/ember-cli/issues/6282
+    let modelSave = this._super.bind(this);
+    run(async () => {
```
balinterdi: @ef4 I have no idea why `run` is needed here. The tests that failed because of the lack of it had a `run` wrapped around them but it didn't seem to matter. I think the need for the run loop is going away soon so it didn't seem like a big deal to add this (temporarily?), but let me know you think it's an issue.

ef4: Yeah, it's because of the async functions. Each time you return from `await` you are starting up with no run loop. It's really only an issue in tests because of the autorun assertion, which is gone in ember 3.4.

You should feel free to update the ember version wherever you see a need for it.

PR #291, opened 2018-09-06
Reload models from API on route transitions

marcoow:
This fixes a problem with models being reused from Ember Data's in-memory store on route transitions. [Ember Data's `findRecord`](https://emberjs.com/api/ember-data/3.3/classes/DS.Store/methods/findRecord?anchor=findRecord) will returns records from its in-memory store if they are present already and not wait for the API request to finish. If these records have synchronous relationships though, it cannot be known for sure whether all of these relationships have been loaded already, potentially leading to the dreaded _"Uncaught Error: Assertion Failed: You looked up the 'xyz' relationship on a 'abc' with id 1 but some of the associated records were not loaded. Either make sure they are all loaded together with the parent record, or specify that the relationship is async (`DS.belongsTo({ async: true })`)"_ error.

This opts out of that behavior (`reload: true`) and always waits until the API request finished and the complete record has been loaded from the API.

closes cardstack/deck#58

### DIFF
--- a/packages/routing/addon/routes/cardstack/common.js
+++ b/packages/routing/addon/routes/cardstack/common.js
@@ -29,7 +29,7 @@ export default Route.extend({
         branch
       });
     } else {
-      promise = this.store.findRecord(mType, slug, { branch });
+      promise = this.store.findRecord(mType, slug, { branch, reload: true });
     }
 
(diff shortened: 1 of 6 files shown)


ef4:
Thanks. 👍

PR #292, opened 2018-09-07
adding implementation for read-time computed field type

mansona:
### Open Questions
- [x] for some reason the field type is not passing eslint. I tried to recreate https://github.com/cardstack/cardstack/blob/9d7a12d8dc9f7284866c9f771a8df8d1c04bf973/packages/core-types/.eslintrc.js but it didn't quite work. Any guidance would be appreciated
- [x] it's unclear how exactly to test this 🤔 I have been looking at some of the other field types and their tests but there doesn't seem to be anything quite like this quite yet
- [x] I have also been looking into the editor to try to make it possible to edit which field this computes the sourceField from, but I've not quite been able to follow that thread


ef4:
The lint issue is fixed in https://github.com/cardstack/cardstack/pull/292/commits/75eb070bb271582b37f4577ba2635608d5fe7e75. The issue was that the package declared itself as supporting node 4, so the linter was refusing to allow things like async functions that don't work in those node versions.

ef4:
To write a test, you can model after https://github.com/cardstack/cardstack/blob/master/packages/models/tests/integration/models-test.js.

To make that kind of test work, you will need a little setup in the `@cardstack/mobiledoc` package:

1. Add `@cardstack/test-support` as a devDependency.
2. Configure a data source in the dummy app by copying what's in https://github.com/cardstack/cardstack/tree/master/packages/models/tests/dummy/cardstack

ef4:
For editing the settings: we don't have any UI yet for editing schema, and this falls into that category.

The schema is all "just data" already, so you could in theory make a `content-type` isolated card and an appropriate set of field editors and you could change the schema. We just haven't designed any of that visually yet.

mansona:
I've tried to recreate as much of the `models` test as I could follow. I'm how hitting a strange error 

```{"errors":[{"title":"Validation error","detail":"field \"fields\" refers to disallowed type(s) \"computed-fields\"","code":400,"source":{"pointer":"/data/relationships/fields"}}]}``` 

which I'm assuming means that I'm missing a little bit of setup to allow for computed fields in this kind of a test? 

If you could give me a point in the right direction that would be great. 😄

File: packages/mobiledoc/.eslintrc.js [hunk 1]
```diff
@@ -1,4 +1,12 @@
...
-  extends: '@cardstack/eslint-config/browser'
+  extends: '@cardstack/eslint-config/browser',
+  overrides: [{
+    files: 'node-tests/**/*.js',
+    rules: {
+      'node/no-unpublished-require': ['error', {
+        'allowModules': ['@cardstack/test-support'],
+      }]
+    }
+  }]
```
ef4: A better way to fix this lint error is to add `node-tests` to `.npmignore`.

PR #294, opened 2018-09-12
[CONCEPT] cs-field-alternatives

jorgelainfiesta:
This PR proposes an implementation for mutually exclusive alternatives field described in https://github.com/cardstack/deck/issues/80

At the moment, the restriction imposed by `cs-field-alternatives` is purely at the UI level. This means that when the user edits these fields, they can choose which group of fields to edit. When they choose an alternative, the other alternative's fields are cleared. 

The purpose of this PR is to show a concept of how a field like this could be implemented. We also want to evaluate if it's desirable to have this field option.

Pending issues:

- Implement a way to show a `placeholder` for `cs-field-alternatives`
- Implement a test for `renderign` and `tools` components that affect this feature.
- Revisit UI styles for the `tools` component.
- Fix mark overlay that renders in a different location.

Considerations:

- Taking this to the schema level could bring a better user experience and ensure that the fields are in fact mutually exclusive. However, it could be relatively complex because we would have to define a mechanism to logically group fields in the schema.

### DIFF
--- a/packages/rendering/addon/components/cs-field-alternatives.js
+++ b/packages/rendering/addon/components/cs-field-alternatives.js
@@ -1,7 +1,7 @@
 import layout from '../templates/components/cs-field-alternatives';
 import { humanize } from '../helpers/cs-humanize';
 import Component from '@ember/component';
-import { computed } from '@ember/object';
+import { computed, get } from '@ember/object';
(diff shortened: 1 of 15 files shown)


ef4:
Closing this in factor of a deeper integration with the constraint system, as we discussed.

PR #303, opened 2018-09-17
[WIP] Card picker

mansona:
This is still very much WIP but I wanted to open the PR early so I can get some feedback to see if the approach is correct so far.

It's mostly working (with some serious issues on the UX for now) 

Example: https://cl.ly/2829a0be7803


mansona:
I have made some more progress 🎉 it now functions like you would expect a file-picker to function, although there seems to be an issue updating the reference in the relationship 🤔 Here is the updated example https://cl.ly/1dde0c43b6a4 and here is the error that I'm getting in the console when I try to set the model as the new relationship on the field: 

```
Assertion Failed: The 'article' type does not implement 'homepage' and thus cannot be assigned to the 'partnersCard' relationship in 'homepage'. Make it a descendant of 'homepage' or use a mixin of the same name.
```

do I need to do something different to be able to set the value for the relationship?

balinterdi:
@mansona I've only seen this error in Ember Data and created an issue for it about the unclear error message: https://github.com/emberjs/data/issues/4633

Unfortunately, I don't think that helps a lot and somebody has fixed it in the meantime. Is there a polymorphic relationship somewhere that might throw this?

ef4:
That error is almost certainly fallout from the way we have implemented polymorphic relationships. They are really only read-only at the moment, from ember-data's perspective.

mansona:
@balinterdi thanks for the link, unfortunately even with the new message I don't 100% understand what's going on 😂 I've never come across this error or any like it before 🤔 

@ef4 I found the bit in the code that's actually producing the model definition (the code-generator in the `@cardstack/models` package) and I tried adding `polymorphic: true,` to the model definition but it didn't work. I understand that this is **super** naive approach but I'm not 100% sure how to proceed. Any pointers that I might look into? I was thinking that we might be able to have all the models depend on a subclass "card" model and implement polymorphism like it's [explained in the docs](https://guides.emberjs.com/release/models/relationships/#toc_polymorphism) but I'm not sure how to implement that in the "compiled" world of this code-generator and I didn't want to go down that route if we didn't think it was worth it.  

I am going to try to test this with a `has-many` relationship now too but I am assuming that we will have the same issue.

mansona:
This PR has now been superseded by #316 #317 #318 and #319 

I split the PR into each logical piece that it included just so we can get each PR merged independently and keep any required conversation focused 👍

Issue #299, opened 2018-09-17
Allow editing fields of owned relationships with the parent


balinterdi:
Since records in owned relationships cannot be edited on their own, we should support a way of editing them when the owner (the parent) is being edited.

We could, for example, list all fields of records in owned relationships in the right sidebar, but in the case of a has-many relationship, it can be challenging to come up with a good UI as we don't want to render all the fields of all the records (that would mean n * m fields)

PR #305, opened 2018-09-17
Add fields from owned relationships to sidebar

balinterdi:
This implements https://github.com/cardstack/cardstack/issues/299


packages/tools/addon/services/cardstack-tools.js
```diff
@@ -37,7 +37,26 @@ export default Service.extend({
...
+        if (fieldModel === activeItemModel) {
+          return true;
+        }
+        let isOwnedField = false;
+        activeItemClass.relationships.forEach(([ relationshipDefinition ]) => {
+          let { meta } = relationshipDefinition;
+          let { owned } = meta.options;
+          if (owned) {
+            if (meta.type === fieldClass.modelName) {
+              isOwnedField = true;
```
balinterdi: It would be great to jump out of the `forEach` here but that gives an  "unsyntactic break" linting error and there are no other methods on the map backing the relationship definitions that allow to cycle through it. Not a biggie, I think, but worth mentioning.

balinterdi:
True, good point, I'll extend this.

balinterdi:
@ef4 Could you review it again? I've addressed the issues and the tests also pass. Thank you.

balinterdi:
@ef4 The test actually brought out a bug that I've fixed and I'm also happy with the test. I set ember-data 3.4 as the dependency. Please merge if you deem it mergeable.

ef4:
Nice, looking good.

In general, feel free to update any or all cardstack packages to latest ember, ember-cli too.

PR #309, opened 2018-09-19
fixing eslint-config on mobiledoc package by using npmignore for node-tests

mansona:
This is just based on a comment on a previous PR that has been merged https://github.com/cardstack/cardstack/pull/292/files#r218638387


ef4:
Thanks for following up.

PR #310, opened 2018-09-19
Image s3 storage

alexspeller:


packages/hub/searchers.js
```diff
@@ -91,6 +91,27 @@ class Searchers {
     return authorizedResult;
   }
 
+  async getBinary(session, branch, type, id) {
+    // look up authorized result to check read is authorized by going through
+    // the default auth stack for the JSON representation. Error will be thrown
+    // if authorization is not correct.
+    await this.get(session, branch, type, id);
```
alexspeller: look at includePaths

packages/image/cardstack/field-type.js
```diff
@@ -1,5 +1,16 @@
...
+    if (Array.isArray(value.data)) {
+      return 'accepts only a single resource, not a list of resources';
+    }
+    if (relatedTypes && value.data && !relatedTypes[value.data.type] ) {
+      return `refers to disallowed type "${value.data.type}"`;
+    }
     return true;
   },
+  default: { data: null },
+  isRelationship: true
```
alexspeller: check for hardcoded belongs-to / has-many

packages/jsonapi/middleware.js
```diff
@@ -55,6 +57,20 @@ function jsonapiMiddleware(searcher, writers, indexers, defaultBranch) {
       return;
     }
 
+    let handler = new Handler(searcher, writers, indexers, ctxt, options);
+
+    let contentType = ctxt.request.headers['content-type'];
+    let isMultipart = contentType && contentType.includes('multipart/form-data');
```
alexspeller: check for not json api

packages/s3/cardstack/writer.js
```diff
@@ -0,0 +1,63 @@
...
+  constructor({ dataSource, branches }) {
+    this.dataSource         = dataSource;
+    this.branches           = branches;
+  }
+
+  async s3Upload(branch, options) {
+    let config = this.branches[branch];
+    log.debug(`Uploading file to S3 key ${options.Key}`);
+
+    let s3 = new AWS.S3({
```
alexspeller: Check if this being null is enough for default values to work

packages/s3/cardstack/writer.js
```diff
@@ -0,0 +1,63 @@
...
+    log.debug(`Upload of ${options.Key} successful`);
+  }
+
+  async prepareBinaryCreate(branch, session, type, stream) {
+    let finalizer = async (pendingChange) => {
+      let Key = `${pendingChange.finalDocument.id}`;
+      // encode metadata as json because values are required to be strings and
+      // we want to store various JSON values
+      let Metadata = {
+        jsonAttrs: JSON.stringify(pendingChange.finalDocument.attributes)
```
alexspeller: Change this

alexspeller:
https://github.com/cardstack/deck/pull/124/files#diff-fc34b375ed39e1636b2eed1140cbe0d0

ef4:
@alexspeller can you help me figure out where we left this PR? What's left to do so we can merge this?

alexspeller:
@ef4 so this should be ready to go except the tests, which actually is something weird, it looks like lerna bootstrap is installing a published version of @cardstack/jsonapi in the local node_modules folder instead of linking together correctly.

The only weird thing in this implementation is that [the consuming app has to define the cs-file model](https://github.com/cardstack/cardstack/pull/310/files#diff-80ddc3f73122219e1a5f39c14cead658R6)

I think this makes sense for now - eventually you'd have a @cardstack/file addon that would define the file model and you could tell it which data source it should be stored in, but AFAICT you can't have an addon provided content type and have the consuming app tell it to put it in an arbitrary data source at config time

PR #313, opened 2018-09-20
use codegen to generate github auth config instead of fetch

habdelra:


ef4:
Yay, this was bugging me. 🎉 

We should remove the remaining `exposeConfig` support in the authentication package.

PR #321, opened 2018-09-21
GitHub user permissions support and user caching

habdelra:
This is broken out into 2 commits for easier review: one commit for the user caching, and the other commit for the permissions support.

### DIFF
--- a/packages/github-auth/node-tests/searcher-test.js
+++ b/packages/github-auth/node-tests/searcher-test.js
@@ -11,46 +11,193 @@ const {
 describe('github-auth/searcher', function() {
   let env, searchers;
 
-  async function setup() {
-    let factory = new JSONAPIFactory();
-
-    factory.addResource('data-sources', 'github').withAttributes({
-      sourceType: '@cardstack/github-auth',
-      params: {
-        'client-id': 'mock-github-client-id',
-        'client-secret': 'mock-github-client-secret',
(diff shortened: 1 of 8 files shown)


packages/github-auth/node-tests/searcher-test.js
```diff
@@ -5,52 +5,399 @@ const {
...
+      factory.addResource('data-sources', 'github').withAttributes({
+        sourceType: '@cardstack/github-auth',
+        params: {
+          'client-id': 'mock-github-client-id',
+          'client-secret': 'mock-github-client-secret',
+          token: 'mock-github-token',
+          permissions: [
+            { repo: 'cardstack/repo1', permission: 'read' },
+            { repo: 'cardstack/repo1', permission: 'write' },
+            { repo: 'cardstack/repo1', permission: 'admin' },
```
habdelra: @ef4 I'm wondering if I made the permissions to complex of an object. maybe instead of representing permissions as an array of objects, it would be easier to setup groups if they were an array of strings? like: 
```js
[ 'cardstack/repo1:read', 'cardstack/repo1:write', 'cardstack/repo1:admin']
```

what do you think?

ef4: Yeah, seems good. Try to write some actual groups to be sure.

PR #322, opened 2018-09-24
Git remote init for git datasource

mansona:
This allows you to use git-clone instead of git-init to initialise a repo in the case that the configured repo does not exist. The existing behaviour is that it will initialise an empty repo for you, but this allows you to clone instead so that you can start with some initial data.

### Open Questions
- [ ] are we ok with this implementation (for now) that is essentially a read-only way to initialise data from a remote?
- [ ] are we ok with a test that relies on github being up? (see `git/indexer with remote`)
- [ ] can we create/move this example data git repo to the cardstack org so that it doesn't rely on my personal repo 😂

### DIFF
--- a/packages/git/indexer.js
+++ b/packages/git/indexer.js
@@ -82,7 +82,9 @@ const {
   Repository,
   Reference,
   Branch,
-  Commit
+  Commit,
+  Clone,
+  Cred,
 } = require('@cardstack/nodegit');
 
 const Change = require("./change");
@@ -94,11 +96,12 @@ const log = require('@cardstack/logger')('cardstack/git-indexer');
 module.exports = class Indexer {
   static create(params) { return new this(params); }
 
-  constructor({ repo, basePath, branchPrefix }) {
+  constructor({ repo, basePath, branchPrefix, remote }) {
     this.repoPath = repo;
     this.branchPrefix = branchPrefix || "";
     this.basePath = basePath ? basePath.split('/') : [];
     this.repo = null;
+    this.remote = remote;
   }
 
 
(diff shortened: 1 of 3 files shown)


habdelra:
@mansona I've actually been using nock to mock github. might not be a bad idea to use that instead of relying on actual github. https://github.com/cardstack/cardstack/blob/master/packages/github-auth/node-tests/authenticator-test.js#L50-L79

mansona:
@habdelra in this case I'm not relying on the "github" part of this equation, I'm more-so relying on having access to a git+ssh remote repository 🤔 from my quick look at the api of nock it's more for mocking out the github http api

I was thinking of creating some convoluted setup where we're providing a local ssh server in the tests but I didn't want to limit the platforms that the tests could run on 😖 but I do share the concern about relying on a third party for tests like this

habdelra:
I wonder if maybe creating a docker for a git repo might be a good idea then, and then spinning that up as part of the tests like we do for elastic search, postgres, and our private blockchain...

ef4:
I'm fine with a docker-based test.

This is a great start, but there's more to do to make it practically usable on any real site. The goal here is to make the remote repo the authoritative data store. The local repo on disk becomes an implementation detail.  It should be throwaway.

The git writer should not return success until the new commit is on the _remote_ repo. In practice, this will require working on a temp branch in the local repo, then `git push the_remote the_temp_branch:the_real_branch`. 

Why a temp branch? Because one-way data flow. We don't want to get into a state where we're resolving differences between the local `master` and the one on github. So `master` only gets updated by _pulling_. Local changes all happen on separate throwaway branches, which can't conflict with each other except at the moment when you're trying to push them to the remote master.

ef4:
Along with the above, it stops making sense to have separate `repo` and `remote` parameters. It's just the `repo` parameter, which is something a local path and sometimes a remote repo. 

If there's a second parameter, it would be optional and would be something like `tempDirectory`, which would be the local path where we clone remote repos into.

PR #324, opened 2018-09-25
Mutually exclusive restriction

jorgelainfiesta:
This PR adds a new restriction that allows users to indicate mutually exclusive attributes in their schema.


ef4:
Nice!

PR #333, opened 2018-10-02
Remote git datasource

mansona:
This PR is now built on top of #339 to keep things a bit cleaner 👍 #339 is safe to merge because all that it is doing is updating the tests. If you are happy with this PR then you can merge this directly and it will contain all commits from #339 

This PR now contains the new branching logic that @ef4 had mentioned that he wanted. I would like some assistance in coming up with some effective test cases to test the conflict resolution but I am confident that this is now good enough to be reviewed and for us to consider merging 🎉 

### Todo
- [x] publish `cardstack/git-test` docker image
- [x] figure out a way to get the repo config into a `Change`
  - Currently the only way I can see is to add it to the constructor params. I feel a bit off about that because there are so many already
- [x] do the branch strategy that @ef4 mentioned in the [original discussion on the issue](https://github.com/cardstack/cardstack/pull/322#issuecomment-424395243)


habdelra:
@mansona so based on that discussion that we had during the demo. I think the identity of the committer actually is here: https://github.com/cardstack/cardstack/blob/59e952181d021d06f0e83bb196d1f0dfbbadf4fb/packages/git/writer.js#L157-L160

the `github-user` actually has a `name` attribute instead of `full-name` attribute. Albiet, I wonder how we should adopt this to an arbitrary user model, as I'm not sure we should always assume there is a `full-name` attribute.

A general observation not directly related to this PR:
@ef Also, in terms of the committer, github uses the email in the commit header to link to a github user account. it would be great if we could stay consistent with github and not use the username@host of the remote repo for the email that we set as part of the git committer https://github.com/cardstack/cardstack/blob/59e952181d021d06f0e83bb196d1f0dfbbadf4fb/packages/git/writer.js#L31 perhaps this could originate from a attribute in github-user that the git data source can be configured to look at (much like deriving the `name` attribute above)?

packages/git/writer.js
```diff
@@ -71,7 +95,15 @@ module.exports = class Writer {
 
     await this._ensureRepo();
     return withErrorHandling(id, type, async () => {
-      let change = await Change.create(this.repo, document.meta.version, this.branchPrefix + branch);
+      const branchName = this.branchPrefix + branch;
+      let change;
+
+      if(this.remote) {
+        change = await Change.createRemote(this.repo, branchName, this.fetchOpts);
```
ef4: I don't think this can be correct. It's ignoring `document.meta.version`, which is really important.

In general, I think the distinction between `create` and `createRemote` is probably not needed / actively harmful.

packages/git/indexer.js
```diff
@@ -107,25 +114,27 @@ module.exports = class Indexer {
 
   async _ensureRepo() {
     if (!this.repo) {
+      if (this.remote) {
+        let tempRepoPath = await mkdir('cardstack-temp-repo');
+        this.repo = await Clone(this.remote.url, tempRepoPath, {
```
ef4: If I understand correctly, the indexer always clones and never fetches. I don't think that can be correct.

If somebody makes changes to the remote repo (manually, bypassing the hub), and indexing runs, their changes need to get indexed.

mansona:
@ef4 this is now (finally) ready for detailed review 🎉 

I was able to further generalise the pushCommit/applyCommit implementation as we discussed on Thursday only in that I have moved the error backoff into the `finalize()` function. This solves the other issue you brought up in that now the `delayTime` is **not** taken up by the time taking to fetch changes 👍 There was a potential to make `applyCommit` also deal with the `pushCommit` functionality internally but I think this is a bit cleaner for now.

Let me know if you have any questions 😄

PR #334, opened 2018-10-02
Fetch dropdown options by content type not field name

bagby:


ef4:
This looks good.

If somebody tries to use it on an unrestricted polymorphic relationship, it will see `meta.type === 'cardstack-card'`. Which is actually kinda correct, and it will try to `findAll('cardstack-card')`, which isn't implemented but could be. Really though that is probably not a good use case for this field-editor, and "pick any card" should instead use the bottom edge card chooser.

Issue #301, opened 2018-09-17
Show validation errors for built-in field editors


balinterdi:
Some constraints are implemented on the server-side in the packages/core-types/constraints folder and  fields can use these constraints. If a save is attempted for a resource with invalid fields, the server will reject it with a proper validation error message. However, the UI is not wired up at all to do anything with the errors received from the server.

On the longer term, we envision a server-side endpoint which would return the validation errors (if any) without attempting to save/update the resource. If we hit that endpoint in the background, we could display errors to the user without them needing to hit the Update button. That'd save us from having to write the validations on the client-side, too.

balinterdi:
Closed via #335

PR #335, opened 2018-10-03
Show validation errors for built-in field editors

balinterdi:
See #301 

* [x] Restore `invalid` class name for `cs-collapsible-section`
* [x] Add `validate` method to CS Model class and call it when the validation needs to be made
* [x] Refactor validation so that it doesn't use ED's `errors` object or its `save` method
* [x] Also validate records in owned relationships
* [x] Write test for validation errors
* [x] Make sure built-in editors trigger validation when leaving them/changing the selected value
* [x] Test `cs-active-composition-panel` with a `grouped` fieldMark
* [x] Test "type" checks (a string value cannot be passed as an integer, for example) and whether the error that is returned is properly displayed
* [x] Modernize integer-editor (still uses manual `update`, `invalid`, etc.)
* [x] If a (custom-built) editor doesn't call `validate` when a value changes, make sure the error is at least displayed when updating the model
* [ ] Can we generalize the validation triggering by not calling it from within the individual editors?


balinterdi:
Meeting notes from discussion with @ef4 

* **Step 1 : Generalize the current solution and implement it in built-in editors**
  * A few properties could be passed into field-editors so that authors can just use those component arguments (`hasValidationError` (or `invalid`) and `errorMessage`)
  * The panel around each field could be styled differently to show invalidity. That'd do away with the need to open invalid fields.
  * We could also display the error message before calling the field-editor component so that displaying error messages are consistent and authors don't have to do anything
* **Step 2:  Show errors before submitting to the back-end**
  * Establish new API endpoint to hit regularly:  `POST /validations/documents(/:id)`
  * As a first step, instead of a new endpoint, we could reuse the existing ones and just add a `onlyValidation=true` QP to be able to play around with it
  * Simple approach: show errors as they are produced
  *Note: there is a "typed" check for certain type fields which are verified by the server. For example, a number can only be a number
* **Step 3: Find out (and implement) when to show the errors to users**
  * Start with the most blunt one: show all errors as they are produced
  * Implement a more refined "show validation error" strategy (probably showing the error when the user leaves the field)

packages/tools/addon/templates/components/cs-active-composition-panel.hbs
```diff
@@ -16,14 +16,16 @@
...
-      class=(concat "cs-toolbox-section" (if (has-errors fieldMark.model) 'invalid')
-      OR hasErrors=(has-errors model=fieldMark.model)
-  --}}
   {{#each fields key="id" as |fieldMark|}}
     {{#cs-collapsible-section
-      class="cs-toolbox-section"
+      class=(concat "cs-toolbox-section "
+        (if (cs-has-errors
+              fieldMark.model
+              fieldMark.model.content.isValid
```
balinterdi: @ef4 The 2nd and 3rd parameters are only passed to the helper so that they trigger a re-computation of `cs-has-errors` when they change. That's unfortunate (and potentially confusing to the readers of the template/helper) but I cannot think of another way (that's not even worse, like creating a stateful helper and setting up an observer to trigger `recomputed` manually).

balinterdi:
After discussing with Ed, the new direction is to build a separate `api-validation` namespace in the jsonapi middleware and add a `validate` method to the `cardstack-data` service that calls into that namespace.

The validation endpoints would reflect the `api` namespace endpoints in their URL structure and would return the validation errors with a 422 status code if there are errors and an empty 200 response if all is good. The error response will have to contain the field names that are in error. The component making the validation request could then pass down the validation error to each respective field editor so that field editors don't have to know anything about how they should validate (and not even that they are validated – they just call the passed-in task when their value changes).

balinterdi:
@ef4 The only item left from the above todolist is `Can we generalize the validation triggering by not calling it from within the individual editors?`.

The idea here (we discussed this together during one of our calls) was to have a way to trigger the validation from the `cs-field-editor` component which calls the appropriate editor (via the dynamic `component` call) instead of the editor components themselves.

I can't think of a way that's DDAU-compatible (and thus doesn't potentially cause weird async issues ) as the action would have to be fired from "the bottom" and different editors have different events when they should fire the validation (inputs have `on-blur`, a dropdown has `on-change`, and so on).

The current solution passes down a `validate` task an `onchange` parameter into the specific field editors which the editor calls when appropriate. Custom editors should call it, too, but even if they don't, the model will be validated before saving.

Let me know if you can think of a better way.

ef4:
One option is to change the way we call all editors so they are forced to be DDAU. Instead of two-way binding to `(get content field)`, they would get a value down and a `setValue` action:

```diff
{{component
      editorName
-     content=content
      field=field
+     value=(get content field)
      editorOptions=editorOptions
      tagName=""
      enabled=enabled
      disabled=disabled
+     setValue=(action setValue content field) 
}}
```

Benefits of this are that we can call the `validate` task (and also some other things, like our mandatory `watchRelationships`) inside our own `setValue` action in the cs-field-editor component.

The question is: does this actually work OK for all our existing editors? Is it practical to force them to be fully one-way bound?

A second option is to modify the above to make two-way binding also possible, by keeping `content=content`, and perhaps offering an additional `didSetValue` action that is distinct from `setValue`. This is closer to your current implementation.

A third option is to stick with what you have, although in that case I would prefer to make `onchange` a plain function, not an ember concurrency task (it can invoke the task inside itself, but we can abstract that away from all the field editors).

Issue #331, opened 2018-09-28
Add @cardstack/data service


habdelra:
From @marcoow :

We need a mechanism to make sure all required fields and associations are loaded for a card that is rendered in its embedded version as part of a relationship of another card. Often cards have relationships to other cards that are not restricted in terms of the types of the related cards so it is not possible to list the required includes for the related cards up-front. Instead, the related cards need to have a mechanism to define the required fields and relationships for their embedded version themselves and then the routing layer needs to make sure to load these fields and relationships.

### Idea
The idea is to add a data service to `cardstack/cardstack` that reads a property for all loaded cards, that says which fields and relationships are needed for the particular cards. That service can then be used by `@cardstack/routing` to automatically load only (but also all of) the required fields and relationships for the isolated card that is being loaded and also for all of the embedded cards that that isolated card embeds.

### TODO
This is only the first step for implementing the above idea. We'll introduce a new fieldsets property to the cards (or content types really) that lists the fields and relationships that are needed both for the isolated and embedded version of the card:

```
fieldsets: {
  isolated: [
    'title',
    'hub.category'
  ],
  embedded: [
    'title'
  ]
}
```

- [ ] add a client data service that has a `load(branch, type, id)` method that looks at the content-type's fieldsets property and loads only the fields and relationships listed there for the card
- [ ] emit the code modules for the loaded cards as part of the `load()` mechanism, like the dynamic import aspect of https://github.com/ef4/ember-auto-import

habdelra:
From @marcoow 

> ember-data does not actually support sparse field sets apparently except through this addon: https://github.com/BookingSync/ember-data-partial-model. The problem with that is though that it requires defining all the attributes that might not always be loaded specifically so it's a) not super flexible and b) would require us to do a bunch of additional things in the ember-data models backing the cards.

>I suggest we treat the `fieldsets` as lists of to-be-included relationships for now and either convert them to proper sparse field sets if we actually decide to switch to Orbit.js eventually or check again whether it's worth introducing the above addon (or just rename `fieldsets` to `includesets` or so ).

habdelra:
yeah, along a similar vein, @ef4 and I were chatting last week, and sparse fieldsets are also not implemented yet in our jsonapi plugin. In the interest of pushing this forward in a timely manner, we're gonna focus on a client-side only approach to this as a first pass on this feature; meaning we'll load all the fields for each relationship. Then we'll refine this over time to add sparse fields.

PR #338, opened 2018-10-03
#331 @cardstack/data load service

habdelra:
Addresses issue #331 

This is a first pass thru a client implementation of the cardstack/data service which provides a mechanism to load all the related models when retrieving data for a particular card in a specific format. Note that I have separated out the ember addon boiler plate code so feel free to ignore that commit when reviewing. 

Still need to think thru a bunch of items:
- [x] don't reload a record that you have already loaded
- [x] parallelize record loads that are not related to one another
- [x] more tests around `defaultIncludes` vs `fieldsets` clashing


packages/models/cardstack/code-generator.js
```diff
@@ -37,6 +37,7 @@ define('@cardstack/models/generated/{{modelName}}', ['exports', '@cardstack/mode
      value: true
    });
    exports.default = _model.default.extend({
+     type: "{{modelName}}",
```
ef4: This unlocks a bunch of other simplifications in places where we're using cs-model-type or model.constructor.name.

packages/data/addon/services/cardstack-data.js
```diff
@@ -0,0 +1,96 @@
...
+export default Service.extend({
+  store: inject(),
+
+  init() {
+    this._super();
+    this._loadedRecordsCache = {};
+    this._contentTypeCache = {};
+  },
+
+  async load(type, id, format) {
```
ef4: This API will eventually also want `branch` as an argument.

ef4:
This looks good to me.

packages/data/addon/services/cardstack-data.js
habdelra: I'll add that parameter in there now so it wont be a mess to deal with in the future

habdelra: or well... when we perform `store.findRecord()` how do we incorporate `branch` into that request? or is that not currently supported?

habdelra: Nevermind, found it: https://github.com/cardstack/cardstack/blob/master/packages/routing/addon/routes/cardstack/common.js#L32

habdelra:
@ef4 I'm thinking we should also add this service in the @cardstack/router 's commonModelHook.

habdelra:
or well I can do that in a subsequent PR. I'm seeing that this works really nicely in our own cardstack app.

PR #339, opened 2018-10-05
Moving git remote tests to docker

mansona:
This is to fix the issue that I introduced by adding tests that read from a github repo for the remote git datasource test. I will continue my work on the datasource in another PR but I figured it was useful to merge this sooner rather than later 👍 

We now have a docker image that we can use to run the git remote datasource tests against.

I don't know exactly how to publish this docker image so that people don't need to build it every time they need to use it 🤔


habdelra:
we've pushed a few things like this into docker hub

habdelra:
Here's a link https://docs.docker.com/docker-hub/repos/#pushing-a-repository-image-to-docker-hub

ef4:
We have a cardstack organization on docker hub. You need to make accounts before I can invite them to join the organization. Please do and I will add you.

mansona:
@ef4 I have created my account with my simplabs email address. Username chrissimplabs 👍

mansona:
@ef4 I have published cardstack/git-test now and this PR is no longer building its own docker images locally 👍 

I think we're good to review and merge 🎉

PR #342, opened 2018-10-05
Fix missing repository urls (for Ember Observer)

localpcguy:
Noticed a few @Cardstock/[name] repos missing their repository URLs on Ember Observer, this should fix that.

- models
- email-auth
- search
- edges

see: https://emberobserver.com/lists/invalid-repo-url


ef4:
Thanks!

PR #355, opened 2018-10-12
Spaces

habdelra:
@ef4 This is a first pass at a very simplistic backend for spaces. I separated it out into separate commits for ease of review. The commit you should pay special attention to is the last commit. The rest is refactoring from another PR that this PR is branched off of and plugin boilerplate.

The way I approached this was as introducing spaces as a new datasource. For the very simplistic first pass, the ID of the space is the `type`/`id` of the primary card. By using a datasource for spaces, I can create a searcher that is able to derive the space based on the ID, and hopefully in the future we can introduce more sophisticated logic on how a space derives its primary card and consumes segments of the passed in URL, like looking up routing rules from other cards in the index and/or routing rules defined in its datasource config (which could even be codegen'ed into something that the ember router can consume).

Additionally, I also introduced a new content-type attribute: `fieldset-expansion-format`. This allows us to specify that for a specific content type (i.e. `spaces` content type), it should always use a fieldset format when building the json:api. In this case the `isolated` format. This is how the client can ask for a spaces instance without having to add extra stuff to the API to tell the hub that it wants to expand the relationships using the `isolated` fieldset format.

I added some additional logic in the DocumentContext to honor this content-type attribute (which is actually pleasingly a small amount of logic). This also illustrates that the decision we made to not assume that all fieldset formats below an `isolated` format should be `embedded` was a good idea, as the spaces content type is processed using an `isolated` format, and it in turn needs to have its primary card also processed using an `isolated` format.

Hopefully this is the right direction for this, and adding logic into our hub routing to get data via spaces should work pretty naturally using this approach.


habdelra:
@ef4 the `fieldsetExpansionFormat` is used to indicate that the document for a resource should include all the relationships (recursively) for a particular format. So for a `spaces` resource, it's json:api will be embellished with all the relationships necessary for the `isolated` format. Otherwise the client would need to have some special way of telling the API that it wants to get a document for a particular format. simply using `include` is not enough as that doesn't consider the polymorphic nature of potential relationships

PR #357, opened 2018-10-15
Implements loading spinner after update version control

### LINKED ISSUE #346: Show some sort of status to the user while a card is being created or updated
Now that we have a remote git-datasource that does not respond until the data is committed remotely it is very clear that we need to give some feedback to the user when creating or updating a card.

In my personal testing I have seen a case of multiple seconds before the network responds but there is currently no `loading` message once you submit the form in the tools.

jorgelainfiesta:
Fixes #346


ef4:
This is nice, thanks.

PR #358, opened 2018-10-16
Using spaces within routing for more efficient relationship loading

habdelra:
Depends on PR https://github.com/cardstack/cardstack/pull/355. This PR optimizes the data loading int he hub generated routes so that a single API call is issued in order to get the graph of all relationships for the primary-card in a space.

### DIFF
--- a/packages/routing/cardstack/searcher.js
+++ b/packages/routing/cardstack/searcher.js
@@ -1,30 +1,61 @@
-module.exports = class SpacesSearcher {
+const { declareInjections } = require('@cardstack/di');
+
+module.exports = declareInjections({
+  searchers: 'hub:searchers',
+  schema: 'hub:current-schema',
+},
+
+class RoutingSearcher {
   static create(...args) {
     return new this(...args);
   }
-  constructor(opts) {
-    let { dataSource } = opts;
+  constructor({ dataSource, searchers, schema}) {
     this.dataSource = dataSource;
+    this.searchers = searchers;
+    this.schema = schema;
   }
 
   async get(session, branch, type, id, next) {
     if (type === 'spaces') {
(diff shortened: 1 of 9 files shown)


ef4:
Does this actually embed all the needed cards in the response? I don't see where that happens. The searcher loads the primaryCard but then only responds with its type and id.

habdelra:
it does. The fact that this test still passes demonstrates that it does: https://github.com/cardstack/cardstack/pull/358/files#diff-62dc4706b86fe924e4e73c82d682566cR97 The `fieldsetExpansionFormat` you were asking me about in the other PR is responsible for this.

habdelra:
(or at least the tests used to pass. fixing them...)

habdelra:
weird, the CI tests passed in travis but github is still showing them as "pending"....

packages/routing/cardstack/searcher.js
```diff
@@ -34,11 +65,11 @@ module.exports = class SpacesSearcher {
         attributes: { 'url-segment': `/${cardType}/${cardId}` },
         relationships: {
           'primary-card': {
-            data: { type: cardType, id: cardId }
+            data: { type: card.type, id: card.id }
```
ef4: My one remaining concern is here.

At this point, we just went through the trouble of loading `card`. And it may have a ton of `included` cards already embedded inside it, if we've done a good job optimizing the index.

But we throw away everything other than `type` and `id`, which will force the `DocumentContext` to reload our card and any other includes.

I think we can build a better response here that `includes` the `card` as well as anything that was included inside it.

habdelra: ah, i see what you mean. cool thanks for the feedback.

PR #363, opened 2018-10-18
Scrollable create new panel

jorgelainfiesta:
Fixes https://github.com/cardstack/deck/issues/157


marcoow:
> Fixes cardstack/deck#157

Technically this alone doesn't fix it but we'd also have to up the dependency in cardstack/deck…

jorgelainfiesta:
Good point, I've updated the description. I'm not sure how releases work, but after the next `tools` release that includes this fix I'll update the dependency in `deck`.

PR #366, opened 2018-10-19
[CONCEPT] Mailer route

jorgelainfiesta:
For the "Contact Us" card, it was agreed that we would use the email plugin to send an email every time a person completed the form. However, `@cardstack/email` is a low-level interface, consumed mainly by `email-auth` through a `messenger`. 

This PR implements a simple middleware that exposes a new route on the server which sends an email using a `messenger`. It relies on `plugins-configs` to be set for `@cardstack/email` and a JSON-API payload.

Open Questions:

- [ ] Is this a correct implementation of the intended feature or is it quite hacky? 
- [ ] Is a generic email middleware appropriate to use for the contact us form, or should a more specific plugin be implemented? 
- [ ] Ed was mentioning we could implement this as a data-source. For the writing part: I assume we send an email everytime we get a write operation. But what would reading this data-source mean?

### DIFF
--- a/packages/email/middleware.js
+++ b/packages/email/middleware.js
@@ -1,0 +1,44 @@
+const { declareInjections } = require('@cardstack/di');
+const route = require('koa-better-route');
+
+module.exports = declareInjections({
+  messengers: 'hub:messengers'
+},
+
+class CodeGenMiddleware {
+  constructor() {
+    this.before = 'authentication';
+  }
+
+  middleware() {
+    //TODO: Get these from config
+    let messageSinkId = 'the-sink';
+    let defaultMailTo = 'contact@cardstack.com';
+
+    return route.post('/email/send', async (ctxt) => {
+      let { req } = ctxt;
+      let body = '';
+      req.on('data', chunk => {
+        body += chunk.toString();
+      });
+      req.on('end', () => {
+        console.log(body);
+        res.end('ok');
+      });
+      
+      await this.messengers.send(messageSinkId, {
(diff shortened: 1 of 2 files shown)


marcoow:
The general approach makes sense to me. However, as of cardstack/deck#198, I think we should be able to move the middleware into the `contact-page` card relatively easily?

habdelra:
@marcoow that is _definitely_ the right approach. I'm so happy that that refactor has immediately usefulness 🎉

ef4:
Yes, I think it's always fine to add features directly to the cards that need them. Just like writing an Ember addon is a better choice than trying to get every feature directly into ember, even if the longer-term goal is to standardize and bring it into core.

I do think this kind of feature belongs in the core framework, but yes, I would prefer it as a `writer`, not a `middleware`. I don't know any reason that a write-only data source wouldn't work, though we haven't made one yet.

(It's also possible that "messengers" stop needing to be a separate concept, and they can be modeled as write-only data sources.)

marcoow:
So the plan is to implement this as a write-only data-source in the card and then potentially move it into hub in some form in a follow-up step?

ef4:
I meant you can ship whatever you want in the card, so if a middleware works and makes sense, it’s fine to go with that. 

Basically the level of experimentation and decoherence can be much higher in a specific card than in the shared framework.

jorgelainfiesta:
I'm closing this PR because the solution will not be in this fashion in cardstack. I'll open a similar PR to add a middleware to the contact us card.

PR #367, opened 2018-10-19
add useful test helpers that cards can leverage for testing

habdelra:


ef4:
Right now there's nothing to prevent this code from getting included in production builds. 

The standard thing to do is to rename `addon` to `addon-test-support`, and those files are importable under `your-addon-name/test-support`. That will look a little silly for this addon: `@cardstack/test-support/test-support`.

There's a Known Hack™️ to avoid the extra `test-support` dir: https://github.com/kaliber5/ember-window-mock/blob/ada959b92f104c875066140be3c3ca7f34110f7d/index.js#L8

habdelra:
Cool thanks! I’ll fix that

PR #371, opened 2018-10-21
handle indexing a resource that has a relationship to itself

habdelra:


ef4:
Thanks, good catch.

There may still be a case that this doesn't handle: a cycle that doesn't include the topmost document. 

`this._references` seems like it's already the right state for detecting circularity. If a document is in there, we have already crawled it.

habdelra:
i'll test for that as well

habdelra:
@ef4 `this._references` almost works for detecting circularity, but I did find a test where it is too eager. Specifically the test that you recommended where the cycle does not involve the root resource: https://github.com/cardstack/cardstack/pull/371/files#diff-6abe8eb403a274232d666329802a9485R335

It mostly just comes down to the order in which things occur, so I used a separate field to track the relationships that we have followed.

PR #372, opened 2018-10-21
Run the find-bad-deps script in CI to ensure bad deps don't get committed to the master branch

alexspeller:
Already found one in master :)

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,8 @@
   "engines": {
     "node": ">= 7.7"
   },
-  "devDependencies": {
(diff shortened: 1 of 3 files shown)


ef4:
GitHub still seems to be behind. I pushed [an additional commit to this branch](https://github.com/cardstack/cardstack/commit/bb0c38eba23410e3fccb00af507c4c8c05b932d0), but I'm not seeing it yet here and CI has not triggered on either mine or your latest.

PR #374, opened 2018-10-22
validate ephemeral models when hub is launched in non-production environment

habdelra:


packages/hub/main.js
```diff
@@ -53,6 +50,33 @@ async function wireItUp(projectDir, encryptionKeys, dataSources, opts = {}) {
   return container;
 }
 
+async function startIndexing(environment, container) {
+  // some datasources are dependent upon a sync at boot for index of pristine system
+  await container.lookup('hub:indexers').update();
+
+  if (environment !== 'production') {
+    let ephemeralStorage = await container.lookup(`plugin-services:${require.resolve('@cardstack/ephemeral/service')}`);
```
ef4: I think this will throw if you don't happen to have ephemeral installed?

packages/hub/main.js
```diff
@@ -53,6 +50,33 @@ async function wireItUp(projectDir, encryptionKeys, dataSources, opts = {}) {
...
+    let searchers = await container.lookup(`hub:searchers`);
+    let controllingBranch = await container.lookup(`hub:controlling-branch`);
+    let models = await (await container.lookup('config:initial-models'))();
+
+    await ephemeralStorage.validateModels(models, async (type, id) => {
+      let result;
+      try {
+        result = await searchers.get(Session.INTERNAL_PRIVILEGED, controllingBranch.name, type, id);
```
ef4: There is a `searchers.getFromControllingBranch()` so you don't need to lookup `hub:controlling-branch` yourself.

packages/hub/main.js
habdelra: good point

packages/hub/main.js
habdelra: nice. thanks!

habdelra:
@ef4 ok, I've resolved the issue that you raised. hopefully this is ready for merging now.

PR #380, opened 2018-10-25
dont crawl duplicate modules when loading plugins

habdelra:


packages/hub/plugin-loader.js
```diff
@@ -114,22 +115,23 @@ class PluginLoader {
 
   async _crawlPlugins(dir, outputPlugins, seen, includeDevDependencies, breadcrumbs) {
     log.trace("plugin crawl dir=%s, includeDevDependencies=%s, breadcrumbs=%j", dir, includeDevDependencies, breadcrumbs);
-    if (seen[dir]) {
-      if (seen[dir].attributes && seen[dir].attributes.includedFrom) {
+    let realdir = await realpath(dir);
+    let packageJSON = path.join(realdir, 'package.json');
+    let json = require(packageJSON);
```
ef4: I think this eliminated an optimization -- we're going to re-parse the package.json for directories we've already seen, where before we hit the early `return`and skipped that.

habdelra:
make sense. will do

packages/hub/plugin-loader.js
habdelra: thanks, i'll fix that

habdelra:
Actually I wonder if throwing will continue to make it just as brittle as it is now. Keep in mind that when we have minor misalignments in @cardstack/* versions, it can cause hundreds of duplicate modules to be loaded, which can cause healthcheck failires on start up. How about a warning message instead, so the hub can be resilient, but still be transparent that not everything is ok.

ef4:
Ok, but lets make the warning an error when it happens inside tests.

habdelra:
sure that's fine. to clarify, when you say "error" do you want to throw and error, or log at an error level?

ef4:
Throw an error. Something that will make the tests fail.

habdelra:
@ef4 when running our ember-tests in this repo, I'm running into dupe modules being loaded like this: ```Error: The plugin module name '@cardstack/codegen' has already been loaded from the module path /Users/hassan/codez/cardstack/packages/codegen, conflict with load of module at path /Users/hassan/codez/cardstack/node_modules/@cardstack/codegen.``` (and because its the test env, we blow up). How do you think I should handle these?

ef4:
I think one of those paths is a symlink to the other? In that case we should run them through fs.realpath before comparing.

habdelra:
cool--thanks! I'll give that a shot

PR #381, opened 2018-10-26
Use exact search for content-type by default

### LINKED ISSUE #302: Query to fetch all instances of a type also returns related types in `data`
As realized in https://github.com/cardstack/deck/pull/91#discussion_r217103524 , a query to fetch "all"  instances of some types returns not only instances of that type but those of the `defaultIncludes` ones, too.

For example, a call to `GET /api/documents?page%5Bsize%5D=1000` returns document-chapters in the `data` key, too, and `document-chapters` can be found in the `defaultIncludes` definition of the `document` type.

bagby:
You can pass in a `fuzzy` query param if you want to use fuzzy content-type search, i.e. `/api/articles?fuzzy=true`

Fixes #302

### DIFF
--- a/packages/jsonapi/middleware.js
+++ b/packages/jsonapi/middleware.js
@@ -112,7 +112,7 @@ class Handler {
     return this.prefix === 'api-validate';
   }
 
-  filterExpression(type, id) {
+  typeFilterExpression(type, id) {
     let filter = this.query.filter;
     if (!filter) {
       filter = {};
@@ -121,7 +121,9 @@ class Handler {
       // This overwrites any additional type filter the user tried to
       // provide in their query, which is reasonable (the endpoint
       // name itself is taking precedence).
-      filter.type = type;
+      filter.type = {
+        exact: type
+      };
     }
     if (id != null) {
       filter.id = id;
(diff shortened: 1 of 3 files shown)


bagby:
@ef4 this seems fine to me, but I wouldn't mind a quick review.

ef4:
Thanks. Making the automatic type filter be exact is good and I want to merge that part.

But I would prefer to drop the new `?fuzzy` query param, it doesn't really work like any of the other operators. I want to implement a `fuzzy` that has the same syntax as today's `exact`, so you can pick which field you're making fuzzy. 

At which point you could do `GET /api?type[fuzzy]=articles` to get the effect you've implemented here. But you could also do more complex things, like `GET /api?type[0]=articles&type[1]=posts&body[fuzzy]=my+keywords`

bagby:
@ef4 I removed the `fuzzy` thing, it was sort of an afterthought anyway.

habdelra:
@bagby so I'm having difficulty using the fuzzy search in the way that @ef4 is describing above for the search-results card in deck. I'm issuing queries like: `http://localhost:3000/api/video?title%5Bfuzzy%5D=chris` to "fuzzy" search for the word "chris" in the title field of a video, but the tests are not passing. Can you point me to a test showing how the fuzzy search works now, since it's an opt in process?

ef4:
@habdelra the thing I suggested is not implemented yet, it would be great if you want to work on it. Right now all the filters are still fuzzy by default, so `http://localhost:3000/api/video?title=chris` is fuzzy.

Issue #298, opened 2018-09-17
Allow primary interaction with fields in edit mode


balinterdi:
When some markup is wrapped in a `cs-field` and we're in edit mode, an overlay is rendered above the markup. That's exactly what we want in most cases, but in some cases the underlying markup has event handlers that cannot be triggered because of this.

The example where this came up is the Table of Contents for a document being wrapped in a cs-field which prevents clicking on each chapter title to load the related chapter content on the UI.

![screen shot 2018-09-17 at 10 45 44](https://user-images.githubusercontent.com/5022/45613503-e9fe7c80-ba66-11e8-8e01-786d00fac0f5.png)

To make this possible, we should mark such fields in some way (`allowHandlers`). When such a field is first interacted with in Edit mode, it should allow the interaction "to go through" (by setting `pointer-events: none` on the overlay. A subsequent 2nd click should trigger the current behavior, bringing up the editor for that field.

(See https://github.com/cardstack/deck/pull/91#issuecomment-420333736)

balinterdi:
> To make this possible, we should mark such fields in some way (allowHandlers). When such a field is first interacted with in Edit mode, it should allow the interaction "to go through" (by setting pointer-events: none on the overlay. A subsequent 2nd click should trigger the current behavior, bringing up the editor for that field.

@ef4 Is this how we want it, or the other way around: the 1st click should focus the editor and the second, subsequent click should allow the interaction to go through?

ef4:
When a field is selected, its overlay should either render the inline-field-editor (if there is one) or set itself to `pointer-events: none` so that the underlying content is accessible.

To make this distinction clearer, we should re-style the hover state so it has a semi-transparent background. 

An example interaction would be:

1. Hover over some content. That causes an overlay to appear, which has a semi-transparent fill making it clear that you can't (yet) interact with the content behind the overlay. (In this step, the only difference from the current behavior is the semi-transparent fill.)

2. Click on the overlay. That causes the field to be selected for editing, which does all the things it does now (like opening the field-editor on the right edge). Assume the field in question has no inline-field-editor. In that case, the overlay now becomes `pointer-events: none` so that the content is interactive.

balinterdi:
> 1. Hover over some content. That causes an overlay to appear, which has a semi-transparent fill making it clear that you can't (yet) interact with the content behind the overlay. (In this step, the only difference from the current behavior is the semi-transparent fill.)

[...]

PR #398, opened 2018-11-01
Allow interacting with fields in Edit mode

balinterdi:
See #298.

### DIFF
--- a/packages/tools/addon/components/cs-field-overlays.js
+++ b/packages/tools/addon/components/cs-field-overlays.js
@@ -5,7 +5,6 @@ import { pluralize } from 'ember-inflector';
 import { get } from '@ember/object';
 import Component from '@ember/component';
 import layout from '../templates/components/cs-field-overlays';
-import { equal } from '@ember/object/computed';
 
 export default Component.extend({
   layout,
@@ -15,17 +14,8 @@ export default Component.extend({
   cardstackRouting: service(),
   router: service(),
 
-  selectedOverlayState: null,
-  isOverlayOpen: equal('selectedOverlayState', 'open'),
-  isOverlayPassthrough: equal('selectedOverlayState', 'passthrough'),
-
   actions: {
     openField(which) {
-      if (!which) {
-        this.set('selectedOverlayState', 'closed');
-      } else {
-        this.set('selectedOverlayState', this.isOverlayOpen ? 'passthrough' : 'open');
-      }
       this.get('tools').openField(which);
     },
 
--- a/packages/tools/addon/components/cs-field-overlays.js
+++ b/packages/tools/addon/components/cs-field-overlays.js
@@ -24,7 +24,6 @@ export default Component.extend({
       if (!which) {
         this.set('selectedOverlayState', 'closed');
       } else {
-        //TODO: Fields with inline-editors cannot be set to passthrough
         this.set('selectedOverlayState', this.isOverlayOpen ? 'passthrough' : 'open');
       }
       this.get('tools').openField(which);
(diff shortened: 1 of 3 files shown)


packages/tools/addon/styles/cs-field-overlays.css
```diff
@@ -1,20 +1,31 @@
...
+.cs-field-overlay.passthrough .overlay-border {
+  border-color: var(--tools-highlight);
+  background-color: transparent;
+}
+
+.cs-field-overlay.passthrough {
+  pointer-events: none;
+}
+
+.cs-field-overlay.passthrough .overlay-scrim {
```
balinterdi: @ef4 As the `.overlay-scrim`s  are children of the `cs-field-overlay` and we set `pointer-events: none` on the latter, I couldn't click on any of the scrims (=anywhere on screen) to dismiss the overlay since that css rules applied to them, too.

That's why I added this rule but I'm not wholly satisfied with it as it crosses boundaries between add-ons (`@cardstack/tools` and `ember-overlays`) and thus regressions are easier to introduce.

ef4:
Oh, I wasn't expecting there to be an explicit option for "passthrough". I was assuming it should just always be true for any field that doesn't have an inline field editor component.

Is there a reason why we wouldn't always want that?

packages/tools/addon/styles/cs-field-overlays.css
ef4: Let's change this selector to `.cs-field-overlay.passthrough > *`. That is the actual intent, I think.

balinterdi:
@ef4 Unless I misunderstand you, there's no explicit option for passthrough. Which code snippet makes you think that?

packages/tools/addon/styles/cs-field-overlays.css
balinterdi: The problem with that is that the overlay scrims (`.overlay-scrim`) are not direct descendants of the `.cs-field-overlay` because they are wrapped in a `liquid-if` (see [here](https://github.com/ef4/ember-overlays/blob/master/addon/templates/components/create-overlay.hbs#L16-L21))

ef4: But I think pointer-events is inherited. So if we disable them for .cs-field-overlay and then re-enable them on each direct descendant, only `.cs-field-overlay` itself will have them disabled.

packages/tools/addon/components/cs-field-overlays.js
```diff
@@ -14,10 +15,20 @@ export default Component.extend({
...
+  selectedOverlayState: null,
+  isOverlayOpen: equal('selectedOverlayState', 'open'),
+  isOverlayPassthrough: equal('selectedOverlayState', 'passthrough'),
+
   actions: {
     openField(which) {
+      if (!which) {
+        this.set('selectedOverlayState', 'closed');
+      } else {
+        this.set('selectedOverlayState', this.isOverlayOpen ? 'passthrough' : 'open');
```
ef4: I don't think we should keep this state (`selectedOverlayState`) in this component. It's redundant with state elsewhere, which means it's at high risk for bugs. I also don't think it's supposed to be shared by every overlay, which is the current effect.

Instead, simplify to:

```
class=(concat "cs-field-overlay "
              (if hasInlineEditor "with-inline-editor ")
              (if (eq mark.id tools.openedFieldId) "opened"))
```

Then you can do something like:

```css
.cs-field-overlay.with-inline-editor.opened {
  pointer-events: none;
}
```

balinterdi:
@ef4 Thank you. I've made the changes you suggested so feel free to merge if you think it's now good.

PR #402, opened 2018-11-05
Upgrade `nodegit` to v0.23.0

Turbo87:
This PR is upgrading the `nodegit` dependency and "unforks" it. Let's hope this works... 🤞

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,13 @@ sudo: required
 services:
  - docker
 
+addons:
+  apt:
+    sources:
+      - ubuntu-toolchain-r-test
+    packages:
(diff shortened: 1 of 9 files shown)


Turbo87:
applying the instructions from https://github.com/nodegit/nodegit/issues/853#issuecomment-172116071 (and their current README) seems to have resolved the original issue on CI without any negative effects on CI time.

is this sufficient or do we need to test anything else?

ef4:
The only possible problem I see is that apps may also now need to make the same kind of adjustment you made in travis.yml. I think when we integrate this with the deck app, for example, we will need to update its Dockerfile to add these libraries.

PR #405, opened 2018-11-05
Update `ember-simple-auth` to v1.7.0

Turbo87:
This removes the remaining nested dependencies on `ember-cli-babel@5`


ef4:
I will be glad not to see that deprecation message anymore. 😀

PR #407, opened 2018-11-05
allow the ember test runner to run against arbitrary folder so that it can be re-used by cardstack applications to test cards

habdelra:


packages/test-support/ember-test-runner.js
```diff
@@ -29,16 +29,16 @@ function testPackage(package) {
   });
 }
 
-async function run() {
-  let packages = glob.sync(path.join(__dirname, '..', '*', 'ember-cli-build.js')).map(p => path.dirname(p));
+async function run(cardsDirectory='cards') {
+  let packages = glob.sync(path.join(__dirname, cardsDirectory, '*', 'ember-cli-build.js')).map(p => path.dirname(p));
```
Turbo87: it seems a little strange that `..` is relative to the file path instead of being relative to cwd 🤔

habdelra: yeah, I was musing about the same thing.... we should update this...

Issue #300, opened 2018-09-17
Render all fields, not just those rendered on the page


balinterdi:
Models might have fields that we also want to render in the sidebar in Edit mode even if they were not explicitly marked for editing in the template (by wrapping them in a `cs-field` or a `{{content.field}}`).

Marking a field as `sticky` (other name suggestions: `pinned`, `editorial`) means that field is always rendered for editing, no matter if it shows up on the page. Care should be taken not to render such fields twice (which can be the case if it's also rendered explicitly in the template).

ef4:
This is indeed the solution we discussed, though I am beginning to wonder if we should just always have a section in the sidebar to show _all_ the fields that are present but weren't rendered in the template.

Perhaps all we really need is an intelligent way to order and/or organize the fields, so that the ones relevant to the template appear first, but you can still find the others further down.

balinterdi:
I think that would make sense though I wonder what the original intent behind only showing the ones rendered on the page was. For example,

>  if we should just always have a section in the sidebar to show all the fields

would mean that we show all related fields of all associated records in "owned" relationships. If a resource has n related resources and each related resource has m fields, that would mean n*m fields in the sidebar which can be pretty overwhelming (cluttered). In the case of documents and chapters, that would mean all chapter titles and bodies are listed (not to mention images which I think would be another owned relationship).

However, we have to provide a way to edit fields in owned relationships as by their definition they cannot be edited in their content-type (as they don't have an isolated view).

I think two concepts are currently coupled which should not be: the inclusion of a field in the editor sidebar and rendering that field in the page. One idea to decouple them and solve the above problem is to introduce a new CS helper/component that would only include the fields in the editor, without rendering them, something like:

```handlebars
{{#cs-editor-fields-for content as |documentFields|}}
  {{documentFields 'thumbnail'}}
  {{documentFields 'headerBackgroundImage'}}
  {{documentFields 'attachments'}}
{{/cs-editor-fields}}

{{#cs-editor-fields-for selectedChapter as |chapterFields|}}
  {{chapterFields 'readingTime'}}
{{/cs-editorFields}}
```

We could also just wrap them in a div that we set to `display: none` but that would feel hacky, and we might be able to add other functionality in the `cs-editor-fields-for` helper.

ef4:
[...]

PR #408, opened 2018-11-05
Show all fields in sidebar, not just rendered ones

balinterdi:
See #300 

Todos:

* [x] ~~Consider DRYing the rendering of `renderedFields` and `modelFields`~~


balinterdi:
@ef4 This implements Phase 1 of what we discussed over at https://github.com/cardstack/cardstack/issues/300 .

No fields are rendered twice and rendered fields come before "model" fields (fields on the model that are not rendered on the page) in the editor.

PR #409, opened 2018-11-05
adding div that ember-power-select requires (as part of core-types) to template for edges

habdelra:


ef4:
The edges package supports `registerTopLevelComponent` so that this kind of thing can live in the package that needs it (core-types).

habdelra:
cool, thanks! I'll update this

PR #413, opened 2018-11-06
Upgrade `ember-cli-postcss` to v4.1.1

Turbo87:
This PR follows #404 and upgrades `ember-cli-postcss` to the latest release, which includes Node 10 compatibility.


Turbo87:
@ef4 I'm not sure where this postcss stuff is actually being used in the end. Is there some way to manually test this change? 🤔

ef4:
> I'm not sure where this postcss stuff is actually being used in the end.

It's giving us things like CSS variable support in the stylesheets that go in the tools and core-types packages. If anything breaks we will notice when we notice -- the visual design for all that stuff is in the process of being redone.

PR #414, opened 2018-11-06
Update `fsevents` subdependency to v1.2.4

Turbo87:
... to make it compatible with Node 10


Turbo87:
since I didn't see any objections I'm assuming this is good to go

Issue #416, opened 2018-11-06
Make sidebar editor grant aware


balinterdi:
One case that would solve is that computed fields would not be a special case – they wouldn't have write grant and thus their value couldn't be edited.

balinterdi:
Duplicate of #448, closing.

PR #421, opened 2018-11-07
Make sidebar editor grant aware

balinterdi:
Implements #416 

* [x] Merge https://github.com/cardstack/cardstack/tree/permissions-searcher into this one
* [x] ~~Do the front-end parts: requesting permissions from back-end, enrich editor with permissions loading state, disable fields (inputs, dropdown-choices, etc.) that don't have a write 
permission~~
* [x] Add special handling of getting the grants for permission resources
* [x] Fix current solution for creating new resource (it now only works for update). The acceptance test in `@cardstack/image` fails because of that
* [x] Write acceptance test for tools dummy app
* [x] ~~Permissions for fields in owned relationships are not fetched and thus marked as read-only in the editor sidebar~~ (see below)
* [x] Fix regression: rendered fields which do not belong to the resource-under-edit (RUE) are not editable as permissions are only queried and returned for the RUE
* [x] ~~Indicate editor is loading permissions(?)~~
* [x] Add error handling for permissions searcher (see todos in code)
* [x] Finalize getting `writable-fields` (maybe it's just a matter of confirming everything works)
* [x] Fix tests


balinterdi:
@ef4 What I bumped into is that after the permissions searcher retrieves the document correctly, `applyReadAuthorization` throws a 404 because the permissions resource itself cannot be read. It's this line that throws:

https://github.com/cardstack/cardstack/blob/master/packages/hub/schema/content-type.js#L242

The permissions resource only has a `groups/@carstack/hub` real that allows reading it and none of the user's realms have it.

I hacked around it for the time being so that I can continue building the front-end but I wonder how to add a read grant to the permission resource (and which group to add it to).

balinterdi:
Probably a band aid but that's what I came up with in the meantime:

https://github.com/cardstack/cardstack/pull/421/commits/4d808d2ebfb19c9d16cb8d63aa4f727a204e3316#diff-51623a8a8654d57b6a248d3c38c411da

ef4:
Hmm. Yeah, I suppose the grants that govern access to `permissions` itself might need to be special in order to not be annoying.

I think you should be able to read a `permissions` resource as long as you have `mayReadResource` on the corresponding document.

balinterdi:
@ef4 Is this an acceptable way to implement that special case?
https://github.com/cardstack/cardstack/pull/421/commits/ea6b0f17269d43567c69d6cd5f2e8367faf77534

balinterdi:
@ef4 So I think making permissions work for fields in owned records is a long shot away. Let me explain.

For a field to be editable in the editor, it has to be 1) rendered in the editor (duh) and 2) have a writable permission in `writable-fields` (and also a resource-level `may-update-resource`).

Issue #408 adds un-rendered fields to the editor of the resource being edited *but not of the owned related resources as we agreed that it's a good phase 1 implementation.

So even if I extended the permissions part to include field names of related content types, it wouldn't make a difference because the fields which those permissions relate to are not rendered in the editor.

What I suggest is that I work on Phase 2 of #408 first and then come back and implement Phase 2 of this issue.

Does that seem a good path to you?

balinterdi:
As https://github.com/cardstack/cardstack/issues/448 will take more time because we also need to come up with a design, I thought about how to have this merged without introducing the regression of not being able to edit rendered fields on owned resources (see my previous comment about why that would be the case).

I think a simple and acceptable solution would be to not disable the field if it's for another resource than the RUE. The server-side will catch any disallowed writes anyways which is the current state of affairs.

balinterdi:
@ef4 If you agree with my above comments, this is yours for review and merge.

packages/hub/searchers.js
```diff
@@ -73,8 +73,10 @@ class Searchers {
         upstreamDoc: { data: resource, meta },
         read: this._read(branch)
       });
-      let pristineResult = await (documentContext.pristineDoc());
-      authorizedResult = await schema.applyReadAuthorization(pristineResult, { session, type, id });
+      let pristineResult = await documentContext.pristineDoc();
+      if (pristineResult) {
+        authorizedResult = await schema.applyReadAuthorization(pristineResult, { session, type, id });
```
ef4: This makes me see a refactor we should do.

`applyReadAuthorization` would be simpler if we moved it out of Schema and into DocumentContext.

The DocumentContext already has access to schema, session, type, and id, and it's already configured with the ability to `read` other documents, which would mean we could eliminate the new injections you needed in Schema.

balinterdi: We could do that but one of the auxiliary methods, `_readAuthorizationForPermissions` uses `this.searchers` and `this.controllingBranch` that are accessed via DI and I don't think are available in DocumentContext. So they'd have to be injected there, too.

https://github.com/cardstack/cardstack/pull/421/files#diff-ed3aae8cfdd1353fe73e9430bea36cd8R476

Not saying I cannot do that, I'm just still not very knowledgeable about what should go where so I'm bringing this up because maybe the refactoring loses some of its rationale (its raison d'être, if I want to be a snob).

balinterdi: The other tasks you pointed out are completed.

balinterdi: Oh, wait, I glossed over one of your sentences ("and it's already configured with the ability to read other documents"), I'll take a look at how that works.

PR #423, opened 2018-11-07
Routing feature

habdelra:
This PR introduces the routing feature for our cards. The documentation below has been added to the documentation site: https://github.com/cardstack/cardstack-documentation/pull/3

The idea is that a card can define a series of routes that it supports, and that the routes can cascade, such that if a card's router routes a path to a particular card, that card can then in turn route to another card. We use the term "routing-card" for a card that has a router. And specifically the top level routing card is called the "application card".

To support this capability, cards can specify a router in their `<card module folder>/cardstack/router.js` file. An example `router.js` looks like this:

```js
module.exports = function (routingCard) {
  return [{
    // Note that because this query does not use `:id` nor `:friendly_id`, we would never 
    // use this route as a canonical path for a card in the API response.
    path: '/latest-article',
    query: {
      filter: {
        type: 'articles'
      },
      sort: '-publish-date'
    }
  },{
    // This is an example of a using a "slug" or human friendly ID in the URL to identify a card
    // as opposed to the actual card's ID (which may not be easy for a human to remember or type).
    path: '/good-reads/:friendly-id',
    query: {
      filter: {
        type: 'articles',
        slug: { exact: ':friendly_id' }
      }
    }
  },{
    // This is an example of using a query param in the path to make a routing decision.
    // Note that because this query does not use `:id` nor `:friendly_id`, we would never 
    // use this route as a canonical path for a card in the API response.
    path: '/most-popular?since=:date',
    query: {
      filter: {
        type: 'articles',
        'publish-date': {
          range: { gt: ':date' }
        }
      },
      sort: '-popularity'
    }
  },{
    // This is an example of a static mapping route, where the type and ID provided in the URL
    // map directly to the type and ID of the card to route to.
    path: '/:type/:id',
    query: {
      filter: {
        type: ':type',
        id: { exact: ':id' }
      }
    }
  },{
    // This is an example of a vanity URL, where you have a URL that points to a particular card.
    // In this example we use the data of the "routing card" to fashion the URL, which happens
    // to be the routing card itself.
    path: '/',
    query: {
      filter: {
        type: routingCard.data.type,
        id: { exact: routingCard.data.id }
      }
    }
  }];
};
```

The router is leveraged when the client makes a request to get a `space`. A `space` is retrieved by URL. So if the client wants to enter the route `/latest-article`, it makes a request to the API: `GET https://<hub api domain>/api/spaces/%2Flatest-article`. The API then uses the router above to match the path to a particular route whose query will be used to find the card to display to the user. The router will match the path in the order that the routes appear in the router, so the first pattern that matches the path is the one that will be used in the case where potentially two routes match match a path.

Each Cardstack application has an "application card" that is the top level routing card for the cardstack application. Any card can serve as the application card in the cardstack application. If no application card is specified, the hub uses a default, "Getting Started", application card. Additionally if the application card does not specify a router, the hub provided a default router (`@cardstack/routing`) that includes a static mapping route and a vanity URL of "/" to the application card itself. Cardstack applications can specify their application cards in the `plugin-configs/@cardstack-hub` configuration. 

In order to specify the name of the router to use for the application card, the cardstack developer can specify an attribute on the application card's content-type `router`. This is set to the feature name of the feature whose router that you want to use. The feature name is the npm module name of the feature. So typically this means that we would specify the npm module name of the card in its content-type's router attribute if we want the card to have a router.

```js
// <card-module-folder>/cardstack/static-model.js for a card that has npm module name of "@acme-corp/acme-application"
// the corresponding router would live in <card-module-folder>/cardstack/router.js within the same npm module

factory.addResource('content-types', 'acme-applications')
  .withAttributes({ router: '@acme-corp/acme-application' })
  .withRelated('fields', [
    factory.addResource('fields', 'application-name').withAttributes({
      fieldType: '@cardstack/core-types::string'
    })
  ]);

```

Additionally, when the router is unable to find a card for the provided path, it will return an error card. A system error card is provided. Cardstack application developers can provide their own custom error cards by creating a content-type that uses an `*-errors` suffix of the routing card's content type. So if the application card's content-type is `acme-applications`, a custom error card for this application card would be the content type of `acme-applications-errors`, and you should have at least one instance of this type with the ID of 'not-found' that is returned when the router cannot find a card for the path provided (this can be established in the `static-model` feature of your custom error card). You can then provide a custom template for this card in the same way a you provide for any card. You should also set a world read grant for this content type, so that errors are not inadvertently hidden for API clients.

When a routing card triggers an error because a card cannot be found for the path, the error will bubble up through all the routing cards that where involved with the routing of the path. The custom error card that is closest to the router that was unable to resolve the path will be the error card that is returned from the server.

Additionally in this PR we have updated the `DocumentContext`, such that if `meta.routing-card`, `meta.routing-id` is provided (as well as hub:plugins so the router can search for features) DocumentContext will add `links.self` to all the resources in the json:api document that it constructs. These are the canonical paths for the resources based on the specified routing card's router. The way that we derive a canonical path for a resource is to introspect the routing card's router and look for (in the following order):

1. A query that matches the specified resource specifically. This is a vanity URL to the resource, and has the highest precedence when deriving a canonical path to the resource.
2. A query that can match the specified resource using a `:friendly_id` in the router. The `:friendly_id` is a special replacement tag in the router that is the developer's signal to the hub that the field should be considered as a identifier for the content. This is what we have previously called the "routing field". This has the second highest precedence when deriving a canonical path to the resource.
3. A static mapping route, where the type and ID are specified as replacement tags in the path, and are used in the query to find a specific resource by type and ID.

The resulting `links.self` property is used by the ember client in order to transition to a particular card's canonical path. Note that it is possible for a router to be fashioned that results in not all cards being routable. In such cases, the `{{cardstack-url}}` helper and various tools that depend on getting the route for a card will be unable to operate for non-routable cards.

The `GET https://<hub api domain>/api/spaces/<path of card>` spaces json:api response has an attribute `query-params` that represents the unconsumed (not used by the router to make a routing decision) query params. These query params are then fed to the resulting card component with the `params` property. Card components can use the `params` property to access their query params. For query params that are used by the router, for example the route `/most-popular?since=:date` in the example router above, these query params must be name-spaced in the actual URL in order for the router to identify and consume the `date` query param. In this example, an actual URL for this route would look like:
```
https://<application domain>/most-popular?since=acme-applications[date]=2018-01-01&highlight-terms=foo
```
where `acme-applications` is the content-type of the routing card that the query param pertains to. The query parameter `highlight-terms` is not specified in the router and as a result is not consumed by the router, and rather, passed to the resulting card. The `space` for this particular path would specify a `query-params` attribute of `"?highlight-terms=foo"`, which would then be passed to the `articles` card's component as the object `{ 'highlight-terms'='foo' }` in the `params` property of the `articles` card component.


packages/hub/plugin-loader.js
```diff
@@ -133,12 +134,13 @@ class PluginLoader {
     let packageJSON = path.join(realdir, 'package.json');
     let json = require(packageJSON);
     if ((dupeModule = Object.values(seen).find(i => i.id === json.name))) {
-      let msg = action => `The plugin module name '${json.name}' has already been loaded from the module path ${dupeModule.attributes.rawDir}, ${action} load of module at path ${dir}.`;
-      if (this.environment !== 'test') {
-        log.warn(msg('skipping'));
-      } else {
-        throw new Error(msg('conflict with'));
-      }
+      // DONT FORGET TO UNCOMMENT THIS!!!
```
ef4: ☝️

habdelra: lol thanks

packages/hub/model.js
```diff
@@ -12,6 +12,20 @@ module.exports = class Model {
     return priv.get(this).contentType;
   }
 
+  getId() {
+    let { jsonapiDoc } = priv.get(this);
+    if (jsonapiDoc) {
+      return jsonapiDoc.id;
+    }
+  }
```
ef4: I guess this exists because `getField('id')` is async, and id doesn't ever really need to be async?

habdelra: I was just assuming that `id` was not a field and using an accessor for it. but now that I look at Model more closely I see that getField can handle this. I'll clean this up.

packages/routing/addon/templates/common.hbs
```diff
@@ -1 +1,6 @@
-{{cardstack-content content=model format="isolated"}}
+{{!-- TODO Let's eventually refactor this "if" condition out --}}
+{{#if isNewContent}}
+  {{cardstack-content content=model format="isolated"}}
+{{else}}
+  {{cardstack-content content=model.primaryCard format="isolated" params=params}}
```
ef4: I think cards need the ability to modify the query params in the URL too. I mean, they can choose to do a whole transition and go through the routing system and the spaces over again, and _sometimes_ you want that, but other times you just want to change the URL to reflect the new state you're already in (like the user just typed some text into a search field and you want to update `?q=`).

Perhaps in addition to `params=params` here we could have `setParam=setParam` (and the resulting action handler should be smart enough to only let you set your own known query params).

packages/hub/model.js
habdelra: and also to be consistent, we should probably take the same approach for `meta`

packages/routing/addon/templates/common.hbs
habdelra: sounds good 👍

habdelra:
ok, so for the query params, lets do the inverse of what we are doing now, and pass to the card component only the query params that it has defined in the router (and we can remove the namespacing in the params that we pass to the component to make life easier for the developers). and I'll allow routes to not have a query property as well, which as you suggest just means return the routing card itself--i agree that is a really clean and simple way for a card to declare that they use consume a query param.

marcoow:
Couldn't there potentially be conflicts when multiple cards define the same routes?

habdelra:
@marcoow definitely, also you could have patterns that stomp on each other, just like ember or rails. Like the route `/article` and `/:slug` could stomp on each other if the hub matches `:/slug` before `/article`. So the idea is that we match the routes deterministicly. The first route in the array has higher precedence. So in that case you’d want to have `/article` appear before `:slug`, otherwise route `/article` would be unreachable.

habdelra:
In terms of multiple routing cards having the same routes. That is a different matter entirely. As I was mentioning in the beginning of my PR description, the routing cards cascade. So there is a singe top level routing card, called the application card, and it can route to to another routing card. 2 routing cards that define the same route would not conflict because you they would have different path segments above them in the application card. The ember analogy would be nested routes. Multiple routing cards can be thought of as nested ember routes. 
```js
Router.map(function (){
  this.route(‘articles’, function () {
    this.route(‘foo’);
  });
  this.route(‘events’, function () {
    this.route(‘foo’);
  });
});

Is analogous to the content type `articles‘ having the route `/foo` and the content type `event` having the route `/foo`. Even though two cards have the same route, they don’t collide because the application card needs to be considered. Every routable card has a unique path, and as a result the path to land on a particular event card would always be different than the path to land on a particular article card. You’d could even go farther because all event cards would inherit their routes from the event content type, but even in that case they are not duplicated within event content types themselves, because each event instance has a unique path just to get to it. 

Does that make sense?

**Comment by marcoow** [comment]
Ok, so every card can only define its own routes *within* the route it is mapped to by the application route? So in the above example both the `articles` and `events` cards route themselves to `/foo` but since the application route maps them to `/articles` and `/events` respectively, the real routes end up being `/articles/foo` and `/events/foo`?

Does that mean that every card used in an application has to be explicitly routed? I guess just doing

```es6
Router.map(function() {
  this.route(cardstackRoutes);
});
```

wouldn't work anymore in this case?

habdelra:
Yeah, or `/articles/my-article/foo`, `/events/thanksgiving/foo`. Keep in mind that there is a particular instance of a routing card (albeit you can hard core a particular card to a route, which is the “vanity url” scenario).

Every card in the application doesn’t have to have a route, but every *isolated* card should have a route. And yes this means we are getting rid of all the custom routes. So the router actually will look like this now:

```js
Router.map(cardstackRoutes);
```

So that the hub will take over all our routing going forward. `catdstackRoutes`, above, in this case *is* the mechanism that you use to expose the routing feature to your Cardstack app in this PR

marcoow:
Hm, I still don't really get it - I'll have another look tomorrow. Would you import `cardstackRoutes` from every card you're using then?

```js
import { cardstackRoutes as cardARoutes } from 'card-a';
import { cardstackRoutes as cardBRoutes } from 'card-b';
```

Otherwise, I don't really understand where the nesting that prevents conflicts you're describing above is happening when multiple cards that have no knowledge about each other can define the same routes? Does that happen in `@cardstack/router` then? But based on what rules?

🤯

habdelra:
No there is only one top level application card, and from there all the applications routes can be derived. The routing request for a particular url is first processed by the application card. Consider the path `/events/thanksgiving/foo` that we were taking about. 

The application card (if we use the default router which is located at @cardstack/routing/cardstack/router.js) will match the first 2 segments of the path `/events/thanksgiving` to the route `/:type/:id`. The default router has a query that just returns the card for the type and id that have been provided in the dynamic segments of this route. The router will then look to see if the events card that it found with the route’s query (with the id of `thanksgiving`) has a router. If it does it will then use the “unconsumed” part of the path `/foo` to match to a route in the events card. We will then match to the `/foo` route of the events card, which itself has a query to return a particle card that corresponds to this route. 

Does this help to illustrate it better? We also start at the single application card and then descend thru the tree of possible routes by matching the path to the route, obtaining the card from the route’s query, matching the remaining path to the route, obtaining the next card from the route’s query, recursing in this manner until there is no more path to consume.

habdelra:
At the very least this discussion is illustrating to me I need better documentation to describe our routing. Perhaps a diagram would help.

marcoow:
I think I get it now (sorry for the many questions 🙏) - so the `/:type/:id` prefix is mandatory for all cards and they can only define vanity URLs below that scope which thus prevents conflicts. I somehow thought that this was supposed to support routing cards to top level paths, e.g. `http://localhost:4200/contact` that we route the `contact-form` card to in the deck. I assume the intended way for defining a route like that for a card would be defining that route on the application level.

habdelra:
You can route `http://localhost:4200/contact to the contact-form card, however that route rule does not live on the contact-form card. All those “root” level rules would be defined in the application card’s router. This router lives at the root level. And as you point out, typically that would be a `/:type/:id`. But the app developer may want some vanity URLs as well (as we do in the deck app).

habdelra:
@ef4 just an FYI, all your comments have been addressed and the latest documentation is available here. https://github.com/cardstack/cardstack-documentation/pull/3 I'm still working thru integrating this with our app...

packages/hub/indexing/document-context.js
```diff
@@ -23,6 +25,8 @@ module.exports = class DocumentContext {
     this._realms = [];
     this._pendingReads = [];
     this._followedRelationships = {};
+    this._routingCardId = get(upstreamDoc, 'meta.routing-card.data.id');
+    this._routingCardType = get(upstreamDoc, 'meta.routing-card.data.type');
```
ef4: Am I understanding correctly that the meta fields are only present when the upstream document is a `spaces` resource?

If yes, I would consider moving them into being `attributes` of the `spaces` type.

packages/routing/cardstack/computed-field-types/http-status.js
```diff
@@ -0,0 +1,16 @@
...
+exports.compute = async function(model) {
+  let primaryCard = await model.getRelated('primary-card');
+  let meta = primaryCard.getMeta();
+
+  if (meta && meta['is-error-card']) {
+    return errorCardMapping[await primaryCard.getField('id')] || 500;
+  }
+
+  return 200;
+};
```
ef4: This computed field type is neat.

packages/hub/indexing/document-context.js
```diff
@@ -196,6 +203,22 @@ module.exports = class DocumentContext {
     return searchTree;
   }
 
+  async _addSelfLink(jsonapiDoc) {
+    if (this._routingCardId && this._routingCardType && this.plugins) {
+      let routingCard = await this.read(this._routingCardType, this._routingCardId, true);
```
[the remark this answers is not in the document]
habdelra: I'm using our read cache to read routing cards. In this case there is no actual reference from the routing card to the card that it is routing too, but I need to get its data in order to populate the self links.

packages/hub/plugin-loader.js
```diff
@@ -276,6 +279,10 @@ class ConfiguredPlugins {
         copied.attributes.enabled = true;
       }
       this._plugins[copied.id] = copied;
+
+      if (copied.attributes['is-root-plugin']) {
+        this.rootPlugin = copied;
```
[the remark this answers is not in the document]
habdelra: It is playing into the decision for the the topmost router, but it's not hardcoding it. The order of precedence is:
1. Use the router associated with the application card defined by the `plugin-configs/@cardstack/hub`
2. If there is no router associated with the application card defined by the `plugin-configs/@cardstack/hub`, then use the router associated with the cardstack application (the top most plugin)
3. If there is no router associated with the cardstack application, then use the "default" router, which is the routing feature in @cardstack/routing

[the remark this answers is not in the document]
habdelra: (and I have tests for these 3 different scenarios in the hub/node-tests/routers-test.js)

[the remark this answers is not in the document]
habdelra: I opted for rule #2 above, as it makes it easy to get up and running for a cardstack app by just having a `router.js` in your cardstack application as opposed to having to create a new module just for the purposes of having a router.

packages/hub/plugin-loader.js
```diff
@@ -172,6 +173,7 @@ class PluginLoader {
       attributes: {
         dir: moduleRoot,
         rawDir: dir,
+        'is-root-plugin': breadcrumbs.length === 0,
```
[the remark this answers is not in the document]
habdelra: it's more for the hub's routing machinery than for use by cards. As I mentioned above, it is part of the order of precedence in terms of where we search for the topmost router.

packages/hub/indexing/document-context.js
habdelra: sure, sounds good.

packages/hub/plugin-loader.js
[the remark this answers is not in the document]
habdelra: I started creating a `cards/application-router` module in the deck project, and then stopped because that felt like too much boilerplate when all i wanted was a `router.js` for my cardstack application.

packages/hub/indexing/document-context.js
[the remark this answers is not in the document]
habdelra: we'll need to use the routing card data for each of the embedded cards in the spaces response. so it definitely was data that should be cached, and we already have a cache.... maybe I could do a better job of sharing the cache without this parameter `notAReference`?

ef4:
Thanks for all the work on this, overall it is looking great and I am happy to land it ASAP.

habdelra:
We're pretty close, but i do wanna address the things we were chatting about in slack first.

habdelra:
@ef4 I've refactored this PR to leverage the router map, as we discussed. The `hub:routers` will initialize a router map by crawling all the possible routes starting from the application card. From that point forward, the router map is leveraged for client routing requests (i.e. queries for `spaces` documents). The prevents arbitrary route discovery when the client requests a route, as all the routes are known up-front and organized to retrieve the specific route first. The secret sauce is that as I derive the route map, I also construct a "route stack" that is a stack of all the antecedent routes for a particular route. The route stack allows me to resolve all the replacement tags in a manner that leverages the context of each specific routing card within the stack of routing cards that lead to a particular primary card; so `:card:my-field` makes sense when I combine all the path segments for a route from its antecedent routing cards. Additionally, I can use this route stack to reverse the query for looking up a card's path, such that I can reverse not only the card I'm looking up, but all of its antecedent routing cards' queries. The stack of cards that we routed through to arrive at the primary card is also an attribute of the spaces document.

I still need to work on the new `params` that we discussed, but I think that it's probably good for you review this again. I'll commit the updated `params` work to this PR on Monday (which should be really straightforward work).

habdelra:
@ef4 Ok, I'm done will all my planned updates to this PR. The refactored params work is now additionally ready for review as part of this PR.

packages/hub/plugin-loader.js
```diff
@@ -276,6 +278,10 @@ class ConfiguredPlugins {
         copied.attributes.enabled = true;
       }
       this._plugins[copied.id] = copied;
+
+      if (copied.attributes['is-root-plugin']) {
+        this.rootPlugin = copied;
```
ef4: This is the only remaining place I see `is-root-plugin`, is this one a leftover?

packages/hub/routers.js
```diff
@@ -0,0 +1,280 @@
...
+
+module.exports = declareInjections({
+  currentSchema: 'hub:current-schema',
+  searchers: 'hub:searchers',
+},
+
+class Routers {
+  // Lazily loading this as running this in the constructor runs into the
+  // situation where the searchers are not ready yet. Is there a better way to know when
+  // the searcher is ready so we can set this up outside of a user request?
```
ef4: I think there's not a better way right now, but we should add one. I [filed a separate issue](https://github.com/cardstack/cardstack/issues/496), because there's already other code like this that should move to the same solution.

packages/hub/plugin-loader.js
habdelra: Oops! Yeah that is unnecessary now. I’ll clean it up.

PR #430, opened 2018-11-12
Adding Node 10 to Travis tests

mansona:
Thanks to https://github.com/cardstack/cardstack/pull/402 we should now be able to use Node 10 for cardstack 🎉 

This is just a PR to kick off Travis and to open the discussion.

## To Do
- [x] Update ember-cli-postcss See https://github.com/cardstack/cardstack/pull/413
- [x] Figure out why the truffle tests are failing

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,7 @@
 language: node_js
 node_js:
   - "8"
+  - "10"
 
 sudo: required
 services:


Turbo87:
@mansona FYI https://github.com/cardstack/cardstack/pull/413, and I have a few more commits that bring us closer to Node 10 compat

habdelra:
@mansona The PR https://github.com/cardstack/cardstack/pull/548 should fix the etheruem plugin failures in node 10

mansona:
@habdelra I assume we're ok with merging this now that it's passing? Also, it might be worth running a release that includes #548 so that we can make sure that downstream projects are using a version that definitely works in Node 10 👍

habdelra:
@mansona totally, let’s do it!

PR #433, opened 2018-11-12
Use `prettier` integrated with ESLint

Turbo87:
see https://prettier.io/


ef4:
I'm fine with this, I just want to find a time to merge it that won't generate a lot of conflict resolution work for people who have PR branches in progress. I suppose there is no perfect time.

Turbo87:
> I just want to find a time to merge it that won't generate a lot of conflict resolution work for people who have PR branches in progress. I suppose there is no perfect time.

yeah, that might be hard... 😅 

we could do it package by package, but I'm not sure if that will actually be any better than just doing it all at once

mansona:
@Turbo87 if we could do it package-by-package I think that might be significantly better 🤔 there are at least some packages that haven't had much movement recently and we could get them merged without much risk of conflicts

Turbo87:
@mansona then we'd need a separate ESLint config and migrate the packages over one by one, not sure if that's the right way forward though

Turbo87:
@ef4 any chance we can land this soon? I believe the package-by-package approach is more complicated due to the additional ESLint configs than just biting the bullet and merging it all at once.

PR #436, opened 2018-11-13
Make data service's type argument optional

### LINKED ISSUE #375: CardData query and queryCard should treat `type` as optional
The `@cardstack/data` service has these methods:

```
  async query(type, format, opts={}) 
  async queryCard(type, format, opts={}) 
```

But `type` should really be optional here.  The query APIs are new enough that we should probably just break them to make it make more sense:

```
  async query(format, opts={}) 
  async queryCard(format, opts={}) 
```

where `type` moves into the `opts`.

In terms of implementation: heterogeneous queries are already implemented on the Hub (via `GET /api` opposed to `GET /api/${type}`). There is also already a base ember-data model named 'cardstack-card' that all the other models inherit from. Probably our Adapter should get smarter about seeing that type and changing the URL to make a heterogeneous query.

marcoow:
This makes the `type` argument optional for the data service's `query` and `queryCard` methods so that we can run heterogeneous queries.

closes #375

### DIFF
--- a/packages/data/tests/integration/query-test.js
+++ b/packages/data/tests/integration/query-test.js
@@ -48,7 +48,7 @@ module('Integration | @cardstack/data service query()', function (hooks) {
 
     test("it can load relationships for a query in a particular format", async function (assert) {
       let dataService = this.owner.lookup('service:cardstackData');
-      let pages = await run(() => dataService.query('page', 'isolated', { filter: { title: 'Homepage' } }));
+      let pages = await run(() => dataService.query('isolated', { filter: { title: 'Homepage' }, type: 'page' }));
 
       let homepage = pages.get('firstObject')
       let articles = homepage.get('articles').toArray().map(i => i.toJSON());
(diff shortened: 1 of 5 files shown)


packages/data/tests/integration/query-test.js
```diff
@@ -135,6 +135,13 @@ module('Integration | @cardstack/data service query()', function (hooks) {
       });
     });
 
+    test("it can load cards without a type", async function (assert) {
+      let dataService = this.owner.lookup('service:cardstackData');
+      let pages = await run(() => dataService.query('isolated', { filter: { title: 'Homepage' } }));
```
habdelra: Make sure to assert how many results you’d expect here. If there is only one homepage card, then assert `pages.data.length` is 1, because when filtering doesn’t work you end up with a result set  that includes all the cards. And your test doesn’t actually prove filtering works in its current form as you always just grab the first item.

habdelra:
I had just one minor comment, aside from that, this looks good.

packages/data/tests/integration/query-test.js
marcoow: Yeah, makes sense - added an assertion on result length.

marcoow:
@habdelra: addressed your comment

habdelra:
LGTM, merge when green

Issue #403, opened 2018-11-05
Refactor cs-files content type


ef4:
The `cs-files` content type should get defined in its own `@cardstack/files` package.

Let's rename it to "cardstack-files", none of our content types use "cs-" yet, and even where we're using that in templates it is a necessary evil until we can namespacing via Ember's module unification.

Something like this:

```js
// The user would write this to configure their file storage
factory.addResource('data-sources').withAttributes({
  sourceType: '@cardstack/files',
  params: {
    storeFilesIn: { type: 'data-sources', id: 'whatever-data-source-goes-here' }
  }
});
```

```js
// That would mean the files package has an opportunity to emit static 
// models based on that parameter:
// 
// packages/files/cardstack/static-models.js
module.exports = function({ storeIn }) {
  return [
    {
       type: 'content-types',
       id: 'cardstack-files',
       attributes: {
          // etc
       },
       relationships: {
          'data-source': {
             data: storeFilesIn
          },
          // etc
       }
    }
  ];
}
```

PR #437, opened 2018-11-13
Refactor cs-files content type to be called cardstack-files and to live in its own addon

alexspeller:


jorgelainfiesta:
Fixes https://github.com/cardstack/cardstack/issues/403

PR #438, opened 2018-11-14
DocumentContext refactor

bagby:
Cleans up some stuff around how we create `DocumentContext`s, based on conversations with @ef4 and @habdelra.

### DIFF
--- a/packages/pgsearch/client.js
+++ b/packages/pgsearch/client.js
@@ -3,7 +3,6 @@ const { Pool, Client } = require('pg');
 const Cursor = require('pg-cursor');
 const migrate = require('node-pg-migrate').default;
 const log = require('@cardstack/logger')('cardstack/pgsearch');
-const DocumentContext = require('@cardstack/hub/indexing/document-context');
 const Session = require('@cardstack/plugin-utils/session');
 const { declareInjections } = require('@cardstack/di');
 const EventEmitter = require('events');
@@ -14,7 +13,7 @@ const { upsert, queryToSQL, param } = require('./util');
 const config = postgresConfig({ database: `pgsearch_${process.env.PGSEARCH_NAMESPACE}` });
 
 module.exports = declareInjections({
-  controllingBranch: 'hub:controlling-branch',
+  controllingBranch: 'hub:controlling-branch'
 },
 
 class PgClient extends EventEmitter {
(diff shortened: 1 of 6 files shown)


packages/pgsearch/client.js
```diff
@@ -296,54 +292,62 @@ class Batch {
...
-          read
+  async _invalidations() {
+    let branches = Object.keys(this._touched).map(key => {
+      log.info('_touched: %s', key);
+      return key.split('/')[0];
+    }).filter((v, i, a) => a.indexOf(v) === i);
+
+    for (let branch of branches) {
+      await this.client._iterateThroughRows(
+        'select id, type from documents where expires < now()', [], async({ id, type }) => {
```
ef4: This is going to run the same query on every iteration of the loop.

I think we want `select id, type, branch from documents` instead, and then there's no need for the outer loop that iterates over branches.

packages/pgsearch/client.js
```diff
@@ -296,54 +292,62 @@ class Batch {
...
+          this._touched[`${branch}/${type}/${id}`] = this._touchCounter++;
         });
+      await this.client.query('delete from documents where expires < now()');
+      await this.client.docsThatReference(branch, Object.keys(this._touched), async (doc, refs) => {
```
ef4: I think we drop the `branch` argument to `docsThatReference` because it's now part of `this._touched` instead.

bagby:
@ef4 I updated the invalidations as you suggested (I think), if you wouldn't mind taking a look 👀 

Still a few test failures around including `included`s (no pun intended) but I'm working my way through it via stepping through tests with the debugger. On the plus side I'm learning a whole lot about the guts of the system 💯

bagby:
@ef4 of particular note is the pgsearch client's `docsThatReference` function -- in the `this._touched`  the keys are `${branch}/${type}/${id}` but in the `refs` column in the DB we are storing `${type}/${id}` so I'm stripping off the branch to enable the query to work. Not sure if this is correct or if we actually want to store `${branch}/${type}/${id}` in the DB, I opted to not go that route since we already have a `branch` column.

Also there's some stuff I need to clean up (logging, etc) plus squash the commits.

ef4:
This is looking good.

I think it would be reasonable to store `${branch}/${type}/${id}` in the `refs` column, it might even simplify things a little. But either way is correct (at least so long as we don't have docs that depend on docs in other branches).

PR #445, opened 2018-11-16
[WIP] Field searchability 

### LINKED ISSUE #396: Implement includeInDocumentSearch flag for fields to control inclusion in full-text search
**Edit:** This issue has been updated to reflect the conversation in https://github.com/cardstack/cardstack/pull/445 and other use-cases surrounding searchability have been extracted into the issues #465 and #464 

Each field should have an `includeInDocumentSearch` boolean attribute that indicates whether it should be included in the combined full-text search. In this context combined full-text search is different from [field-level full-text search](https://github.com/cardstack/cardstack/issues/464) in that it is just powered by the query implementation of a searcher `{ query: 'some words' }` and searches recursively through relationships. 

Our combined full-text search index (the `q` column) should only include fields that are marked with the `includeInDocumentSearch` tag. It should _recursively_ include searchable fields, meaning that _if and only if_ a relationship field is searchable, than the searchable fields from the related resource(s) are also included.

### LINKED ISSUE #465: Implement configuration to be able to chose which fields are `searchable`
This work has been started in https://github.com/cardstack/cardstack/pull/445 and this issue is intended to cover the functionality of that work-in-progress implementation which was originally intended to solve #396

This issue concerns the ability to be able to **filter** a document based on the value of its attributes. The ability to filter a document by a particular attribute is governed by the presence of that attribute in the `searchDoc` in the DocumentContext.

## Todo: 
- [ ] make sure that if the `searchable` flag of an attribute is false it does not make into the searchDoc (see the tests in the linked PR)
- [ ] make sure that the `searchable` flag affects relationships as well as it does other field types
- [ ] make sure the functionality of the `searchable` flag works as expected with nested relationships

mansona:
Fixes #465 

This is still considered WIP because it doesn't cover enough test cases (see the pending tests in the code).

I want to get some feedback on the implementation approach **and** the testing approach. @habdelra and I had a long discussion about the test on slack and I was able to get it working without the need for a full Cardstack environment. Instead, I'm testing the DocumentContext directly. This feels a bit better to me because it seemed a bit overkill to set up a full hub environment just to test something that almost exclusively lives in the DocumentContext class.

As for the implementation. The tests pass as a result of this really small change, but instead of assuming that this safe to do I wanted to check in to see if anyone could possibly see any unintended consequences from this change? If we think there might be any issues it might be prudent to make some more tests that fail to prove the issue 👍

### DIFF
--- a/packages/hub/indexing/document-context.js
+++ b/packages/hub/indexing/document-context.js
@@ -124,24 +124,38 @@ module.exports = class DocumentContext {
 
   // copies attribues appropriately from jsonapiDoc into
   // pristineDocOut and searchDocOut.
-  async _buildAttributes(contentType, jsonapiDoc, userModel, pristineDocOut, searchDocOut) {
+  async _buildSearchAttributes(contentType, jsonapiDoc, userModel, searchDocOut) {
     for (let field of contentType.realAndComputedFields.values()) {
       if (field.id === 'id' || field.id === 'type' || field.isRelationship) {
         continue;
       }
+
+      if(!field.searchable) {
+        continue;
+      }
+
       if (contentType.computedFields.has(field.id) ||
           (jsonapiDoc.attributes && jsonapiDoc.attributes.hasOwnProperty(field.id))) {
         let value = await userModel.getField(field.id);
-        await this._buildAttribute(field, value, pristineDocOut, searchDocOut);
+        // Write our value into the search doc
+        searchDocOut[field.id] = field.searchIndexFormat(value);
       }
      }
   }
 
-  async _buildAttribute(field, value, pristineDocOut, searchDocOut) {
-    // Write our value into the search doc
-    searchDocOut[field.id] = field.searchIndexFormat(value);
-    // Write our value into the pristine doc
-    ensure(pristineDocOut, 'attributes')[field.id] = value;
+  async _buildPristineAttributes(contentType, jsonapiDoc, userModel, pristineDocOut) {
+    for (let field of contentType.realAndComputedFields.values()) {
+      if (field.id === 'id' || field.id === 'type' || field.isRelationship) {
+        continue;
+      }
+
+      if (contentType.computedFields.has(field.id) ||
+          (jsonapiDoc.attributes && jsonapiDoc.attributes.hasOwnProperty(field.id))) {
+        let value = await userModel.getField(field.id);
+        // Write our value into the pristine doc
+        ensure(pristineDocOut, 'attributes')[field.id] = value;
+      }
+     }
   }
 
   async _buildRelationships(contentType, jsonapiDoc, userModel, pristineDocOut, searchDocOut, searchTree, depth, fieldsets) {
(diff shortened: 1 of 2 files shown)


habdelra:
I feel pretty comfortable that if all the tests pass then this is a pretty safe change to make. Additionally, I'm really looking forward to your DocumentContext testing, as I have a bunch of stuff that I want to move into it.

ef4:
There are two distinct concepts that I think are being handled as one in this PR.

One (I'll call it "filterability") is whether we want a field to appear in the search attributes. That controls whether it's possible query things like `{ filter["some-field"]="some-value" }`.

The second concept (I'll call it "searchability") is which fields get included in the full-text search, so that `{ query: 'some words' }` will only check in fields that are searchable.

Today these concepts are conflated, because _everything_ in the searchDoc goes into the full-search search. 

The test case here is to create a field that is _not_ full-text searchable, such that `{ query: 'some words' }` doesn't match words in the field, while simultaneously still allowing the field to be filtered explicitly.

Our database design is intended to support this pattern, the changes are localize to the point where we construct `q`.

ef4:
Yeah, I think you're right that we should clarify the configuration API by picking better names.

And I realize now there might even be a third concept to consider, which is whether we would build a postgres full text index against any particular field in the search doc as well, in order to enable full-text searches within a particular field.

Strawman ideas:
 - `searchable`: whether a field goes into the searchDoc at all
 - `fullTextSearchable`: whether you get a full-text postgres index against the field's particular place in the searchDoc (which isn't something we've implemented yet, but will probably want to)
 - `includeInDocumentSearch`: whether the field goes into `q`

mansona:
@ef4 can you clarify what you mean by the `fullTextSearchable` description 🤔 As it stands I don't understand the distinction between that and `includeInDocumentSearch`

ef4:
Sure. This is probably confusing because I'm planning ahead to optimizations we haven't implemented yet.

We support both exact filtering and full-text filtering. When you do `?filter[title]=ember` you get a full-text search against the title field, and you will match titles like "About ember".  If you want an exact search, you have to say `?filter[titlte][exact]=ember`, which would _not_ match "About ember" and would only match "ember".

We plan to change this API by making exact searching the default, so that `?filter[title]=ember` does an exact match and we would introduce `?filter[title][contains-text]=ember` for when you really want the full-text search. The only reason full-text is the default is because of the earlier iteration where our search index was Elasticsearch. But that is not super relevant to this issue, in either case, we want both kinds of searches to be possible.

To run efficient full-text searches, you really want to have precomputed indices for them in postgres. There are two flavors of this. The one we're already using for the `q` column is to explicitly store your own tsvector. The other is probably what we would want for speeding up full-text searches of specific fields in the searchDoc, and it would look something like:

```
CREATE INDEX search_doc_title ON documents USING GIN (to_tsvector('english', search_doc->>'title'));
```

This has the benefit of being able to index arbitrary queries against the search doc, for example we can support `?filter[author.name][contains-text]=chris` with an index like:

```
CREATE INDEX search_doc_author_name ON documents USING GIN (to_tsvector('english', (search_doc->>'author')->>'name'));
```

We can create these indices automatically if we know from the schema which fields support full-text-searchability. We don't really want to create them speculatively for all fields, because they are not free -- each one imposes some cost at document writing time, and takes up storage space.

ef4:
Closing this because it's been sitting in a "WIP" state.

Issue #434, opened 2018-11-12
Can't input valid value into integer-editor 


balinterdi:
The value of an integer field is rejected by the backend saying that '"51"' is an invalid value for an integer field:

![screen shot 2018-11-12 at 17 20 37](https://user-images.githubusercontent.com/5022/48360396-70b68b00-e69f-11e8-9056-6f68d34ab518.png)

balinterdi:
Another good example is the "Underestimated read time" field in the tools dummy app.

balinterdi:
Closed via #449

PR #449, opened 2018-11-19
Fix integer editor setting string values

balinterdi:
See #434

### DIFF
--- a/packages/core-types/addon/components/field-editors/integer-editor.js
+++ b/packages/core-types/addon/components/field-editors/integer-editor.js
@@ -8,10 +8,10 @@ export default Component.extend({
     update(value) {
       let content = this.get('content');
       let field = this.get('field');
-      if (value == null || value === '') {
+      if (!value) {
         content.set(field, null);
       } else {
-        content.set(field, +value);
+        content.set(field, Number(value));
       }
     }
   }
--- a/packages/core-types/addon/components/field-editors/integer-editor.js
+++ b/packages/core-types/addon/components/field-editors/integer-editor.js
@@ -3,4 +3,16 @@ import layout from '../../templates/components/field-editors/integer-editor';
 
 export default Component.extend({
   layout,
+
+  actions: {
+    update(value) {
+      let content = this.get('content');
+      let field = this.get('field');
+      if (value == null || value === '') {
+        content.set(field, null);
+      } else {
+        content.set(field, +value);
+      }
+    }
+  }
 });
--- a/packages/core-types/tests/integration/components/field-editors/integer-editor-test.js
+++ b/packages/core-types/tests/integration/components/field-editors/integer-editor-test.js
@@ -1,7 +1,7 @@
 import { module, test } from 'qunit';
 import { setupRenderingTest } from 'ember-qunit';
-import { render, triggerEvent, waitFor } from '@ember/test-helpers';
-import Object from '@ember/object';
+import { render, triggerEvent, fillIn } from '@ember/test-helpers';
+import EmberObject from '@ember/object';
 import hbs from 'htmlbars-inline-precompile';
 
 module('Integration | Component | field editors/integer editor', function(hooks) {
@@ -9,7 +9,7 @@ module('Integration | Component | field editors/integer editor', function(hooks)
 
   test('it renders', async function(assert) {
     this.setProperties({
-      model: Object.create({ rating: 3 }),
+      model: EmberObject.create({ rating: 3 }),
       onchange() {}
     });
     await render(hbs`
@@ -19,12 +19,12 @@ module('Integration | Component | field editors/integer editor', function(hooks)
         onchange=onchange
       }}`);
     assert.dom('input[type=number]').hasValue('3', 'number input has correct value');
-    assert.dom('input').isNotDisabled('input is not disabled');
+    assert.dom('input').isNotDisabled();
   });
 
   test('it updates the value correctly', async function(assert) {
     this.setProperties({
-      model: Object.create({ rating: 3 }),
+      model: EmberObject.create({ rating: 3 }),
       onchange() {}
     });
     await render(hbs`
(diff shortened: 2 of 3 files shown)


jorgelainfiesta:
Could we perhaps add an integration test for `integer-editor` now it was changed? there are tests for the other field editors but not for this one

packages/core-types/addon/components/field-editors/integer-editor.js
```diff
@@ -3,4 +3,16 @@ import layout from '../../templates/components/field-editors/integer-editor';
...
   layout,
+
+  actions: {
+    update(value) {
+      let content = this.get('content');
+      let field = this.get('field');
+      if (value == null || value === '') {
+        content.set(field, null);
+      } else {
+        content.set(field, +value);
```
Turbo87: is there a reason for using `+value` over something more readable like `Number(value)`?

packages/core-types/addon/components/field-editors/integer-editor.js
```diff
@@ -3,4 +3,16 @@ import layout from '../../templates/components/field-editors/integer-editor';
 
 export default Component.extend({
   layout,
+
+  actions: {
+    update(value) {
+      let content = this.get('content');
+      let field = this.get('field');
+      if (value == null || value === '') {
```
Turbo87: why not `!value`? I'm assuming `value` can only be something nullish or a string?

packages/core-types/addon/components/field-editors/integer-editor.js
balinterdi: I could use `Number(...)`, it's just that I saw this casting "trick" recently and really liked it. But I think you're right, as it's not as readable and widely used, it's probably better to use `Number`.

packages/core-types/addon/components/field-editors/integer-editor.js
balinterdi: I restored this piece of code from the previous version of the integer editor where it was written that way. I don't think it makes a difference, though.

packages/core-types/tests/integration/components/field-editors/integer-editor-test.js
```diff
@@ -0,0 +1,57 @@
+import { module, test } from 'qunit';
+import { setupRenderingTest } from 'ember-qunit';
+import { render, triggerEvent, waitFor } from '@ember/test-helpers';
+import Object from '@ember/object';
```
Turbo87: this overrides the `Object` global. it's better to `import EmberObject from '@ember/object';`. I think ESLint should also have a rule for this, but maybe it's not enabled? 🤔

packages/core-types/tests/integration/components/field-editors/integer-editor-test.js
```diff
@@ -0,0 +1,57 @@
...
+      onchange() {}
+    });
+    await render(hbs`
+      {{field-editors/integer-editor
+        content=model
+        field="rating"
+        onchange=onchange
+        disabled=true
+      }}`);
+    assert.dom('input').isDisabled('integer input is disabled');
```
Turbo87: the `'integer input is disabled'` messages is somewhat redundant. `qunit-dom` will automatically generate a message like `Element "input" is disabled` if no explicit message is provided. IMHO these messages are most often good enough, but have the advantage of not cluttering up the testing code.

packages/core-types/tests/integration/components/field-editors/integer-editor-test.js
```diff
@@ -0,0 +1,57 @@
...
+      onchange() {}
+    });
+    await render(hbs`
+      {{field-editors/integer-editor
+        content=model
+        field="rating"
+        onchange=onchange
+      }}`);
+    this.set('model.rating', 5);
+    await triggerEvent('.field-editor > input', 'input');
```
Turbo87: I'm not sure I understand what is being tested here. Are you testing a user input? maybe the `fillIn()` or `typeIn()` test helpers might be better suited here?

packages/core-types/tests/integration/components/field-editors/integer-editor-test.js
balinterdi: I kind of felt it wasn't the thing I wanted :)

packages/core-types/tests/integration/components/field-editors/integer-editor-test.js
balinterdi: True, I was so focused on triggering an `input` event (since that's what the component reacts to) that I lost sight of the fact that `fillIn` does the same and reads better.

packages/core-types/tests/integration/components/field-editors/integer-editor-test.js
```diff
@@ -0,0 +1,56 @@
...
+    await render(hbs`
+      {{field-editors/integer-editor
+        content=model
+        field="rating"
+        onchange=onchange
+      }}`);
+    await fillIn('.field-editor > input', '5');
+    assert.dom('input[type=number]').hasValue('5', 'input is updated');
+    assert.dom('input').isNotDisabled();
+    assert.equal(this.get('model.rating'), '5', 'model attribute is updated');
```
Turbo87: I think this should be:
```js
assert.strictEqual(this.get('model.rating'), 5, 'model attribute is updated');
```

as in: use `strictEqual` to make a `===` comparison, and use `5` instead of `'5'` to make sure it's actually a number

packages/core-types/tests/integration/components/field-editors/integer-editor-test.js
```diff
@@ -0,0 +1,56 @@
...
+      model: EmberObject.create({ rating: 3 }),
+      onchange() {}
+    });
+    await render(hbs`
+      {{field-editors/integer-editor
+        content=model
+        field="rating"
+        onchange=onchange
+      }}`);
+    await fillIn('.field-editor > input', '5');
```
Turbo87: can we use just `'input'` here? otherwise we'd be using 3 different selectors for the same thing

packages/core-types/tests/integration/components/field-editors/integer-editor-test.js
balinterdi: Sure, you're right!

Issue #450, opened 2018-11-19
Issue with ember-source version


jorgelainfiesta:
With @mansona, we observed very unexpected behaviors happening with `cardstack/edges` when trying to use it in a dummy app for `cardstack/card-picker`. The edges would not be rendered, but no errors would be yielded.  After a lot of experimentation, we discovered that the cause of problems was the version of `ember-source` being used, which was `~3.4.0`. If the app uses `ember-source` `~3.2.0`, the edges work as expectedly. 

When isolating this problem in a new branch, the errors that we get are:

 ```
Uncaught Error: Could not find module ember-glimmer/syntax
```
and
```
Uncaught Error: Could not find module `ember-resolver` imported from `dummy/resolver`
```

This issue can be reviewed in this branch https://github.com/simplabs/cardstack/tree/ember-source-issue by running `yarn start` in `packages/tools`.

Turbo87:
the second issue is related to https://github.com/thefrontside/ember-let/issues/49 and can be fixed by moving to `ember-let-polyfill` instead of using `ember-let`, which seems to be safe as we don't use the inline let form at all.

PR #451, opened 2018-11-19
Close Bottom edge

### LINKED ISSUE #410: Editor: card search panel can never be hidden
Once the card search panel comes up, there is no way to hide it again - even when leaving the edit mode, the panel does not disappear:

bildschirmfoto 2018-11-06 um 12 37 51

The simplest way to fix that would be adding a _"close [x]"_ button to the top right or so. @dnaidionov did you have anything specific in mind?

### TODO

- [ ] add a _"close [x]"_ button to the card selector that closes the panel
- [ ] auto-close the card selector when the user leaves edit mode

jorgelainfiesta:
Fixes #410


packages/card-picker/package.json
```diff
@@ -42,12 +42,13 @@
     "ember-cli-sri": "^2.1.1",
     "ember-cli-template-lint": "^1.0.0-beta.1",
     "ember-cli-uglify": "^2.1.0",
+    "ember-data": "^3.5.0",
```
Turbo87: why is Ember Data needed for this PR?

packages/card-picker/package.json
```diff
@@ -42,12 +42,13 @@
...
     "ember-disable-prototype-extensions": "^1.1.3",
     "ember-export-application-global": "^2.0.0",
     "ember-load-initializers": "^1.1.0",
     "ember-maybe-import-regenerator": "^0.1.6",
     "ember-resolver": "^5.0.1",
-    "ember-source": "~3.4.0",
+    "ember-source": "~3.2.0",
```
Turbo87: I'm assuming this is because of #450, but how is that related to this PR?

packages/card-picker/tests/dummy/config/optional-features.json
```diff
@@ -1,3 +1,3 @@
 {
-  "jquery-integration": false
+  "jquery-integration": true
```
Turbo87: why do we need jQuery for this?

packages/card-picker/tests/acceptance/close-button-test.js
```diff
@@ -0,0 +1,29 @@
+import { module, test } from 'qunit';
+import { visit, click, waitFor, waitUntil, find} from '@ember/test-helpers';
+import { setupApplicationTest } from 'ember-qunit';
+
+module('Acceptance | close button', function(hooks) {
+  setupApplicationTest(hooks);
```
Turbo87: any particular reason for using an application test over a rendering/integration test?

packages/card-picker/addon/templates/components/card-picker-toolbox.hbs
```diff
@@ -1,10 +1,20 @@
...
+  </button>
 </div>
 
 <div class="cards-list">
   {{#cardstack-search query=query as |content|}}
-    <div role="button" onclick={{action "select" content}}>
+    <div
+      role="button"
+      onclick={{action "select" content}}
+    >
```
Turbo87: unrelated change?

ef4:
I'm ok with merging after @Turbo87's questions are addressed and conflicts are resolved.

packages/card-picker/package.json
mansona: When you open the bottom edge it (eventually) instanciates the `cardstack-query-editor` from `@cardstack/search` which uses the `ember-data` store.

Looking at the `@cardstack/search` it seems that `ember-data` is not a dependency 🤔 I guess that should be changed.

@ef4 or @habdelra any idea why we're using a particular sha as an ember-data dependency on `@cardstack/search`? https://github.com/cardstack/cardstack/blob/master/packages/search/package.json#L44

packages/card-picker/package.json
mansona: Yes this was in relation to #450, more so the `ember-elsewhere` version bump than the `ember-let` issue. We can revert once we get a release of `@cardstack/edges`

Turbo87: shouldn't that be unnecessary due to the use of `lerna`?

mansona: eh... I don't think so 🤔 just because it explicitly depends on a particular version of edges in this package.json https://github.com/cardstack/cardstack/pull/451/files#diff-d6818a352bbaab3106246009a428b3e7R25

but then again I don't know if that is true 😂

packages/card-picker/tests/dummy/config/optional-features.json
mansona: This is because we need ember-data and it's still needed for now 👍

packages/card-picker/tests/dummy/app/router.js
```diff
@@ -7,6 +7,7 @@ const Router = EmberRouter.extend({
 });
 
 Router.map(function() {
+  this.route('close-button');
```
mansona: @jorgelainfiesta can we just get this changed to the application template please? Just so that when people spin up the dummy app for this it shows them something without needing to go to the `close-button` route

packages/card-picker/addon/templates/components/card-picker-toolbox.hbs
Turbo87: how did it pass before this PR?

packages/card-picker/package.json
habdelra: I’m unsure, @ef4 would have to answer that.

packages/card-picker/package.json
jorgelainfiesta: I tried to set `ember-source` back to `3.4.0` and do `lerna bootstrap` however the bottom edge simply stops working without throwing any errors.

packages/card-picker/addon/templates/components/card-picker-toolbox.hbs
jorgelainfiesta: I actually don't know, I guess it was before template linting was enabled?

packages/card-picker/tests/dummy/app/router.js
jorgelainfiesta: Sure, I've moved everything to the application route.

packages/card-picker/package.json
Turbo87: make sure to rebase. this should have been fixed by #453

Issue #389, opened 2018-10-29
Convert built-in editors to follow DDAU


balinterdi:
In #335 we got some weird test flakiness related to built-in editors. One suspicion ([that Ed shared](https://github.com/cardstack/cardstack/pull/335/#issuecomment-433456162)) is that this might have something to do with some of the built-in editors being two-way bound.

Turning the built-in editors to follow DDAU would probably do away with those issues.

balinterdi:
Implemented in #452

PR #452, opened 2018-11-19
Convert field editors to DDAU

balinterdi:
See #389

### DIFF
--- a/packages/core-types/tests/integration/components/field-editors/date-editor-test.js
+++ b/packages/core-types/tests/integration/components/field-editors/date-editor-test.js
@@ -4,53 +4,62 @@ import { render, triggerEvent } from '@ember/test-helpers';
 import EmberObject from '@ember/object';
 import hbs from 'htmlbars-inline-precompile';
 
+let input;
+
 module('Integration | Component | field editors/date editor', function(hooks) {
   setupRenderingTest(hooks);
+
+  hooks.beforeEach(() => {
+    input = '.field-editor > input';
+  });
 
   test('it renders', async function(assert) {
     this.set('model', EmberObject.create({
       expiration: '2030-01-01'
     }));
+    this.set('onchange', () => {});
     await render(hbs`
       {{field-editors/date-editor
         content=model
         field="expiration"
+        onchange=onchange
       }}`);
-    assert.dom('input[type=date]').hasValue('2030-01-01', 'date input has correct value');
-    assert.dom('input').isNotDisabled();
+    assert.dom(input).hasValue('2030-01-01', 'date input has correct value');
+    assert.dom(input).isNotDisabled();
   });
 
   test('it renders with an invalid date', async function(assert) {
     this.set('model', {
       expiration: 'pizza'
     });
+    this.set('onchange', () => {});
     await render(hbs`
       {{field-editors/date-editor
         content=model
         field="expiration"
+        onchange=onchange
       }}`);
-    assert.dom('input[type=date]').hasNoValue('date input has no value');
-    assert.dom('input').isNotDisabled();
+    assert.dom(input).hasNoValue('date input has no value');
+    assert.dom(input).isNotDisabled();
   });
 
   test('it can be disabled', async function(assert) {
     this.set('model', EmberObject.create({
       expiration: '2030-01-01'
     }));
+    this.set('onchange', () => {});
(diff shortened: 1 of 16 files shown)


packages/core-types/tests/integration/components/field-editors/date-editor-test.js
```diff
@@ -1,34 +1,64 @@
...
+        disabled=true
+      }}`);
+    assert.dom('input').isDisabled();
+  });
+
+  test('onchange is called when the field is left', async function(assert) {
+    let changed;
+    this.set('onchange', () => {
+      changed = true;
     });
```
Turbo87: an alternative pattern is this:
```js
this.set('onchange', () => assert.step('change'));
// ...
assert.verifySteps(['change']);
```

this has the advantage or verifying that the action was only called exactly once

packages/core-types/tests/integration/components/field-editors/integer-editor-test.js
```diff
@@ -40,17 +32,29 @@ module('Integration | Component | field editors/integer editor', function(hooks)
...
+      changed = true;
+    });
+    this.set('model', EmberObject.create({ rating: 3 }));
+    await render(hbs`
+      {{field-editors/integer-editor
+        content=model
+        field="rating"
+        onchange=onchange
+      }}`);
+    await triggerEvent('input', 'blur');
```
Turbo87: there is a dedicated [`blur()`](https://github.com/emberjs/ember-test-helpers/blob/master/API.md#blur) helper, which has a bit more code in it to make it work properly on all the browsers

packages/core-types/tests/integration/components/field-editors/date-editor-test.js
balinterdi: I love that. It was passive knowledge but it's so much better.

packages/core-types/tests/integration/components/field-editors/integer-editor-test.js
balinterdi: Didn't know about this but it'd be great as I've had problem with triggerEvent in the past. For some reason, `await blur('input')` doesn't work, though, it seems like it doesn't trigger the event.

balinterdi: Ignore this, the selector (`input`) is not specific enough so it finds another one. Interesting that it has worked with `triggerEvent` though.

balinterdi: There's still a timing issue in there. The test fails by default but if I put in a breakpoint in the `blur` method (I did this to check if the event is fired on the right element) it passes.

balinterdi: Yep, that works!

packages/core-types/tests/integration/components/field-editors/date-editor-test.js
```diff
@@ -1,34 +1,74 @@
...
 import hbs from 'htmlbars-inline-precompile';
 
+let input;
+
 module('Integration | Component | field editors/date editor', function(hooks) {
   setupRenderingTest(hooks);
 
+  hooks.beforeEach(() => {
+    input = '.field-editor > input';
+  });
```
Turbo87: why do you set that in `beforeEach`? would be easier to make it `const INPUT = '.field-editor > input';`, right?

packages/tools/addon/components/cs-field-editor.js
```diff
@@ -13,6 +13,13 @@ export default Component.extend({
   enabled: true,
   errors: null,
 
+  init() {
+    this._super(...arguments);
+    if (!this.onchange) {
+      this.onchange = () => {};
+    }
```
Turbo87: what's the advantage of this over just defining `onchange() {}` in the class itself?

Turbo87: what's the advantage of this over defining `onchange() {}` on the class itself?

balinterdi: Nothing, at one point I had a bug that I tried to trace down and forgot to refactor. Thanks!

balinterdi:
@Turbo87 If the tests pass, could you review once more and then merge (if there are no further issues)?

PR #460, opened 2018-11-22
Update `lerna` to v3.4.3

Turbo87:
see https://github.com/lerna/lerna/blob/master/CHANGELOG.md#300-2018-08-10


mansona:
So... I'm not sure the fact that Travis is passing is a valid test to see if this is working with the latest lerna 🤔 only just because I think that Tavis does not use lerna to install packages... I could be wrong.

Maybe @ef4 or @habdelra can weigh in on this one?

Turbo87:
I've used `lerna bootstrap` locally and it seems to still do the same things as before. Not sure what else to test or what else we use this for.

PR #461, opened 2018-11-22
Simplify `@cardstack/test-support` import paths

Turbo87:
There is no reason for us to use relative paths to reach into the `node_modules` of a different folder when we can import from `@cardstack/test-support` directly.

### DIFF
--- a/packages/hub/node-tests/computed-field-test.js
+++ b/packages/hub/node-tests/computed-field-test.js
@@ -1,9 +1,9 @@
-const JSONAPIFactory = require('../../../tests/sample-computed-fields/node_modules/@cardstack/test-support/jsonapi-factory');
+const JSONAPIFactory = require('@cardstack/test-support/jsonapi-factory');
 
 const {
   createDefaultEnvironment,
   destroyDefaultEnvironment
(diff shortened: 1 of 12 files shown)


mansona:
@Turbo87 I would be interested to know if this actually works (apart from the current eslint fail) because I wonder if it is affected by the same thing as is described in this comment: https://github.com/cardstack/cardstack/blob/master/packages/ephemeral/node-tests/ephemeral-test.js#L2 We may need to make sure that we don't cause issues with circular dependencies 🤔

ef4:
Yeah, this funny indirection was intentional.  It should only be necessary in packages that are themselves dependencies of the test-support package.

Turbo87:
@ef4 I'm wondering if there is a better way to do this. Would this be solved if the `test-support` package used peer dependencies instead?

ef4:
These things really aren’t peer deps of test-support. Plugin authors are supposed to be able to just add test-support and have that result in a working test environment.

Turbo87:
looks like lerna is not interested in supporting these kinds of situations: https://github.com/lerna/lerna/issues/1198#issuecomment-442278902 😕

PR #466, opened 2018-11-26
WIP: Fix unknown fields validation issues

Turbo87:
Resolves https://github.com/cardstack/deck/issues/250

In the problematic situation the schema is changed without adjusting the data, then the next patch request comes in which uses the new schema, but the patch data is merged with the old data from the original document, making the final document include fields from the old schema. This fails "unknown fields" validation on the final document.

There are multiple potential solutions to this:

### 1) adjust the `before + patch = after` behavior to only include fields that are known to the schema

the patching happens in each of the individual `Writer` classes, which don't have access to the schema, so this solution seems hard to implement

### 2) run "unknown fields" validation on the patch that comes in with the request, and not run it inside of `Schema#validate()`

it is not clear where else `Schema#validate()` might be used and it would need to duplicate quite a bit of validation logic

### 3) adjust the "unknown fields" validation to look at `before` *and* `after`

currently the validation only considers fields in the final document, but by adjusting the logic we can check if they were already included in the original document and not fail validation then. since the client will only get fields that are in the schema it should not be possible for it to send fields that are no longer part of the schema so this should only be a theoretical problem AFAICT.

@ef4 we have implemented the third solution for now, but we are unsure about the right way forward on this. would be good to get some feedback.

### DIFF
--- a/packages/hub/node-tests/unknown-fields-validation-test.js
+++ b/packages/hub/node-tests/unknown-fields-validation-test.js
@@ -60,7 +60,9 @@ describe('unknown fields validation', function() {
     expect(article.data.attributes).to.have.property('header', 'bar');
   });
 
-  it('backend has old field, request has old field', async function () {
+  // This is an artificial test case that should never actually happen in the app
+  // as the JSONAPI will only return fields that are included in the schema
+  it.skip('backend has old field, request has old field', async function () {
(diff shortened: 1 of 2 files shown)


ef4:
This is nice, good work. 👍

PR #471, opened 2018-11-26
Fix writable-fields in permissions response

balinterdi:


packages/data/addon/services/cardstack-data.js
```diff
@@ -136,7 +136,11 @@ export default Service.extend({
...
   getCardMeta(card, attribute) {
     if (attribute === 'human-id') {
-      return [`${capitalize(getType(card))} `, card.id]
+      let humanType = getType(card)
+        .split('-')
+        .map((s) => capitalize(s))
+        .join(' ');
+      return [`${humanType} `, card.id]
         .join('#');
     }
```
Turbo87: do we have any tests for this method? it's not directly obvious to me what it does 🤔

balinterdi: Because then I should come up with another type related to posts that has a compound name (so Comment won't do). I can do it, though, I was just lazy or, to put it in a more favorable light, didn't see as much value in testing this very case as to be worth it to add a further type. I think you're right, though, and I should do it.

Issue #448, opened 2018-11-19
Show unrendered fields from owned relationships in editor


balinterdi:
This is Phase 2 of #300 . Phase 1 has implemented showing unrendered fields of the resource-under-edit in the sidebar editor. This issue is about adding the unrendered (= not in the template) fields in records of owned relationships, too.

As @ef4 [pointed out](https://github.com/cardstack/cardstack/issues/300#issuecomment-435901143), this may need some design help as rendering all the fields of all records in a has-many relationship (to give an example) can make the editor pretty messy.

balinterdi:
Closed via #483

PR #479, opened 2018-11-28
Show unrendered fields from owned relationships

balinterdi:
Implements #448


balinterdi:
@ef4 This PR adds un-rendered fields from owned resources to the editor:



The next logical step is to fetch and use permissions for those records (we currently only fetch permissions for the "main" record of the page).

PR #481, opened 2018-11-28
[WIP] Include in document search

mansona:
Fixes https://github.com/cardstack/cardstack/issues/396

### Todo 
- [ ] Look into either updating chai or using a workaround for the older version of Chai for the `deep.includes` for objects
- [ ] Discuss potential performance hits as a result of this PR
  - do we have any performance benchmarks for the hub?
- [ ] Finish the last test of deeply nested relationships

### DIFF
--- a/packages/hub/bootstrap-schema.js
+++ b/packages/hub/bootstrap-schema.js
@@ -36,7 +36,7 @@ const models = [
           { type: 'fields', id: 'default-at-create' },
           { type: 'fields', id: 'default-at-update' },
           { type: 'fields', id: 'caption' },
-          { type: 'fields', id: 'searchable' },
+          { type: 'fields', id: 'include-in-document-search' },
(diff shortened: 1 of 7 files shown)


ef4:
I don't think we need `fullTextDoc`. It's not a format that we want to consume anywhere. The [only place its used](https://github.com/cardstack/cardstack/pull/481/files#diff-ca6fb7ebcbd4ab7ad3b0e7205c532d9aR208) immediately flattens the document down to its text, so we are spending a lot of effort generating structure that's never used.

I am picturing a method on DocumentContext that extracts the text directly as a string or list of strings. It can work off the searchDoc and extract the values from fields that are `field.includeInSearchDoc`, recursing down into related documents for which the relationship field is `field.includeInSearchDoc`.

ef4:
Closing until we're reading to work on finishing this.

PR #483, opened 2018-11-28
Allow editing un-rendered fields in owned resources

balinterdi:
This is done by fetching –and correctly setting– permissions for owned resources.

Todos:

* [x] Write test

### DIFF
--- a/packages/data/addon/services/cardstack-data.js
+++ b/packages/data/addon/services/cardstack-data.js
@@ -143,6 +143,9 @@ export default Service.extend({
       return [`${humanType} `, card.id]
         .join('#');
     }
+    if (attribute === 'uid') {
+      return `${getType(card)}/${card.id}`;
+    }
     return '';
   },
 
--- a/packages/tools/addon/components/cs-active-composition-panel.js
+++ b/packages/tools/addon/components/cs-active-composition-panel.js
@@ -20,7 +20,17 @@ export default Component.extend({
   },
 
   fetchPermissions: task(function * () {
-    let permissions = yield this.get('data').fetchPermissionsFor(this.model);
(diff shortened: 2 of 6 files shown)


balinterdi:
@ef4 I'd like to hear your take on whether the way I implemented this is good. I mostly think about my introducing the `cs-uid` helper which creates a unique id from the type and id of the resource (I need this to get the permissions for each model edited in the sidebar editor). It seems like `cs-uid` should either be an important feature of the framework or maybe it's not something we want to encourage (I need a way to get the same key for the same model object).

I originally wanted to use a WeakMap with the models as keys (I never used a WeakMap before and this seemed like a good use case :) ) but turns out the `get` template helper can't take objects as keys.

ef4:
I think adding `uid` to metadata is OK. Another option in this particular case is to use `WeakMap` to associate the permission with each model.

Am I correct that we render before the permissions have loaded, and while the permissions are loading the fields will appear as not-editable? If so, we are effectively already using a promise-proxy-like API, where the template refers to permissions that might not be available yet. In that case, we could encapsulate the permissions loading in a helper.

```diff
-(get permissions (cs-uid fieldModel.content))
+(cardstack-permissions fieldModel.content)
```

The value that comes back from the helper would update itself after the permission are available.

balinterdi:
> Am I correct that we render before the permissions have loaded, and while the permissions are loading the fields will appear as not-editable? If so, we are effectively already using a promise-proxy-like API, where the template refers to permissions that might not be available yet. In that case, we could encapsulate the permissions loading in a helper.

Yes, that's correct. We kick off fetching the permissions in `willInsertElement` so the permissions response might arrive before or after the component (the editor) has rendered. I think most of the time the component will render first but that's not too important.

PR #490, opened 2018-11-29
hub/schema: Use `realFields` in `withOnlyRealFields()`

Turbo87:
From what I understand `realFields` is not the opposite of `computedFields` because unknown fields are neither of those two. Since the function is called `withOnlyRealFields()` we should only include the `realFields`, or rename the function to `withoutComputedFields()`.

This code path was discovered during the work on #466.


Turbo87:
it seems that the test suite does not agree with this change. maybe renaming the function is the right way forward?

ef4:
I think your change was good and the test had a bug.

PR #491, opened 2018-11-29
eslint-config: Add `eslint-plugin-mocha` to default test config

### LINKED ISSUE #467: Use `eslint-plugin-mocha` for `node-tests` directories
see https://github.com/lo1tuma/eslint-plugin-mocha/

this can catch things like `describe.only()` early (see https://github.com/lo1tuma/eslint-plugin-mocha/blob/master/docs/rules/no-exclusive-tests.md) and also has other useful lint rules

Turbo87:
Resolves #467

### DIFF
--- a/packages/authentication/package.json
+++ b/packages/authentication/package.json
@@ -61,6 +61,7 @@
     "ember-try": "^0.2.23",
     "eslint": "^5.8.0",
     "eslint-plugin-ember": "^5.0.0",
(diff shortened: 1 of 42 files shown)


Turbo87:
I think `eslint-plugin-mocha` only adds rules, but no globals. But what we could certainly do is replace the globals list with `env: { mocha: true }`, which automatically applies those globals.

PR #492, opened 2018-11-30
Error in calculating modelFields

burieberry:
@ef4 @balinterdi  I came across this error when running the "deck" with the yarn-linked `restyling-right-edge` branch: `Uncaught TypeError: model.relatedOwnedRecords(...) is not a function or its return value is not iterable`. Below change seems to work... If not, feel free to take over.

### DIFF
--- a/packages/tools/addon/services/cardstack-tools.js
+++ b/packages/tools/addon/services/cardstack-tools.js
@@ -31,7 +31,7 @@ export default Service.extend({
     let model = this.get('activeContentItem.model');
     if (!model) { return []; }
 
-    let records = [model, ...model.relatedOwnedRecords()];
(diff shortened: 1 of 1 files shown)


ef4:
Thanks @burieberry!

balinterdi:
Duplication removed in #493 

@burieberry Could you check if this branch works for you? (again, master should also work fine and I'm not sure how you got the error but I want to make sure). Thank you.

burieberry:
> Duplication removed in #493
> 
> @burieberry Could you check if this branch works for you? (again, master should also work fine and I'm not sure how you got the error but I want to make sure). Thank you.

@balinterdi  Yes it works when I try this branch, thanks

balinterdi:
Thank you, I'm then closing this.

PR #495, opened 2018-12-03
Update ember-cli-qunit and @ember/test-helpers

balinterdi:


packages/tools/package.json
```diff
@@ -56,7 +57,7 @@
     "ember-cli-dependency-checker": "^3.0.0",
     "ember-cli-htmlbars-inline-precompile": "^1.0.3",
     "ember-cli-inject-live-reload": "^1.8.2",
-    "ember-cli-qunit": "^4.3.2",
+    "ember-cli-qunit": "^4.4.0",
```
Turbo87: while we're on it we might want to replace `ember-cli-qunit` with `ember-qunit`

packages/tools/package.json
```diff
@@ -26,6 +26,7 @@
     "@cardstack/plugin-utils": "0.11.9",
     "@cardstack/rendering": "0.11.9",
     "@cardstack/routing": "0.11.10",
+    "@ember/test-helpers": "^1.0.1",
```
Turbo87: this should not be added as an explicit dependency

balinterdi: Right, but the thing is, the latest ember-cli-qunit would've only pulled in an @ember/test-helpers version that only differs in a patch version (0.7.x).

I'll check what the deal is if I use ember-qunit directly. I'm starting to get the sense that I shouldn't ever use the `ember-cli-*` add-ons, I should just stick to the `ember-*` ones :)

PR #497, opened 2018-12-04
adding cardstack-catalog as a shareable component

mansona:
We originally added the `catalog` component and route to the deck project for ease of development. As we are now looking at building other apps using cardstack it makes sense for us to share the carstack-catalog helper component between instances.

I have this working on a new project by just dropping the following into a route template: 

```
// app/templates/catalog.hbs
{{cardstack-catalog}}
```


ef4:
This never had passing tests and has sat a long time. We can readdress if it becomes a priority.

PR #527, opened 2018-12-11
Add condition on rendering all editor fields

burieberry:


packages/tools/addon/templates/components/cs-active-composition-panel.hbs
```diff
@@ -71,35 +71,37 @@
...
-              (contains field.name fieldModelPermissions.writableFields)
-            )
-            onchange=(action "validate")
-            errors=(get validationErrors field.name)
-          }}
-        </div>
-      {{/cs-collapsible-section}}
-    {{/let}}
-  {{/each}}
+  {{#unless model.hideEditorField}}
```
marcoow: Is `hideEditorField` a good name as it disables **all** fields from being listed automatically?

burieberry: I'm open for suggestions. it doesn't hide all fields, just the ones that either aren't explicitly on the template or that are not editable

marcoow: Hm, maybe I'm mis-understanding what `modelFields` is - I read this as unless `model.hideEditorField` is `true`, iterate over all of the model's fields and show the respective editors?

burieberry: there's another block above this one that renders `renderedFields` on the editor panel. what I see when I inspect `renderedFields` vs. `modelFields`:

on Master Recording card where we have 4 fields:
`renderedFields`: 4 items
`modelFields: []`

on Master Artist Assignments List card:
`renderedFields`: only Artist Assignments field
`modelFields`: the other 7 items (that we don't want to display)

marcoow: Ah sorry, I get it then. Maybe something like `disableAutoEditorFields` or so would be more appropriate then?

burieberry: Works for me, changed it

marcoow: bit of a nitpick but I think `disableAutoEditorField**s**` would be better as it applies to multiple fields, not just one

burieberry: right. I couldn't remember if it was fine to pluralize in the schema since it's not a `has-many` relationship

Issue #518, opened 2018-12-10
Clear up cs-active-composition-panel template


balinterdi:
The template is getting long and though it's not obvious how to break it to smaller pieces (mostly because the difference between how we render `renderedFields` and `modelFields` there are a few things we can do to make it more readable:

* Create a helper for concating the `errors` class
* Move boolean logic into the component (e.g `cs-field-editor`)
* Stretch goal: move fetching permissions into the component where it's needed. Currently we fetch permissions to the "content" and all related, owned records and pass down the appropriate permissions object for those records from the top. If we did it in lower-level components, we might be able to get rid of the `cs-uid` setting, too.

balinterdi:
@ef4 Other than that, #528 is ready for review.

mansona:
@balinterdi can you create a separate issue that covers the "third bullet point" above please? I guess we can discuss the merits the implementation in that new Issue and we can close this one 👍

balinterdi:
Good idea, I did just that: https://github.com/cardstack/cardstack/issues/566

PR #528, opened 2018-12-12
Clear up cs-active-composition-panel

balinterdi:
See #518


mansona:
@balinterdi can you rebase or fix the conflicts on this one? I would like to get it merged 👍 

I am also assuming that if the tests pass in this case we can consider this ready to merge? I'm essentially asking: do we think that there is significant enough test coverage that we should feel confident in merging this refactor if it doesn't break any of our existing tests?

balinterdi:
@mansona I just rebased it and I'm confident that the tests will catch any eventual regression I managed to sneak in :)

balinterdi:
@mansona Can you merge this if it looks good to you?

Issue #482, opened 2018-11-28
Refactor cardstack-data#validate to use `relatedOwnedRecords`


balinterdi:
https://github.com/cardstack/cardstack/pull/479 implements a method on all cardstack model objects called `relatedOwnedRecords` that returns all of its related records for the model. That should be used to clear up the `validate` method in the `cardstack-data` service where we basically reimplement `relatedOwnedRecords` inside the function.

dnaidionov:
done!

balinterdi:
This should be fixed by #493 (when that's merged).

PR #529, opened 2018-12-12
Use relatedOwnedRecords method for validation

balinterdi:
See #482


packages/tools/tests/integration/components/cs-version-control-test.js
```diff
@@ -16,7 +14,12 @@ module('Integration | Component | cs version control', function(hooks) {
     // looking for its metadata, and ember-resource-metadata works
     // through ember-data's identity map.
     this.owner.register('model:thing', DS.Model.extend());
-    model = EmberObject.create({
+    // It's unfortunate we use a mock object for a full-fledged
```
ef4: I agree, these tests are fragile, next time we're in here we should refactor to generate actual models in the right state.

PR #530, opened 2018-12-13
Fix adding logged-in user details to git commits

mansona:
screenshot 2018-12-13 at 11 00 24

This implementation was already there, it just was accessing the wrong location of the session user (`user.attributes` instead of `user.data.attributes`).

### DIFF
--- a/packages/git/writer.js
+++ b/packages/git/writer.js
@@ -135,9 +135,16 @@ module.exports = class Writer {
 
   async _commitOptions(operation, type, id, session) {
     let user = session && await session.loadUser();
+    let userAttributes = {};
+    try {
+      userAttributes = user.data.attributes;
+    } catch (e) {
+      // do nothing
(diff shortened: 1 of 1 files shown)


packages/git/writer.js
```diff
@@ -135,9 +135,16 @@ module.exports = class Writer {
 
   async _commitOptions(operation, type, id, session) {
     let user = session && await session.loadUser();
+    let userAttributes = {};
+    try {
+      userAttributes = user.data.attributes;
+    } catch (e) {
+      // do nothing
+    }
```
marcoow: I don't think this is a good use of `try`/`catch` tbh. Why not

```js
let userAttributes = user && user.data && user.data.attributes;
```

instead?

marcoow:
> Currently there are no tests that cover the authentication/authorisation

😢

PR #531, opened 2018-12-13
fixing the implementation of the branch queryParam

mansona:
This may be a bit of a naive fix but it happens to work 😂 I've tested this locally by linking all the things and it's working as expected

We have decided to add testing to this functionality in a follow up issue


mansona:
I've discussed this with @habdelra and we will add a follow-up issue to cover this change with tests 👍 we don't want to block this feature on figuring out the test harness

mansona:
The follow-on issue to track how to test these features is here: https://github.com/cardstack/cardstack/issues/532

PR #533, opened 2018-12-13
running learna bootstrap on a clean system resulted in a new yarn.lock

habdelra:


Turbo87:
are we not using those dependencies anymore?

habdelra:
I think these are older versions that we no longer use

habdelra:
love my avatar on the commit here. it's the git version of "new computer, who dis?" 😆

PR #534, opened 2018-12-13
Branch picker

mansona:
This now fully implements the Branch Picker 👍 

It requires https://github.com/cardstack/cardstack/pull/535 to be able to get the current branches to choose from and https://github.com/cardstack/cardstack/pull/531 for any changes to be reflected in the hub

This branch is safe to merge before either of the other two, It just wouldn't make too much sense to 😂


packages/tools/addon/components/cs-branch-control.js
```diff
@@ -1,29 +1,51 @@
...
-  previewing: computed('enabled', 'tools.branch', function() {
-    return this.get('enabled') && defaultBranch !== this.get('tools.branch');
+  willInsertElement() {
+    this._super(...arguments);
+    this.get('fetchBranches').perform();
+  },
+
+  fetchBranches: task(function * () {
+    // checking if the branches() fucntion is available on this version of the data service
+    if((typeof this.get('data').branches) === 'function') {
```
marcoow: Do we actually need this?

mansona: well, that's a good question 😂 This is my attempt to work around the fact that this package will depend on the features in https://github.com/cardstack/cardstack/pull/535 and won't work until there is a version of the data package **released** and then that is the version that the current app has installed. Essentially depending on a feature like this across packages will cause there to be a **minimum version** of the `cardstack-data` package that is supported. Any other versions would cause a runtime error because of a `undefined is not a function` error.

Maybe I'm being a bit too "safe" here but I couldn't think of a way to get this working in a satisfactory way otherwise 🤔  I would love to get @habdelra and/or @ef4's opinions on this, I'm sure this idea of "minimum supported versions" must have come up in the past at some point.

marcoow: Isn't the way to make this safe, just updating the `tools`' packages [`@cardstack/data` dependency`](https://github.com/cardstack/cardstack/blob/master/packages/tools/package.json#L25) to the version that includes the `branches` method?

marcoow: 🤦‍♂️ never mind what I said above - the service is of course coming from the app's dependency as you say…

mansona: I'm going to merge this PR with https://github.com/cardstack/cardstack/pull/535 because they are going to need to be merged and released at the same time anyway

mansona:
Closing in favour of https://github.com/cardstack/cardstack/pull/535/files

I'm merging these 2 PRs together since they will need to be released together anyway 👍

PR #535, opened 2018-12-13
expose branches on the API

mansona:
This is a bit of a rudimentary implementation (that does actually work) designed to spark some conversation about the right way to achieve this same thing.

Currently, with these changes active I am able to make this request: 

```
curl 'http://localhost:3000/api/branches'
```

and get this response: 

```json
{
  "data": [{
    "id": "master",
    "type": "branches",
    "attributes": {}
  }, {
    "id": "face",
    "type": "branches",
    "attributes": {}
  }, {
    "id": "master-update",
    "type": "branches",
    "attributes": {}
  }, {
    "id": "sandbox-master",
    "type": "branches",
    "attributes": {}
  }]
}
```

Which corresponds to the branches on my local git repo: 

```bash
➜  application-data git:(face) git branch | cat
* face
  master
  master-update
  sandbox-master
```

### DIFF
--- a/packages/data/addon/services/cardstack-data.js
+++ b/packages/data/addon/services/cardstack-data.js
@@ -138,6 +138,20 @@ export default Service.extend({
     }
   },
 
+  async branches() {
+    let url = `${hubURL}/api/branches`;
+    let response = await fetch(url, {
+      headers: this._headers()
+    });
+    let { data } = await response.json();
+    let { attributes, relationships } = data;
+    return {
+      mayUpdateResource: attributes['may-update-resource'],
+      writableFields: relationships['writable-fields'].data
+        .map((field) => camelize(field.id))
+    }
+  },
+
   getCardMeta(card, attribute) {
     if (attribute === 'human-id') {
       let humanType = getType(card)
--- a/packages/jsonapi/middleware.js
+++ b/packages/jsonapi/middleware.js
@@ -250,6 +250,13 @@ class Handler {
   }
 
   async handleCollectionGET(type) {
+    if(type === 'branches') {
+      let branches = await this.indexers.branches();
+
+      this.ctxt.body = { data: branches.map(branch => ({id: branch, type: 'branches', attributes: {}}))};
+      return;
+    }
+
     let { data: models, meta: { page }, included } = await this.searcher.search(this.session, this.branch, {
       filter: this.filterExpression(type),
       sort: this.query.sort,
--- a/packages/hub/indexers.js
+++ b/packages/hub/indexers.js
@@ -42,6 +42,7 @@ const log = require('@cardstack/logger')('cardstack/indexers');
 const { declareInjections, getOwner } = require('@cardstack/di');
 const bootstrapSchema = require('./bootstrap-schema');
 const RunningIndexers = require('./indexing/running-indexers');
+const _ = require('lodash');
 
 module.exports = declareInjections({
   schemaLoader: 'hub:schema-loader',
@@ -100,6 +101,20 @@ class Indexers extends EventEmitter {
       { forceRefresh, hints },
       { singletonKey: 'hub/indexers/update', singletonNextSlot: true, expireIn: '2 hours' }
     );
+  }
+
+  async branches() {
+    let sources = this._dataSourcesMemo && this._dataSourcesMemo.dataSources;
+
+    let promises = Array.from(sources.values()).map((dataSource) => {
+      if(dataSource.indexer) {
+        return dataSource.indexer.branches();
+      }
+    });
+
+    let results  = await Promise.all(promises);
+
+    return _.chain(results).compact().flatten().uniq().value();
   }
 
   async _setupWorkers() {


habdelra:
seems reasonable. I think it makes sense to have this be part of the API as opposed to codegen'd, as the branches available change over time as branches are created for documents that are saved

habdelra:
one thought: do we care about permissions here? can any client make a request to see what all branches exist regardless of session?

habdelra:
Also, I wonder if we should provide schema for the branch content-types in the bootstrap schema. Doing that will automatically create the ember data models (via the codegen package) that the client can use to query for branches

mansona:
@ef4 so I've given it a go and this implementation "works" but I'm sure there are some things here that I'm missing 🤔

The most stark/important point of this new design is that when we're inside the branch-update a lot of the infrastructure is set up to make the change on the **current** branch that we're updating. As it doesn't really make sense to have the list of branches each on their respective branch I tried to cater for the system to always check the "controlling branch" for this stuff. 

Because of the things I needed to do to make it work, I feel like it maybe shouldn't be **inside** the branch-update itself, but maybe if we could tack-on a branch creation just outside here: https://github.com/cardstack/cardstack/pull/535/files#diff-cd32d350f2df14cc39ad9f2d171841d2R71 then it might make a bit more sense? 

Do you have any thought's on this implementation?

packages/data/addon/services/cardstack-data.js
```diff
@@ -138,6 +138,16 @@ export default Service.extend({
...
 
+  async branches() {
+    let url = `${hubURL}/api/branches`;
+    let response = await fetch(url, {
+      headers: this._headers()
+    });
+    let { data } = await response.json();
+
+    return data.map(branch => branch.id);
+  },
```
marcoow: Why is this not simply

```js
this.store.findAll('branch');
```

?

habdelra: that's a good point, if you emit schema via the static-models, the hub framework will codegen an ember data `branch` model for you.

mansona: I've updated this now to use the store directly 👍

marcoow: Is the `return branches.map(branch => branch.id);` still necessary then?

ef4: Yes, I think we should simplify this to just return the list of branch models.

packages/hub/node-tests/indexers-test.js
```diff
@@ -154,6 +154,11 @@ describe('hub/indexers', function() {
       });
 
       indexers.on('add', model => {
+        // ignore add events for branches
```
ef4: Why do we need this special case?

packages/jsonapi/middleware.js
```diff
@@ -250,7 +251,14 @@ class Handler {
   }
 
   async handleCollectionGET(type) {
-    let { data: models, meta: { page }, included } = await this.searcher.search(this.session, this.branch, {
+    let branch = this.branch;
+
+    if(type === 'branches') {
```
ef4: Putting a special case here is unlikely to be correct, because jsonapi is not the only place somebody might ask about branches.

If this special case is really needed, we should move it down into `hub/searchers.js`. Then it will also work for (for example) a plugin that is calling the searchers directly, or for a future GraphQL middleware plugin.

mansona: I have removed this special case and "moved it one layer down" but I think we have opened up a different set of unintended consequences as a result 🙈 I'll comment on the issue so I can bring it to your attention 👍

packages/hub/searchers.js
```diff
@@ -137,10 +152,18 @@ class Searchers {
...
     if (arguments.length < 3) {
       throw new Error(`session is now a required argument to searchers.search`);
     }
+
+    let branch = branchRequest;
+
+    // only ever get the branches off the controlling branch
+    if (get(query,'filter.type.exact') === "branches") {
+      branch = this.controllingBranch.name;
+    }
```
mansona: I have a strong feeling that this is almost certainly **not** what we want 😖 essentially the issue here is that the way search works and how it is related to the query builder. There are theoretically many (infinitely?) more queries that could be made that people might expect to include branches, but the only way to return branches in this implementation is to have a query with the `filter.type.exact` set to `branches`. There are other public functions on this class like `searchFromControllingBranch()` that implies that the "special case" implementation might be better suited one level up, like in the `@cardstack/jsonapi` package where it was originally. I accept and agree with your assertion that it didn't feel right in the jsonapi package, especially if we are expecting other public APIs in the future like graphql, but I guess I don't see a clear way forward from here 🤔

mansona: @ef4 do you have any thoughts?

packages/hub/indexing/running-indexers.js
```diff
@@ -88,6 +88,15 @@ module.exports = class RunningIndexers {
...
     await this._loadSchemaModels();
+
+    await Promise.all(Object.keys(this.branches).map((branch) => {
+      return this.branches[this.controllingBranch].add(
+        'branches',
+        branch,
+        { data: {id: branch, type: 'branches', attributes: {}}},
+      );
+    }));
+
```
habdelra: @mansona  this specific chunk of code is what is causing the test failures that you were asking me about.
```
error: database "pgsearch_test_1547739697309" does not exist
```
I don't know why, though--perhaps @ef4 does....

packages/hub/searchers.js
ef4: I think you are right to wonder about this very specific query matching. In general it's probably intractable to find all the ways that people could ask about branches and special case them.

I filed https://github.com/cardstack/cardstack/issues/610 as a followup to come back to this and make it better.

PR #541, opened 2018-12-14
Field editor style fix

burieberry:


packages/tools/addon/styles/cs-field-editor.css
```diff
@@ -108,4 +92,4 @@
 .cs-field-editor .ember-power-select-trigger:focus {
     border-color: var(--active-foreground);
     outline: none;
-}
+} */
```
habdelra: we should just delete this if we dont need it any more

PR #543, opened 2018-12-17
Move applyReadAuthorization to DocumentContext

balinterdi:


packages/hub/indexing/document-context.js
```diff
@@ -107,6 +108,203 @@ module.exports = class DocumentContext {
...
+          output = { data: document.data };
+          if (document.meta) {
+            output.meta = document.meta;
+          }
+        }
+        output.included = safeIncluded;
+      }
+    }
+
+    if (output !== document && output.included) {
```
habdelra: @balinterdi did you actually mean to do a deep equal here? because this is just testing to see if the memory references are the same, which is not the same as deep equality (you need to use lodash for that)

packages/hub/indexing/document-context.js
```diff
@@ -107,6 +108,203 @@ module.exports = class DocumentContext {
...
+        // is a modified copy, we need to bring the original included
+        // list along. If our document is not a copy, it already has
+        // the original included list.
+        if (output !== document) {
+          output.included = document.included;
+        }
+      } else {
+        // we need to modify included. First copy the output document
+        // if we didn't already.
+        if (output === document) {
```
habdelra: @balinterdi did you mean a deep equal here? because this is not testing for the object equality

packages/hub/indexing/document-context.js
```diff
@@ -107,6 +108,203 @@ module.exports = class DocumentContext {
...
+        data: authorizedResource
+      };
+      if (document.meta) {
+        output.meta = document.meta;
+      }
+    }
+
+    if (document.included) {
+      let safeIncluded = this._readAuthIncluded(document.included, userRealms);
+      if (safeIncluded === document.included) {
```
habdelra: @balinterdi did you mean a deep equal here? because this is not testing for the object equality

habdelra: (and also since this is `included` you might actually want to test for `Set` equality, as opposed to `Array` equality.

packages/hub/indexing/document-context.js
```diff
@@ -107,6 +108,203 @@ module.exports = class DocumentContext {
...
+        // the include list didn't need to be modified. If our output
+        // is a modified copy, we need to bring the original included
+        // list along. If our document is not a copy, it already has
+        // the original included list.
+        if (output !== document) {
```
habdelra: @balinterdi did you mean a deep equal here? because this is not testing for object equality

packages/hub/indexing/document-context.js
```diff
@@ -107,6 +108,203 @@ module.exports = class DocumentContext {
...
+  // This takes an arbitrary JSON:API document and makes it safe to
+  // show within the given context. If it can't be shown at all
+  // (because the primary resource is not readable), we return
+  // undefined.
+  //
+  // Otherwise, we go through each included resource and field to
+  // ensure there are valid grants for them. We return a new sanitized
+  // document that strips out any included resources or fields that
+  // were lacking grants.
+  async applyReadAuthorization(document, context={}) {
```
habdelra: @balinterdi @ef4 why does this operate on an arbitrary document as opposed to the document contained in this DocumentContext instance? I'm concerned about sharing the DocumentContext internals outside of its context. Maybe we should be calling `applyReadAuthorization()` without passing a document to it, so that `applyReadAuthorization()` operates on its `this.upstreamDoc`.

packages/authentication/cardstack/middleware.js
```diff
@@ -262,8 +262,13 @@ class Authentication {
...
 
   async _applyReadAuthorization(session, user) {
+    let branch = this.controllingBranch.name;
     let schema = await this.currentSchema.forControllingBranch();
-    let authorizedUser = await schema.applyReadAuthorization(user, { session });
+    let documentContext = this.searcher.createDocumentContext({
+      branch,
+      schema,
+    });
+    let authorizedUser = await documentContext.applyReadAuthorization(user, { session });
```
habdelra: @balinterdi this doesnt make sense to me--why do you create a DocumentContext with no document? it seems as though the only reason to create a document context it to leverage the "ad hoc" read authorization against a document

packages/hub/searchers.js
```diff
@@ -161,14 +161,15 @@ class Searchers {
...
         branch,
         schema,
         includePaths,
         upstreamDoc: result,
-      }).pristineDoc());
+      });
+      let pristineResult = await documentContext.pristineDoc();
 
-      let authorizedResult = await schema.applyReadAuthorization(pristineResult, { session });
+      let authorizedResult = await documentContext.applyReadAuthorization(pristineResult, { session });
```
habdelra: @balinterdi we should not have the ability to apply read authorization against any document--rather applying read authorization for a context should just apply it against it's own document. it seems like this pattern that we have can be easily misused by using document contexts to apply read authorization against documents that have nothing to do with the contexts

packages/hub/searchers.js
```diff
@@ -86,7 +86,7 @@ class Searchers {
       });
       let pristineResult = await documentContext.pristineDoc();
       if (pristineResult) {
-        authorizedResult = await schema.applyReadAuthorization(pristineResult, { session, type, id });
+        authorizedResult = await documentContext.applyReadAuthorization(pristineResult, { session, type, id });
```
habdelra: @balinterdi same question around applying document context to an aritrary document vs the pristine document already contained in the document context. I think we are abusing document context by allowing it apply read authorization against documents that are not part of its context.

packages/hub/writers.js
```diff
@@ -80,7 +81,7 @@ class Writers {
       if (pending) { await pending.abort();  }
     }
 
-    return schema.applyReadAuthorization(pristine, { session });
+    return context.applyReadAuthorization(pristine, { session });
```
habdelra: @balinterdi same question around applying document context to an arbitrary document vs the pristine document already contained in the document context. I think we are abusing document context by allowing it apply read authorization against documents that are not part of its context.

packages/hub/writers.js
```diff
@@ -120,7 +122,7 @@ class Writers {
...
```
habdelra: @balinterdi ditto here too

packages/hub/indexing/document-context.js
balinterdi: This was copied verbatim from package/hub/schema/index.js and I'm not sure about the reasons it was done that way then. I'll call this Answer No. 1 so that I can refer to it in my other comments :)

packages/hub/indexing/document-context.js
balinterdi: See Answer 1.

packages/hub/indexing/document-context.js
balinterdi: See Answer 1.

packages/hub/indexing/document-context.js
balinterdi: See Answer 1.

packages/hub/indexing/document-context.js
balinterdi: See Answer 1.

packages/authentication/cardstack/middleware.js
balinterdi: Yes, good point, I'm not sure. We discussed moving the the `_applyReadAuthorization` method here with @ef4 and at the time I understood some of the reasons. We don't have a document here that we could create a document context for, do we? Maybe `user` is one?

packages/hub/searchers.js
balinterdi: Maybe, but do you think the behavior has changed between the old and new version? The main difference I see is that we now call `applyReadAuthorization` on the document context (which we also had in the previous version) whereas we now call it on the DC itself.

packages/authentication/cardstack/middleware.js
habdelra: I think `user` is the document in this case

ef4: Yes, this code is applying read auth to the `user` document.

packages/hub/searchers.js
habdelra: well the purpose of DC is a bit different. it represents saved work for a specific document. I dont think we are using it wrong now, but the API makes it really easy to abuse this.

ef4:
(Already discussed with hassan, but for posterity:)

The equality checks here are because we do copy-on-write internally while sanitizing documents. If nothing needs to be changed we return the original object. So deepEquals is not the intent.

balinterdi:
Ok, so after having read @ef4's comments, too, I think I see what needs to be done. The document shouldn't passed in to `applyReadAuthorization` as it opens up the API for abuse and, since the document is part of the DC, it can be accessed (and used) from there.

@habdelra I'll rewrite this on my next working da. Or have you already started the refactoring because you need it for what you're working on?

habdelra:
yeah--its intimately related to what I'm working on. I've already started working thru it. it would be nice if you let my refactor handle this as the merge conflict will be nightmarish otherwise.

balinterdi:
Sure thing, thank you!

PR #546, opened 2018-12-18
[WIP] Finish saving related records before saving self

bagby:
Basically makes everything in the `save()` chain `async`. Uses `ember-concurrency` tasks to simulate async `computed`s, which seems a little janky but is a pattern I've seen recommended elsewhere.

**[WIP]** because I haven't even looked at tests and whether or not they're passing.

### DIFF
--- a/packages/tools/addon/services/cardstack-tools.js
+++ b/packages/tools/addon/services/cardstack-tools.js
@@ -11,6 +11,7 @@ import injectOptional from 'ember-inject-optional';
 import { defaultBranch } from '@cardstack/plugin-utils/environment';
 import { guidFor } from '@ember/object/internals';
 import { get } from '@ember/object';
+import { task } from 'ember-concurrency';
 
 export default Service.extend({
   overlays: service('ember-overlays'),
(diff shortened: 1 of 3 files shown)


packages/tools/addon/services/cardstack-tools.js
```diff
@@ -48,10 +50,10 @@ export default Service.extend({
       });
       return fields;
     });
-    return modelFields.reduce((flattened, fields) => {
+    this.set('modelFields', modelFields.reduce((flattened, fields) => {
       return flattened.concat(fields);
-    });
-  }),
+    }));
+  }).observes('_renderedFieldNames', 'activeContentItem.model').on('init'),
```
Turbo87: is there no other way to implement this that doesn't use observers?

bagby: my understanding is that E-C doesn't use Ember observers under the hood, it has its own mechanism for handling this. Documentation for E-C `observes` is sparse, however, so don't hold me to that 🙈

Turbo87: > my understanding is that E-C doesn't use Ember observers under the hood, it has its own mechanism for handling this

http://ember-concurrency.com/api/task-property.js.html indicate that it's using `import { addObserver } from '@ember/object/observers'` under the hood. which is actually a little worrying to me as it does not seem to import `removeObserver` in that file... 🤔

bagby: hmmm this could explain why there are failing `tools` tests 😆I will investigate further.

bagby:
@Turbo87 I think I have a much simpler way of doing this without using `ember-concurrency` at all.

bagby:
Superceded by #547

PR #548, opened 2018-12-18
Ethereum plugin care and feeding

habdelra:
This PR allows the ethereum plugin to work in node 10 (which I think means this project is now 100% node 10 compatible). Additionally some cleanup on attr names to be more consistent.

in my local dev env, after making this change the ethereum plugin is now passing in node 10.


tests/ethereum-computed-fields/node_modules/@cardstack/test-support
```diff
@@ -0,0 +1 @@
+../../../../packages/test-support
```
Turbo87: is that a committed symlink?

habdelra: Yes, that is correct.

Turbo87: was that intended?

habdelra: Yes that is intended. The module itself is the thing being tested. In that test/ folder you’ll lots of modules that we are testing like this. And we use symlinks like this for the module to refer back to the @cardstack modules.

Turbo87: doesn't that make it impossible to use this repo on Windows? or does Windows support this too?

Turbo87: and why do we use a symlink instead of relying on yarn workspaces or lerna to resolve that `devDependency` correctly? if the answer is "cyclic dependency issues" when why not use relative paths to import from it like we do in some of the other tests?

habdelra: So all the node_modules folders under test/ are _very_ special. They are special enough we probably need a readme in this directory. The contents of these   node_modules are not artifacts from npm, but rather test data (that is committed into our repo) to test that cards that are provided as modules can load into the hub correctly (via the plugin loader). We don’t want to yarn or npm install these things, as they are the actual test. I’m struggling to find the right way to describe this, maybe @ef4 can do a better job.

habdelra: If you’ll recall the cards in dotbc and deck are node modules, the modules in the test/ folder are sample cards in that same structure where they have a package.json and a “manufactured” node_modules folder for each card that we use to test that the card loading of external cards provided as npm modules can work correctly. Is that making sense?

habdelra: Here’s other examples: https://github.com/cardstack/cardstack/tree/master/tests/stub-project/node_modules
https://github.com/cardstack/cardstack/tree/master/tests/plugin-loader-test-project/node_modules
https://github.com/cardstack/cardstack/tree/master/tests/pgsearch-test-app/node_modules
And so on...
Every test cardstack application in the tests/ folder follows this approach, where we setup a test cardstack app with artificial node_nodules

habdelra: In terms of windows compatibility, @bagby was using windows for awhile. I think he reported being able to run the tests in this repo. Although I think he had much more difficulty with docker in Windows TBH.

bagby: @Turbo87 you can indeed use symlinks in Windows, but it requires that every process is using elevated Administrator privileges, which mainly amounts to your editor and terminal (I used VSCode and ConEmu/gitbash primarily), which *by default* do not run in admin, you have to jump through several hoops to make admin the default. But it can be done.

bagby: And yes, as @habdelra said, I was able to run all the tests in this monorepo in Windows. As to whether or not I would *recommend* Cardstack development (or any Ember/node development in general) in Windows...Docker and nodegit were particular sources of misery, as well as Windows' insistence on continually scanning the entire filesystem for changes, which as you can imagine, did not play nice with Broccoli.

PR #549, opened 2018-12-18
Remove Ember.Map deprecations

bagby:
This mainly pertains to running tests, it was proving very difficult to find things with all the deprecation noise.


ef4:
I think there are still a bunch of other packages that depend on ember-data? If anybody adds one of those, it has a chance of winning.

habdelra:
Closing since this is addressed by https://github.com/cardstack/cardstack/pull/676

Issue #478, opened 2018-11-28
Add mechanism for acceptance-testing the interplay of a card and the editor


marcoow:
While working on cardstack/deck#284 we realized there is currently no good (or at least no officially sanctioned) way that we were aware of for testing the interplay of a card and its templates with the editor, in particular the correct usage of `cs-field` and `cs-field-group`.

These tests should probably be some sort of acceptance tests as they require multiple parts of the application (the card itself, its backing data as well as the editor). It seems like the key thing to figure out would be how the tests would interface and assert on element of the editor - e.g. if a test would want to assert the presence of an editor input field for a particular content element it does not seem like a good idea to assert on the presence of an input field with a particular ID as that makes the card test dependent on internals of the editor so that it could break when updating to a newer cardstack version that contains changes to the editor UI while in fact everything still works. We might require some sort of interface that assertions could be based on or a set of test helpers that allow this kind of assertion and that are kept in sync with the internals of the editor UI.

habdelra:
I had one note around the routing. for the integration tests we actually run `renderCard()` against the default router (@cardstack/routing/cardstack/default-router.js). For acceptance tests we'd want to also make sure that we are doing something similar, specifically mounting the cardstack routes in the ember router. Might not be a bad idea to have a card blueprint the framework itself to set this in the dummy app for acceptance tests. Specifically it should look like this: https://github.com/cardstack/cardstack/blob/master/packages/routing/tests/dummy/app/router.js

marcoow:
I'm not sure why we wrote _"acceptance test"_ above - I don't think we need a full acceptance test for this as the only thing that we need to be active in the test in addition to the card itself are the tools. If we can render something like

```hbs
{{#cardstack-edges}}
  {{cardstack-content content=card format=format params=params }}
{{/cardstack-edges}}
```

in `renderCard`, that should be sufficient for an integration test that tests the interplay between the card and the editor. I guess we could just add a flag to `renderCard` that tell it to also render the tools?

habdelra:
That seems perfectly reasonable

habdelra:
Oh also you’ll need to pass a session too (since the tools leans heavily on grants). We have a privileged session available for qunit, let me double check how we leverage that ...

habdelra:
[...]

PR #550, opened 2018-12-19
test-support: Add more test helpers

Turbo87:
This PR primarily introduces the `setupCardTest()` helpers as a replacement for `setupRenderingTest()`, which it still uses under the hood. It also sets up a `cardstack-card-test` component that can be used to render cards in a more composable way compared to `renderCard()` (which this PR deprecates)

mostly extracted from https://github.com/cardstack/dotbc/pull/161, resolves https://github.com/cardstack/cardstack/issues/478


packages/test-support/addon-test-support/test-helpers.js
```diff
@@ -39,4 +83,52 @@ export function renderCard(type, id, format, options = {}) {
...
+    });
+
+    CardTestComponent.reopenClass({
+      positionalParams: ['type', 'id'],
+    });
+
+    this.owner.register('component:cardstack-card-test', CardTestComponent);
+
+    this.owner.register('template:components/cardstack-card-test',
+      hbs`{{#if card}}{{cardstack-content event-isolated content=card format=format params=params}}{{/if}}`);
```
marcoow: I assume `event-isolated` is copied from somewhere and could either be removed or should be dynamic?

Turbo87: It's copied from the existing `renderCard()` helper. I don't know what it does and would prefer not to touch it for now.

Turbo87:
> This looks good to me but I'm not sure it resolves #478

true, it doesn't resolve all of it yet. I've adjusted the PR description accordingly.

packages/test-support/addon-test-support/test-helpers.js
marcoow: I think you can just remove it - `cardstack-content` doesn't have any positional parameters anyway and I think this is simply ignored.

Turbo87: then why is `renderCard()` using it? /cc @habdelra

habdelra: think @marcoow is right, it looks like a copy/paste mistaketo me.

Turbo87: remove with the latest commit

PR #551, opened 2018-12-19
Handle case when related records are missing

balinterdi:


ef4:
Let's move this filter inside the `relatedRecordsFor` function. Otherwise this same bug will strike the other place that uses it.

PR #556, opened 2018-12-20
Implement more test helpers

Turbo87:
This should resolve https://github.com/cardstack/cardstack/issues/478


Turbo87:
> Is there any way we could have tests using these helpers in the `test-support` package itself?

I have not found a way to do that yet... :(

ef4:
One way we have solved a similar problem (using the test-support package in tests in packages that are themselves used by the test-support package) is to create a stub application under /tests to use as the context for the test (the directory passed to `createDefaultEnvironment`).

Turbo87:
I'd be happy to look into that, but I'd prefer to get this merged for now so that we can take advantage of these helpers in dotbc already, while I figure out how to test them.

PR #557, opened 2018-12-21
Adding a fixed decimal data type

bagby:
This is meant for things like percentages, where we want accuracy but we don't want floating point nonsense, i.e. `13.999999999999031`


packages/core-types/cardstack/field-types/fixed-decimal.js
```diff
@@ -0,0 +1,8 @@
+module.exports = {
+  valid(value) {
+    return typeof value==='number';
+  },
+  buildQueryExpression(sourceExpression, name){
```
bagby: I honestly don't know what this function does.

bagby:
@ef4 this nearly broke my brain, particularly the `postgresql` part of it. Would ❤️ it if you could take a gander at it.

packages/postgresql/row-to-doc.js
```diff
@@ -46,6 +46,7 @@ function convertValue(pgValue, fieldType) {
   case '@cardstack/core-types::boolean':
   case '@cardstack/core-types::date':
   case '@cardstack/core-types::integer':
+  case '@cardstack/core-types::fixed-decimal':
```
bagby: I literally spent hours looking for this.

packages/core-types/cardstack/field-types/fixed-decimal.js
ef4: This is needed because of how types work in postgres. If somebody is trying to write a `filter` or `sort` that depends on a fixed-decimal field, we need to know how to exact that value out of a JSON document.

`sourceExpression` is some postgres expression that results in the JSON for a model and `name` is the name of a fixed-decimal field that we want to extract from that model.

The [`->>` operator is documented here](https://www.postgresql.org/docs/10/functions-json.html#FUNCTIONS-JSON-OP-TABLE). It extracts the value from the json, but as text. So it also needs to be cast to a number, which you're doing with `::real`.

Postgres does have a native `decimal` type that might also work here instead of `real`.

packages/postgresql/row-to-doc.js
ef4: Ugh, sorry, I realize this is a megatroll. Maybe we can make it throw? The intention was to not break if a table has some field we don't know how to handle.

ef4:
This is looking good, my only concern is that the Javascript representation for this still going through float. Your default field renderer will always round to look right, but as soon as people start doing math with these values they can still get into floating-point trouble.

If we want to truly handle these as fixed-decimal, our representation needs to be based on integers only. This is the same idea as storing prices in cents, so that you never get into floating point rounding issues.

One option is to represent the values as rationals, so 62.45 would be `[6245, 100]`. As in, the actual value you get off the model is an array of two integers. `buildQueryExpression` would still cast that to a true postgres `decimal` so sorting & filtering work right, and the default field renderer would know how to display the value as it displays now.

bagby:
@ef4 the issue with `numeric`/`digital` type is that `pg-types` automagically converts it to a string to [avoid loss of precision](https://github.com/brianc/node-pg-types/issues/28#issuecomment-111779909), so the number will be stored in the index as a string. As pointed out in the link, we could force it to _not_ do that, using something like:

```javascript
const { types } = require('pg');

types.setTypeParser(1700, function(val) {
  return parseFloat(val);
});
```

although I'm not convinced that's better than just leaving it as a string.

> One option is to represent the values as rationals, so 62.45 would be [6245, 100]

Interesting idea. For my current use case, I don't need to do math-y stuff (yet), perhaps we could make that a separate data type for when we need precision?

ef4:
> @ef4 the issue with numeric/digital type is that pg-types automagically converts it to a string to avoid loss of precision, so the number will be stored in the index as a string.

This is what `buildQueryExpression` is for. You don't need to let pg-types see the actual numeric.

>  For my current use case, I don't need to do math-y stuff (yet), perhaps we could make that a separate data type for when we need precision?

I really don't want to have a decimal type in the core framework that can't actually do decimal math.

If you're blocked and want to use this code, I think you should just move it into your app. It would all work the same, it doesn't need to live in core-types.

bagby:
> If you're blocked and want to use this code, I think you should just move it into your app. It would all work the same, it doesn't need to live in core-types.

Fair point, I'll probably just do that. Closing this for now, we can revisit in the future.

Issue #519, opened 2018-12-10
Merge cardstack-data#getCardMetadata to getCardMeta


balinterdi:
We have two methods for the same thing, living on the same object (the cardstack-data service). We should rewrite the only place `getCardMetadata` is called from to use `getCardMeta` as that's the method we want to use going forward.

See #521

marcoow:
I'm not sure these methods are for the same thing although they are similarly named of course. `getCardMeta` seems like being used for getting a string identifier for a card (something like a custom `toString`?) while `getCardMetadata` is meant to be the way to get arbitrary metadata for a card. The current implementation is really just a placeholder but the idea is that as we have that now, we can change how this works under the covers, potentially with an implementation as sketched out here: https://github.com/cardstack/dotbc/issues/156#issue-391796369

balinterdi:
We'd discussed this with @ef4 and we want a single method on the service that restrains the "public API" of cards.

`getCardMetadata` gives you anything which is undesired while `getCardMeta` only allows a set of meta attributes (that should be extended as we have requirements). We still reach into cards in some places and we should get rid of these and move them to `getCardMeta`.

ef4:
Yes, these two methods are supposed to be doing the same thing. They just take very different approaches to filling in an initial basic implementation.

It's better if we can be explicit about what metadata fields are in use (more like the current getCardMeta), but if we need to fall back to the card's own field of the same name (like the current getCardMetadata) as a temporary measure that would be OK.

balinterdi:
> It's better if we can be explicit about what metadata fields are in use (more like the current getCardMeta), but if we need to fall back to the card's own field of the same name (like the current getCardMetadata) as a temporary measure that would be OK.

And that's exactly what the PR does in #521 (providing the fallback with the comment that this should only be temporarily allowed).

PR #559, opened 2018-12-21
Use cardstack data service to get model title

burieberry:


burieberry:
Actually WIP: this PR will need to change after #519 via #521

balinterdi:
@burieberry Please use `getCardMeta` instead of `getCardMetadata`. As an ongoing effort, we'd like to 1) get rid of `getCardMetadata` as it's a duplicate method and 2) only expose certain meta properties of cards instead of allowing reaching into them for whichever property.

There's already an issue (#519) and a PR (#521) that merges the two methods into `getCardMeta`. There you'll also find two examples of exposing meta properties (`human-id` and `uid`). You can add `model-title` (or maybe just title?) as a third one.

Let me know if you have any questions regarding this. Thank you.

burieberry:
@balinterdi did as you said

packages/tools/addon/components/cs-composition-panel-header.js
```diff
@@ -1,9 +1,18 @@
...
 
 export default Component.extend({
   layout,
+  cardstackData: service(),
   tagName: 'header',
   classNames: ['cs-composition-panel-header'],
-  classNameBindings: ['editingEnabled:enabled:disabled']
+  classNameBindings: ['editingEnabled:enabled:disabled'],
+
+  getModelTitle: task(function * () {
```
balinterdi: I'm not sure we need to launch a task on init here, we could just use a CP:

```js
title: computed('model', function() {
  return this.cardstackData.getCardMeta(this.model, 'title');
})
```
(that's simpler and should `model` ever change, it will be recomputed)

burieberry: Sounds good, thanks!

balinterdi:
@burieberry Let me know if you'd like me to (formally) review.

burieberry:
@balinterdi sure, thank you!

balinterdi:
As #521 has been merged, I've merged this one, too.

PR #562, opened 2018-12-28
Ethereum address indexing

habdelra:
This PR provides support for indexing contract agnostic ethereum addresses, and providing support for associating ethereum addresses to user sessions.

Ethereum transactions that do not involve a smart contract are discovered in a fundamentally different manner than transactions that do involve smart contracts. Specifically non-contract based transactions do not fire events. In order to detect transactions against specific addresses, the indexer must introspect mined blocks to discover transactions that involve addresses of interest. There are a couple heuristics that we use to lessen the burden on the indexer in order to discover transactions of interest.

The primary approach here is that we are introducing a new indexer class within the `@cardstack/ethereum` plugin that is used when the ethereum data-source is configured to index addresses instead of contracts. An important requirement to note here is that the addresses that are indexed need to be dynamically specified at run time, so that as users indicate they are interested in indexing an address, the hub can begin searching for these addresses of interest. Likewise, when users indicate they are no longer interested in an address, the hub will stop paying attention to the address. To support this capability, we introduce a content type called `tracked-ethereum-address`. The id for documents of this content type is simply the ethereum address to track. Additionally, since the ethereum data source is read-only, in order to persist the tracked ethereum addresses, the data source to persist the tracked ethereum addresses is specified in the datasource config. So in effect, a writable data source becomes the backing store for the ethereum data source in order to persist the tracked ethereum addresses. 

Behind the scenes the ethereum data source uses a writer that intercepts requests to create and delete `tracked-ethereum-addresses`, this serves as a shim that gives us an obvious place to kick off indexing for a new address or to cleanup the index for an address that we no longer want to track. The ethereum plugin actually uses a `proxied-tracked-ethereum-addresses` content type in the writeable data source -- but that is meant to be an implementation detail. The public API for cards that want to track ethereum addresses is the `tracked-ethereum-addresses` content type. 

Likewise for associating sessions with tracked ethereum addresses we use the same approach. The `user-ethereum-addresses` content-type is one that any user can create or delete (a grant exists that allows only "self" CRUD operations). When a new `user-ethereum-addresses` is created, the ethereum writer will create a new `tracked-ethereum-addresses` document. Computed properties are then used to surface the indexed `ethereum-addresses` and the associated `ethereum-transactions` for the tracked address to the `user-ethereum-addresses` document. Additionally, when a `user-ethereum-addresses` document is deleted, if it is the last `user-ethereum-addresses` document to point to a particular ethereum address, then its associated `tracked-ethereum-addresses` document will be removed, thus stopping the indexing for the address and cleaning up any artifacts for the indexed address in the index.

Cards can track ethereum addresses for all sorts of different reasons. For the scenario where cards wish to track ethereum addresses because they are associating users to addresses (which is probably a really obvious reason for tracking address) we provide this additional capability. For cards that consume the capabilities exposed in this PR, there are basically 2 specific content-types for cards to leverage.
1. For cards that are interested in tracking ethereum addresses and their associated txn's but that don't care about relating users to addresses, these cards should leverage the `tracked-ethereum-addresses` content-type as their interface to this data source.
2. For cards that are interested in tracking ethereum addresses by user, they should leverage the `user-ethereum-addresses` content-type as their interface to this data source.

For both of these 2 approaches, they will trigger the content types `ethereum-addresses` and `ethereum-transactions` to appear in the index (or disappear from the index, as the case may be). Cards using approach #&#x2060;2 can see the ethereum addresses and transactions by following the `user-ethereum-addresses.address-data` relationship in their hub features and components. Cards using approach #&#x2060;1 are operating a bit "closer to the metal" and will need to search for the `ethereum-addresses` that they care about in their hub features and components based on their own specific requirements around why they are tracking ethereum addresses.


habdelra:
just testing discord webhook

habdelra:
An amendment to the PR description: with the commit https://github.com/cardstack/cardstack/pull/562/commits/758912e7cacefda75cfce5071e5a13db19e1381c, we refactor the indexing of the tracked ethereum addresses to hook into the pgsearch indexing events. This eliminates the need to have the "proxy" content types. Rather, the ethereum plugin's address indexing can be configured to use any content type's field as the field that designates tracked ethereum addresses. This also means, that it is no longer necessary to provide user -> ethereum address relationships in the framework, as you can easily configure any content type that is related to the user model to be the address tracking content type. (@ef4 ready for another review)

packages/pgsearch/client.js
```diff
@@ -237,12 +237,14 @@ class Batch {
...
     let { branch, type, id } = context;
+    let { rows } = await this.client.query('select branch, type, id, source, generation, upstream_doc as "upstreamDoc" from documents where branch=$1 and type=$2 and id=$3', [branch, type, id]);
+    let [ eventContext={} ] = rows;
 
     this._touched[`${branch}/${type}/${id}`] = this._touchCounter++;
     let sql = 'delete from documents where branch=$1 and type=$2 and id=$3';
 
     await this.client.query(sql, [branch, type, id]);
-    await this.client.emitEvent('delete', { type, id });
+    await this.client.emitEvent('delete', eventContext);
```
ef4: I would prefer not to leak all the internal implementation details of the rows here. Let's enumerate only the parts that are needed.

habdelra:
@ef4 yes that is correct. BYO-tracked-ethereum-content-type

packages/pgsearch/client.js
habdelra: ok. in this case i need to read the document to see what addresses are being deleted, so i'll just add that to the event.

packages/pgsearch/client.js
```diff
@@ -236,16 +236,18 @@ class Batch {
...
+    let { branch, type, id, doc } = context;
+    let { rows } = await this.client.query('select branch, type, id, source, generation, upstream_doc as "upstreamDoc" from documents where branch=$1 and type=$2 and id=$3', [branch, type, id]);
+    let [ eventContext={} ] = rows;
 
     this._touched[`${branch}/${type}/${id}`] = this._touchCounter++;
     let sql = 'delete from documents where branch=$1 and type=$2 and id=$3';
 
     await this.client.query(sql, [branch, type, id]);
-    await this.client.emitEvent('delete', { type, id });
+    await this.client.emitEvent('delete', eventContext);
```
ef4: I don't think your last commit did what you intended it to do. We're stilling emitting the whole row in the event here.

habdelra: my bad! I missued this

PR #565, opened 2019-01-03
Bump ember-fetch versions to 6.4.0

balinterdi:


yarn.lock
```diff
@@ -6358,6 +6392,24 @@ ember-factory-for-polyfill@^1.3.1:
     node-fetch "^2.0.0-alpha.9"
     rollup-plugin-babel "^3.0.7"
 
+ember-fetch@^6.4.0:
```
Turbo87: this seems problematic as we're now running with two conflicting versions of `ember-fetch` 🤔

balinterdi: That shouldn't be the case.

Above bumping up dependencies for the sake of better maintainability, this came up because in deck the `@cardstack/auth` package installed ember-fetch 5 while ESA 1.8 installed ember-fetch 6 (for some reason, I know it also supports ember-fetch 5) so I wanted to resolve the conflict.

However, I now see that ESA 1.7 (the current dependency of @cardstack/auth) doesn't support ember-fetch 6 so indeed both ember-fetch 5 and 6 will be installed.

What about bumping the ESA dependency to `^1.8.0` (https://github.com/simplabs/ember-simple-auth/releases/tag/1.8.0)?

Turbo87: > What about bumping the ESA dependency to `^1.8.0` (https://github.com/simplabs/ember-simple-auth/releases/tag/1.8.0)?

I guess that would be appropriate if we want to bump `ember-fetch`

Turbo87: (ideally `ember-fetch` should be a `peerDependency` in `ember-simple-auth`, there really is no need for it to be brought in as a direct dependency)

balinterdi: By "ideally", I guess you meant "in a world where peer dependencies worked reliably in npm" so I went ahead and updated ESA to ^1.8.0.

balinterdi: @Turbo87 Could you merge if you think it's good now?

PR #572, opened 2019-01-04
Improve hasDirty* CP definitions

balinterdi:
In more complex object graphs, `hasDirtyOwned` somehow gets to be `undefined` and thus the dirtiness of that object's owned relationships is not tracked properly.


balinterdi:
I'd realized what the error is and described it in detail in the linked issue (that prompted this).

I should try to revert the change that unpacks the `hasDirtyOwned` CP definition as the previous version was easier to read and the problem probably lies elsewhere (as described there)

PR #586, opened 2019-01-15
Test branch query param

### LINKED ISSUE #532: Figure out a way to test that the `?branch=` queryParam is being passed correctly to the server
This is a follow-on issue from https://github.com/cardstack/cardstack/pull/531

The problem that we face is that the adapter `@cardstack/models/adapter` is not resolvable directly in a test because it is used as the source of codegen for hub types. We would need to do some sort of acceptance test so that a real adapter is generated for us and then ensure that the branch query param is included on the network request 👍

jorgelainfiesta:
Fixes #532


packages/models/tests/integration/models-test.js
```diff
@@ -85,6 +84,8 @@ module('Integration | Models', function(hooks) {
         type: 'pages'
       },{
         type: 'authors'
+      },{
+        type: 'fields'
```
marcoow: Why is this necessary?

jorgelainfiesta: When I was debugging I kept running into errors about duplicated fields but might have been related to the fact that I was running the tests on server mode. I'll check if it's really needed, probably not.

mansona: so as @jorgelainfiesta mentioned, if you're running this in server mode (i.e. restarting test every time you save the file) then the test will fail on the second execution of the test with an error along the lines of `Could not create field with id "blah" because it already exists`

This indicates that the types described in the destroy definition were not expansive enough 👍 so this should get committed

marcoow: ok, cool 👍

Issue #566, opened 2019-01-03
Fetch permissions in editor components


balinterdi:
We currently fetch permissions in the top-level `cs-active-composition-panel` component of the right sidebar, for the "activeContent item" and all related, owned records. We then pass down the appropriate section of that permissions object to each editor.

That works fine but it might be better to do the permissions fetching in each editor component. That means a lot more network requests (though I assume the hub caches most of that so it's not really costly) but then the permissions are always there locally with the editor which makes the template a lot clearer (see the 3rd point of https://github.com/cardstack/cardstack/issues/518) and might also resolve other issues like https://github.com/cardstack/cardstack/issues/563 .

PR #591, opened 2019-01-16
Enable editing of newly added, owned record

balinterdi:
Issue: #591

### DIFF
--- a/packages/models/addon/model.js
+++ b/packages/models/addon/model.js
@@ -109,7 +109,7 @@ function createHasDirtyForRelationship(model, name, kind) {
 }
 
 function createHasDirtyOwned(model, ownedRelationships, properties) {
-  defineProperty(model, 'hasDirtyOwnedRelationships', or(properties.join(',')));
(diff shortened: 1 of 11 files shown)


balinterdi:
@ef4 You might remember that currently we fetch the permissions for the "complete object graph" in `cs-active-composition-panel` and then pass down the appropriate chunk of that for each field editor. 

That works for records that exist when the editor is rendered but fails for records the user adds when editing.

My idea (that this PR implements) is to pass down an action to each field editor that they can use to fetch the permissions they need (= for their field model) *if* they need it. In other words, if the passed permissions from the top-level are missing.

I think that strikes a middle ground between completely moving fetching the permissions in the field editors (because then each permission would be fetched as many time as there are fields for that object) and not being able to edit fields of newly added records.

balinterdi:
If having each field editor fetch its own permissions is not wasteful (or we can deal with that trade-off), we can also move permissions fetching in the field editors, like this issue suggests: https://github.com/cardstack/cardstack/issues/566.

It would make things simpler but I'm not sure if we mind the perf hit (if there's one) and the clutter on the dev console.

PR #598, opened 2019-01-18
Making sure static models exist on all branches

mansona:
Fixes https://github.com/cardstack/dotbc/issues/163

This is a relatively small fix for something that was creating a relatively large issue. If you had defined your schema in static-models then no branch other than the controlling branch (master) would get indexed correctly. This is because the schema models were not available during the indexing process on other branches and any document contexts that were generated for other branches would return empty search documents.

### DIFF
--- a/packages/hub/indexing/branch-update.js
+++ b/packages/hub/indexing/branch-update.js
@@ -36,9 +36,11 @@ class BranchUpdate {
   }
 
   addStaticModels(schemaModels, allModels) {
-    if (!this.updaters['static-models']) { return; } // static models updaters only run on the master branch
-
+    // static models must be added to all branches
     this.schemaModels = this.schemaModels.concat(schemaModels);
+
+    // we only keep track of the static-models updater on master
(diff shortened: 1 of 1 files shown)


ef4:
Thanks. I think it's not just schema models that should go on all branches. Probably we happen to be using static-models only for schema right now, but it's allowed to contain anything.

mansona:
@ef4 I have changed the implementation of this now so that it doesn't just implement **schema** for static models 👍 

When I actually tested this end-to-end with https://github.com/cardstack/cardstack/pull/535 it wouldn't actually allow me to view other branches because the **grants** that were defined in the static-models were not included in the branch that I was trying to view. 

I have now tested it end-to-end locally and I am confident that this is working, it's just a matter of knowing if you're happy with this way of including static models in each BranchUpdate 👍

ef4:
> because the grants that were defined in the static-models were not included in the branch that I was trying to view.

If this is true it's a serious bug. Only the grants on the controlling branch are supposed to actually control permissions.

PR #599, opened 2019-01-19
always try searching for new tracked addresses instead of waiting for 10 minute cycle to pickup changes made outside of hub

habdelra:


habdelra:
I dont wanna do actually merge this, but it makes me feel good to reopen this PR anytime i see the ethereum addresses not indexing quickly enough. and each time I reopen it, the indexing completes. That has literally happened 3 times now. This is my good luck PR, reopening it causes the indexing to complete

PR #614, opened 2019-02-01
Hyperledger updates

alexspeller:
Integrate hyperledger directly into git plugin and take advantage of new caching behaviour


bagby:
LGTM 👍

mansona:
Is there a reason why the hyperledger stuff was merged into the `@cardstack/git` repo? I thought the idea was to keep things that were functionally separate well isolated 🤔 

This seems like a big digression and it seems quite strange that someone that might be using git without hyperledger should have any of this code involved in their system. And it has already caused issues because of how some of the assumptions and design requirements have differed  https://github.com/cardstack/cardstack/pull/618

alexspeller:
@mansona the actual hyperledger code is stored seperately in https://github.com/cardstack/sawtooth-git - this is just the integration into the cardstack git plugin.

Initially this integration started out as a seperate git-hyperledger data source, but that turned out not to make sense. It had to use monkeypatching in order to integrate properly, and as the integration needed to change it reached a point where this was no longer viable.

Therefore I had to bring it in to the git plugin. This does make sense to some degree architecturally, as it is just an extra feature that the git plugin also can do.

Perhaps a better approach would be adding some kind of hook mechanism in to data sources so that this kind of integration can be written in a more isolated fashion, but that's a bigger bit of work and will probably need @ef4 's input I imagine

mansona:
@alexspeller could you write an issue that covers the "better approach" to the best of your understanding? I know these things will most likely require some architectural input from @ef4 but I want to make sure that the issue isn't lost and we forget about it.

alexspeller:
@mansona I'm not really sure that my proposal would be "better". It certainly allow more isolation in terms of moving these few lines of code out of the git plugin, but the tradeoff would be adding a bunch of complexity in terms of hooks, and a whole extra plugin that actually does very little.

Personally I would bias towards the less complex option, especially as these hooks are likely to be used _only_ for this for the forseeable future - I'm not aware of any speculation of any other thing that would need to plug in to the git infrastructure in this way.

So I guess what I'm saying is that I'm not convinced either way whether there _should_ be an issue or not, and if there should be, I'm not in a position to design the better architecture enough in order to write an issue about it

PR #618, opened 2019-02-04
Fix gitchain integration to work with remote repo config

alexspeller:
With remote repos, a local repo path is not passed in to the config, and instead
a tmpdir is set up later on


bagby:
LGTM, merge when green.

PR #621, opened 2019-02-06
Fix fetching permissions on model change

balinterdi:


balinterdi:
@ef4 Would be great to merge this in asap as this bug was brought to surface by https://github.com/cardstack/dotbc/issues/324 and so to ship that, this needs to be merged.

balinterdi:
Turns out that doing the same in `didReceiveAttrs` was necessary, after all (as evidenced by the failing test). Fingers crossed it now passed on Travis, too (it's passed locally).

ef4:
Looks good, thanks.

PR #624, opened 2019-02-06
Editor panel cancel button

burieberry:
Resets changes to last saved state and closes editing mode

### DIFF
--- a/packages/tools/addon/components/cs-version-control.js
+++ b/packages/tools/addon/components/cs-version-control.js
@@ -163,7 +163,7 @@ export default Component.extend({
     this.get('tools').setEditing(false);
 
     if (this.get('anythingDirty')) {
-      let model = this.get('model');
+      let model = yield this.get('model');
 
       model.rollbackAttributes();
       model.rollbackRelationships();
--- a/packages/tools/addon/components/cs-version-control.js
+++ b/packages/tools/addon/components/cs-version-control.js
@@ -160,7 +160,12 @@ export default Component.extend({
   }).keepLatest(),
 
   cancel: task(function * () {
-    yield this.get('tools').setEditing(false);
+    if (this.get('anythingDirty')) {
+      let model = this.get('model');
+      let upstreamModel = yield this.get('store').findRecord(model.type, model.id);
+      upstreamModel.rollbackAttributes();
+    }
+    this.get('tools').setEditing(false);
   }),
 
   delete: task(function * () {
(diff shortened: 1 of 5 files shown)


packages/tools/addon/components/cs-version-control.js
```diff
@@ -158,6 +159,15 @@ export default Component.extend({
     yield model.save();
   }).keepLatest(),
 
+  cancel: task(function* () {
+    if (this.anythingDirty) {
+      let model = this.get('model');
+      let upstreamModel = yield this.get('store').findRecord(model.type, model.id);
```
marcoow: I think this could be synchronous and use `peekRecord` as only models that are already in the store could even be dirty at all.

burieberry: When I use `peekRecord`, the cancel button doesn't undo the changes. Any idea?

marcoow: Hm, that seems strange. I just realized you should also be able to simply do `model.rollbackAttributes` though…

burieberry: That's what I initially tried as well, but simply doing that didn't revert the changes. Even though the changes aren't actually saved, visually the changes remain until page is refreshed. Only when I fetch the record, it's reset to the formerly saved version without needing to refresh the page.

burieberry: Also realized changes in the dropdown menu selection (ie. selecting a different Artist using the dropdown) doesn't mark the field as dirty. Will take a look at that. Multi select dropdown is fine (Genre selection).

packages/tools/addon/components/cs-version-control.js
```diff
@@ -158,6 +159,28 @@ export default Component.extend({
...
+        relationNames.forEach(relation => {
+          let relationItems = model.get(relation);
+          relationItems.forEach(item => {
+            item.rollbackAttributes();
+            item.rollbackRelationships();
+          });
+        });
+      }
+    }
+  }),
```
burieberry: This version works for reseting all types of dirty fields except the titles editor on dotbc. Need help with writing tests, though. Getting `Uncaught TypeError: model.rollbackAttributes is not a function` error when I try to use the rollback functions in `cs-version-control-test` file.

ef4:
This looks good. 

Can we move the code from the `cancel` method that rolls back all the things into `packages/models/addon/model.js` and rename is something like `cardstackRollback()`? That way all the detailed knowledge of things like `dirtyTrackingRelationNames` can stay internal to that module.

burieberry:
@ef4 sure, will do now. thanks for reviewing

PR #625, opened 2019-02-07
forgive node-pg-migrate migration already running errors

habdelra:


ef4:
Perhaps we can use our queueing system here. I think it doesn't rely on this pgsearch database, it has its own for coordinating.

habdelra:
cool, I'll try that out

habdelra:
@ef4 I'm seeing that we actually have an oversight in our queue logic. it looks like our `queue.publishAndWait()` never actually resolves when the node process that submits the job isn't first in the queue. i think it comes down to how the job ID's are assigned in pg-boss. It seems that from the documentation that the job ID is null for jobs that are waiting when you are using the singleton key. https://github.com/timgit/pg-boss/blob/master/docs/configuration.md#unique-jobs. I'm unsure why that is, it doesn't really make sense to me.

habdelra:
I think if the job id is null, we need to start polling pgboss until we get a job id....

habdelra:
(the pgboss tests also reinforce what the documentation is saying https://github.com/timgit/pg-boss/blob/master/test/singletonTest.js#L37)

PR #632, opened 2019-02-12
Handle non-JSON requests as binary

balinterdi:


balinterdi:
This came up as requests to `.flac` audio files streamed through our sftp proxy didn't launch as expected because the hub treated the requests (launched from the `src` attribute of the `audio` tags) as JSON and returned JSON for them.

See https://github.com/cardstack/dotbc/issues/353

ef4:
What `Accept` header is the browser sending when the request comes from an `<audio>` tag? I would have expected we could make this work based on that and not file extension.

alexspeller:
I know I went back and forth a few times with @ef4 about the criteria for whether we should consider a request binary or not.

I think that respecting the file extension, if present, is a good addition because it allows the client to easily specify what it wants in the case that Accept headers aren't viable such as this, or in other cases such as `<video>` or `` tags.

I'm somewhat surprised that [Safari and Chrome](https://developer.mozilla.org/en-US/docs/Web/HTTP/Content_negotiation/List_of_default_Accept_values#Values_for_audio_resources) have such bad defaults for this, but they do 🤷‍♂️

alexspeller:
I guess we could [use fetch](https://medium.com/@bitshadow/play-audio-via-react-fetch-post-request-92a901d0bb7f) if we really wanted too although that's definitely a PITA on the client side

alexspeller:
I guess we could [use fetch](https://medium.com/@bitshadow/play-audio-via-react-fetch-post-request-92a901d0bb7f) if we really wanted too although that's definitely a PITA on the client side

balinterdi:
@ef4 They send `*/*` (= accept anything) which made me extract the extension and rely on that.

I'm open to feedback as I'm a bit concerned about how bullet-proof the above is.

ef4:
I don't think we forbid `.` inside IDs, meaning that `{ type: 'posts', id: '3.14159' }` would need to serve a jsonapi response at `/api/posts/3.14159`. If I'm reading this PR correctly, that would break here regardless of what Accept header you tried to use.

What breaks if we force jsonapi to always have a json or jsonapi accept header? Is it just some tests that aren't setting any header at all?

Can we treat a _missing_ `Accept` header as `Accept: application/vnd.api+json` while otherwise being strict about only serving jsonapi if you explicitly ask for it? Presumably the browsers at least all include _some_ Accept header? This lets us continue to be lazy in debugging and testing, without leaving a broken case behind.

alexspeller:
> Can we treat a _missing_ `Accept` header as `Accept: application/vnd.api+json` while otherwise being strict about only serving jsonapi if you explicitly ask for it? Presumably the browsers at least all include _some_ Accept header? This lets us continue to be lazy in debugging and testing, without leaving a broken case behind.

I _think_ that this will resolve both the `audio` issue and the issue #382 was fixing but I'm not 100% sure

balinterdi:
> What breaks if we force jsonapi to always have a json or jsonapi accept header? Is it just some tests that aren't setting any header at all?

I'm not sure I understand. Where would we force this? The browser sends an accept header of `*/*` if coming from the `src` attribute of an `audio` tag (and there might be similar cases, like a `video` , or even an `img` tag). If we force this to `accept: jsonapi`, we'll not serve what we should, binary data.

I probably misunderstand you.

ef4:
I'm saying that we should only serve JSON:API if there is `Accept: application/vnd.api+json`.

That is what the JSON:API spec demands anyway. And it would solve your problem here, because the browser's audio tag is not asking for jsonapi.

balinterdi:
That would be great. Provided all of our JSON API requests are sent with the `Accept: application/vnd.api+json` header, that would be a simple and clean solution. I'll check that, thanks!

balinterdi:
@ef4 That really simplifies our branching logic. I looked at whether we always set the `Accept: Accept: application/vnd.api+json` when we ask for JSON:API data and there were a few places where we don't, we only set `Content-Type: application/vnd.api+json`.

That's not the same though, as the `Content-Type` refers to the type of the request itself – not a demand for the type of the response. So I've added the `Accept` header where it was missing.

PR #641, opened 2019-02-18
Include CPs in doc when checking permissions

balinterdi:


packages/hub/node-tests/permissions-test.js
```diff
@@ -0,0 +1,91 @@
...
+  it("attempting to fetch permissions for a resource the user doesn't have access to", async function() {
+    try {
+      await searchers.get(env.session, 'local-hub', 'permissions', `articles/improve-chess`);
+    } catch (err) {
+      expect(err.message).to.match(/No such resource master\/articles\/improve-chess/);
+    }
+  });
+
+  it("attempting to fetch permissions for a resource the user has access to", async function() {
+    allReadable();
```
ef4: This has no effect, because it needs to happen before `createDefaultEnvironment`.

ef4: @balinterdi apologies for dropping this PR for a while, I thought I had already reviewed it.

It doesn't actually do the right thing, or at least its tests don't.

packages/hub/node-tests/permissions-test.js
```diff
@@ -0,0 +1,91 @@
...
+  function allReadable() {
+    factory.addResource('grants').withAttributes({
+      mayReadResource: true,
+      mayReadFields: true
+    }).withRelated('who', [everyone]);
+  }
+
+  it("attempting to fetch permissions for a resource the user doesn't have access to", async function() {
+    try {
+      await searchers.get(env.session, 'local-hub', 'permissions', `articles/improve-chess`);
```
ef4: This test doesn't really work. You need to insert a 

`throw new Error('should never get here')` after the `await searchers.get`. If you do, you will see that throw.

bagby:
@ef4 I assume we can close this?

PR #642, opened 2019-02-18
Add option to hide model fields from editor

burieberry:


balinterdi:
I like the approach but it would be great if we could default the `editorModelField` to true so that we don't have to explicitly set it everywhere (assuming we want to retain the current functionality of showing the field in the editor by default).

burieberry:
@balinterdi yeah I can do that

balinterdi:
@burieberry Looks good to me. The only thing I'd change is to only hide one normal and one computed field. We don't gain by hiding most of them and they can be useful to test other things.

And a question: why the `echo-` prefix? :)

burieberry:
@balinterdi You're right, removed the extra tests. Ha I've seen someone else use the prefix in some other tests on this repo

balinterdi:
> Ha I've seen someone else use the prefix in some other tests on this repo

Seem redundant/potentially confusing, can you remove them?

burieberry:
Changed the field names. I added two new fields to test this because I didn't want to disturb the existing tests and it's good to have the examples in the dummy app.

Issue #635, opened 2019-02-14
fetchPermissions doesn't check response.status


ef4:
If [this request](https://github.com/cardstack/cardstack/blob/3a4872e69cb5cd9c1a080ca8cb8dd5c61751c7fa/packages/data/addon/services/cardstack-data.js#L108) doesn't return a 200, we try to destructure an undefined value.

cc @balinterdi

balinterdi:
Good catch, I'll tackle that next week.

balinterdi:
@ef4 I submitted [the PR](https://github.com/cardstack/cardstack/pull/655) but I wonder if that doesn't add a guard that's never needed. I think a request to `/api/permissions/:path` can only do one of two things:

1) Return a 200 OK response with `may-update-resource` and `writable-fields` attributes in a JSON:API doc so the original, non-guarded code is correct.
2) Throw an error (with a status of 404) if the content-type doesn't exist, in which case control doesn't get to the next line – so the guard is unneeded.

PR #655, opened 2019-02-20
Fix fetchPermissions response destructuring for non-200 case

balinterdi:
See #635

### DIFF
--- a/packages/data/addon/services/cardstack-data.js
+++ b/packages/data/addon/services/cardstack-data.js
@@ -109,7 +109,8 @@ export default Service.extend({
       headers: this._headers()
     });
     if (response.status !== 200) {
-      return;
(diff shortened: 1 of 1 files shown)


ef4:
Does this move the problem up to the caller?

It may still be reasonable to reject the promise instead, but with a more sensible error message than the current runtime exception.

balinterdi:
I can do that, I have only one slight concern.

If I return a rejected promise with a descriptive error message (in this case informing the caller that the content-type doesn't exist) doesn't that give away (slightly) confidential information about what content-types we have? (and thus increase the surface area of a potential attack).

I'm reasonably sure that trade-off is worth it (app developers will benefit) but I just wanted to mention this.

ef4:
I agree that information leakage is a concern, but this is client-side code, it's not relevant here.

balinterdi:
@ef4 Could you check again?

Issue #593, opened 2019-01-16
Improve caption for fields of newly created records


balinterdi:
Captions for new records appear as "Type #: Field" where normally the `id` follows the `#` (but new records don't usually have that yet).

![](https://d2ddoduugvun08.cloudfront.net/items/200e2B0x160I3S080a0B/%5Beef0eb39eda29371e00a6ab8e472a61f%5D_Image%25202019-01-16%2520at%25205.49.26%2520PM.png)

I suggest rendering "New Type" instead if the record doesn't have an id.

PR #667, opened 2019-02-25
Improve caption for newly created records

balinterdi:
See #593

### DIFF
--- a/packages/data/addon/services/cardstack-data.js
+++ b/packages/data/addon/services/cardstack-data.js
@@ -122,13 +122,14 @@ export default Service.extend({
   },
 
   getCardMeta(card, attribute) {
+    let hasId = !!card.id;
     if (attribute === 'human-id') {
       let humanType = getType(card)
         .split('-')
-        .map((s) => capitalize(s))
+        .map((s) => hasId ? capitalize(s) : s)
         .join(' ');
-      return [`${humanType} `, card.id]
-        .join('#');
+
+      return card.id ? [`${humanType} `, card.id].join('#') : `New ${humanType}`;
     }
     if (attribute === 'uid') {
       return `${getType(card)}/${card.id}`;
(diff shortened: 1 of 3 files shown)


habdelra:
Could we actually default to first trying to display the field that the `field-editor/dropdown-choices-editor` uses? That would be most human friendly. this is either `title` or       `Field.editorOptions.displayFieldName`.

packages/data/addon/services/cardstack-data.js
```diff
@@ -122,13 +122,14 @@ export default Service.extend({
...
     if (attribute === 'human-id') {
       let humanType = getType(card)
         .split('-')
-        .map((s) => capitalize(s))
+        .map((s) => hasId ? capitalize(s) : s)
         .join(' ');
-      return [`${humanType} `, card.id]
-        .join('#');
+
+      return card.id ? [`${humanType} `, card.id].join('#') : `New ${humanType}`;
```
habdelra: `#` should only be used if you know for sure the thing you are displaying is an ordinal number--otherwise it's just confusing.

balinterdi: That's a good point but then how should we distinguish between two resources that have the same type? You'd see multiple "Comment: Body" labels and furthermore they might not even be rendered on screen to use the overlay for the visual telling apart.

One idea I had is to pass in an index to this method (well, to the entry point to this method as I recall this is not called directly) but that would increase the surface area for this method (and of all the methods on the call chain) and would make the caller too smart.

balinterdi:
> Could we actually default to first trying to display the field that the field-editor/dropdown-choices-editor uses? That would be most human friendly. this is either title or Field.editorOptions.displayFieldName.

Do you mean this for the `title` meta property? (human-id doesn't have such concerns, it only displays "Comment # 24" or "New comment".

ef4:
Closing because this has sat with open questions a long time and it's not even clear if this is the right direction.

balinterdi:
Ok, but that means we'll keep displaying "Category #: Title" as the editor label when creating a new category (or other new resources, of course).

Issue #579, opened 2019-01-09
Dont display fields that are not readable in the right edge


habdelra:
In this screenshot, the `password-hash` field is actually not readable to the user, however the right edge is leaking the existence of this field (although there is no value displayed in it, and it is not editable). We need to hide fields that are not readable in the right edge.

<img width="970" alt="screen shot 2019-01-09 at 11 21 32 am" src="https://user-images.githubusercontent.com/61075/50913314-16694a80-1402-11e9-8501-226a0aa8b06a.png">

balinterdi:
Indeed. My only concern is there might be cases where we still want to show fields that are not editable.

A computed field that displays some useful information might be such a case, like the review status of a blog post or the number of words in a post (or comment).

Also, in some cases it might be useful for the user to see there's such a field even if they don't have permission to edit it. The review status of a blog post might be a good example, here, too. If I want to be able to change it and I see I  currently can't, it gives me a clue about how to fix that. Whereas if I don't even see it, I might be stomped about why my post doesn't appear on the blog.

I think a better way would be to add a field-level attribute (`hideInEditor`) so that it never appears in the editor.

@habdelra I now realize you didn't say we have to use `permissions` to implement this, it's just what popped into my mind as I've worked with it recently. I also think we don't currently have another method for disabling fields. We have the `disableAutoEditorFields` sledgehammer/shotgun that suppresses the rendering of all "model fields" (the ones that are not explicitly rendered in the template) but I don't think it could be used to do that.

@ef4 What do you think?

ef4:
This is about non-readable fields, not about non-writable fields.

balinterdi:
@ef4 Ah, I totally overlooked that, thank you.

So the implementation should be extending what we currently have in the `/permissions` response by a `readable-fields` attribute that lists all the readable fields (just like we currently do for `writable-fields`). If a field is not in `readable-fields` we shouldn't even render it.

ef4:
Ideally, we shouldn't leak the existence of non-readable fields to the client at all. So there would be nothing to ask about. So no, I'd prefer not to make readability something clients can even ask about in the authorization API.

They're currently leaking because we don't take permissions into account when we generate the code for the ember data models. There will be ember data attrs for all fields, even ones that aren't readable. Doing this correctly involves running codegen later, and more dynamically.

[...]

PR #669, opened 2019-02-26
Don’t code generate DS.attrs for unreadable attributes

balinterdi:
See #579


ef4:
Closing this because it's not the right solution to this problem. This is still generating the DS.Model during the ember build. It needs to happen later, at runtime, per client so that different users can get different results.

PR #672, opened 2019-02-26
Fetch all from remote git in indexer branches

robustdj:


marcoow:
@robustdj can you add some description? Is this what breaks branch creation when you're using a git repo with a remote?

mansona:
So it looks like this is essentially going to be fetching the branches directly from the remote **every** time we request them 🤔 I'm pretty sure this is **not** what we want because it entirely avoids the cache

Right now we are relying on the indexing system to update the listed branches but if this is for the "create branch" middleware I have a feeling we would want to do some sort of "write through" on the cache somehow 🤔 I don't know if this is something that we can do from the outside (i.e. not in core)

I agree with @marcoow, we need some more explanation as to **why** this change was needed and then we should probably check with @ef4 for any unintended consequences 👍

ef4:
The bug here was that a new branch in the remote repo would not be discovered until the _second_ time you ran indexing. Because the list of branches was being discovered _before_ we ran `repo.fetchAll`. The bugfix moves the `fetchAll` so it happens before we emit the list of branches.

I think this change actually makes fetching more efficient, not less. During indexing, the `branches` method is called first, and then we call `updateContent` for each of the branches that was returned by the `branches` method.

Prior to this PR, we did the `fetchAll` in `updateContent`, which means we actually do it more times than necessary since `updateContent` runs once per branch.

We definitely do not want to hit the cache here, since `branches` is part of indexing and is supposed to be getting the latest upstream state.

PR #681, opened 2019-03-04
Always enable fields in newly created records

bagby:


habdelra:
@bagby i'm also seeing this same issue when I edit a newly saved document https://github.com/cardstack/cardstack/issues/689

Issue #564, opened 2019-01-03
Allow seeding images (when using the `cardstack-image` type)


balinterdi:
Replacing a URL-based image (where simply the src URL is given as a string) with the 1st class image support via the `cardstack-image` type greatly enhances the user experience but unfortunately we can't yet seed an image that way as it's not just a matter of using a URL.

To enable this, we'd have to upload the image during boot / set up, probably by calling `handleCollectionPostBinary` in the jsonapi middleware:

https://github.com/cardstack/cardstack/blob/master/packages/jsonapi/middleware.js#L284-L299

bagby:
Implemented by #706

PR #706, opened 2019-03-11
Add ability to seed and index binary data in ephemeral storage

bagby:
This allows you to seed binary data (images, audio, etc) into ephemeral data source for use in development. Also implements indexing binary data in ephemeral storage.

An example of how to seed data in your Ember app:
```
const { createReadStream, readdirSync } = require('fs');
const path = require('path');

let models = [];

if (process.env.HUB_ENVIRONMENT === 'development') {
  readdirSync(path.join(__dirname, './images')).forEach(filename => {
    let readStream = createReadStream(path.join(__dirname, 'images', filename));

    readStream.type = 'cardstack-files';
    readStream.id = filename.replace(/\..+/, '');

    models.push(readStream);
  });
}

module.exports = models;
```


packages/hub/main.js
```diff
@@ -67,6 +67,16 @@ async function startIndexing(environment, container) {
...
+    let branch = 'master';
+    let writers = container.lookup('hub:writers');
+
+    // write cardstack-files to data source
+    for (let model of models) {
+      if (model.readable) {
+        await writers.createBinary(branch, Session.INTERNAL_PRIVILEGED, 'cardstack-files', model);
+      }
+    }
+
```
ef4: Why are we writing out models in `startIndexing`? This seems like a data destruction bug.

The ephemeral data source is the only one that _always_ loads seed data automatically. The others only load seed data if you run the hub:seed command. But they will go through `startIndexing` and had better not overwrite their contents with the seed models when they do.

packages/hub/node-tests/indexers-test.js
```diff
@@ -88,6 +90,46 @@ describe('hub/indexers', function() {
...
+      let factory = new JSONAPIFactory();
+
+      factory.addResource('content-types', 'cardstack-files')
+        .withRelated('fields', [
+          factory.addResource('fields', 'created-at').withAttributes({fieldType: '@cardstack/core-types::date'}),
+          factory.addResource('fields', 'content-type').withAttributes({fieldType: '@cardstack/core-types::string'}),
+          factory.addResource('fields', 'sha-sum').withAttributes({fieldType: '@cardstack/core-types::string'}),
+          factory.addResource('fields', 'file-name').withAttributes({fieldType: '@cardstack/core-types::string'}),
+          factory.addResource('fields', 'size').withAttributes({fieldType: '@cardstack/core-types::integer'})
+        ]);
```
ef4: I think you will get the schema for cardstack-files automatically if you make sure `@cardstack/files` is a dependency of stub-project.

packages/ephemeral/service.js
```diff
@@ -48,7 +51,27 @@ module.exports = declareInjections({
...
-          storage.store(model.type, model.id, model, isSchema);
+          if (model.readable) {
+            // store binary data
+            let id = model.id || crypto.randomBytes(20).toString('hex'); // if model has id, use it, otherwise generate one
+            let blob = await streamToPromise(model);
+            let storedDocument = {
+              type: 'cardstack-files',
+              id,
+              attributes: {
+                'created-at':   new Date().toISOString(),
```
ef4: This may be better as `statSync(model.path).ctime`

ef4: (Assuming stat is OK in general here, I'm going to write another comment about that.

ef4:
The `statSync` in service.js worries me a little. What does it do when the readable stream in question is being provided by the webserver, because somebody is doing an upload and saving it into ephemeral?

Basically I'm not sure every readable stream that will be sent in here actually has a stat-able file associated with it.

bagby:
> The `statSync` in service.js worries me a little. What does it do when the readable stream in question is being provided by the webserver, because somebody is doing an upload and saving it into ephemeral?
> 
> Basically I'm not sure every readable stream that will be sent in here actually has a stat-able file associated with it.

The only things that call `findOrCreateStorage` are the ephemeral indexer and searcher. The searcher doesn't get to that part of the code, and the indexer only includes `initialModels` when seeding. Getting to `storeBinary` from the browser follows a different path.

packages/hub/node-tests/indexers-test.js
bagby: do you mean putting it in cardstack/packages/hub/node-tests/stub-middleware-extra/package.json? I tried that and it didn't work.

ef4: No, I mean `${__dirname}/../../../tests/stub-project/package.json`. That's why `createDefaultEnvironment` takes a path as an argument, it needs to know where to find the package.json to know what plugins to use.

bagby: ok I tried that _and_ added a `@cardstack/files` data source in `bootstrap-schema.js`, but the `default` ephemeral data source isn't created until you call `createDefaultEnvironment`. So it's sort of a Catch-22.

balinterdi:
@bagby That's great that we now have this. A while ago I created https://github.com/cardstack/cardstack/issues/564 that I think this PR also implements. If that's correct, can you close #564?

PR #724, opened 2019-03-28
git: fix finalise retry logic

mansona:
This is an attempt to fix the merge retry logic in the git datasource. Previously if the remote was out of sync it wouldn't update the local representation of the branch before retrying.

Also because of a missed `continue` statement it would continue as if it was a success even when it failed 🙈 This PR fixes that.

### DIFF
--- a/packages/git/change.js
+++ b/packages/git/change.js
@@ -85,27 +85,38 @@ class Change {
 
     let delayTime = 500;
     let mergeCommit;
+    let needsFetchAll = false;
+
     while (delayTime <= 5000) {
       mergeCommit = await this._makeMergeCommit(newCommit, commitOpts);
 
       try {
-        if(this.fetchOpts) {
+        if (this.fetchOpts) {
+          // needsFetchAll only gets set to true if the retry block has failed once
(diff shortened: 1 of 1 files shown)


packages/git/change.js
```diff
@@ -89,23 +89,30 @@ class Change {
...
         } else {
           await this._applyCommit(mergeCommit);
         }
       } catch (err) {
-        await delay(delayTime);
         log.warn('Failed to finalize commit "%s"', err);
 
+        if (this.fetchOpts) {
+          // pull remote before allowing process to continue
+          await this.repo.fetchAll(this.fetchOpts);
```
ef4: This is a network request that is likely to throw exceptions sometimes. If it does, we will break out of the retry loop entirely.

Perhaps we can move the fetchAll so it happens before `_makeMergeCommit`, but only when a `needsFetchAll` flag is set. The flag would start out as false but set to true inside the `catch` block.

mansona: Yep that sounds like a good idea 👍 I'll do that now

PR #726, opened 2019-03-29
Fixes #682 Refactor to incorporate DocumentContext in document authorization

### LINKED ISSUE #682: Unable to use computed fields as part of field based grants
In the scenario where you have a grant that uses a field dependent `who` relationship, and that field is a computed field, the grant will not be evaluated correctly, as the grant validation is not taking into account computed fields. 

From @ef4:

The root cause bug is that this code doesn’t go through Model#getField: https://github.com/cardstack/cardstack/blob/2d7f04ff1bdf4d9ccb150b96083006f32e147ebe/packages/hub/schema/grant.js#L277-L278

habdelra:
Fixes #682 

This refactor pushes the creation of a DocumentContext for a document as soon as it is possible to actually create a DocumentContext--which is in the PendingUpdate. By moving the creation of DocumentContext as early as possible, we are able to leverage it as part of the schema's authorization of the document. This means that we can leverage `Model.getRelated()` to look up fields which will resolve computed fields (which currently cannot be used when performing document authorization).

***This is a breaking change***

The various document authorization schema methods which were previously synchronous, are now async (to facilitate index reads that are required for looking up relationships from the Model class). Additionally, these methods now require you to pass in the DocumentContext for a resource instead of just the raw resource. Finally the `PendingUpdate` ~~requires that the writer that is using it must supply a searcher, schema, branch, and data source ID to it so that it can fashion a DocumentContext for the pending document~~ needs to be retrieved from the `hub:writers.createPendingChange()` so all writers that use this need to be updated.


habdelra:
@ef4 I’d love to get your feedback on this PR.

packages/drupal/writer.js
```diff
@@ -6,13 +6,19 @@ const { isEqual } = require('lodash');
 const { drupalToCardstackDoc, cardstackToDrupalDoc } = require('./lib/document');
 const pendingChanges = new WeakMap();
 const log = require('@cardstack/logger')('cardstack/drupal');
+const { declareInjections } = require('@cardstack/di');
```
ef4: I'm fine with deleting modules we aren't actively using like `@cardstack/drupal`. We could always resurrect them out of source control next time we actually want them.

packages/plugin-utils/pending-change.js
```diff
@@ -1,11 +1,48 @@
...
+    finalizer,
+    aborter,
+    searchers,
+    branch,
+    sourceId,
+    schema
+  }) {
+    if (!branch || !searchers || !schema) {
+      throw new Error(`PendingChange requires 'branch', 'searchers', and 'schema' arguments.`);
+    }
```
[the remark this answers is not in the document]
habdelra: cool. I like the first suggestion of constructing the empty PendingChange instance in hub/writers.js, as we need to be able to have DocumentContext in the writer finalizers in order to validate the potential write.

 I'll go down that path

habdelra:
@ef4 ok, I made your recommended changes. The `PendingChange` is now a class that you obtain from the `hub:writers`, and it will be built with all the stuff necessary so that it can internally create a DocumentContext, and not have to rely on each specific writer to supply this stuff. This means that I was able to use the original 4 parameters for this, plus optionally `branch` (which all writers will have immediate access to). 

Also I removed the drupal modules.

packages/plugin-utils/pending-change.js
```diff
@@ -1,25 +0,0 @@
...
-  }
-  async abort() {
-    let aborter = this._aborter;
-    this._finalizer = null;
-    this._aborter = null;
-    if (aborter) {
-      return aborter.call(null, this);
-    }
-  }
-};
```
ef4: Oh, I like that we don't need this as a dependency of every writer plugin anymore. 👍

packages/ephemeral/writer.js
```diff
@@ -9,6 +8,7 @@ const pendingChanges = new WeakMap();
 
 module.exports = declareInjections({
   indexers: 'hub:indexers',
+  writers: 'hub:writers',
```
ef4: If every Writer is going to need this parameter, perhaps we can give it to them automatically. They get instantiated here https://github.com/cardstack/cardstack/blob/474abbd05df120fd7e71f2c06737fc53d8a53e3a/packages/hub/schema/data-source.js#L38

habdelra:
awesome suggestions!

habdelra:
@ef4 ok, I made your additional suggestions--that ended up eliminating more of the writer's dependencies, and further simplified the writer's APIs

PR #731, opened 2019-04-03
add filter condition for bool field type, also relaxed the test timeouts

habdelra:
it was very odd that this was missing--but its not now


ef4:
Our default filter behavior is still fuzzy rather than exact. If you do an exact filter the boolean case is already correct, as far as I know:

```js
{
  filter: { 
    theField: { exact: true } 
  }
}
```

The real issue here is that we still need to change the default from fuzzy to exact.

PR #736, opened 2019-04-08
Query based relationships

bagby:
This is the "first pass" at implementing query-based relationships. It allows you to have query based relationships in seed data, using the `withRelatedLink` factory function.

Example:
```javascript
// create the schema
factory.addResource('content-types', 'boards')
  .withAttributes({
    defaultIncludes: [
      'published-articles'
    ],
    fieldsets: {
      isolated: [
        { field: 'published-articles', format: 'embedded' },
      ],
    }
  })
  .withRelated('fields', [
    { type: 'fields', id: 'title'},
    factory.addResource('fields', 'published-articles').withAttributes({
      fieldType: '@cardstack/core-types::has-many'
    })
  ]);

factory.addResource('content-types', 'articles')
  .withAttributes({
    ...
  }) // etc.

// add the query relationship
factory.addResource('boards', 'community').withAttributes({
  title: 'Community'
}).withRelatedLink('published-articles', `/api?${qs.stringify({
  filter: {
    type: { exact: 'articles' }
  },
  sort: '-published-date'
})}`);
```

which, in our index, will create a `board` that looks something like:
```javascript
{
  "data": {
    "id": "community",
    "type": "boards",
    "attributes": {
      "title": "Community"
    },
    "relationships": {
      "published-articles": {
        "data": [],
        "links": {
          "related": "/api?filter%5Btype%5D%5Bexact%5D=articles&sort=-published-date"
        }
      }
    }
  }
}
```

When you `GET` the `board` via API, the `relationships.published-articles.data` will be populated with `articles` based on the query.

### DIFF
--- a/packages/hub/indexing/document-context.js
+++ b/packages/hub/indexing/document-context.js
@@ -1,13 +1,14 @@
 const authLog = require('@cardstack/logger')('cardstack/auth');
 const log = require('@cardstack/logger')('cardstack/indexing/document-context');
-const Model = require('../model');
+const { Model, privateModels } = require('../model');
 const { getPath } = require('@cardstack/routing/cardstack/path');
 const { merge, get, uniqBy } = require('lodash');
 const Session = require('@cardstack/plugin-utils/session');
+const qs = require('qs');
 
 module.exports = class DocumentContext {
 
-  constructor({ read, routers, schema, type, id, sourceId, generation, upstreamDoc, includePaths }) {
+  constructor({ read, search, routers, schema, type, id, sourceId, generation, upstreamDoc, includePaths }) {
     if (upstreamDoc && !upstreamDoc.data) {
       throw new Error('The upstreamDoc must have a top-level "data" property', {
         status: 400
@@ -22,6 +23,7 @@ module.exports = class DocumentContext {
     this.upstreamDoc = upstreamDoc;
     this.includePaths = includePaths ? includePaths.map(part => part ? part.split('.') : null).filter(i => Boolean(i)) : [];
     this._read = read;
+    this._search = search;
     this._realms = [];
     this._pendingReads = [];
     this._followedRelationships = {};
@@ -55,13 +57,13 @@ module.exports = class DocumentContext {
         let contentType = this.schema.types.get(doc.type);
         if (!contentType) { throw new Error(`Unknown content type=${doc.type} id=${doc.id}`); }
 
-        return new Model(contentType, doc, this.schema, this.read.bind(this));
+        return new Model(contentType, doc, this.schema, this.read.bind(this), this.search.bind(this));
       });
     } else {
       let contentType = this.schema.types.get(this.type);
       if (!contentType) { throw new Error(`Unknown content type=${this.type} id=${this.id}`); }
 
-      this._model = new Model(contentType, this.upstreamDoc.data, this.schema, this.read.bind(this));
+      this._model = new Model(contentType, this.upstreamDoc.data, this.schema, this.read.bind(this), this.search.bind(this));
     }
 
     return this._model;
@@ -119,6 +121,7 @@ module.exports = class DocumentContext {
 
     let cached = this.cache[key];
     if (cached) {
+      log.debug(`document with key ${key} is in cache, fetching...`);
       return await cached;
     }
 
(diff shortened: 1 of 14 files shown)


packages/hub/indexing/document-context.js
```diff
@@ -429,8 +438,14 @@ module.exports = class DocumentContext {
         let data = related ? { type: related.type, id: related.id } : null;
         ensure(pristineDocOut, 'relationships')[field.id] = Object.assign({}, value, { data });
       }
+    } else if (value.links && value.links.related) {
+      let isCollection = field.fieldType === '@cardstack/core-types::has-many';
+      let data = await this.relatedLinks(qs.parse(value.links.related.split('/api?')[1]), isCollection);
```
habdelra: I think that be evaluating this inside of `DocumentContext` we actually end up materializing the result of the query in the index since we use `DocumentContext` to generate the `pristinedoc` column in the index DB (you should be able to confirm this by looking directly at the documents table)--which will likely be out of date since we are not invalidating models based on their appearance in queries that are relationships. I think you need to move the materialization of these relationships wholly outside of `DocumentContext` until we actually do support this type of invalidation.

packages/hub/searchers.js
```diff
@@ -215,6 +241,7 @@ class Searchers {
       upstreamDoc,
       includePaths,
       routers: this._getRouters(),
+      relatedLinks: this._getRelatedLinks.bind(this),
```
habdelra: based on my earlier comment, I would not pass this function into `DocumentContext` until we can support invalidation of models that appear in the query based relationships. but rather materialize these queries after getting the pristine doc from the `DocumentContext.

packages/hub/indexing/document-context.js
habdelra: A test that would reveal this problem would be a test that has a computed that is based on a query based relationship. Changing the underlying models that effect the output of a query based relationship should invalidate a computed field that consumes a query based relationship. My guess is that such a test would fail if we are saving the materialized resources from a query based relationship in the index.

bagby: Ahh yes, not sure why I didn't notice I was putting it in the index before. I'll write the test you suggested.

packages/hub/searchers.js
bagby: 👍

packages/hub/model.js
```diff
@@ -27,6 +27,13 @@ module.exports = class Model {
     }
   }
 
+  // TODO watch out its probblyok for a relationship to have both links and data...
+  getLinks(fieldName) {
```
ef4: Instead of adding this method, we should change `getField` so that for relationships it returns the whole `{ data, links }`.

ef4:
One place to implement the ignoring of linked relationships during indexing is here:

https://github.com/cardstack/cardstack/blob/6c46481c6e26b94eac49029b282d9e6f362f47ad/packages/hub/indexing/document-context.js#L361

The `searchTree` is the complete description of which relationships we will traverse down into. If you prune it, that should solve the whole problem neatly.

ef4:
That also gives me an idea for avoiding needing a new option on DocumentContext. We already have `includePaths`, which lets you optionally customize the includes (which is what happens if you pass `?include=` on JSON:API).

We could add explicit special values for `includePaths`. We would pass `includePaths=DEFAULT_USER_INCLUDES` from the searchers and we would pass `includePaths=DEFAULT_INDEX_INCLUDES` from the indexers.

habdelra:
oooh, that's a nice idea

packages/hub/model.js
bagby: I like this idea, but `getField` already returns `relationships.${fieldName}.data` for relationships, and I would assume there is code that relies on this behavior. Hopefully it shouldn't be too large of a refactor.

bagby:
@ef4 all of this sounds good, the only remaining issue I can think of is that, since we are skipping over indexing for query-based relationships, we can't have computed fields that are based on them, since the point at which we load those relationships in the searcher comes _after_ the computed fields are, er, `compute`d.

habdelra:
Thinking about it more, there is a subset of query based relationships that could be _volatile_ (like return all the articles published in the last 24 hours). In that case I think we would have to always use the approach we are discussing right now, because we wouldn't have a reliable way to invalidate a volatile query based relationship. I wonder if in this case the onus is on the developer to tell the system that the query is volatile, or if we could statically determine that...

ef4:
@bagby you're right that other computed fields that depend on query-based-relationships will also need to become lazy and get recomputed at searcher time. We can make that work.

I hope that computed fields already do the right thing in terms of using the search index as a cache, but recomputing on demand if they're not in the cache. If that's not already true, we should fix that first.

Then we can make sure that any computed that reads a volatile field becomes volatile itself, and we don't save any volatile fields into the search index.

habdelra:
@ef4 so the `read()` function that is used by `Model` to get a relationship (as well as the `read()` used to get the data to instantiate the `Model` itself) does actually use caching (this is the `read()` that comes from `DocumentContext`, who in turn received it from hub:searchers). is this the caching that you are talking about?

bagby:
> I hope that computed fields already do the right thing in terms of using the search index as a cache, but recomputing on demand if they're not in the cache. If that's not already true, we should fix that first.

@ef4 I'm unsure what you mean by "using the search index as a cache", are you referring to `DocumentContext`'s `this.cache`? I might need a little handholding from you and/or @habdelra to walk through this.

bagby:
@ef4 I started on the changes we discussed yesterday, would appreciate if you could take a look and see if it's more or less in line with what we talked about. FWIW there's only 2 `computed-field` tests failing (including the new one I added), I'll take that as a positive sign :)

packages/hub/indexing/document-context.js
```diff
@@ -135,6 +138,40 @@ module.exports = class DocumentContext {
...
   }
 
+  async search(query) {
+    let key = `/api?${qs.stringify(query)}`;
+
+    log.debug(`Searching for records via ${key}`);
+
+    this._references.push(`${key}`);
+
+    // if (get(this, 'upstreamDoc.data.id') === id && get(this, 'upstreamDoc.data.type') === type) {
```
bagby: @ef4 I wasn't sure what to do with this (`search` started as a copy paste of `read`). It didn't _seem_ relevant for searching so I commented it out 🤷‍♂️

ef4:
I came back and reviewed this anyway and it all looks great, thanks @bagby.

Issue #733, opened 2019-04-05
Model-based encapsulation


habdelra:
- no more global name space (i'm looking at you, fields!)
- data service is aware of which card is asking for information so that we know how to scope the response.
- the line between models and cards is blurry. Need to make a clearer distinction between what is a model and what is a card (generally today you are dealing with cards not models). make this more formal in the system.
- data service arguments to load a card are:
  - npm package (replaces type),
  - id,
  - version (aka branch). there are defaults. the right default depends on who is asking. 
  - data source (interchangeable with repo)

- relationships will need to be modified to use the tuple above. need to synthesize the type and id based on this tuple (figure out our rule for doing this).
- made a new type called cards?
  - the new type cards will use the new id format that includes this tuple

- making a developer oriented store for optimizing modifying the data for a card. think of thing that is more "card" oriented compared to how we are storing models in the git data source. 
- when serialized the card's package/id/version/repo can be derived from repo and file folder. (version is actually only realizable when exported from git).

- start with looking at where we are loading things by type and id. are we loading a "card" or a "model" (mostly its a card). let's make 2 new methods on searchers.get getModel() and getCard() and update all callers to use the more explicit API.

- When you want to get a model that needs to happen in the context of a card now (need to know whose asking).

- split version off of the API so that we are not fashioning relationships with the version bound into the ID. 

- when a document is saved we save a "snapshot"  of the document with the specific version in the index which is different than the "current" version (which is just an update of the current document's row). 
  - Need to also make sure that this process is both incremental and full--so that if the index is destroyed and the recreated we have all the snapshots. indexers can be amended to also emit history to support this full case (which might be a background step via jobs). 
  - Also we need the package version when saving the snapshot (in the meta). think about the developer use case, where a developer is constantly updating the package version--think about a feature where the version of the package is whatever the developer has checked out.
  - also, think about how we need to migrate card data between package versions.

- stop using static-models for defining card definitions (schema). schema is a new feature for cards. 
  - we care write it the same as we do for static-models, it's a new feature. 
  - probably very similar to how static models loads. this unlocks the ability to index data sources separately.

balinterdi:
I very much welcome this issue as it addresses issues of clarification/simplification that I've bumped into several times.

For example, 
> the line between models and cards is blurry. Need to make a clearer distinction between what is a model and what is a card (generally today you are dealing with cards not models). make this more formal in the system.

was one of the first things I encountered and I'm happy to see it's covered, too.

[...]

PR #738, opened 2019-04-09
Refactor searchers get

habdelra:
***BREAKING CHANGE***

This is part of the work for #733.  Using newer API for `hub:searchers.get` and `hub:searchers.search`. After going thru the homework of seeing where all `Searchers.get()` and `Searchers.search()` is being called from, it becomes evident that from the outside of `Searchers`, for the most part only cards are actually requested. A few notable exceptions:
- schema: for the time being we are going to treat schema as cards until that doesn't make sense
- spaces: spaces are inherently not cards. To resolve this I've added a `Searchers.getSpace()` method that we can embellish as time goes on.
- binary data: There exists a `Searchers.getBinary()` that allows us to retrieve the binary data that can back a model. This represents what we'd consider an internal model of a card. We should follow up with another PR that allows access to the `Searchers.getBinary()` only from an internal model API endpoint that is namespaced with a card (e.g. `GET /api/${repo}/${package}/${cardId}/${modelType}/${modelId}`). Additionally, from inside the hub, you should also have to provide the `repo`, `package`, and `cardId` to the `Searchers.getBinary()` method.

I've updated the `Searchers.get` method signature to look like this (`package` is actually a reserved keyword in javascript when in `strict` mode--hence `packageName`):
```
  async get(session, packageName, id, opts={})
```
Where opts can include: `version` and `includePaths`. Omitting the `version` would imply that you want the current card.

I've updated the `Searchers.search` method signature to look like this:
```
  async search(session, query, version)
```
I'm a little unsure about the `version`. right now it is the placeholder for what we are currently calling the `branch`--but honestly it might make more sense to include the `version` in the query itself.

For the searcher hub feature, I have left that unchanged. This implies that the searcher hub feature is used to return cards. I opted not to split the searcher feature to supply separate hooks for returning models and cards, since it looks like in practice are are not consuming models outside of hub:searchers--based on how everything shakes out, we may revisit that. 

Finally, if you call `Searchers.search` or `Searchers.get` in what looks like to be the old way, you'll get a helpful deprecation exception informing you of the new way these are called.

### DIFF
--- a/packages/ethereum/cardstack/transaction-indexer.js
+++ b/packages/ethereum/cardstack/transaction-indexer.js
@@ -274,6 +274,7 @@ class TransactionIndexer {
     let batch = this.pgsearchClient.beginBatch(this.schema, this.searchers);
     for (let address of addresses) {
       let document = await this.searchers.get(Session.INTERNAL_PRIVILEGED,
+        'local-hub',
         'ethereum-addresses',
         address,
         ['transactions.from-address', 'transactions.to-address']);
@@ -313,7 +314,7 @@ class TransactionIndexer {
 
     let addressResource;
     try {
-      addressResource = await this.searchers.get(Session.INTERNAL_PRIVILEGED, 'ethereum-addresses', address.toLowerCase());
+      addressResource = await this.searchers.get(Session.INTERNAL_PRIVILEGED, 'local-hub', 'ethereum-addresses', address.toLowerCase());
     } catch (err) {
       if (err.status !== 404) { throw err; }
     }
--- a/packages/ethereum/cardstack/transaction-indexer.js
+++ b/packages/ethereum/cardstack/transaction-indexer.js
@@ -273,7 +273,7 @@ class TransactionIndexer {
 
     let batch = this.pgsearchClient.beginBatch(this.schema, this.searchers);
     for (let address of addresses) {
-      let document = await this.searchers.getFromControllingBranch(Session.INTERNAL_PRIVILEGED,
+      let document = await this.searchers.getCurrentCard(Session.INTERNAL_PRIVILEGED,
         'ethereum-addresses',
         address,
         ['transactions.from-address', 'transactions.to-address']);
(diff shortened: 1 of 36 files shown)


habdelra:
@ef4 this is ready for you to look at now

habdelra:
Based on conversation with @ef4: let's add the `source` parameter to these calls. And so that we can easily refactor laster, let's use a well-known string to represent the "local" source.

habdelra:
@ef4 I've updated the `Searchers.get()` API to include a `source` param as well so that it reflects the ideal API for retrieving a card.

packages/ethereum/cardstack/transaction-indexer.js
```diff
@@ -313,7 +314,7 @@ class TransactionIndexer {
 
     let addressResource;
     try {
-      addressResource = await this.searchers.getFromControllingBranch(Session.INTERNAL_PRIVILEGED, 'ethereum-addresses', address.toLowerCase());
+      addressResource = await this.searchers.get(Session.INTERNAL_PRIVILEGED, 'local-hub', 'ethereum-addresses', address.toLowerCase());
```
habdelra: @robustdj here is an example of what the new searchers.get() looks like. there is no more `getFromControllingBranch()` if you see any of those change them to just `get()`, and provide a parameter called `local-hub` as the 2nd parameter (we don't use this yet, but this is how we'll be able to target foreign hubs when getting stuff form the searcher).

packages/git/node-tests/config-test.js
```diff
@@ -88,7 +88,7 @@ describe('git/config', function() {
 
     env = await createDefaultEnvironment(`${__dirname}/..`, factory.getModels());
 
-    let response = await env.lookup('hub:searchers').search(env.session, 'master', { filter: { type: 'articles' } });
+    let response = await env.lookup('hub:searchers').search(env.session, { filter: { type: 'articles' } });
```
habdelra: @robustdj for the `searchers.search()` we've removed the branch parameter. just delete these.

packages/git/node-tests/indexer-remote-test.js
```diff
@@ -201,7 +201,7 @@ describe('git/indexer cloning', function() {
     await start();
     await indexer.update();
 
-    let contents = await searcher.get(env.session, 'master', 'events', 'event-1');
+    let contents = await searcher.get(env.session, 'local-hub', 'events', 'event-1');
     let jsonapi = toResource(contents);
```
habdelra: @robustdj for the `searchers.get()` we no longer take a branch parameter, instead of that, we now just pass the hard coded string `local-hub`, which will be useful in the future to specifying getting a model from a foreign hub instance. for now we only support your local hub instance.

PR #740, opened 2019-04-11
Use error card (instead of application card) as primary card when route not found

### LINKED ISSUE #590: Routes with path '/' seem to map single segments paths as well
It seems like routes defined for `path: '/'` map single segment paths as well, e.g.:

```js
{
  path: '/',
  query: {
    filter: {
      type: { exact: ':card:type' },
      id: { exact: ':card:id' }
    }
  }
}
```

will actually catch `/fdsfdsfdsf`.

see discussion in cardstack/dotbc#247

robustdj:
Fixes #590

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -121,8 +121,8 @@
                 "packages/test-support/bin/run.js",
                 "--timeout",
                 "200000",
-                // "--grep",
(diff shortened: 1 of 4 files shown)


habdelra:
nice!

PR #742, opened 2019-04-12
Schema feature

habdelra:
This PR addresses #733 and introduces a `schema` hub feature that cards can use to define their schema. Up until now, we've been using the `static-models` feature to define both schema and static models for cards. This PR will make it such that if a schema model is specified in `static-model.js` the hub will throw an error. Likewise, if a non-schema static model is defined in `schema.js` the hub will throw an error. So schema and static models are separated in to different features as part of this PR.


ef4:
I don't think we should land the new schema feature until it has the actual API we know we want it to have. This is not that API.

If we manage to land the next stage quickly it wouldn't be a big deal, but if for any reason we get pulled in another direction and this work stalls, we end up in a broken intermediate state where every card is forced to update into a format we don't actually want to support.

habdelra:
sure. I'm gonna add a pointer to our schema refinement gist here as well so we can bring these two efforts together: https://gist.github.com/ef4/b3d92ccfcf91b747081707c0bd99eb55

habdelra:
@ef4 This is ready for another review. Hopefully the PR now represents the API of the card schema feature that we want to have. There are a couple card schema file examples in our tests, you can find that here https://github.com/cardstack/cardstack/pull/742/files#diff-3692b6da2f43674dede70505e1e3a9b7 and here https://github.com/cardstack/cardstack/pull/742/files#diff-098412965e618af9d5ccded9f17b6ae6, which are tested here: https://github.com/cardstack/cardstack/pull/742/files#diff-f0fb1c576963e8458635f112f8779396

For convenience, this is what the article-card's `schema.js` looks like:
```js
const JSONAPIFactory = require('@cardstack/test-support/jsonapi-factory');
module.exports = function({sourceId, packageName}) {
  // Article Card Schema
  let factory = new JSONAPIFactory(sourceId, packageName);

  factory.addResource('card-definitions')
    .withRelated('model', factory.addResource('content-types', 'articles')
      .withAttributes({ name: 'articles' })
      .withRelated('fields', [
        factory.addResource('fields', 'title').withAttributes({
          name: 'title',
          'is-metadata': true,
          'needed-when-embedded': true,
          'field-type': '@cardstack/core-types::string'
        }),
        factory.addResource('fields', 'body').withAttributes({
          name: 'body',
          'is-metadata': true,
          'field-type': '@cardstack/core-types::string'
        }),
        factory.addResource('fields', 'categories').withAttributes({
          name: 'categories',
          'is-metadata': true,
          'needed-when-embedded': true,
          'field-type': '@cardstack/core-types::has-many'
        }).withRelated('related-types', [
          // this is modeling an enumeration
          factory.addResource('content-types', 'categories').withAttributes({ name: 'categories' })
        ]),
        factory.addResource('fields', 'author').withAttributes({
          name: 'author',
          'is-metadata': true,
          'needed-when-embedded': true,
          'field-type': '@cardstack/core-types::belongs-to'
        }).withRelated('related-types', [{ type: 'content-types', id: 'cards' }])
      ])
    );
  return factory.getModelsFor('card-definitions');
};
```

packages/hub/node-tests/cards-test.js
```diff
@@ -0,0 +1,241 @@
...
+
+  async function teardown() {
+    if (env) {
+      await destroyDefaultEnvironment(env);
+    }
+  }
+
+  describe('card schema', function () {
+    let searchers;
+    //TODO also test: field instructions, field placeholder, field constraints
```
habdelra: I'm still adding these but this should be far enough along to be worth reviewing

tests/stub-cards/valid-cards/node_modules/article-card/schema.js
```diff
@@ -0,0 +1,39 @@
...
+      .withAttributes({ name: 'articles' })
+      .withRelated('fields', [
+        factory.addResource('fields', 'title').withAttributes({
+          name: 'title',
+          'is-metadata': true,
+          'needed-when-embedded': true,
+          'field-type': '@cardstack/core-types::string'
+        }),
+        factory.addResource('fields', 'body').withAttributes({
+          name: 'body',
```
habdelra: seems duplicative to set the resource "id" (it's not the actual ID because the ID is really a composite) and field's "name" to the same value. I think we can add some additional smarts to the `JSONAPIFactory` to add the "name" attribute for the field content type when the `JSONAPIFactory` has knows about the card definition context (because you specified that in it's constructor).

tests/stub-cards/valid-cards/node_modules/article-card/schema.js
```diff
@@ -0,0 +1,39 @@
...
+          'field-type': '@cardstack/core-types::string'
+        }),
+        factory.addResource('fields', 'categories').withAttributes({
+          name: 'categories',
+          'is-metadata': true,
+          'needed-when-embedded': true,
+          'field-type': '@cardstack/core-types::has-many'
+        }).withRelated('related-types', [
+          // this is modeling an enumeration
+          factory.addResource('content-types', 'categories').withAttributes({ name: 'categories' })
```
habdelra: same comment around being duplicative for setting the `name` attribute on `content-types` to be the same as the "id" of the resource (which is not the actual ID because the ID is really a composite)

tests/stub-cards/valid-cards/node_modules/article-card/schema.js
ef4: Can we just leave the ID blank? `addResource` doesn't require that you pass one.

packages/hub/bootstrap-schema.js
```diff
@@ -693,6 +813,55 @@ const models = [
       'field-type': '@cardstack/core-types::object'
     }
   },
+  {
+    type: 'fields',
+    id: 'member-name',
```
ef4: What's the deal with the `member-name` field on `content-types` and `fields`?

Is it derived state, or something people are really allowed to set in their schema.js?

If it's internal derived state, it doesn't belong in bootstrap schema, and would only be a property on the actual `Field` and `ContentType` classes we use internally.

If it's the public API for setting the name of a field or content-type, then it should be the *only* public API for setting them. And we could call it `name` instead (remember, even the bootstrap schema will now be scoped! Field names in here should not collide with any user fields).

packages/routing/cardstack/static-model.js
```diff
@@ -1,99 +1,5 @@
 module.exports = function() {
   return [{
-    type: 'content-types',
-    id: 'spaces',
```
ef4: Why did this stuff need to move into bootstrap-schema?

packages/hub/plugin-loader.js
```diff
@@ -21,6 +21,7 @@ const featureTypes = [
   'computed-field-types',
   'writers',
   'searchers',
+  'schemas',
```
ef4: People don't always agree on what the plural of "schema" should be, but this is fine.

packages/hub/schema/data-source.js
```diff
@@ -65,8 +74,39 @@ module.exports = class DataSource {
...
+      card indexing at the same time as well. The next step here is to remove this grandfathering so that everything uses
+      the new schema feature and as a result, everything is forced to use the new card boundaries. This can be done byr
+      uncommenting the line below and unskipping the unit test that goes along with this.
+
+      throw new Error(`The datasource '${this.id}' defines static-models that includes schema. Schema is not allowed in static-models. Found schema models: ${JSON.stringify(schemaModels)}`);
+    */
+    }
     return this._staticModels;
   }
+  get staticSchemaModels() {
```
ef4: This seems problematic. We don't want schema.js to be a per-data-source feature. It is unlike static-model.js in that crucial way.

We need to load the schema for a card any time the card is part of our dependency graph. We don't want to need to configure the card as a data-source first.

Our loading of the `schemas` feature should work similar to how we load `middlewares` or `field-types`. Which is unlike how we load `writers` or `static-models`.

packages/test-support/addon/jsonapi-factory.js
```diff
@@ -1,20 +1,37 @@
...
+    this.sourceId = sourceId;
+    this.packageName = packageName;
   }
 
-  addResource(type, id=null) {
-    if (id == null) {
-      id = idGenerator++;
+  addResource(type, _id=null) {
+    let id = _id;
+    if (type === 'card-definitions' && this.sourceId != null && this.packageName != null) {
```
ef4: The fact that JSONAPIFactory needs to know about the composite id scheme and needs to get it correct is a problem. 

The factory is a convenience DSL only, people can author their schema.js in pure json too (and that will become the more common thing to do as soon as we have tools for manipulating the schema, instead of hand writing it). But this means that people will be free to author invalid schema.js files because they don't follow the composite id format, which is not a thing I want to expose to users.

And while we can easily detect invalid ids and bully users into typing out correct ones, that's unnecessary when we definitely have enough information to do the right thing automatically.

It would be better if we could forbid users from picking IDs at all in here. But to express that in json:api, we would need to use json:api operations format, rather than a straight-up json:api document. That may be the way to go. Alternatively, we can just state that the ids you put into schema.js are local to that file only and you don't get to expect them to mean anything other than expressing relationships inside that file.

packages/test-support/addon/jsonapi-factory.js
```diff
@@ -25,39 +42,49 @@ export default class JSONAPIFactory {
     return new ResourceFactory(resource);
   }
 
-  getModels() {
+
+  getModelsFor(type, id) {
```
ef4: The name of this method confused me. But also, it returns the wrong format (based on where you're calling this, as the return value from schema.js files). A schema.js file should not return an array. It should return a complete json:api document, with the single card definition as `data` and all the related stuff in `included`.

tests/stub-cards/card-definition-id-does-not-match-package-name/node_modules/person-card/schema.js
```diff
@@ -0,0 +1,21 @@
+const JSONAPIFactory = require('@cardstack/test-support/jsonapi-factory');
+module.exports = function() {
+  let factory = new JSONAPIFactory();
+  // Person Card Schema
+  factory.addResource('card-definitions', 'not-person-card')
+    .withRelated('model', factory.addResource('content-types', 'people')
+      .withRelated('fields', [
+        factory.addResource('fields').withAttributes({
+          name: 'name',
```
ef4: Now I'm even more confused about `member-name`, because we already have `name`.

packages/hub/node-tests/cards-test.js
```diff
@@ -0,0 +1,247 @@
...
+        try {
+          await createDefaultEnvironment(`${__dirname}/../../../tests/stub-cards/non-schema-models-in-schema-file`, factory.getModels());
+        } catch (e) {
+          error = e;
+        }
+
+        expect(error.message).to.match(/Non-schema models are not allowed in schemas/);
+      });
+
+      it('throws when card schema models are not scoped to the source and package that they originate from (which can happen when JSONAPIFactory is not given source and package context)', async function () {
```
ef4: This error becomes impossible if we switch to local-only (non-composite) ids.

packages/hub/node-tests/cards-test.js
```diff
@@ -0,0 +1,247 @@
...
+        try {
+          await createDefaultEnvironment(`${__dirname}/../../../tests/stub-cards/jsonapi-factory-missing-card-context`, factory.getModels());
+        } catch (e) {
+          error = e;
+        }
+
+        expect(error.message).to.match(/has schema models that are not scoped to this data source and package/);
+      });
+
+      it('throws when card-defintions id does not match source::package of the schema file name', async function () {
```
ef4: This error becomes impossible if we switch to local-only (non-composite) ids.

packages/hub/node-tests/cards-test.js
```diff
@@ -0,0 +1,247 @@
...
+        try {
+          await createDefaultEnvironment(`${__dirname}/../../../tests/stub-cards/card-definition-id-does-not-match-package-name`, factory.getModels());
+        } catch (e) {
+          error = e;
+        }
+
+        expect(error.message).to.match(/that is not scoped for this source::package/);
+      });
+
+      it('throws when there are mulitple card definitions in schema file', async function () {
```
ef4: This error becomes impossible if we switch to returning a json:api document from schema.js. Or rather, the error condition would be slightly different (it would just need to check that `data` is singular and has type `card-definitions`).

packages/hub/node-tests/cards-test.js
```diff
@@ -0,0 +1,247 @@
...
+        try {
+          await createDefaultEnvironment(`${__dirname}/../../../tests/stub-cards/multiple-card-definitions-in-same-schema-file`, factory.getModels());
+        } catch (e) {
+          error = e;
+        }
+
+        expect(error.message).to.match(/has more than one card-definitions model/);
+      });
+
+      it('throws when there are no card definitions in schema file', async function () {
```
ef4: Similarly, this particular error condition becomes impossible if we switch to returning a json:api document from schema.js.

tests/stub-cards/valid-cards/node_modules/article-card/schema.js
habdelra: sure

packages/hub/bootstrap-schema.js
habdelra: Yeah, it;s the public API for setting the name of the field. I was hedging on using the word "name" for this attribute, and wanted to be a bit more descriptive (JSON:API itself calls this thing the member name). But I can go back to using `name`.

packages/routing/cardstack/static-model.js
habdelra: sorry, that was a change that I made when I was working thru throwing errors when schema stuff appeared in the static-model feature, but then realized that its too soon to start enforcing that. I'll move these things back.

tests/stub-cards/card-definition-id-does-not-match-package-name/node_modules/person-card/schema.js
habdelra: this was me not being complete in switching to `member-name`

packages/test-support/addon/jsonapi-factory.js
habdelra: got it 👍

packages/test-support/addon/jsonapi-factory.js
habdelra: what about all the existing uses of `JSONAPIFactory` for arbitrary test setup? should I refactor all that to use just plain objects? or should we provide a means that you can assert an ID in the resource creation for the time being, until we actually start using the new schemas in earnest?

habdelra:
@ef4 in regards to making id's local to the schema.js file, I'm not too sure how to reconcile that with the id's that get created in the `documents` table for the content types and fields that we are defining in the schema.js file. Are you saying that we can use the composite ID's for `content-types` and `fields`, but that we need to derive those outside of `JSONAPIFactory` (and what does that mean for the JSON:API that is emitted from JSONAPIFactory for in the schema.js file)? or are you saying we shouldn't use composite ID's at all for these `content-types` and `fields` as they are represented in the `documents` table--in which case I'll need some further guidance in terms of how we do want these schema models to look in the documents table.

packages/hub/schema/data-source.js
habdelra: Ok, so I think this means that in `SourcesUpdate.add()` I should look at the document that is being added to the index, and if it is a document that depends on a card for which we have a schema.js file, that is when we load the schema (before the document being added to the index is actually added). And as part of that I should be smart about loading the schema such that I don't reload a card's schema that I've loaded previously (like we do when loading the middleware stack). Is this the correct approach?

habdelra: oops, I was rereading this, when you said "dependency graph" you mean the package dependency graph (I was originally thinking that you meant we'd load do some kind of JIT card schema loading when an indexer encountered a card instance it didn't know about). So i think this is pretty straight forward: just move the logic that we use to load the schema feature models out of the `RunningIndexers._activateDataSource()`, since as you say, we should load the schema regardless of whether the data source is activated.

habdelra: (this also answered another question I had in a comment questioning why we need to activate data sources in order to use card schemas)

habdelra:
@ef4 I have resolved all the points that you raised in your review. 
- the return of the JSONAPIFactory is now a document
- I'm using the `name` attribute for the name of the field/content-type
- the ID's within schema.js are local to that file, and the JSONAPIFactory is not doing anything special (aside from being able to write out jsonapi documents).
- the card schema feature do not require the loading the card as a data source--rather we just load any card schemas that are in the package dependency graph.
- plus misc cleanup items that you identified.

Could you take another look and let me know if things look ok to merge?

packages/data/tests/helpers/test-data-setup.js
```diff
@@ -116,13 +116,13 @@ export default function(factory, opts) {
 
   let vanGogh = factory.addResource('puppies', 'vanGogh')
     .withAttributes({
-      name: 'Van Gogh',
+      nickname: 'Van Gogh',
```
ef4: Presumably this is avoiding a collision, but the collision is only a temporary thing until we can refactor all tests to go through the new schema feature?

packages/hub/indexing/running-indexers.js
```diff
@@ -49,16 +51,43 @@ module.exports = class RunningIndexers {
...
+  get cardSchemas() {
+    if (this._cardSchemas) { return this._cardSchemas; }
+
+    let schemas = [];
+    let plugins = this.seedSchema.plugins;
+    let features = this.seedSchema.plugins.featuresOfType('schemas');
+    for (let feature of features) {
+      let { id:featureName } = feature;
+      let factory = plugins.lookupFeatureFactory('schemas', featureName);
+      // It's not clear to me how you could load card schema from different sources if we load card schema outside of a data-source context...
```
ef4: You're correct that they all come from one source right now (the on-disk node_modules tree of cardstack plugins).

But this won't remain true for long. We plan to move away from having a static set of pre-installed cards.

packages/data/tests/helpers/test-data-setup.js
habdelra: confirm

PR #747, opened 2019-04-17
removed the concept of "branches" from cardstack

habdelra:
***BREAKING CHANGE***

Removed the concept of "branches" from cardstack. Branches will soon be replaced with a more canonical concept of cards that facilitates exchanging cards in a multi-hub scenario. This is an intermediate step to that future.

Note that the @cardstack/git module still retains the idea of branches insofar as a means to index a git datasource. But within our index there is no more idea of a branch. Also as part of this I'm removing the elastic search module--which we can resurrect in the future if that becomes important.


habdelra:
@ef4 This is ready for your review now

habdelra:
cool, that sound good

packages/ephemeral/indexer.js
```diff
@@ -5,11 +5,7 @@ module.exports = declareInjections({
   loadInitialModels: 'config:initial-models'
 }, class Indexer {
 
-  async branches() {
-    return ['master'];
-  }
```
habdelra: @robustdj this is an example of a change that should be made to any indexers. (we have a couple indexers in the various cards of portfolio). you'll need to delete the `branches()` hook that you find in any indexers.

packages/ephemeral/indexer.js
```diff
@@ -5,11 +5,7 @@ module.exports = declareInjections({
...
-
-  async beginUpdate(/* branch */) {
+  async beginUpdate() {
```
habdelra: @robustdj additionally, no more need to pass in the `branch` parameter in the `beginUpdate()` of any indexers that you encounter in the cards or portfolio.

packages/ethereum/cardstack/event-indexer.js
```diff
@@ -13,9 +13,8 @@ function attachMeta(model, meta) {
 
 module.exports = declareInjections({
   indexer: 'hub:indexers',
-  controllingBranch: 'hub:controlling-branch',
```
habdelra: @robustdj controllingBranch is no longer a thing. delete these (and downstream stuff) where every you see it.

packages/ethereum/cardstack/event-indexer.js
```diff
@@ -64,13 +62,12 @@ class EthereumEventIndexer {
   async _indexRecord(batch, record) {
     log.debug('indexing model in pgsearch %j', record);
     let { id, type } = record;
-    let schema = await this.schema.forControllingBranch();
+    let schema = await this.currentSchema.getSchema();
```
habdelra: @robustdj I'm unsure if `current-schema` is injected into any of the modules we use in portfolio, but it it is, this has been updated to return a schema from the `getSchema()` function. (this used to only get a schema for a specific branch)

packages/git/node-tests/writer-test.js
```diff
@@ -111,7 +111,7 @@ describe('git/writer', function() {
 
   describe('create', function() {
     it('saves attributes', async function () {
-      let { data:record } = await writers.create('master', env.session, 'articles', {
+      let { data:record } = await writers.create(env.session, 'articles', {
```
habdelra: @robustdj this is an example of a writer change. we need to remove the branch parameter from any `writers.create()` calls being made (I think there is some middleware in one of the cards that is doing this).

packages/git/node-tests/writer-test.js
```diff
@@ -311,7 +311,7 @@ describe('git/writer', function() {
 
     it('requires id in body', async function() {
       try {
-        await writers.update('master', env.session, 'articles', '1', {
+        await writers.update(env.session, 'articles', '1', {
```
habdelra: @robustdj same for `writers.update()` we need to remove any branch parameters being passed into this function. we might have some of these within the cards of portfolio.

PR #750, opened 2019-04-22
Update Ember packages

jenweber:
Changed:
- update packages when possible
- small js linting applied by running codemods
- resolving template linting errors

Packages that need work before they can be upgraded are those that rely on liquid fire or the edges, plus core-types needs to have ember-let removed & refactored to use the core API.


jenweber:
This is ready for review now. Anyone can jump in.

ef4:
Tests are failing. For example, the `image` package is throwing some exception about jquery.

PR #754, opened 2019-04-26
Make editor attributes panel dynamic

burieberry:
- Editor panel display changes depending on what the user clicks on the rendered template
- User can return to main panel ("global attributes") by clicking elsewhere on page
- Stop trying to show overlay for hidden fields
- Adjust colors and z-indices on tools


packages/tools/addon/styles/cs-field-overlays.css
```diff
@@ -37,3 +42,15 @@
 .cs-field-overlay.passthrough > * {
   pointer-events: auto;
 }
+
+/* Overriding style attribute that sets `visibility` to `hidden`.
+   This is so that when user can exit focus on field by
+   clicking elsewhere on the page */
+.cs-field-overlay.focused > .liquid-child {
+  visibility: visible !important;
```
ef4: This is likely to cause animations to flicker. The liquid-child is only supposed to be hidden temporarily, during the start of an animation. If it's staying hidden, that's a bug, and we should fix that instead.

packages/tools/addon/templates/components/cs-active-composition-panel.hbs
```diff
@@ -9,62 +9,25 @@
...
-              </div>
-            {{/let}}
-          {{/each}}
-        </div>
-
-        {{#each renderedFields key="id" as |fieldMark|}}
-          {{! template-lint-disable attribute-indentation }}
+  {{#if openedFieldId}}
+    <div class="cs-active-composition-panel--parameters">
+      {{#each renderedFields key="id" as |fieldMark|}}
```
ef4: This template file was already at or beyond the maximum acceptable complexity. The copy-paste duplication introduced by this PR pushes it way beyond understandable. So we need to factor the shared pieces out into one or more separate components.

packages/tools/addon/styles/cs-field-overlays.css
burieberry: Yes, something is off with the scrim behavior, I'll investigate further.

packages/tools/addon/templates/components/cs-field-overlays.hbs
```diff
@@ -6,20 +6,25 @@
       mark.model.content
       mark.model.caption
       (eq mark.id tools.openedFieldId)
+      (eq mark.id tools.detailedViewFieldId)
       (cs-field-editor-for mark.model.content mark.model.name variant="inline")
-    as |fieldName content caption isOpenedField editorName|}}
+    as |fieldName content caption isOpenedField isDetailedViewField editorName|}}
       {{#let (is-component editorName) as |hasInlineEditor|}}
+        <div hidden=true onclick={{action "openField" mark}} data-test-overlay-button={{dasherize fieldName}}>
```
burieberry: What's a better way to do this? Trying to click on the field overlay below doesn't work during testing so I had to add a hidden `div` with the same click action on it.

packages/tools/addon/templates/components/cs-active-composition-panel.hbs
burieberry: Any better now?

packages/tools/addon/templates/components/cs-field-overlays.hbs
jenweber: Do you have any ideas why the click doesn't work in testing?

I would add an Esc key action, for both keyboard nav and testing. In testing you'd have to still set the focus on the overlay though. See "closing the dialog" on https://www.nomensa.com/blog/2014/how-improve-accessibility-overlay-windows-part-1

packages/tools/addon/templates/components/cs-active-composition-panel.hbs
jenweber: @ef4 can you take another look at this?

burieberry:
I'm closing this since the changes in tools package are currently irrelevant. We can reopen in the future if necessary.

PR #761, opened 2019-05-06
[WIP] Model encapsulation refactor

habdelra:
@ef4 @jenweber I'd like to direct your attention to the `@cardstack/hub/indexing/document-context.js` and the `@cardstack/pgsearch/client.js`. Right now all the `@cardstack/pgsearch/node-tests/indexer-tests.js` are working.

This approach, as we discussed, creates a row in the `documents` table per card. card internal models are encapsulated solely in the `included` of each card's row.


habdelra:
@ef4 @jenweber  I'm ready to hand off the model encapsulation work now. I've updated this branch to use the current documents table schema (no schema changes being made now), and to create one row per model in the index (where cards are automatically created and removed from the index based on adding or deleting the primary model for the card).

In the `cardstack/@pgsearch`'s  `indexer-tests.js` you'll find all the original "legacy model" tests as well as all the new card based tests. currently they are all passing (I have 2 card tests being skipped right now, but they are much more esoteric tests that seem to be leaking). I'm unsure 

A large chunk of the work that is not done is everything around handling metadata relationships as card relationships (as opposed to objects in card metadata). As part of this we need to make sure to handle the `default-metadata-includes` paths that are set on the `card-definitions` (similar to `default-includes` on content-types). Ideally building card metadata should look really similar to how we build fields (both attrs and relationships) for content-types within the `DocumentContext`. Also, make sure that the `searchdocs` for the card metadata (including the card's metadata relationships) is constructed correctly.

packages/pgsearch/node-tests/indexer-test.js
```diff
@@ -6,17 +6,133 @@
...
+  let env, writer, indexer, searcher, currentSchema, changedModels;
 
-  before(async function() {
+  // Note that we are rebuilding the env between each test as the tests otherwise leak between the card tests and legacy model tests
+  beforeEach(async function () {
     this.timeout(5000);
+
+    let cards = [];
+    let factory = new Factory();
+    cards.push(addContextForCardDefinition(sourceId, 'puppy-card', factory.getDocumentFor(
```
habdelra: @ef4 here is a new schema example

Issue #764, opened 2019-05-08
Refactor codegen to work with Embroider


jenweber:
> Module not found: Error: Can't resolve '@cardstack/plugin-utils/environment' in '/private/var/folders/9k/l7x9d45s373_nchwp07mkrqr0000gn/T/embroider/900d16/Users/jenweber/cardstack/cardstack/packages/codegen'

To-do:
- [x] change environment codegen to modules
- [x] change models codegen to modules
- [x] change ethereum codegen to modules
- [x] change github-auth codegen to modules
- [x] change live-queries codegen to modules
- [x] change test-support codegen to modules
- [x] ~~figure out how to test ethereum codegen (there were no node/ember tests for it)~~ removed
- [x] figure out how to test github-auth codegen (no test coverage of generated code) - added it to an existing test
- [x] test live-queries codegen. created new node-tests.
- [x] remove generateCode use

jenweber:
Steps to show generated code:

From `packages/cli`, run:
```
CARDSTACK_DEV=true node ./bin/cardstack.js run
```

cd into the `tmp dir` printed to the console, then one more level into cardstack-standard-app. From there:

```
ember hub:start
```

Visit `http://localhost:3000/codegen/foo` to see generated code. Not all codegens are used by the CLI's app so some generators will not be visible, and instead you'll just need to run the tests.

jenweber:
We removed the ethereum codegen because we would likely need to replace it in the future anyway.

jenweber:
Closed by #768

PR #768, opened 2019-05-13
Codegen refactor 764

jenweber:
Partially addresses #764

## What it does

generateCode is replaced by generateModules. The generated content is a better match for the work with Embroider.

## What changed
- convert hub/code-generators to TS
- Did some additional configuration of gitignore so eslint doesn't go too deep
- removed generateCode
- added generateModules
- added dependency on babel
- removed ethereum codegen


.gitignore
```diff
@@ -1,4 +1,4 @@
-/node_modules
+/**/node_modules
```
ef4: This may lead to trouble. We *do* check in some code that happens to live inside folders named node_modules. This will make it confusing if somebody tries to add a new file there.

See for example https://github.com/cardstack/cardstack/blob/master/tests/stub-middleware/node_modules/stub-middleware-extra/middleware.js

The lint errors that I think motivated this change:

```
/home/travis/build/cardstack/cardstack/tests/ephemeral-test-app/node_modules/@cardstack/hub/code-generators.d.ts
  2:1  error  Expected indentation of 2 spaces but found 4  @typescript-eslint/indent
```

are happening because we have a rule to ignore `/packages/hub/code-generators.d.ts` but we don't have a rule to ignore `/tests/ephemeral-test-app/node_modules/@cardstack/hub/code-generators.d.ts`, which is really the same file, but via a symlink. 

Probably we could ignore everything below the symlink like:

```
/tests/ephemeral-test-app/node_modules/@cardstack/hub/**/*
```

jenweber: Makes sense. I removed the `**` and replaced it with specific rules.

Issue #757, opened 2019-05-02
Epic - enable a quickstart experience for writing Cards


jenweber:
With a lot of the infra and planning in place, it's time to bring the Cardstack "quickstart" experience to life. What should a developer do to get started making their own cards?

### To-do:
- [x] Create a CLI with basic commands (run, for example)
- [ ] create a temporary cache/persistence so that the data people use in cards while they are working locally does not get destroyed by container restarts
- [x] Cards will render within a virtual app with cardstack run
- [ ] Establish a skeleton of what card files should be (i.e. not an entire addon filetree, because it's too much boilerplate)
- [x] Incorporate Embroider into the build
- [x] Handle issues with dynamic code generation. Embroider expects to know the deps.
- [ ] Incorporate Hassan's work on model schema into the Card filesystem and build
- [ ] Add more Card-focused CLI commands, like creating a card, for example

### How we know we did it

We can write a real tutorial where someone installs the Cardstack CLI, creates a card with templates and model schema, starts it with cardstack run, and can see the results served locally

jenweber:
```
Uncaught (in promise) TypeError: Found non-callable @@iterator
    at new ComputedProperty (metal.js:3174)
    at new TaskProperty (-task-property.js:411)
    at task (index.js:83)
    at eval (common.js:44)
```

jenweber:
Ember-cli resolves multiple copies of a lib by picking one. Embroider doesn't. We added a https://github.com/embroider-build/embroider/pull/179

```
webpack:////private/var/folders/9k/l7x9d45s373_nchwp07mkrqr0000gn/T/embroider/bde525/Users/jenweber/cardstack/cardstack/node_modules/ember-inflector/lib/ext/string.js?:11 Uncaught (in promise) TypeError: Cannot redefine property: pluralize
    at Function.defineProperty (<anonymous>)
    at eval (webpack:////private/var/folders/9k/l7x9d45s373_nchwp07mkrqr0000gn/T/embroider/bde525/Users/jenweber/cardstack/cardstack/node_modules/ember-inflector/lib/ext/string.js?:11)
    at Module.../../../../../../../../../Users/jenweber/cardstack/cardstack/node_modules/ember-inflector/lib/ext/string.js (chunk.bae4
```

jenweber:
```
[...]

PR #771, opened 2019-05-17
update ember-concurrency and power-select

jenweber:
This PR updates ember-concurrency and ember-power-select (which depends on EC).

The upgrades are needed for us to use Octane in a Card. When we turned on Embroider, it resolved dependencies in such a way that we were using an older version of ember-concurrency.

Co-authored-by: Edward Faulkner <edward@eaf4.com>


jenweber:
Part of work for #757

Issue #785, opened 2019-06-04
Refactor cardstack-modal-target to use ember-elsewhere `params`


jenweber:
In `cardstack-modal-target.hbs`, we have a from-elsewhere that accesses a component from a hash. This causes issues with embroider because the build can't know that a value like that is safe:

```hbs
{{#from-elsewhere name="cardstack-modal" as |args|}}
  {{#if args.content}}
    <div class="cardstack-modal-backdrop"></div>
    <div 
      role="button" 
      class="cardstack-modal-container"
      onclick={{action (cs-exact-target ".cardstack-modal-container" args.outsideClick) }}
    >
      <div class="cardstack-modal-dialog" >
        {{component args.content}}
      </div>
    </div>
  {{/if}}
{{/from-elsewhere}}
```

When the [params feature](https://github.com/ef4/ember-elsewhere/issues/31) lands in ember-elsewhere, we can refactor this to render the component from the `send` and utilize the rest of the hashed args under the `params`.

jenweber:
The branch `elsewhere-785` includes this fix. It can't be merged until a release of `ember-elsewhere` with the `outsideParams` feature, and we version-bump all uses of `ember-elsewhere` in the mono-repo.

jenweber:
This issue needs someone to take a quick look and see if it's still a thing to fix after the new-hub merge.

PR #789, opened 2019-06-10
upgrade ember-elsewhere and refactor usage

jenweber:
#785


jenweber:
This has an acceptance test in it that will only run headless. It was like that on master, but I'm going to debug it some to see if I can fix it together with this PR.

PR #795, opened 2019-06-11
upgrade to latest ember-auto-import

habdelra:


habdelra:
This may also require that the cardstack app polyfills nodejs builtins from webpack (specifically web3 and ethereum based cardstack apps will want to polyfill their nodejs builtins).

PR #800, opened 2019-06-13
Allow ethereum searcher to search by id

robustdj:
Closes https://github.com/cardstack/portfolio/issues/186

### DIFF
--- a/packages/ethereum/cardstack/searcher.js
+++ b/packages/ethereum/cardstack/searcher.js
@@ -38,6 +38,9 @@ class TransactionSearcher {
       query.filter.or.every(i => get(i, 'type.exact') === 'ethereum-transactions' ||
                                  get(i, 'type') === 'ethereum-transactions')) {
       return await this._queryTransactions(query);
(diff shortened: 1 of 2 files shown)


packages/ethereum/cardstack/searcher.js
```diff
@@ -38,6 +38,9 @@ class TransactionSearcher {
       query.filter.or.every(i => get(i, 'type.exact') === 'ethereum-transactions' ||
                                  get(i, 'type') === 'ethereum-transactions')) {
       return await this._queryTransactions(query);
+    } else if (get(query, 'filter.id.exact') && get(query, 'filter.type.exact') === 'ethereum-transactions') {
+      let transaction = await this._getTransaction(query.filter.id.exact);
+      return { data: [transaction.data] };
```
habdelra: it looks like if there is no transaction for the ID provided _getTransaction() returns undefined. can you make sure we convert this to an empty array? something like:
```js
let { data: transaction } = await this._getTransaction(query.filter.id.exact) || {};
return { data: transaction ? [ transaction ] : [] };
```

PR #801, opened 2019-06-14
CLI - Stabilize app generation dependencies

### LINKED ISSUE #799: CLI - Stabilize app generation
- have the blueprint output be output into a special folder, and a command to regenerate it (pre-build)
- create the app in the tmpdir, and on success, copy into the directory
- cardstack run should keep using the tmpdir, and if it doesn't already exist, copy the prebuilt app

jenweber:
Closes #799 

## Added
- a command `cardstack pre-build` that runs file generators within a tmpdir, then copies the result into `/blueprints/` on success.
- a blueprints directory containing the output of the generators

## Changed
- cardstack run has fewer responsibilities. It copies the contents of blueprints into a tempdir and yarn installs.
- exec is in its own file, since it's common functionality needed between run and pre-build
- I modified the gitignore so that blueprint js files are checked in. They were ignored because we had made this whole package use TypeScript, and so we ignored all `.js` previously.

## To test

- delete the blueprints directory. Do `cardstack pre-build` to see the generated app recreated.
- modify a file like the README
- `cardstack run` and check that the resulting change shows up in the tmpdir app


jenweber:
I'm not sure of the best way to get eslint to ignore the Ember app. I want _some_ linter to run on it.

packages/cli/blueprints/ember-cli-build.js
```diff
@@ -0,0 +1,63 @@
+/*eslint-disable node/no-missing-require */
```
jenweber: I couldn't get this to pass even when adding the embroider packages as deps in the cli. Let me know if you have any other ideas.

ef4: It's checking for the actual installed packages in node_modules, which won't be present yet since we don't run `yarn install` until after copying the blueprint elsewhere.

We could probably disable `no-missing-require` for the whole blueprints directory. That still gives us the `no-extraneous-require` rule, which is nearly as good and only compares with the contents of package.json.

PR #831, opened 2019-07-26
[DOC] write some API docs for the git plugin, closes #830

### LINKED ISSUE #830: Write API docs for git
Related to https://github.com/cardstack/cardstack-documentation/issues/38

jenweber:
I read through the tests and determined that this seems to be the public API for using the git plugin. This documentation is needed so that I can link to it from the Guides/Tutorial

### DIFF
--- a/packages/git/README.md
+++ b/packages/git/README.md
@@ -1,7 +1,109 @@
-This is a Cardstack data source plugin for reading and writing to Git.
+# @cardstack/git
 
-Building native nodegit dep on osx
---------------------------------
+This is a Cardstack data source plugin for reading and writing to Git.
+Always have access to see how data has changed across time.
+
+Features:
+
+- Just like using git for code, your data is versioned. A deleted or changed record can be retrieved!
+- Card data can be saved to either local or remote git repositories
+- Data can be stored in either the same repository as the project code,
+or a separate repository
+- Data for multiple projects can be saved to the same repository
+
+## Installing in a Cardstack project
+
+To use a git as a data source, include `@cardstack/git` in the `devDependencies`
+of `cardhost/package.json`
+
+```bash
+yarn add --dev @cardstack/git
+yarn install
+```
+
+See the "Troubleshooting" section of this README if you encounter any issues installing.
+
+## Configuration
+
+Data plugins are configured in `cardhost/data-sources`.
+The only required configuration is to set either the `repo` or `remote`.
+
+| Parameter | Type |  Description |
+|-----------|------|-------------|
+| `remote` | object |  Use a remote git repository, such as one hosted on Github. The object should contain a `url` and `privateKey` as strings. Example: `remote: { url: 'some-url', privateKey: process.env.GIT_PRIVATE_KEY}`|
+| `repo` | string | The path to a local git repository that already exists on disk |
+| `branchPrefix` | string | optional - a prefix to prepend to the git branch when saving data. This is especially useful when you are storing data in the same repository as your project, or you have one repository that holds code for multiple projects. Example: if you set `basePath: 'cs-'`, data will be stored at `cs-master`|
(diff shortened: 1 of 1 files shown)


packages/git/README.md
```diff
@@ -1,7 +1,109 @@
...
+  sources.push({
+    type: 'data-sources',
+    id: 'default',
+    attributes: {
+      'source-type': '@cardstack/git',
+      params: {
+        branchPrefix: 'cs-',
+        remote: {
+          url: 'git@github.com:cardstack/some-git-repo.git',
+          privateKey: process.env.GIT_PRIVATE_KEY,
```
skarabocu: I think the overall flow of this is really good. One thing that I can add is to explain more about how to set 'process.env.GIT_PRIVATE_KEY" and also 
'''
attributes: {
      'plugin-config': {
        'application-card': { type: 'app-cards', id: 'cardhost' }
      }
''' part, because someone can understand why 'data-sources' and 'plugin-configs' are there for, but 'application-card' type isn't that straightforward.

ef4:
Maybe we can use a more simplified example config. One that doesn’t distinguish between production and development for example. 

Also we can use comments to explain which parts of the config connect to each other (the default data source in the hub config points at the git datasource we are creating).

ef4:
At a more meta level, we really don’t want to need to teach people to hand write jsonapi configs like this, with or without JSONAPIFactory. It’s abstract and not very intuitive, because it’s optimized to be manipulated by tooling.

packages/git/README.md
```diff
@@ -26,12 +26,12 @@ See the "Troubleshooting" section of this README if you encounter any issues ins
...
 
 Data plugins are configured in `cardhost/data-sources`.
-The only required configuration is to set either the `repo` or `remote`.
+The only required configuration is to set either `repo` or `remote`.
 
 | Parameter | Type |  Description |
 |-----------|------|-------------|
 | `remote` | object |  Use a remote git repository, such as one hosted on Github. The object should contain a `url` string and the absolute path to your `privateKey` as a string. Example: `remote: { url: 'some-url', privateKey: process.env.GIT_PRIVATE_KEY}`|
-| `repo` | string | The path to a local git repository that already exists on disk |
+| `repo` | string | The path to a local git repository that already exists on disk, and has an initial commit |
```
ef4: I would have guessed the initial commit wasn't needed. Maybe that is a bug we can fix.

jenweber:
I think we should still provide these examples for our own benefit, but definitely work towards abstracting this away.

I will also simplify the project-template to not use split env configs, and just kind of leave it empty with a comment, like how `environment.js` in Ember does it.

packages/git/README.md
```diff
@@ -44,9 +44,8 @@ GIT_PRIVATE_KEY="/path/to/your/key/id_rsa"
 ### Example configuration
 
 For example, the configuration below is set in `cardhost/data-sources/default.js`.
-It would use a local git repository for local development,
-and point to a repository on GitHub once it is deployed to production.
-Remember, keep your private keys safe and do not commit them!
+The Hub has a `default-data-source` relationship that points to `@cardstack/git`.
```
ef4: Perhaps instead:

> This config creates a `@cardstack/git` data source named `default` and configures it as the Hub's `default-data-source`.

packages/git/README.md
jenweber: Agreed, I was surprised as well. I was getting "cannot find branch master" (or prefix-master) without it. I opened an issue here https://github.com/cardstack/cardstack/issues/832 and will add a note that we need to come back and adjust this part and our git guide when we fix it.

packages/git/README.md
jenweber: Changed to:


> For example, the configuration below is set in `cardhost/data-sources/default.js`.
> This config creates a `@cardstack/git` data source named `default` and configures it 
> as the Hub's `default-data-source`.
> Now, if a Card does not have its own data source defined, it will use git.

PR #834, opened 2019-08-07
creating card bootstrap models, creating sample card using desired API, stubbing out test for card loading

habdelra:


habdelra:
cool make sense.

Yeah I was wondering about an end-to-end test for the browser too, but I was trying to scope this PR to a bite size chunk of work (I have tendency to bite off more than I can chew)...

ef4:
Yeah, definitely subsequent PRs.

packages/hub/node-tests/cards-test.js
```diff
@@ -1,165 +1,63 @@
...
+    // This data source will monitor a a glob for card json:api documents.
+    // When it detects a new card document, the data source will indexing the card,
+    // generate the necessary schema, serialize templates, css, js to disk in a
+    // folder that ember-cli is watching, such that ember-cli will dynamically
+    // build and serve the new card assets
+    factory.addResource('data-sources', 'card-loader')
+      .withAttributes({
+        sourceType: '@cardstack/hub::card-loader',
+        params: {
+          // this is the glob that we watch in order to discover cards
```
ef4: On closer inspection, this is an area we will want to change. Card discovery needs to be done by indexers and/or searchers. It's not a special code path.

Which is good because it means you can load the JSON:API documents representing cards using all the existing tools, and implementing a "card NPM package searcher" is a separable piece of work that's not blocking to any of the other work.

In other words, loading code out of any card always goes through these steps:

1. Some indexer or searcher gives you a json:api document of type card. This is a complete description of the card, no other inputs needed. 
2. When you want to incorporate the card's code into the app (either server or client code), the hub will need to convert the json:api document to actual files on disk so that 
    1. yarn install can run and ensure all dependencies are present.
    2. the card's code is present in conventionally-located files on disk that can be `require`ed by the hub or `import()`ed into the ember app.

As an optimization, it would be good to be able to avoid writing new files in step 2 when the original data source already happens to be based off local files on disk. But that is strictly an optimization and things must work when it isn't true.

habdelra: so maybe like a `hub:card-loader` service that indexers and searchers can use when they have received a card that they want to load into the index?

habdelra: I guess also, actually creating test card node modules might be a little premature--as i wouldn't need to have a module in order to write a test--simply the card document would do. (but I could at perhaps build the card document in a manner that would look like how a developer would interact with the JSONAPIFactory for creating a card, and imbue the document with the compound ID's using test helpers)

habdelra: just thinking out load, i wonder if the new wrapper API for dealing with cards and the card loaded actually live in the same service, maybe like `hub:card-services`? which would include functions for loading cards and retrieving cards

ef4: Yeah I think that makes sense. Then even things like the retrieving a computed property implementation would go through the card service to load the code.

packages/hub/node-tests/cards-test.js
```diff
@@ -1,165 +1,133 @@
...
-        try {
-          await createDefaultEnvironment(`${__dirname}/../../../tests/stub-cards/schema-returns-non-card-definitions-document`);
-        } catch (e) {
-          error = e;
-        }
 
-        expect(error.message).to.match(/defines a schema document that is not of type 'card-definitions'/);
-      });
+  it('can load a card', async function () {
+    await cardServices.loadCard(articleCard);
```
habdelra: @ef4 I've updated the test to load a card from `hub:card-services`

packages/hub/node-tests/cards-test.js
```diff
@@ -1,165 +1,133 @@
...
-        }
+    // The method for getting a card would look like this:
+    //     cardServices.getCard(id:string, format:CardFormats) : Promise<JSONAPIDocument>,
+    //
+    // Do we want to split out the compound ID in the API for getting a card? Feels weird that
+    // callers of this API would need to massage the card ID that they have in order to use this API
+    // if we decide to have the multi-part id's components as separate params. I think outside of
+    // this wrapper API, the ID should be an atomic entity as it is today and callers of the wrapper ID
+    // should not have to manipulate the card ID they have in order to use the wrapper API.
+    let article = await cardServices.getCard('local-hub::article-card::millenial-puppies', 'isolated');
```
habdelra: @ef4 I've updated the test to also use a new wrapper API for getting a card

packages/hub/node-tests/cards-test.js
```diff
@@ -1,165 +1,133 @@
...
 const {
   createDefaultEnvironment,
   destroyDefaultEnvironment
 } = require('../../../tests/stub-searcher/node_modules/@cardstack/test-support/env');
 
+// addCardContext() is a test helper that imbues the compound ID's into the card document
+// based on the repository and module name of the card. This allows our tests to create card documents
+// using the same API that developers would use for creating card documents with the
+// JSONAPIFactory within the respective card modules that a card document originates from.
+let articleCard = addCardContext('local-hub', 'article-card',
```
habdelra: @ef4 here's an example of adding context to a card using a test helper so that the card authoring experience in our tests can mimic the authoring experience that developers would have when writing a card document.

ef4: There should not be two different forms of a card's json:api document. 

There will be an on-disk format for cards, but that is an internal concern of the indexer/searcher that knows how to read & write the on-disk format. The entire rest of the system only knows about the json:api format.

If it turns out nice to write tests using the on-disk format, then the tests should use that indexer/searcher.

ef4: Also, the schema in the on-disk format should be a json file, and the normal way to edit it will be to do it in the UI and let the writer update the json file. Focus on that use case. Up until now we have been doing everything by hand, even though we know that's not the experience we're going for. While it will always be possible to hand-edit these files, we want to focus primarily on the happy path where people are editing schema in the UI.

habdelra:
@ef4 I'm ready for another review on this. I think I was able to do what you wanted here in terms of loading a card at the time of need, which you can see in the tests where we are automatically loading cards that are returned by a searcher or added by an indexer.:

- https://github.com/cardstack/cardstack/pull/834/files#diff-f0fb1c576963e8458635f112f8779396R153
- https://github.com/cardstack/cardstack/pull/834/files#diff-f0fb1c576963e8458635f112f8779396R188
- https://github.com/cardstack/cardstack/pull/834/files#diff-f0fb1c576963e8458635f112f8779396R204

There's still a bunch of work left around validating the card (that it's not trying to reach into the internals of another card or using the schema of another card), read authorization for a card (which is probably just a matter of making sure we are utilizing all the current read auth stuff, and writing tests). I think though, if you are ok with this, it's probably worth merging so we can start tackling the UI end of this (which probably also means thinking through how the hub works with ember CLI to deliver to it the browser assets that it will need).

packages/hub/card-services.ts
```diff
@@ -0,0 +1,233 @@
...
+  currentSchema: 'hub:current-schema',
+},
+
+class CardServices {
+  writers: todo;
+  indexers: todo;
+  searchers: todo;
+  currentSchema: todo;
+
+  cardContextFromId(id: string) {
```
ef4: Can this and `cardContextToId` be private?

packages/hub/card-services.ts
```diff
@@ -0,0 +1,233 @@
...
+      return;
+    }
+
+    let schemaModels = await this.buildCardModelSchema(card);
+    if (!schemaModels) { return; }
+
+    for (let model of schemaModels) {
+      await this.writers.create(Session.INTERNAL_PRIVILEGED, model.type, { data: model });
+    }
+    await this.writers.create(Session.INTERNAL_PRIVILEGED, 'cards', card);
```
ef4: `writer.create()` here is a problem.

Both the card and its internal model should already be real models in some data source. (With namespaced ids and fields, presumably, so they fit into the existing system.) It doesn't make sense to create them here.

And even if these created models are needed as legitimate derived state, it would still be incorrect to write derived state into the real data sources like this.

packages/hub/card-services.ts
```diff
@@ -0,0 +1,233 @@
...
+      modelId,
+    };
+  }
+
+  cardContextToId({ repository, packageName, cardId, modelId }: CardContext) {
+    return [repository, packageName, cardId, modelId].filter(i => i != null).join(cardIdDelim);
+  }
+
+  // TODO should we care about the session of the caller who is requesting to load the card?
+  async loadCard(card: SingleResourceDoc) {
```
ef4: I want to try to get the public API of this service correct, even if the implementation needs hacks as we're in the middle of transitioning. And I don't think this method belongs in the public API.

`loadCard` should take the card's identity and it should find the card document itself via `hub:searchers`.

This implies changing where `loadCard` is being called from as well.

packages/hub/card-services.ts
habdelra: actually i pulled it put into plugin-helpers a subsequent PR as I needed it in a few other places (specifically around field ID `::` handling)

ef4: Yeah, I see that, but that is just another reason I don't think we should be doing that field id handling where we're doing it.

habdelra:
@ef4 so I was only calling `loadCard()` from the searcher and indexer (as well as tests) when they encounter a new style card. It sounds like you want me to pull this out of CardServices (which I can do, no prob). If I move `loadCard()` out of `CardServices` and into its own module that both the indexer and searcher use, would that be the correct way to move forward here?

packages/hub/card-services.ts
habdelra: ok I see, i think i went down this path because I wrote the tests around getting a card before I wrote the tests around the indexers and searchers being able to auto new cards they encounter. I'll re-work this.

habdelra: @ef4 I'm working thru how to load schema discovered by a searcher without writing it to the default data source. specifically how to update the schema that is available via `hub:current-schema`. This ultimately points to hub:indexers.schema(), which is where the schema cache lives. I was thinking that if the searcher encountered a new card, it could use the schema loader to load the new card's schema models, and then call a new function on hub:indexers to update the schema cache with the newly derived schema. does this sound like a reasonable approach?

this would mean that the card schema is not necessarily persisted in the index--but I think that is probably ok, since the card schema would automatically be loaded the next time the hub starts via the same means. (like whenever the card is encountered)

habdelra: I guess then the concern would be if an indexer emitted some new schema via conventional means, would the card schema we inserted into the schema cache be lost by a new schema provided by a `RunningIndexer`? I think you mentioned this way back when we were talking thru this in july, around the idea that the schema would just have to build up over time--i think I need to accommodate preserving discovered card schema with the RunningIndexer provided schema.

habdelra: Sorry for the rubber ducking--it looks like Schema has an `applyChanges()` that lets me add additional schema models. Maybe I can keep track of the discovered card schema during teh lifetime of the hub, and just use applyChanges to inject any discovered card schema when the running indexers return more schema.

habdelra:
@ef4 ok, I updated how and where we emit the schema for the cards so that the datasource that found the card is the datasource that emits the schema into the `Schema` class, as opposed to writing the card schema (and its models) into the default data source via the `loadCard()` function. I think this simplified things lots. You can see the meat of this here:
https://github.com/cardstack/cardstack/pull/834/commits/5a6a28d9940d243a6a4cd4af6477ea99b9fc3c51#diff-cb0d7e1308c4e29a5e4b441a09320c1aR81-R90

habdelra:
@ef4 Ok, I'm ready for another review--the logic that was previously spread all around the hub searchers and indexers is now self contained within the DocumentContext. the lines of code haven't really changed, but the logic is much more contained and more efficient (less trips to the DB since we can make good use of the DocumentContext's caching).

PR #839, opened 2019-08-15
(2 of 6) Field names as valid jsonapi

habdelra:
This PR depends on https://github.com/cardstack/cardstack/pull/834, please merge that PR first and rebase this afterwards. 

In the original PR for loading cards, I punted on the issue where field id's are used as the field names in the generated JSON:API, thus exposing something like this in the card model JSON:API, since the metadata field id's for the card are prepended with the card's repo and package name:
```
{
  data: {
    id: 'local-hub::article-card::millenial-puppies',
    type: 'local-hub::article-card',
    attributes: {
      'local-hub::article-card::title': 'Tips for the Discerning Puppy',
      'local-hub::article-card::author': 'Hassan Abdel-Rahman'
    },
    relationships: {
      'local-hub::article-card::tags': {
        data: []
      }
    }
  }
}
```
In addition to being awkward, having `:` in the attribute name and the relationship name is actually invalid JSON:API--these are considered illegal characters. As well as, this wreaks havok in the search doc for this document in the index.

This change actually uses the last part of the  field's composite ID for the attribute name and relationship name when serializing the document to JSON:API. So that the JSON:API for the same document actually looks like this (both the wire format, and how it is represented in the index--as it all uses the same DocumentContext):
```
{
  data: {
    id: 'local-hub::article-card::millenial-puppies',
    type: 'local-hub::article-card',
    attributes: {
      title: 'Tips for the Discerning Puppy',
      author: 'Hassan Abdel-Rahman'
    },
    relationships: {
      tags: {
        data: []
      }
    }
  }
}
```


habdelra:
what about the search doc? the search doc will have search fields that look like `local-hub::article-card::title` which is fine if that is how we decide our search queries should look as well. But will also be more problematic if we decide we want to search against interface fields that cards from different packages, but share a similar interface use.

habdelra:
But if you want to punt on the whole: how to search against card interface fields until later, and are happy with using the field ID's in the card search queries--then i'll happy close this PR.

ef4:
The search queries passed to *existing* methods would indeed look like that. But we want to wrap all of those places in a new API that can search for new-style cards and translate those queries.

> But will also be more problematic if we decide we want to search against interface fields that cards from different packages, but share a similar interface, use.

This is a thing we're explicitly deciding not to support by making the fields be namespaced. If we could target distinct fields from different packages with one query, that would violate the namespacing.

On the other hand, a good reason to derive from a public schema is so that searches against one of that schema's public metadata fields will find your card.

habdelra:
this is helpful, as it's something that I've been grappling with this morning (before I got your comments). so for the new card services search API, I was making it really similar to the current Searchers.search method signature, where you just provide a session and a query object. For the new card search API, would you like the additional card context to be part of the query object or additional params in the search method?

habdelra:
I'm gonna rework this in a higher layer (the jsonapi layer)

PR #841, opened 2019-08-19
Bugfix: Handle `everyone` group

jenweber:
As part of the review, I would like help figuring out what kind of test this should have. I was not able to call the searcher directly as a unit test, and the [failing test in Cardboard](https://github.com/cardstack/cardboard/blob/6eec2990ee2b661c79357a6627039877e153b18f/cards/article/node-tests/grant-test.js#L141) is not portable.

Added:
- `EveryoneGroup` searcher that returns a document for the everyone group, in order to skip group validations and an unnecessary database query
- an entry into `bootstrap-schema` so the new searcher will run

Changed:
- I had to add a validation exception to `group.js`, since the `everyone` group has no search queries and filters. The constructor function was handling too much after that, so I split it into functions, with no change in behavior for those sections.

Still needs:
- a test

### DIFF
--- a/packages/hub/schema/group.js
+++ b/packages/hub/schema/group.js
@@ -1,44 +1,29 @@
-const { get, isEqual, intersection } = require('lodash');
+const {
+  get,
+  isEqual,
+  intersection
+} = require('lodash');
 
 module.exports = class Group {
   constructor(model, allFields) {
-    let searchQuery = get(model, 'attributes.search-query');
-    if (!searchQuery) {
-      throw new Error(`group must have a search-query attribute: ${JSON.stringify(model, null, 2)}`);
+    if (model.type === 'groups' && model.id === 'everyone') {
+      // Skip validation for the `everyone` group. All users get assigned this group.
+      // See also bootstrap-schema.js and everyone-group.js for special handling.
+      this.id = model.id;
+      this.types = []; // required for downstream typechecks
+    } else {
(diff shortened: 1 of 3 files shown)


packages/hub/schema/group.js
```diff
@@ -51,5 +36,39 @@ module.exports = class Group {
       }
     });
   }
+};
+
+const validatedFieldFilters = function(allFields,  searchQuery, model) {
```
jenweber: I made these pure functions outside of the class, but let me know if they should be functions that are part of the class

packages/hub/schema/group.js
```diff
@@ -1,44 +1,29 @@
...
+      this.id = model.id;
+      this.types = []; // required for downstream typechecks
+    } else {
+      let searchQuery = get(model, 'attributes.search-query');
+      this.types = validatedTypes(searchQuery, model);
+      this._fieldFilters = validatedFieldFilters(allFields,  searchQuery, model);
+      this._allFields = allFields;
+      this.id = model.id;
     }
-    let types = get(searchQuery, 'filter.type.exact');
```
jenweber: This big section got moved, unchanged, into two smaller functions called `validatedTypes` and `validatedFieldFilters`

packages/hub/schema/group.js
habdelra: I personally also prefer pure functions outside of classes, and aggressively move functions out of classes when I can.

PR #843, opened 2019-08-19
Add publishing and release instructions for maintainers

jenweber:
I think this captures the steps that Hassan and I took to do the most recent publishing, except for the `--force-publish` option.

### DIFF
--- a/MAINTAINERS.md
+++ b/MAINTAINERS.md
@@ -8,10 +8,10 @@ The versions of all packages in the mono-repo are updated at the same time, to t
 
 1. Get the latest code on master and create a branch. `git checkout master`, `git pull origin master`, `git checkout -b vX.Y.Z`
 2. Look at the `git log` since the last release, and list anything significant in the `CHANGELOG.md`.
-Determine whether the release should be a major (breaking) or minor (non-breaking).
+Determine whether the release should be a major (breaking), minor, or patch.
 For breaking releases, include upgrade instructions in the changelog.
 3. Commit the changes, make sure your working tree is clean, and push your branch to GitHub.
-4. To update versions in the `package.json` for all packages, use `npx lerna version --force-publish`. `npx` will use the version of `lerna` that is installed as a `devDependency` of the mono-repo. `--force-publish` will update interdependencies as well as the version numbers of the packages themselves.
+4. To update versions in the `package.json` for all packages, use `npx lerna version --force-publish="*" --exact`. `npx` will use the version of `lerna` that is installed as a `devDependency` of the mono-repo. `--force-publish` will update interdependencies as well as the version numbers of the packages themselves.
 5. Open a pull request with your updates
 6. Merge the pull request after version-checking tests finish
 7. Create a tag on GitHub for the version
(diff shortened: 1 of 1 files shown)


MAINTAINERS.md
```diff
@@ -0,0 +1,26 @@
...
+
+This document is for project maintainers who are doing versioning, releases, and publishing for the cardstack mono-repo.
+
+## Updating package versions
+
+The versions of all packages in the mono-repo are updated at the same time, to the same version number.
+
+1. Get the latest code on master and create a branch. `git checkout master`, `git pull origin master`, `git checkout -b vX.Y.Z`
+2. Look at the `git log` since the last release, and list anything significant in the `CHANGELOG.md`.
+Determine whether the release should be a major (breaking) or minor (non-breaking).
```
habdelra: there are actually 3 choices:
1. patch
2. minor
3. major

MAINTAINERS.md
```diff
@@ -0,0 +1,26 @@
...
+For breaking releases, include upgrade instructions in the changelog.
+3. Commit the changes, make sure your working tree is clean, and push your branch to GitHub.
+4. To update versions in the `package.json` for all packages, use `npx lerna version --force-publish`. `npx` will use the version of `lerna` that is installed as a `devDependency` of the mono-repo. `--force-publish` will update interdependencies as well as the version numbers of the packages themselves.
```
habdelra: I think you actually want `--force-publish="*"`, at least that is what I've been using. We've also been using `--exact` instead of allowing a range (i.e. `^`) to be used.

habdelra: should we actually just use `lerna publish` instead of `lerna version`? that's what we've done in the past. It was a bit more yolo, but it was just a single command to do the versioning and publishing. I'm not necessarily opposed to opening PR's for releasing the hub. Maybe @ef4 has thoughts here.

jenweber: Changed to `npx lerna version --force-publish="*" --exact`

jenweber: I think lerna publish would be ok too. It just has fewer guardrails. Eventually I'd like to write something that checks that you have the latest commits on master before publishing. Then it would be safer to do in one command.

MAINTAINERS.md
```diff
@@ -0,0 +1,26 @@
...
+## Updating package versions
+
+The versions of all packages in the mono-repo are updated at the same time, to the same version number.
+
+1. Get the latest code on master and create a branch. `git checkout master`, `git pull origin master`, `git checkout -b vX.Y.Z`
+2. Look at the `git log` since the last release, and list anything significant in the `CHANGELOG.md`.
+Determine whether the release should be a major (breaking), minor, or patch.
+For breaking releases, include upgrade instructions in the changelog.
+3. Commit the changes, make sure your working tree is clean, and push your branch to GitHub.
+4. To update versions in the `package.json` for all packages, use `npx lerna version --force-publish="*" --exact`. `npx` will use the version of `lerna` that is installed as a `devDependency` of the mono-repo. `--force-publish` will update interdependencies as well as the version numbers of the packages themselves.
```
robustdj: Is lerna smart enough to know which version number to set? Or does that correlate with the branch name?

jenweber: It checks the current version, then prompts you to choose major/minor/patch, and displays the options that increment. I'll add a line about this. Thanks for the feedback!

MAINTAINERS.md
jenweber: I added a section with this command:

```sh
npx lerna publish from-git --force-publish="*" --exact
```

@habdelra does that seem correct?

PR #842, opened 2019-08-19
(4 of 6) Card relationships to other cards and card validation

habdelra:
This PR depends on https://github.com/cardstack/cardstack/pull/840, please merge that PR first and rebase this afterwards. 

This PR adds the ability for a card to have a relationship to another card as well as validation of card documents as they are loaded.

PR #844, opened 2019-08-20
(5 of 6) Support for multiple cards from same package (with differing models), and support for cards with no explicit model data specified

habdelra:
This PR depends on https://github.com/cardstack/cardstack/pull/842, please merge that PR first and rebase this afterwards.

This PR introduces support for for multiple cards from same package (with differing models), and support for cards with no explicit model data specified. This basically comes down to checking to see if the card schema models already exists before trying to add the schema, as well as asserting all cards have a model relationship defined, and manufacturing empty model data if need be.


habdelra:
Archiving this. we'll resurrect it using the latest card logic later.

PR #853, opened 2019-08-30
writes cards on disk as a side effect of loading card schema

habdelra:
Adding logic to write discovered cards to the hub's filesystem so we will be well positioned to perform embroider builds from the discovered card's browser assets (also removing the edit format stuff--we can re-add this when we actually tackle custom card edit mode components/templates). Also fixed a bug around how we are caching cards in the DocumentContext.

### DIFF
--- a/packages/hub/indexing/card-utils.ts
+++ b/packages/hub/indexing/card-utils.ts
@@ -1,6 +1,19 @@
+import Error from '@cardstack/plugin-utils/error';
 import { todo } from '@cardstack/plugin-utils/todo-any';
 import { SingleResourceDoc, ResourceObject, ResourceIdentifierObject, CollectionResourceDoc } from "jsonapi-typescript";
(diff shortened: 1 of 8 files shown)


packages/hub/indexing/card-utils.ts
```diff
@@ -1,6 +1,19 @@
...
+import logger from '@cardstack/logger';
+import { join } from "path";
+import {
+  ensureDirSync,
+  pathExistsSync,
+  writeFileSync,
+  removeSync,
+} from "fs-extra";
+
+const cardsDir: string = join(process.cwd(), 'card_modules');
```
ef4: If we can use system temp instead, that would avoid allowing cards to "accidentally" resolve dependencies from the app. Also, it means less things to put into `.gitignore` et al.

habdelra: sounds good!

PR #857, opened 2019-09-04
Adding support for creating new cards from a browser client

habdelra:


packages/hub/indexing/document-context.js
```diff
@@ -17,7 +17,7 @@ const qs = require('qs');
 
 module.exports = class DocumentContext {
 
-  constructor({ read, search, routers, schema, type, id, sourceId, generation, upstreamDoc, format, includePaths }) {
+  constructor({ read, search, routers, schema, type, id, sourceId, generation, upstreamDoc, included, format, includePaths }) {
```
ef4: What's the meaning of this `included`? Is it different from having an `included` inside `upstreamDoc`?

ef4: This argument list has grown unwieldy. Maybe some of these arguments are things that could be computed inside DocumentContext on demand instead of being passed into it? Maybe some of them can become arguments to specific methods (`documentContext.asSearchDoc({ format, included })`) instead?

We are being limited by the current implementation which is very "do all the work up front at once". All the methods are backed by the monolithic `_buildCachedResponse`. Things will get simpler if we can refactor so that things like `searchDoc` and `pristineDoc` and `realms` compute on demand, which shared caches so that subsequent calls to the other methods don't duplicate effort.

packages/hub/writers.js
```diff
@@ -265,6 +276,7 @@ class PendingChange {
     this.serverProvidedValues = new Map();
     this._finalizer = finalizer;
     this._aborter = aborter;
+    let { included } = opts;
```
ef4: I don't understand why included is getting special handling here. We pass whole jsonapi documents to writers and they pass whole documents back, so both cases necessarily already support included. Writers are free to not do anything with the included we give them, and they can choose to pass back anything they want to, and in either case the document context should still be able to do the right thing -- if present, we can use the included models directly without needing to load them, if missing and wanted, we will load them.

packages/hub/indexing/document-context.js
habdelra: So the challenge here is that the documents coming out of ephemeral and the git plugins (and I'm assuming the other writers) don't include the `included` part of the document. I wanted to avoid making deeper changes in all of our writers, and instead have the hub:writers.create just pass the additional includes directly into the DocumentContext so that we could derive the schema for the card.  Although now that I'm typing this, I wonder what would happen if we dropped the index and reindexed ephemeral or the git data source--since the included part was never written to the backing store, we don't actually know what the card schema is....

https://github.com/cardstack/cardstack/blob/master/packages/ephemeral/writer.js#L31-L37
https://github.com/cardstack/cardstack/blob/master/packages/git/writer.js#L78-L84

Maybe the correct way to do this is to write all the included resources in the json:api document to the backing store--not just the main `cards` resource

packages/hub/writers.js
habdelra: the issue is like what I mentioned above, the writers are only saving the `data` portion of the document. they are actually ignoring the `included` part of the document. I can update the existing writers to actually care about the `included` part of the document, if that's the direction you'd like to go.

habdelra: Or maybe the write approach is to invoke the writers for each resource in the included part of the card document (so effectively this means that we write out the internal models of the document as their own rows in the index, which I think we've gone back and forth with)...

habdelra:
Summarizing a conversation between Ed and I on how to move forward: 
> Ed: Just thinking... I'm realizing why this is hard, it feels bad that the card and its primary model are two resources instead of one.
But that follows from us trying to use the existing system without a big rewrite

> Ed: Maybe we can avoid the wrinkle. In the new world, we represent the card and card's model as two jsonapi resources with a relationship. That's still nice because it helps keep their schemas distinct. When we translate that to the old world, we can make a single resource, which has the content type of the internal model plus the fields from "card" itself, suitable prefixed so they can't collide with the internal model's fields.
so then there would be only one write to hub:writers
the special middleware that handles type===card would translate to that card's own particular content type for the rest of the hub to understand

habdelra:
@ef4 I'm ready for another review on this PR. I created an "internal" card format that our card services layer transforms cards into (and transforms internal cards back into external ones). This resulted in a bunch of churn, but I think the net LOC changed is the same as before. And as a result of doing this, we don't have to do any special handling of the `included` in the writers.

PR #859, opened 2019-09-09
(1 of 4) Update & delete card

habdelra:
This PR adds update and delete capabilities to cards. Also I've added acceptance tests for the ember application that will eventually become the card builder.


habdelra:
@ef4 I'm ready for a review of this PR.

packages/hub/indexing/card-utils.ts
```diff
@@ -132,6 +153,19 @@ async function generateCardModule(card: SingleResourceDoc) {
...
+      seenVersions = readJSONSync(join(cardFolder, seenVersionsFile));
+    } catch(e) {
+      if (e.code !== 'ENOENT') { throw e; } // ignore file not found errors
+    }
+    // the PendingChange class will actually create DocumentContext for the old
+    // and new versions of the document when it is being updated, and we dont want
+    // to inadvertantly clobber the latest card on disk with an older version of
+    // the card as a result of processing an older card as part of what PendingChange
+    // needs to do, so we keep track of the versions of the card that we've seen and
+    // only write to disk if the version of the card is not one we have encountered yet.
```
ef4: I think the version should be part of the identity, at least in this particular area. In other words, the path on disk you're writing out includes the version, so that versions never clobber each other. The only extra state needed would be for optimization (like a special dotfile inside each card's folder that indicates we wrote it out successfully, so it's easy to test for that and skip doing it again).

packages/hub/indexing/card-utils.ts
```diff
@@ -113,12 +126,20 @@ async function generateCardModule(card: SingleResourceDoc) {
     data: { ...card.data },
```
ef4: RE the TODO comment above: 

> TODO it's possible for cards originating from the same package to have different templates/components as a specific card instance could have its schema altered.

I think this isn't the right way to think about it. Every card, including every card instance, *is* a package. Two card instances that happen to have the same exact contents other than their individual card data can share one on-disk package for holding their code, *but this is only an optimization*.  In principle, every card instance is a complete package.

packages/hub/writers.js
```diff
@@ -132,7 +123,22 @@ class Writers {
     }
     await this.pgSearchClient.ensureDatabaseSetup();
 
-    let schema = await this.currentSchema.getSchema();
+    let addedModels = [];
```
ef4: I think we can layer this better so it's clear what's the difference between the new outer layer and the original layer. Also, I think the layering can look the same for each of `create()`, `update()`, and `delete()`.

What I mean is, each of those methods should have only two lines dedicated to special behavior for the card type:

```js
if (type == 'cards') {
  return this.handleCardOperations(...);
}
```

And then handleCardOperations is responsible for taking the card and figuring out a list of new calls to `create`, `update`, and/or `delete` to make, and running them.

packages/jsonapi/middleware.js
```diff
@@ -218,15 +218,24 @@ class Handler {
 
   async handleIndividualPATCH(type, id) {
     let data = this._mandatoryBodyData();
-    let record = await this.writers.update(this.session, type, id, data);
+    let record;
+    if (type === 'cards') {
+      record = await this.cardServices.update(this.session, id, data);
```
ef4: Why is this needed? I was under the impression you had already solved the same problem inside hub/writers instead. We should solve it at this layer or that layer but probably not both?

packages/hub/indexing/card-utils.ts
habdelra: ok, I'll refactor this to organize the on disk artifacts by card id instead.

packages/hub/indexing/card-utils.ts
habdelra: makes sense

packages/jsonapi/middleware.js
habdelra: i did, but I thought for consistency that within the middleware we always go through the CardServices layer to interact with cards. (since we do that for the get-invididual and get-collection methods), ditto for delete.

habdelra: Card service's mutation methods just wraps the writer functions for create, update, and delete and assumes a `cards` type

packages/hub/writers.js
habdelra: cool--will do.

packages/jsonapi/middleware.js
habdelra: If you'd prefer not to do this, then no biggie, I can remove these wrapped methods.

ef4: Oh yeah, I see the consistency concern.

I think if we're going to create the new interface at the level of hub/writers, then we also need to do it in hub/searchers. Those two are paired.

The alternative is to do it in jsonapi middleware and then *not* in hub/writers and hub/searchers. But I think that's probably less good, since there are plenty of other use cases (like tests) that want to go directly into hub/writers and hub/searchers.

PR #862, opened 2019-09-10
Bug fix - address cannot find module babel-eslint #861

### LINKED ISSUE #861: Bug - cannot find module babel-eslint in v0.14.22 and v0.14.23
> Cannot find module 'babel-eslint'
Referenced from: /Users/jenweber/cardstack/project-template/cardhost/.eslintrc.js

It seems [this refactoring](https://github.com/cardstack/cardstack/commit/109b7b988390ad6390cf5f0c61d3435635d324a2#diff-b0b9be13694993c7d72b14c28d8226e8) dropped inheriting from node, which it appears was important for existing apps. Here's what the config looked like [before](https://github.com/cardstack/cardstack/blob/v0.14.21/packages/eslint-config/browser.js).

Here's an example of an app that uses these configs: [project-template](https://github.com/cardstack/project-template/find/master)

jenweber:
This commit essentially reverts the `-browser.js` file to look more like `browser.js` at commit 4a09bfdea48b07700f6af4d0c752c74489af150d. It seems the reliance on the node config is important for existing applications.

Closes #861


jenweber:
@habdelra @ef4 you will want to take a look at this to see if it's in harmony with the CLI work you have been doing, since this is a step backward from changes you/we made.

ef4:
I think a better solution is for the apps to add babel-eslint.

It is unfortunately a peerDep of `@cardstack/eslint-config` because of the way eslint works, we can't make it our own dependency, so consumers of `@cardstack/eslint-config` need to change (by adding babel-eslint as their own dependency) to keep up.

ef4:
We should have listed babel-eslint explicitly as a peerDep in `@cardstack/eslint-config`, which might have made this problem easier to see, although it would not have solved it automatically because NPM is terrible at peerDeps.

jenweber:
Closed, addressed by #864 which has upgrade instructions that do not require a release of this package.

PR #840, opened 2019-08-16
(3 of 3) Support for computeds as card metadata

habdelra:
This PR depends on https://github.com/cardstack/cardstack/pull/839, please merge that PR first and rebase this afterwards.

PR #872, opened 2019-09-16
(2 of 4) Computed card metadata

habdelra:
This PR adds support for metadata fields being computed properties. (@ef4 actually approved an older version of this PR https://github.com/cardstack/cardstack/pull/840). This PR is unchanged from that previously approved PR--the only difference is that this PR is rebased on the latest understanding of a card.


packages/cardhost/tests/acceptance/update-card-test.js
```diff
@@ -0,0 +1,142 @@
...
+import { module, test } from 'qunit';
+import { click, visit, currentURL, waitFor } from '@ember/test-helpers';
+import { setupApplicationTest } from 'ember-qunit';
+import JSONAPIFactory from '@cardstack/test-support/jsonapi-factory';
+import Fixtures from '@cardstack/test-support/fixtures'
+import { setCodeMirrorValue, getCodeMirrorValue } from '../helpers/code-mirror';
+
+let factory = new JSONAPIFactory();
+let articleCard = factory.getDocumentFor(
+  factory.addResource('cards', 'local-hub::article-card::millenial-puppies')
```
ef4: General comment about tests:

I think we should be able to state the tests in new-card terminology, whereas I think these are using the internal representation (where field names include explicit namespaces, etc).

tests/stub-card-project/searcher.js
```diff
@@ -5,7 +5,7 @@ module.exports = class StubCardSearcher {
   }
 
   constructor(params) {
-    this.cards = params.cardSearchResults;
+    this.cards = params.cardSearchResults || [];
     this.cardIds = this.cards.map(i => i.data.id);
   }
```
ef4: Why is this searcher needed? Is it a placeholder under some other API exists for getting a card?

packages/cardhost/tests/acceptance/update-card-test.js
habdelra: makes sense. I took a small stab at cleaning up the namespaces, to just make it JSON:API compatible again. But I should get way more aggressive with that, and make it so that the only thing with a namespace is the ID of the card itself (and it's associated model's type and id).

tests/stub-card-project/searcher.js
habdelra: it's used to test that we can implicitly load a card returned by a custom searcher.

habdelra: (we have a similar test for custom indexers too)

packages/cardhost/tests/acceptance/update-card-test.js
habdelra: I was thinking a bit more on this: on the client I think we'll need to be able to disambiguate field names that multiple cards may share (article-card has a `name` field and user-card has a `name` field). If we want to use something like ember data (not saying we have to) that uses a store for the schema models, we'll need to disambiguate the name field for article and the name field for user if we decide to but those schema models in the store. So i think the client needs to understand the namespace for the card field schema. which means I think it is part of the external card format. is your concern more around how we reason about cards specifically from a testing standpoint? I'm trying to make sure I dont introduce a 3rd card format (that is different from our internal format that we use in our index and the external format that the hub:writers accept).

My hope is that as the card building becomes more fleshed out, that we literally just use the UI elements on the page to create the cards. while we wait for those UI elements to be built, I'm falling back to using the external card format via code mirror--but I have every intention of factoring that out.

habdelra: or maybe my thinking is too locked into how we were doing things before.... perhaps our data service is smart enough to keep the schema records associated with the cards (via our own `Map` or something). and it's not like "just dump everything into ember data".

habdelra: in general i'm not sure if we even can use ember data due to the dynamic nature of cards--where a `card` model can have arbitrary (from ember data's perspective) fields.

PR #878, opened 2019-09-19
[DOC] Add more details to the github-auth readme

jenweber:
This PR adds some of the context I had to infer when trying out this plugin. I think it is best for this material to live here on the README rather than be directly in the tutorial, since it is the bare essentials for using the plugin.


jenweber:
Closed, this package is deprecated

PR #883, opened 2019-09-25
New data service - support for creating a new card

habdelra:


habdelra:
Actually, I’m gonna also update the `setField()` so you can also pass in a `Card` instance as well for the card relationships.

habdelra:
cool. I recall that pattern from the Model API that the hub uses. I'll roll that into the inflight work that I have for this in the next PR.

PR #895, opened 2019-10-10
Drag & drop schema editing

bagby:
**Very** rough first pass at card creation UI improvements using drag & drop.

**EDIT** thx to @burieberry for the CSS 😄


packages/cardhost/app/components/card-manipulator.js
```diff
@@ -145,4 +179,46 @@ export default class CardManipulator extends Component {
   delete() {
     this.deleteCard.perform();
   }
+
+  @action
+  beginDragging(field, mousedownEvent) {
+    let dragState;
+
+    set(this, 'finishDrag', (dropEvent) => {
```
habdelra: is there a reason we we add a new function to the component as opposed to just setting this function's contents in the `dropField()` action on line 222?

packages/cardhost/tests/helpers/card-helpers.js
```diff
@@ -28,6 +33,7 @@ export async function createCards(args) {
 }
 
 export async function addField(name, type, isEmbedded, position) {
+  await dragAndDropField(0);
```
habdelra: we are still using the old way to add a field in this test. I was thinking that the drag and drop was going to replace the old way of adding fields. so that you didn't need to click on the `[data-test-field-creator-add-field-btn]` button to add a field.

habdelra:
I was really kinda hoping that you'd get to the point where dragging a field actually changes the underlying `Card` and `Field` instances, as that is likely where the traffic jam around doing the right edge and drag and drop of fields will be.

packages/cardhost/tests/helpers/card-helpers.js
bagby: @habdelra this is basically a stopgap between the current way and what you are mentioning. We could skip the step of using the `<FieldCreator>` if we fake it and give the field some default name, i.e. `field1` or whatever, but until we have a right edge, there won't be a way to edit the `name` or other attributes of the field.

habdelra: I'm not really sure what you are testing here then. like if I were to comment out that line, the test will still pass

habdelra: probably you'll have to make some kind of throw away field to capture the name and embedded check box too...

packages/cardhost/app/components/card-manipulator.js
bagby: this is a copypaste from some experimentation with `ember-animated`, which needed the `field` and the initial `mousedownEvent` in the `drop` event to do math-y stuff. Since we aren't (yet) using animations, this could certainly be simplified as you suggested.

packages/cardhost/tests/helpers/card-helpers.js
habdelra: or continue to use the FieldCreator component for setting the name and embedded fields and ket right edge work deal with refactoring that

packages/cardhost/app/templates/components/card-schema-updator.hbs
```diff
@@ -4,41 +4,97 @@
...
+      <div
+        class="card-manipulator--card--field--drop-zone"
+        role="button"
+        ondrop={{action this.dropField}}
+        ondragover={{action this.dragOver}}
+        ondragenter={{action this.toggleOverDropZone true}}
+        ondragleave={{action this.toggleOverDropZone false}}
+        data-test-card-add-field-drop-zone
+      >
+        Drop fields here to edit/add
```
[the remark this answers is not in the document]
bagby: That is the plan, we'll get there. I'm just trying to do the simplest thing to unblock other efforts.

packages/cardhost/tests/helpers/card-helpers.js
habdelra: Another idea is to use the `<FieldRenderer>` just put it on the page someone for the right edge person to cleanup. For the `<FieldRenderer you could pass to it the field that you created at dragend event, and use the `@mode=schema` in the `<FieldRenderer>`--this is probably the most straight forward approach using an already built mechanism for capturing field names.

habdelra: Ultimately, I'd like to not go backwards in terms of functionality. If you are not sure how to move forwards here then we can pair on it, or if you prefer I can add a commit to this branch to do that.

packages/cardhost/app/components/card-manipulator.js
```diff
@@ -150,4 +222,44 @@ export default class CardManipulator extends Component {
...
+  beginDragging(fieldComponent, mousedownEvent) {
+    let dragState = {
+      usingKeyboard: false,
+      initialPointerX: mousedownEvent.x,
+      initialPointerY: mousedownEvent.y,
+      latestPointerX: mousedownEvent.x,
+      latestPointerY: mousedownEvent.y
+    };
+
+    window.addEventListener('dragend', () => set(this, 'isDragging', false));
```
habdelra: where do we remove these event listeners? I think this will mean that you just accumulate new event listeners every time dragging begins. and then when dragging ends every accumulated event listener will be executed (so this function could be executed dozens of times for each dragend). If the reason we are doing this is for closure, then let's use the new octane trio of action helpers/modifiers (`(on)`,`(action)`, and `(fn)`). I've commented in the template what this should look like.

I think the best practice here is that instead of `window.addEventListner` you should favor using an `{{on}}`, because when you do, ember will take are of all the removing of the event listeners for you which is something that you really need to add here if you are gonna wire the event listeners by hand.

packages/cardhost/app/templates/components/card-schema-updator.hbs
```diff
@@ -4,42 +4,103 @@
...
+    <div class="card-manipulator--left-edge--catalog">
+      <div class="title">Card Catalog</div>
+
+      {{#each this.fieldComponents as |field|}}
+        <div
+          class="card-manipulator--left-edge--catalog-field"
+          role="button"
+          draggable="true"
+          onmousedown={{action this.beginDragging field}}
+          data-test-card-add-field-draggable={{field.type}}
```
habdelra: As I was mentioning in the `card-manipulator.js`, let's use the new helpers and modifiers for octane to add a draggend event handler that is closed over the `CardManipulatorComponent` so `this` resolves to the `CardManipulator` in the action handler. 

```hbs
{{on "dragend" (action this.endDragging)}}
```
(where `CardManipulatorComponent.isDragging()` is a function that takes no params and does the thing you want to do when dragging ends)

This will invoke `CardManipulatorComponent.endDragging()` with the `this` of `endDragging()` bound to the `CardManipulatorComponent`. This prevents the need to add event listeners to the `window`, as well as worry about removing event listeners--as you should always pair an `addEventListener()` with a `removeEventListener()` otherwise you'll have event listener functions firing multiple times for each time you have invoked `addEventListener()`.

In this case I think you just want to set `this.isDragging` to `false` when dragging stops. For which a `mut` will work perfectly fine. So likely this is even more succinct and requires no javascript at all:

```hbs
{{on "dragend" (action (mut this.isDragging) false)}}
```

Here's the reference for `{{on}}`: https://api.emberjs.com/ember/3.11/classes/Ember.Templates.helpers/methods/on?anchor=on

Here's the octane reference for `{{action}}` https://octane-guides-preview.emberjs.com/release/templates/actions/.  `{{action}}` in octane has been _greatly_ simplified so that it's only job is to do the template equivalent of this `() => { ... }`. It doesn't do anything else.

Here's the reference for `{{fn}}` which you can use to pass arbitrary things into the function like perhaps the `{{field}}` that is being emitted from the `{{#each}}`: https://api.emberjs.com/ember/3.11/classes/Ember.Templates.helpers/methods/on?anchor=fn The job of this guy is to pass additional things into your function (currying). So if you want to pass more than just the event, you can use this to add the additional items to pass into the function. Additionally, `{{fn}}` will close over the function so that `this` inside your function will refer to the `CardManipulatorComponent`. You can see an example of this for a bunch of the other `CardManipulatorComponent` actions:  https://github.com/cardstack/cardstack/blob/master/packages/cardhost/app/templates/components/field-renderer.hbs#L76 https://github.com/cardstack/cardstack/blob/master/packages/cardhost/app/templates/components/field-renderer.hbs#L91

packages/cardhost/app/components/card-manipulator.js
```diff
@@ -150,4 +222,44 @@ export default class CardManipulator extends Component {
...
+      set(this, 'isOverDropZone', false);
+    }
+
+    let fieldComponent = this.isDragging;
+    let field = this.card.addField({
+      type: this.fieldTypeMappings[fieldComponent.type],
+      position: this.card.isolatedFields.length,
+      name: this.newFieldName,
+      neededWhenEmbedded: false
+    });
```
habdelra: 👍

packages/cardhost/app/templates/components/card-creator.hbs
```diff
@@ -20,34 +20,100 @@
...
+      <div class="card-manipulator--left-edge--catalog">
+        <div class="title">Card Catalog</div>
+
+        {{#each this.fieldComponents as |field|}}
+          <div
+            class="card-manipulator--left-edge--catalog-field"
+            role="button"
+            draggable="true"
+            onmousedown={{action this.beginDragging field}}
+            data-test-card-add-field-draggable={{field.type}}
```
habdelra: ditto from the comment in `card-schema-updator.hbs` about using `{{on}}` instead of hand wiring `window.addEventListener()` (and having to deal with the bookkeeping of `window.removeEventListener()` when the component is destroyed, or in your case when the drag is done).

packages/cardhost/tests/helpers/card-helpers.js
```diff
@@ -27,22 +32,19 @@ export async function createCards(args) {
...
   if (isEmbedded) {
-    await click('#new_field_embedded');
-  }
-
-  if (position != null) {
-    await fillIn('#new_field_pos', position);
+    await click(`[data-test-field="${name}"] .field-renderer--needed-when-embedded-chbx`);
   }
 
-  await click('[data-test-field-creator-add-field-btn]');
```
habdelra: 👍

packages/cardhost/app/components/card-manipulator.js
bagby: this is a copy-paste of pre-Octane code, will update it.

habdelra:
looks great--let's merge when green

PR #896, opened 2019-10-11
Add hyperledger so it can work independently of githereum

alexspeller:


bagby:
@alexspeller can this be closed?

PR #908, opened 2019-10-21
adding prember to cardhost

habdelra:


ef4:
I am pleasantly surprised that this works with embroider. I had heard bug reports that it doesn't.

habdelra:
well the build doesn't blow up. but I'm just noticing that no `_empty.html` was created...

PR #909, opened 2019-10-21
Multiple drop zones for positioning of fields

bagby:
Added some minimal but functional styling.

### DIFF
--- a/packages/cardhost/app/templates/components/cards/default/isolated.hbs
+++ b/packages/cardhost/app/templates/components/cards/default/isolated.hbs
@@ -2,7 +2,7 @@
   data-test-isolated-card={{@card.id}}
   class="isolated-card default-card"
 >
-  {{#each @card.isolatedFields key="nonce" as | field |}}
+  {{#each @card.isolatedFields key="nonce" as |field index|}}
     {{#if @setFieldValue}}
       <FieldRenderer
         @field={{field}}
@@ -10,6 +10,9 @@
         @setFieldValue={{fn @setFieldValue field.name}}
       />
     {{else if (or @setNeededWhenEmbedded @setPosition @removeField @setFieldName)}}
+      {{#if @dropField}}
+        <DropZone @dropField={{fn @dropField index}} @position={{index}} />
+      {{/if}}
       <FieldRenderer
         @field={{field}}
         @mode={{@mode}}
@@ -24,4 +27,10 @@
     {{/if}}
     <br>
   {{/each}}
+  {{#if @dropField}}
+    <DropZone
+      @dropField={{fn @dropField @card.isolatedFields.length}}
+      @position={{@card.isolatedFields.length}}
+    />
+  {{/if}}
 </div>
--- a/packages/cardhost/app/templates/components/cards/default/isolated.hbs
+++ b/packages/cardhost/app/templates/components/cards/default/isolated.hbs
@@ -2,7 +2,7 @@
   data-test-isolated-card={{@card.id}}
   class="isolated-card default-card"
 >
-  {{#each @card.isolatedFields key="nonce" as |field index|}}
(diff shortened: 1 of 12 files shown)


packages/cardhost/app/templates/components/cards/default/isolated.hbs
```diff
@@ -2,14 +2,17 @@
...
     {{#if @setFieldValue}}
       <FieldRenderer
         @field={{field}}
         @mode={{@mode}}
         @setFieldValue={{fn @setFieldValue field.name}}
       />
     {{else if (or @setNeededWhenEmbedded @setPosition @removeField @setFieldName)}}
+      {{#if @dropField}}
+        <DropZone @dropField={{fn @dropField index}} @position={{index}} />
+      {{/if}}
```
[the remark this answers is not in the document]
habdelra: or wait--nevermind. I see now...

packages/cardhost/app/templates/components/cards/default/isolated.hbs
```diff
@@ -2,14 +2,17 @@
   data-test-isolated-card={{@card.id}}
   class="isolated-card default-card"
 >
-  {{#each @card.isolatedFields key="nonce" as | field |}}
+  {{#each @card.isolatedFields key="nonce" as |field index|}}
```
habdelra: there is no need for this--each `field` has a `position` property now.

packages/cardhost/app/templates/components/cards/default/isolated.hbs
```diff
@@ -2,14 +2,17 @@
...
     {{#if @setFieldValue}}
       <FieldRenderer
         @field={{field}}
         @mode={{@mode}}
         @setFieldValue={{fn @setFieldValue field.name}}
       />
     {{else if (or @setNeededWhenEmbedded @setPosition @removeField @setFieldName)}}
+      {{#if @dropField}}
+        <DropZone @dropField={{fn @dropField index}} @position={{index}} />
```
habdelra: you can use the `field`'s position here like this:
```hbs
<DropZone @dropField={{fn @dropField index}} @position={{field.position}} />
```

packages/cardhost/app/templates/components/cards/default/isolated.hbs
```diff
@@ -24,4 +27,10 @@
     {{/if}}
     <br>
   {{/each}}
+  {{#if @dropField}}
+    <DropZone
+      @dropField={{fn @dropField @card.isolatedFields.length}}
+      @position={{@card.isolatedFields.length}}
```
habdelra: is there a reason why we need to pass the position in twice for the `<DropZone>` Component? why cant we leverage the `this.args.position` in the `dropField()` action? is there  a reason that those positions would ever be different?

packages/cardhost/tests/acceptance/create-card-test.js
```diff
@@ -102,19 +102,18 @@ module('Acceptance | card create', function(hooks) {
...
     assert.equal(currentURL(), '/cards/new');
 
     await setCardId(card1Id);
     await addField('title', 'string', true);
-    await addField('body', 'string', false);
+    await addField('body', 'string', false, 1);
     await addField('author', 'string', false, 1);
 
-    assert.deepEqual([...document.querySelectorAll('[data-test-field]')].map(i => i.getAttribute('data-test-field')),
+    assert.deepEqual([...document.querySelectorAll('[data-test-isolated-card="local-hub::article-card::millenial-puppies"] [data-test-field]')].map(i => i.getAttribute('data-test-field')),
```
habdelra: I made a const for the magic constant:
```js
const card1Id = 'local-hub::article-card::millenial-puppies';
```
you can just use that instead of having to type this all out.

bagby: in my defense I didn't actually type it in :)

PR #911, opened 2019-10-21
Pin node version when using volta

ef4:
This helps ensure you're using the right node version, if you have [volta](https://volta.sh). It's pretty good.


ef4:
Ah, it turns out this doesn't actually work reliably, because they don't support monorepos correctly.

PR #920, opened 2019-10-24
limit test suite scope to things used by cardhost

ef4:


ef4:
This looks like it will save about 7 minutes of actual time per CI run, and about 27 minutes of total Travis job time. 

This is just the lowest hanging fruit, of not testing thing that we aren't actively using. The next step is to do some refactorings that will cut more dependencies out.

PR #925, opened 2019-10-28
Fix drag & drop for cardhost in Firefox, closes #924

### LINKED ISSUE #924: Bug in cardhost: drag and drop does not work in Firefox
I am only able to use drag & drop in Chrome.

We may need to specify onDragStart: https://developer.mozilla.org/en-US/docs/Web/API/GlobalEventHandlers/ondragstart

Example of minimums needed to get basic drag & drop behavior in Firefox: https://codepen.io/jweber/pen/PooKOMb

jenweber:
#924

### DIFF
--- a/packages/cardhost/app/components/card-manipulator.js
+++ b/packages/cardhost/app/components/card-manipulator.js
@@ -267,4 +267,10 @@ export default class CardManipulator extends Component {
 
     this.selectedField = field;
   }
+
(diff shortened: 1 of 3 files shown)


burieberry:
@jenweber We also need to add the same thing to `card-schema-updator` component, since it isn't refactored here yet. The one on `card-creator` LGTM 👍

PR #926, opened 2019-10-29
Right edge styles

robustdj:
TODO:

- [x] Rebase master
- [x] Remove `Export as Metadata` checkbox
- [x] Fix remaining broken tests
- ~Wire up `Required` checkbox~ (will be addressed as a separate ticket)
- [x] Wire up `Display Label`


packages/cardhost/app/components/card-creator.js
```diff
@@ -1,9 +1,18 @@
...
     this.card = this.data.createCard(`local-hub::${id}`);
   }
+
+  get sanitizedType() {
+    return this.selectedField.type.replace(/::/g, '/').replace(/@/g, '');
+  }
+
+  get fieldEditor() {
+    return `fields/${dasherize(this.sanitizedType)}-editor`;
+  }
```
robustdj: I don't think this is being used anywhere. Remove this

habdelra: this is used in teh field renderer--we use a component helper to get the editor for a field.

habdelra: this looks like a dupe of the field renderer code. agreed on removing.

packages/cardhost/app/components/cardhost-right-edge.js
```diff
@@ -0,0 +1,15 @@
+import Component from '@glimmer/component';
+
+const typeToInputTypeMap = {
+  '@cardstack/core-types::string': 'Text Field',
+  '@cardstack/core-types::case-insensitive': 'Text Field',
+  '@cardstack/core-types::boolean': 'Checkbox',
+  '@cardstack/core-types::date': 'Date',
+  '@cardstack/core-types::integer': 'Text Field',
+}
```
habdelra: this seems duplicative with what we have in the field renderer--please lets not redefine this. Rather let's figure out a way to only define this once.

packages/cardhost/app/components/cardhost-right-edge.js
```diff
@@ -0,0 +1,15 @@
...
+
+export default class CardhostRightEdge extends Component {
```
habdelra: There is a good chance that these components will move into the core package--the "cardhost" package is meant to be a lab for developing things. Because of that, let's not prefix this component with the name of "Cardhost"

packages/cardhost/app/templates/components/card-creator.hbs
```diff
@@ -46,21 +46,15 @@
...
-        @setNeededWhenEmbedded={{action this.setNeededWhenEmbedded}}
-        @setFieldName={{action this.setFieldName}}
-      />
-    {{/if}}
-  </section>
-</aside>
+<RightEdge
+  @selectedField={{this.selectedField}}
+  @selectedFieldName={{this.selectedFieldName}}
+  @selectedFieldLabel={{this.selectedFieldLabel}}
```
habdelra: I'm confused by this: `selectedFieldName` and `selectedFieldLabel` being passed in. These are already properties of the `@selectedField`, which already has tracked properties of `@selectedField.name` and `@selectedField.label`. Why are we duplicating this instead of pulling this from the @selectedField arg?

packages/cardhost/app/components/card-manipulator.js
```diff
@@ -102,6 +109,8 @@ export default class CardManipulator extends Component {
   @tracked card;
   @tracked selectedField;
   @tracked isDragging;
+  @tracked selectedFieldName;
+  @tracked selectedFieldLabel;
```
habdelra: We shouldn't have to track these two prperties, as they are already being tracked in the @selectedField's `Field` class

habdelra: like the field's name in the right edge was updating in tests in master when you click on the field in the card, as well as the type and needed when embedded. that all happens because we are already tracking these properties in the `Field` class.

habdelra: I think the problem is actually in the FieldRenderer--we are not treating the field label's input the same as the field name's input when doing one-way binding

habdelra: Most likely we should add a {{did-update}} render modifier for dealing with this...

habdelra:
@robustdj  If you are having difficulty establishing one-way bindings for the field label input in the `FieldRenderer` so that you don't need to pass around the internals of the fields in the template as args, let me know and I can add a commit for this.

packages/core/addon/components/field-renderer.js
```diff
@@ -23,6 +39,7 @@ export default class FieldRenderer extends Component {
     if (this.args.field.nonce !== this.currentNonce) {
       this.currentNonce = this.args.field.nonce;
       this.newFieldName = this.args.field.name;
+      this.newFieldLabel = this.args.field.label;
```
habdelra: @robustdj this is the key line right here that makes the label's `<Input>` one way bound. this is how the upstream state gets set back into the input component.

packages/core/addon/components/field-renderer.js
```diff
@@ -45,6 +62,12 @@ export default class FieldRenderer extends Component {
     this.args.setFieldName(this.args.field.name, this.newFieldName);
   }
 
+  @action
+  updateFieldLabel(newLabel) {
+    this.newFieldLabel = newLabel;
+    this.args.setFieldLabel(this.args.field.name, this.newFieldLabel);
```
habdelra: @robustdj this is where we fire an action to signal that the label input has changed instead of just directly changing the `Field.label` (since it has a getter only--the way to set a `Field.label` is actually to call `Field.setLabel()` which is the action that the `CardManipulator` component is sending down to us.

packages/core/addon/components/field-renderer.js
```diff
@@ -3,16 +3,32 @@ import { dasherize } from '@ember/string';
...
   @tracked newFieldName;
+  @tracked newFieldLabel;
   @tracked currentNonce;
+  @tracked renderNonce;
 
   constructor(...args) {
     super(...args);
 
     this.newFieldName = this.args.field.name;
+    this.newFieldLabel = this.args.field.label;
```
habdelra: @robustdj here we are initializing the property that will hold the local label's input field state. we bind the input field to this property instead of directly to the `Field.label` because changing `Field.label` requires actually calling `Field.setLabel()`.

packages/core/addon/templates/components/field-renderer.hbs
```diff
@@ -1,3 +1,6 @@
+{{! eslint is confused about indentation for else-if }}
+{{! template-lint-disable block-indentation}}
+
 {{this.onFieldUpdated}}
```
habdelra: @robustdj this is how we "pull" on the `onFieldUpdated()` getter so that it will act like a `didRecieveAttrs`. There is actually a better pattern to use here which is the `{{did-update}}` render modifier. At some point I'm going to refactor to use that.

PR #933, opened 2019-10-29
Add field labels to data service

robustdj:


habdelra:
@robustdj After the adoption PR is merged (this morning), I'd like to rebase this branch and add a new commit to rename the `metadata-field-types` to be `metadata-summaries` like we were talking about. Are you ok with that?

PR #936, opened 2019-10-30
Move navigation to mode switcher component

burieberry:
Note: Will use the `<Dropdown>` component when it's available


packages/cardhost/app/components/mode-switcher.js
```diff
@@ -0,0 +1,42 @@
...
+
+  @service router;
+  @tracked currentCardName;
+  @tracked currentMode;
+
+  constructor(...args) {
+    super(...args);
+
+    this.currentCardName = this.args.name;
+    this.currentMode = this.args.mode;
```
habdelra: There is no need to create this property--just use the arg you passed in. you only need to create component properties if it's a value that you need to mutate.

packages/cardhost/app/components/mode-switcher.js
```diff
@@ -0,0 +1,42 @@
...
+    id: 'schema',
+    displayTitle: 'Schema View'
+  }
+];
+
+export default class ModeSwitcher extends Component {
+  modes = modes;
+
+  @service router;
+  @tracked currentCardName;
```
habdelra: you don't need this

packages/cardhost/app/components/mode-switcher.js
```diff
@@ -0,0 +1,42 @@
...
+    this.currentCardName = this.args.name;
+    this.currentMode = this.args.mode;
+  }
+
+  @action
+  selectMode(mode) {
+    if (!mode || !this.currentCardName) { return; }
+
+    this.currentMode = mode;
+    this.router.transitionTo(`cards.${this.currentMode}`, this.currentCardName);
```
habdelra: use `this.args.name` instead of `this.currentCardName`

packages/cardhost/app/components/mode-switcher.js
```diff
@@ -0,0 +1,42 @@
...
+  constructor(...args) {
+    super(...args);
+
+    this.currentCardName = this.args.name;
+    this.currentMode = this.args.mode;
+  }
+
+  @action
+  selectMode(mode) {
+    if (!mode || !this.currentCardName) { return; }
```
habdelra: use `this.args.name` instead of `this.currentCardName`

packages/cardhost/app/components/mode-switcher.js
```diff
@@ -0,0 +1,42 @@
...
+  modes = modes;
+
+  @service router;
+  @tracked currentCardName;
+  @tracked currentMode;
+
+  constructor(...args) {
+    super(...args);
+
+    this.currentCardName = this.args.name;
```
habdelra: There is no need to create this property--just use the arg you passed in. you only need to create component properties if it's a value that you need to mutate.

burieberry:
@habdelra merging this if it looks good now

packages/cardhost/app/components/mode-switcher.js
```diff
@@ -0,0 +1,35 @@
...
+];
+
+export default class ModeSwitcher extends Component {
+  modes = modes;
+
+  @service router;
+
+  constructor(...args) {
+    super(...args);
+  }
```
habdelra: no need for this. if you don't provide a constructor, it will automatically do what you have here. you can just delete the constructor.

PR #939, opened 2019-10-31
Add label to field renderer

robustdj:


habdelra:
@robustdj can you make sure to add tests for this too? like when you click on different fields this will update? that kind of test is actually already being performed here https://github.com/cardstack/cardstack/blob/master/packages/cardhost/tests/acceptance/create-card-test.js#L68-L91

robustdj:
@habdelra I'm planning to write an acceptance test with the new right edge in my `right-edge-css` branch

habdelra:
@robustdj this is the test to reference. The url I provided before was not locked to a commit. https://github.com/cardstack/cardstack/blob/a0856962b16a60dedf5e01a2c9cfe25b15f41445/packages/cardhost/tests/acceptance/create-card-test.js#L68

It is more desirable, tho, that changes you make are accompanied by their tests. It makes it easier for other to reason about how you are changing the system as well as to make sure that no one else breaks the mechanism that you are introducing. Between when you merge this commit, and when you merge the test that actually tests this change, undesirable things can happen. And with so many people working on the same thing at the same time, this just increases the chances of people stomping on each other.

habdelra:
Honestly i think you should just combine this PR and the right-edge PR so that you have all your tests and your code in the same branch, as opposed to having code and tests living in separate PR's.

PR #945, opened 2019-11-04
Fix issue where githereum with a remote git source would not push latest

alexspeller:
The local cached repo was not updated after pushing to the remote, and therefore
githereum, which operates on the local repo, was not pushing the correct latest
commit.

Ensure the repo is up to date in the finalizer, and add a test, also fix two
tests this change caused to fail

### DIFF
--- a/packages/git/change.js
+++ b/packages/git/change.js
@@ -112,6 +112,11 @@ class Change {
         continue;
       }
 
+      if (this.fetchOpts && !this.repo.isBare()) {
+        await this.repo.fetchAll(this.fetchOpts);
+        await this.repo.mergeBranches(this.targetBranch, `origin/${this.targetBranch}`, null, Merge.PREFERENCE.FASTFORWARD_ONLY);
+      }
+
       return mergeCommit.id().tostrS();
     }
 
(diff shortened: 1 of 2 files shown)


packages/git/change.js
```diff
@@ -112,6 +112,9 @@ class Change {
         continue;
       }
 
+      await this.repo.fetchAll(this.fetchOpts);
+      await this._applyCommit(mergeCommit);
```
ef4: I don't think the `_applyCommit` is always safe. It's making the assumption that the local HEAD should be forced to match the new merge commit we created. But really we want the local HEAD to be forced to match whatever the remote HEAD is after `fetchAll`. Which will *usually* be the same thing, but doesn't necessarily need to be, given possible concurrency.

Instead of setting `this.targetBranch` to the value of `mergeCommit`, we want to set it to the value of the `origin/${this.targetBranch}` remote branch. Similar to [this](https://github.com/cardstack/cardstack/blob/ae7477d4a8196b83063d5cc54efe80832d93668b/packages/git/service.js#L125).

An invariant I'm trying to preserve is that (when using a remote repo) the only thing that alters the local branch is merging from remote.

PR #962, opened 2019-11-08
Add assertations about internal fields for adopted cards

alexspeller:


habdelra:
@alexspeller card should be able to inherit their adopted parent's internal fields. While that does mean that the field schema resources are not part of the child car'd document, you should be able to set an adopted internal field attribute in the child card's model resource. Can you reflect that in the tests for the adopted isolated card test?

PR #963, opened 2019-11-08
Adjust z-index of save button so it can be clicked

robustdj:


burieberry:
I'm wondering if it's better to give the top-edge a higher z-index than giving all `a` tags a z-index. What do you think of that? @robustdj

robustdj:
@burieberry So if you look at the markup, those buttons are actually not in the top edge. They're in a separate div container with the class `.card-renderer--top-edge-buttons` which is where I applied the z-index

burieberry:
I see what you mean. That makes sense then @robustdj

PR #996, opened 2019-11-14
integrate monaco-editor component

jenweber:
This has a linting failure that I'm not sure how to fix, which does not show up when I use this same code in another app. (Edit - resolved, refactored to not use the decorator in that way)
Screen Shot 2019-11-13 at 8 05 14 PM


jenweber:
This should be a squash & merge.

PR #1007, opened 2019-11-15
Bugfixes and add monaco editor

jenweber:
## Added
- dependency on monaco-editor, the webpack plugin, and Ember polyfill
- webpack config passed to embroider
- `<CodeEditor>` component and documentation in code comments

## Changed
- update version of embroider (needed css bugfixes)
- Use CodeEditor instead of CodeBlock

## Removed
- Dependency on prism
- CodeBlock component

There is one quirk of monaco where `textContent` of the code editor container only has the visible text content, truncating the JSON. So I store the code block in a `data-test-code-block` attribute and look it up there in tests.

More tests will be added in a separate PR. I wanted to get this in ASAP since it unblocks using `ui-components` in cardhost.


jenweber:
I'll fix the ui-components page and ping you when.  I am done. I agree that we should add tests once we start using those actions, and those will be in the next PR.

The code editor is very configurable and we can do `lineNumbers: false` if we want, along with changes to font size. I am thinking that the next PR could have a generic `options` object that can be used for any config besides the language, code, and readOnly.

jenweber:
@bagby this is ready for another review.

I made the following changes:
- the code container has a minimum height of 50px
- added some bottom margin to code blocks
- actual height is dictated by the number of lines present. This will need some refactoring when we put this into the right edge because we will want to set a max height, and I would like this calculation to go into one place
- I changed the whitespace in the ui-components code text. It looks too weird and badly spaced if we keep the indents. They render in the editor.

Screenshot_2019-11-18 cardhost(1)

jenweber:
Screenshot_2019-11-19 cardhost(1)

Issue #1004, opened 2019-11-15
When creating new or adopting a card, user should stay in Schema view after saving it.


dnaidionov:
Currently, user is switched to Layout view, but we should leave them in Schema

PR #1024, opened 2019-11-18
Don’t redirect to view page when saving

bagby:
Finishes #1004 

Almost entirely text fixes, the change itself was one line :)

### DIFF
--- a/packages/cardhost/app/components/card-manipulator.js
+++ b/packages/cardhost/app/components/card-manipulator.js
@@ -155,7 +155,6 @@ export default class CardManipulator extends Component {
       this.statusMsg = `card ${this.card.name} was NOT successfully created: ${e.message}`;
       return;
     }
-    this.router.transitionTo('cards.view', this.card.name);
   })) saveCard;
 
   @(task(function * () {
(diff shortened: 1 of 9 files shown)


packages/cardhost/app/components/card-manipulator.js
```diff
@@ -155,7 +155,6 @@ export default class CardManipulator extends Component {
       this.statusMsg = `card ${this.card.name} was NOT successfully created: ${e.message}`;
       return;
     }
-    this.router.transitionTo('cards.view', this.card.name);
```
habdelra: @bagby we should still transition here--if only to the newly saved card in the schema view. otherwise we are left on the /cards/new route after the card has an ID which is problematic.

habdelra: so something like:
```js
this.router.transitionTo('cards.schema', this.card.name);
```

bagby: Good point. Along those lines, the "Save" button on the New Card page should really say "Create Card" or something.

habdelra: There is a `card.isNew` property you can use to tell the difference between a card that is new and unsaved if you want to add that message to the button.

PR #1037, opened 2019-11-20
Remove jquery and quash a deprecation warning

alexspeller:


habdelra:
I just made a commit with this change directly into `new-hub` here: 95370dbf437cde6f139049d44c7fe00ada97a2d9

PR #1056, opened 2019-11-20
Basic Editor layout

robustdj:
When visiting `/cards/new/` or the schema view of an existing card with the `?editingCss=true` query param, display the editing layout which is just the card view on the left and the code editor on the right (hide left, right edges, and view dropdown)


packages/cardhost/app/routes/cards.js
```diff
@@ -0,0 +1,11 @@
+import Route from '@ember/routing/route';
+import { inject as service } from '@ember/service';
+
+export default class CardsRoute extends Route {
+  @service cssModeToggle
+  queryParams = { editingCss: {} }
+
+  model({ editingCss }) {
+    this.cssModeToggle.setEditingCss(editingCss);
```
jenweber: This feels like something that will bite us, but good for now until we figure out the real flow we want this to have

packages/cardhost/app/services/css-mode-toggle.js
```diff
@@ -0,0 +1,7 @@
+import Service from '@ember/service';
+
+export default class CssModeToggleService extends Service {
+  setEditingCss(value) {
```
jenweber: `editingCss` should probably be `@tracked`, though we have duplicated state between the query params and this service.

PR #1057, opened 2019-11-21
Let template css be adoptable; card view css fixes

burieberry:


burieberry:
@habdelra Do my changes the core package make sense for passing down a template id for the styles? Let me know your feedback. I'll add some tests

packages/cardhost/app/templates/components/cards/cardstack/base-card/embedded.hbs
```diff
@@ -3,7 +3,7 @@
 {{!-- TODO use a sanitized version of the @card.id in the class name after we have css coming from card documents --}}
 <div
   data-test-embedded-card={{@card.name}}
-  class="embedded-card cardstack_base-card"
+  class="embedded-card cardstack_base-card {{@cardId}} {{@mode}}"
```
habdelra: this isn't going to do the thing that you want. what you _actually_ want is to add the adopted card's ID into the template--otherwise this is going to be a totally different class for every card that exists and there will be no commonality.

Instead you should do this:
```hbs
class="embedded-card {{if @card.adoptedFrom @card.adoptedFrom.name "cardstack_case-card"}} {{@mode}}"
```

packages/cardhost/app/templates/components/cards/cardstack/base-card/isolated.hbs
```diff
@@ -3,9 +3,14 @@
 {{!-- TODO use a sanitized version of the @card.id in the class name after we have css coming from card documents --}}
 <div
   data-test-isolated-card={{@card.name}}
-  class="isolated-card cardstack_base-card"
+  class="isolated-card cardstack_base-card template_{{@templateId}} {{@cardId}} {{@mode}}"
```
habdelra: Same comment as above. Instead you should do this:
```hbs
class="isolated-card {{if @card.adoptedFrom @card.adoptedFrom.name "cardstack_case-card"}} {{@mode}}"
```

packages/cardhost/app/templates/components/cards/cardstack/base-card/embedded.hbs
habdelra: I added `@card.name` in there too because you'll also need to style the template to look like the cards that adopt from the template. so the template itself will need to use the `@card.name` css selector (use the card name, not the id--that way you can have consistent styling across realms when we start supporting that). and then all the cards that adopt from the card name need to be styled exactly the same, so you should also add a `adopts-@card.name` css selector to the same styles that you create for the template.

you should use the card adoption mechanism to create some event cards and you'll see what I mean.

packages/cardhost/app/templates/components/cards/cardstack/base-card/isolated.hbs
```diff
@@ -3,9 +3,14 @@
...
 >
   {{#each @card.isolatedFields key="nonce" as |field|}}
+    {{!-- TODO: This is here temporarily until we have a better solution for rendering images --}}
+    {{#if (eq field.name "image")}}
+      <img src={{field.value}} alt="" width="100%" class="isolated--hero-image">
```
habdelra: my experience has been that for responsive layout, `` is always the wrong answer. you'll need to think about `contain` vs `cover` when people start resizing the page. Definitely you'll want to use a background div here.

packages/core/addon/templates/components/card-renderer.hbs
```diff
@@ -53,13 +55,21 @@ Ember Concurrency into the rendering of the card component here.
         @route="cards.view"
         @model={{@card.name}}
       >
-        <Embedded @card={{@card}} @mode={{this.mode}}/>
+        <Embedded
+          @card={{@card}}
+          @mode={{this.mode}}
+          @cardId={{this.sanitizedCardId}}
```
habdelra: let's not pass this. use @card.name instead

packages/core/addon/templates/components/card-renderer.hbs
```diff
@@ -53,13 +55,21 @@ Ember Concurrency into the rendering of the card component here.
...
         {{#if @showName}}
           <div class="embedded-card-label">{{@card.name}}</div>
         {{/if}}
       </LinkTo>
     {{else}}
-      <Embedded @card={{@card}} @mode={{this.mode}}/>
+      <Embedded
+        @card={{@card}}
+        @mode={{this.mode}}
+        @cardId={{this.sanitizedCardId}}
```
habdelra: please don't pass this in

packages/core/addon/templates/components/card-renderer.hbs
```diff
@@ -27,6 +27,8 @@ Ember Concurrency into the rendering of the card component here.
           <Isolated
             @card={{@card}}
             @mode={{this.mode}}
+            @cardId={{this.sanitizedCardId}}
```
habdelra: please don't pass this in

packages/core/addon/templates/components/card-renderer.hbs
```diff
@@ -27,6 +27,8 @@ Ember Concurrency into the rendering of the card component here.
...
+            @templateId={{this.sanitizedTemplateId}}
```
habdelra: I don't think this is necessary. in general we should not pass in additional card detail into the card renderer. it can figure everything it needs from `@card` no need to add extra coupling to it.

packages/core/addon/components/card-renderer.js
```diff
@@ -36,11 +36,31 @@ export default class CardRenderer extends Component {
...
\ No newline at end of file
+
+  get cardId() {
+    return this.args.card.id;
+  }
+
+  get templateId() {
+    if (!this.args.card.adoptedFrom || this.args.card.adoptedFrom.name === '@cardstack/base-card') {
+      return this.cardId;
+    }
```
habdelra: this is wrong. there will be no commonality amongst cards that adopt from the same parent.

packages/cardhost/app/templates/components/cards/cardstack/base-card/isolated.hbs
habdelra: Also, let's hoist this class to the card renderer so that the user cant break this--as the card template will eventually be edited by the user.

packages/cardhost/app/templates/components/cards/cardstack/base-card/embedded.hbs
habdelra: Also, let's hoist this class to the card renderer so that the user cant break this--as the card template will eventually be edited by the user.

habdelra: let's also make a `(safe-string)` helper too so that you can make safe CSS strings from the card names. this way you don't need to pass the template name around everywhere (and also template is probably the wrong name for this).

habdelra: so maybe that all looks like:
```hbs
class="embedded-card {{if @card.adoptedFrom (safe-string (concat "adopts-" @card.adoptedFrom.name)) "cardstack_base-card"}} {{@card.name}} {{@mode}}"

```

habdelra: If you're not familiar with creating a helper, they are really quite simple. Here is an example of a helper in the core package that adds `1` to the value: https://github.com/cardstack/cardstack/blob/master/packages/core/addon/helpers/inc.js

so you'd use it like:
```hbs
<div>(inc @field.position)</div> 
```
and it would spit out what ever the field's position is plus 1.

packages/core/addon/templates/components/card-renderer.hbs
burieberry: So my thinking was the user might change the card name (when we add that capability), the styles would break, so it’s better to use the card id. The card id creates “template_adopted-card-id card-id” classes. If the card has to adopted parents, the template_adopted-card defaults to its own id. We use card-id class to style the card. If this turns out to be a template and another card adopts from it, the new card has classes “template_adopted-card-id card-id”. The card-id is there so that user can use it to make any custom changes.

I only implemented this for the isolated card and not for the embedded card. There was an issue with embedded card that I thought we can address if we decide we need the template styles in embedded card too, which normally shouldn’t be necessary since embedded card should just be a static snapshot of the isolated version.

packages/cardhost/app/templates/components/cards/cardstack/base-card/isolated.hbs
burieberry: There’s actually `object-fit` and `object-position` css properties which do what bg image properties can do. But I’m fine with using a bg image instead too, though.

packages/core/addon/templates/components/card-renderer.hbs
habdelra: the card name comes form the id :-) it's just the id without the `local-hub::` but ultimately all cards will come with their css baked in. what we are doing is 100% temporary.

habdelra: also there is no need to pass it in. we should strive to pass as little as possible.

PR #1059, opened 2019-11-21
Integration test for the code editor and testing hook

jenweber:


packages/cardhost/tests/integration/components/code-editor-test.js
```diff
@@ -1,18 +1,23 @@
...
+    this.set('editorIsReady', false)
+    let editorIsReady = false
+    this.set('editorReady', () => {
+      editorIsReady = true;
+    })
 
+    await render(hbs`<CodeEditor @editorReady={{action this.editorReady}} @code="card" />`);
+    await waitUntil(() => {
+      return editorIsReady;
+    }, {timeout: 10000})
```
robustdj: This is good for now. Eventually we should abstract this into a test helper so that we can more easily test other editor features later.

PR #1063, opened 2019-11-22
add more cardhost percy snapshots, closes #1030

### LINKED ISSUE #1030: Add percy tests to card builder
PR number one will cover the easy things.

PR number two will account for animations.

jenweber:
I just added a handful of percy tests for card builder. We can add more as we go.

@burieberry can you let me know if any particular views or states that are missing from what we should be evaluating with these diffs?


burieberry:
@jenweber there's also the `/adopted` route, which is slightly different from the `/new` route template. also the event-card template styles.

jenweber:
I added the "adopt" route for testing, but something seems unusual about event-card. I get 409 conflict on save for the test I wrote. In the interest of unlocking visual testing for the other interfaces, I'm `skip`ping the event-card test for now and will open an issue on it.

jenweber:
One known issue is that the catalog layout is a bit weird in the snapshot, I think possibly due to the resolution/dimensions of the snapshot. I will open a ticket for this as well. I'm leaving that test in, since it is better than nothing.

Issue #1033, opened 2019-11-19
Display Right Edge in Edit and Layout views


dnaidionov:
-  [x] In Edit and Layout views Right Edge should be displayed all the time. 
- [ ] ~~When card is selected (no fields are in focus), Right Edge should display general card properties, all sections read-only except for Appearance.~~ needs design
- [x] Appearance section should allow a user to select a style, or edit a custom one if they have permissions. 
- [ ] ~~When a field is selected, it's edit mode related properties are displayed in the Right Edge.~~ needs design

jenweber:
This issue has been partially resolved by PR https://github.com/cardstack/cardstack/pull/1090, because we show the right edge in layout view now, but there is still more styling and state work to do.

If anyone else wants to grab the next steps for this ticket, go for it. I am concentrating on the Themer at the moment.

burieberry:
PR #1176 was handling this ticket along with ticket #1077 until the other ticket was blocked. I'll make #1176 to just be about this ticket so we can close it

PR #1068, opened 2019-11-25
Open the interactive themer

jenweber:
Partially addresses #1033 

## Adds:
- The right edge into the layout view
- A button where clicking it opens up the CSS editor
- a generic class for vertical padding (is this the right move?)
- Tests!

## Changed
- CTA buttons have a width instead of min-width. Otherwise, the "Create a theme" text overruns the button width.
- In testing, resizeable will be false. The loop was blocking acceptance tests.
- The Right Edge can be used as a block component with `{{yields}}`

## Known issues
- The right edge has more options than it should, left over from other layouts. This should be addressed when the new designs are ready.
- The Dropdown menu from ui-components needs styling.
- When you hit the back button, the css editor does not close. This will be addressed in a later PR.
- This is not yet a close match with mockups. Styling comes later.
- The themer overlaps the layout. Derrick is working on these controls in another PR.


jenweber:
Screen Shot 2019-11-25 at 5 41 12 PM

Screen Shot 2019-11-25 at 5 41 03 PM

packages/cardhost/app/components/code-editor.js
```diff
@@ -27,8 +28,8 @@ export default class CodeEditor extends Component {
 
   constructor(...args) {
     super(...args);
-
-    this.resizable = (this.args.resizable === true) ? true : false;
+    let notTesting = ENV.environment !== 'test'; // disable resize loops in testing
+    this.resizable = (this.args.resizable === true && notTesting) ? true : false;
```
robustdj: I'm generally not a fan of putting test code into implementation code, but in this case I'm not sure how to disable the timer when in test. This is for an acceptance test right?

jenweber: Surprisingly, this works:

```
  hooks.beforeEach(function () {
    this.owner.lookup('service:data')._clearCache();
    // any time you visit the editor page, you need to set resizable to
    // false, or tests will time out.
    this.owner.lookup('controller:cards.view').resizable = false;
  });
```

PR #1069, opened 2019-11-25
WIP show card CSS and markup in themer

### LINKED ISSUE #1065: Feature - Ephemeral CSS editing
As part of #998, we should show how making changes to the CSS are reflected on the rendered Card. To close this issue, these changes do not need to be persisted.

jenweber:
closes #1065

This is a draft PR that we can merge together with Derrick's work when it makes sense to. It targets the WIP `1066-dock-editor` branch.

## Added
- code that applies the Card's CSS (if it has any) and renders any changes. 
- Can edit CSS and see an immediate preview
- code that gets the Card markup via query selectors and renders it in an "HTML Viewer" code editor that is read-only
- temporary "save" button, that should be replaced/removed before merging this. Save will persist the CSS changes

## Changed
- pass the card model to the EditorPanel
- pass a saveCss action to the editor panel

### DIFF
--- a/packages/cardhost/app/components/editor-pane.js
+++ b/packages/cardhost/app/components/editor-pane.js
@@ -4,6 +4,17 @@ import { action } from '@ember/object';
 
 export default class EditorPane extends Component {
   @tracked dockLocation = "bottom";
+  @tracked markup;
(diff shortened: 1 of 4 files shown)


jenweber:
closed in favor of #1090

PR #1087, opened 2019-12-02
Use ui components text field

bagby:


burieberry:
@bagby I'm done with updating this branch but there's a test issue we can look into tomorrow. Visually everything is working during testing but the test-helper is having problems with filling in email input fields or something.

bagby:
@burieberry yes let's pair on this tomorrow

PR #1088, opened 2019-12-02
Dock editor to right or bottom

### LINKED ISSUE #1066: Dock code editor to bottom or left side

robustdj:
Closes #1066


packages/cardhost/app/services/css-mode-toggle.js
```diff
@@ -1,10 +1,22 @@
...
+
+  @action
+  dockRight() {
+    this.dockLocation = "right";
+  }
+
+  @action
+  dockBottom() {
+    this.dockLocation = "bottom";
+  }
```
robustdj: This service is now doing more than just handling state of whether or not it's in css editing mode. We should rename this service to something more meaningful. Would love suggestions!

jenweber: I think maybe css-mode-options? But this can be in another PR.

PR #1090, opened 2019-12-03
Save css

### LINKED ISSUE #1065: Feature - Ephemeral CSS editing

jenweber:
closes #1065

## Added
- code that applies the Card's CSS (if it has any) to the layout view
- Can edit CSS and see an immediate preview
- code that gets the Card markup via query selectors and renders it in an "HTML Viewer" code editor that is read-only
- add functionality to the save button in the layout view
- test for closing the editor

## Changed
- pass the card model to the EditorPanell
- bugfix - closing the editor works
- bugfix - resolve type error on the editingCss query param

## How to test

Create a new card, add a field (like `title`), save it, open the Custom Theme editor from the Right Edge in the layout view. Try adding styles, i.e. `.field-title { border: 2px solid red; }`. Close Editor. Save. Refresh.

## Known issues
- no CSS scoping
- Save button has no visual feedback
- Cannot edit CSS of existing templates
- Adding a custom theme does not affect the dropdown (it's just for show, at the moment)
- does not include error handling

### DIFF
--- a/packages/cardhost/app/components/editor-pane.js
+++ b/packages/cardhost/app/components/editor-pane.js
@@ -17,26 +17,7 @@ export default class EditorPane extends Component {
   }
 
   @action
-  preview() {
-    var css = this.css
-    let newStyle = document.createElement('style')
-    newStyle.setAttribute('id', 'card-styles')
-
-    document.querySelector('#card-styles').replaceWith(newStyle)
-
-    newStyle.type = 'text/css';
-    if (newStyle.styleSheet){
-      // This is required for IE8 and below.
-      newStyle.styleSheet.cssText = css;
-    } else {
-      newStyle.appendChild(document.createTextNode(css));
-    }
-  }
-
-  @action
   updateCode(code) {
-    this.css = code
     this.args.model.setIsolatedCss(code);
-    this.preview()
   }
 }
--- a/packages/cardhost/app/components/editor-pane.js
+++ b/packages/cardhost/app/components/editor-pane.js
@@ -1,7 +1,42 @@
 import Component from '@glimmer/component';
 import { inject as service } from '@ember/service';
+import { tracked } from '@glimmer/tracking';
+import { action } from '@ember/object';
 
 export default class EditorPane extends Component {
-  @service cssModeToggle
+  @service cssModeToggle;
+  @tracked markup;
+  @tracked css = this.args.model.isolatedCss;
 
+  constructor() {
+    super(...arguments)
+    // getting markup must be done in the constuctor to guarantee that it
+    // resolves before the code editor is rendered.
+    let cardMarkup = document.querySelector('.card-renderer-isolated--content')
+    this.markup = cardMarkup ? cardMarkup.innerHTML.toString().trim() : "";
+  }
+
+  @action
+  preview() {
+    var css = this.css
+    let newStyle = document.createElement('style')
+    newStyle.setAttribute('id', 'card-styles')
+
+    document.querySelector('#card-styles').replaceWith(newStyle)
+
+    newStyle.type = 'text/css';
+    if (newStyle.styleSheet){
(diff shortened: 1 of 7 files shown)


packages/cardhost/app/components/editor-pane.js
```diff
@@ -1,7 +1,42 @@
...
+  @service cssModeToggle;
+  @tracked markup;
+  @tracked css = this.args.model.isolatedCss;
 
+  constructor() {
+    super(...arguments)
+    // getting markup must be done in the constuctor to guarantee that it
+    // resolves before the code editor is rendered.
+    let cardMarkup = document.querySelector('.card-renderer-isolated--content')
+    this.markup = cardMarkup ? cardMarkup.innerHTML.toString().trim() : "";
```
jenweber: This is where we grab the HTML. The indentation is a little strange when you get it from the DOM.

packages/cardhost/app/controllers/cards/view.js
```diff
@@ -29,4 +30,15 @@ export default class ViewCardController extends Controller {
     this.themerOptions.push({name: 'Custom theme'})
     this.router.transitionTo('cards.view', this.model, { queryParams:{ editingCss: true } })
   }
+
+  @action
+  save() {
+    this.model.save()
```
jenweber: There was a "Save" button on the layout page, but it didn't do anything, so now we add that functionality.

packages/cardhost/app/controllers/cards/view.js
```diff
@@ -29,4 +30,15 @@ export default class ViewCardController extends Controller {
...
+  }
+
+  @action
+  closeEditor() {
+    this.cssModeToggle.setEditingCss(false);
+    this.router.transitionTo('cards.view', this.model, {queryParams:{editingCss: false}})
```
jenweber: I don't know how to clear the query params altogether. Any ideas?

packages/cardhost/app/routes/cards.js
```diff
@@ -6,6 +6,7 @@ export default class CardsRoute extends Route {
   queryParams = { editingCss: {} }
 
   model({ editingCss }) {
-    this.cssModeToggle.setEditingCss(editingCss);
+    let editingBool = editingCss === "true" ? true : false;
```
jenweber: The query param is cast to a string! So we need to turn it back into a boolean. Otherwise `"false"` is truthy and our if helpers in the template do the wrong thing.

packages/cardhost/app/templates/cards/view.hbs
```diff
@@ -82,4 +91,6 @@
       </div>
     </div>
   </RightEdge>
-{{/if}}
\ No newline at end of file
+{{/if}}
+
+<style id="card-styles">{{this.model.isolatedCss}}</style>
```
jenweber: This is how we render the Card's CSS.

packages/cardhost/app/controllers/cards/view.js
robustdj: I believe this works

```
this.router.transitionTo('cards.view', this.model, {queryParams: { editingCss: undefined }})
```

packages/cardhost/app/components/editor-pane.js
robustdj: What's the issue that this resolves?

robustdj: Oh is it so that the code editor actually has the markup? Were you seeing cases where the code editor showed nothing because the `markup` property was set late?

jenweber: I didn't see any issues, but this work has to happen somewhere, and I think it should be the responsibility of the editor pane and not CodeEditor.

Issue #995, opened 2019-11-13
UI - Adopt section of card properties in the Right Edge


dnaidionov:
This ticket is for building the UI component/visuals only. There will be a separate ticket for "real" data driven functionality. This component will be part of the right edge, so it lives in the cardstack mono-repo.

Could see if some of this could be copied over from tally and/or ui-components.

The adoption section of the right edge is called "Template"

It displays currently adopted card, and provides an option to change it, or remove completely. 

![image](https://user-images.githubusercontent.com/1836666/69356374-63889080-0c51-11ea-8232-8ad0ca176462.png)

Adopted card also shows its type - the card that it itself, in turn, adopted (in the screenshot above the current card is adopted from "Party Invite", which is adopted from "Base Card"). This is done so that when changing adopted cards user could easily see cards of the same type - having common parent and thus providing more compatibility and more likely to not require field type conversion (but not necessary). 

When we remove the card, the adoption path resets to Base Card, and we display the Base Card as a parent. If a card is adopted from Base Card, "Remove" button becomes disabled. Base Card doesn't have a parent. 

## Regular card
![image](https://user-images.githubusercontent.com/1836666/69380414-5cc54200-0c80-11ea-9a54-d89dd575f503.png)

Current card is inherited (adopts) from Meetup, which is of the type (adopts) Event Registration

## Card descends directly from Base Card
![image](https://user-images.githubusercontent.com/1836666/69380645-e543e280-0c80-11ea-891b-0e696d33f928.png)

Parent card removed - now this card is adopted from the Base card. "Remove" CTA is disabled

## Base Card
![image](https://user-images.githubusercontent.com/1836666/69380776-35bb4000-0c81-11ea-81c0-91fbe0e5b7ca.png)

We might have hidden entire "Template" section, but we might display other information in there - e.g. list of cards adopted from this card (TBD)

XD file: https://drive.google.com/open?id=1xSnis7586pGMysyWycrjvGHM7GQzbLZm

habdelra:
Cool. this should be able to be picked up by anyone. the data service for the card adoption should be pretty simple for devs to use.

PR #1094, opened 2019-12-04
Adopted card right edge

bagby:
This addresses part of #995. It displays the adoption chain two levels deep.


packages/core/addon/services/data.js
```diff
@@ -196,6 +196,11 @@ class Card {
 
   get adoptedFromId() {
     let internal = priv.get(this);
+
+    if (internal.id === baseCard) {
+      return null;
```
jenweber: I think this needs a test, like `assert.equal(parent.adoptedFromId, null);` added to https://github.com/cardstack/cardstack/blob/master/packages/core/tests/unit/services/data-test.js#L694

bagby: 👍

PR #1112, opened 2019-12-06
install and configure ember-animated

### LINKED ISSUE #1049: Bug - Ember animated container content won't rerender
## Problem:
When animated-if is used inside an AnimatedContainer, the contents don't rerender when values change. This blocks a lot of animation work.

> Assertion Failed: Could not find component named "-dynamic-element" (no component or template with that name was found)

## Escape hatch

Could change static components to false "fixes" the bug

Cons are that we don't see when there are missing files, could reveal other bugs.

## To do
- [x] need to try and see if we can reproduce in a regular Octane app. Reproduced here https://github.com/jenweber/embroider-percy-debug/tree/animated-debug
- [x] Turn embroider off and see if the minimal repro works. Works normally.
- [ ] turn off staticComponents in the cardhost app and see how far that gets us (this helps)
- [x] try classic Ember w/Embroider and see if it works (it does)
- [ ] Where does `-dynamic-element` come from? 
- [ ] The `element` helper is null in AnimatedContainer. Where does that come from? It comes from ember-element-helper dependency https://github.com/tildeio/ember-element-helper

## Debugging tips
search for staticComponentspackages/core/src/app.ts
controls whether imports get pushed onto the stack (appJSAsset)look at the assets file in the build

When static components is on:
resolver.ts, look at astTransformer
Can't easily look at build output bc the template compiling is done by webpack, so use debuggerlook at resolveMustache. Adds imports to templates. ~~Seems to be including the component because it's not missing, just broken~~ AnimatedContainer is broken on initial render

Things to look for
- animated if being used by itself- renders correctly on the first try, but not on changes- seems like something is weird about component imports
- check for multiple copies of ember-animated. Should be ok, but could check- use will's branch. the thing that is not working is responsibility of animated-if, used inside an AnimatedContainer. Contents of if don't rerender when values change!- 

ruled out:
- infinite animation- look at redneredChildren in animated-each. Doesn't seem to be running
- positional params seemed ok when Ed looked at it

jenweber:
Includes package rules for its dependency on ember-element-helper.
Pins the version of ember-animated to a SHA that uses
ember-element-helper v0.2.0.
Closes #1049

### DIFF
--- a/packages/cardhost/ember-cli-build.js
+++ b/packages/cardhost/ember-cli-build.js
@@ -102,6 +102,14 @@ module.exports = function(defaults) {
             },
           },
         },
+        {
+          package: 'ember-element-helper',
+          addonModules: {
+            'helpers/-element.js': {
+              dependsOnComponents: ['{{-dynamic-element}}', '{{-dynamic-element-alt}}'],
+            },
+          },
+        },
       ],
     });
   })();
--- a/packages/cardhost/package.json
+++ b/packages/cardhost/package.json
@@ -50,7 +50,7 @@
     "@glimmer/component": "^0.14.0-alpha.13",
     "babel-eslint": "^10.0.2",
     "broccoli-asset-rev": "^3.0.0",
-    "ember-animated": "^0.10.0",
+    "ember-animated": "github:jenweber/ember-animated#1f812240be876d91361ad9434adbf0cc78178fff",
     "ember-auto-import": "^1.4.0",
     "ember-cli": "3.13.0",
     "ember-cli-app-version": "^3.2.0",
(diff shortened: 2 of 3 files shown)


jenweber:
We are pinning to a SHA until the next release of Ember Animated.

jenweber:
Tests are failing because of a typescript/babel issue. Ed is working on a fix.

ef4:
I'm trying to execute the new github actions tests against this pr, it doesn't seem to have picked up automatically so I'm going to try closing this one and making a separate pr.

ef4:
PR continues at https://github.com/cardstack/cardstack/pull/1112

PR #1115, opened 2019-12-09
Scope custom theme css using card id

### LINKED ISSUE #1110: Scope user-generated CSS in the Card Themer
Right now, someone can write themer CSS that affects the whole card builder app.

jenweber:
Closes #1110

Added
- helper that prefixes CSS rules
- test for the helper
- dependency on `css-scope`

Changed
- scope custom isolatedCss before rendering


packages/cardhost/app/helpers/scope-css.js
```diff
@@ -0,0 +1,14 @@
+import { helper } from '@ember/component/helper';
+import scope from 'scope-css';
+
+// Prefixes all CSS rules with the parent class name
+// Requires two params, the CSS and a class name or ID to add to the front of every rule.
+
+export default helper(function scopeCss(params) {
+  if (!params[0] || !params[1]) {
```
robustdj: In this case, should we throw an error? Wondering if we should be strict about making sure parent selectors are passed in so that styles are scoped.

packages/cardhost/app/helpers/scope-css.js
```diff
@@ -0,0 +1,14 @@
...
+    return '';
+  }
+
+  // local-hub::my-card becomes my-card
+  return scope(params[0], `.${params[1]}`);
```
robustdj: Should we also do any light validation of the class name that is passed in? I'm thinking about the cases where a developer might blindly pass in a class name that has special characters in it that are not valid in css. We could also just throw an error when an invalid class name is passed in

packages/cardhost/tests/integration/helpers/scope-css-test.js
```diff
@@ -0,0 +1,15 @@
...
+import { setupRenderingTest } from 'ember-qunit';
+import { render } from '@ember/test-helpers';
+import hbs from 'htmlbars-inline-precompile';
+
+module('Integration | Helper | scope-css', function(hooks) {
+  setupRenderingTest(hooks);
+
+  // Replace this with your real tests.
+  test('it renders', async function(assert) {
+    await render(hbs`{{scope-css ".some-class { font-size: 10px; }" "my-card"}}`);
```
robustdj: Just a suggestion to have the test example css have 2 styles to demonstrate that it prefixes both of them with the parent selector. I didn't realize that we had a plugin helping us until I looked a bit further into the implementation.

packages/cardhost/app/helpers/scope-css.js
jenweber: I was thinking that this shouldn't break our code if something doesn't have an ID yet, so I prefer the graceful failure of just not applying any CSS.

packages/cardhost/tests/integration/helpers/scope-css-test.js
jenweber: Great point. I'm adding this!

packages/cardhost/app/helpers/scope-css.js
jenweber: I don't think so, for the same reasons that we don't validate the CSS itself either. Browsers are pretty resilient towards "bad" CSS and we are already leaning on that functionality.

PR #1124, opened 2019-12-11
Entire card is drop zone when there are no fields

### LINKED ISSUE #1074: If the card has no fields, the entire card should be a drop zone for adding new fields.
As soon as the first field is added, card drop zones should return to regular behavior

Only do this if it is an easy/quick fix. Otherwise, throw it back in the queue.

bagby:
Closes #1074


burieberry:
@bagby just saw this. this is how it needs to look like, so we need to update the text and the icon as well:

drop-zone

bagby:
🤦‍♂ I'll fix it up

PR #1125, opened 2019-12-11
Fix card text label from getting bigger on hover

### LINKED ISSUE #1104: Text label shouldn't scale on hover on launcher screen
On launcher screen there's a card transition - card is shrinking while platter is growing. Currently card's lable is growing with platter (see screen recording). It should remain fixed.
https://www.loom.com/share/5e72368bfba74ec5896c24ba9d4517dc

robustdj:
Closes #1104

### DIFF
--- a/packages/cardhost/app/styles/index.css
+++ b/packages/cardhost/app/styles/index.css
@@ -161,9 +161,10 @@
   font-size: 13px;
   text-align: center;
   transform: none;
-  transition: padding var(--ch-card-transition-duration) ease-in-out calc(var(--ch-card-transition-duration) / 3);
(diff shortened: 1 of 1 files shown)


robustdj:
@burieberry The text seems to move down a half pixel or so. Do you think this is something where we need to change the markup in core to have the text stay absolutely still?

burieberry:
@robustdj I see what you mean. It would probably be good to do so, but maybe whoever works on the embedded card UI can handle that @dnaidionov. Anything that only belongs to the catalog page should be on the catalog page and not on the embedded card renderer. And their styling needs to be less entangled.

Issue #1109, opened 2019-12-06
Bug: Card builder should have lightweight validations for field and card id


jenweber:
To fix:
- [ ] Spaces in field names break cards permanently. Disallow spaces. This also impacts CSS styles.
- [ ] when you backspace to clear the field label, it autofills too eagerly with the id

Note that if you use ui-components, it has validation hooks built-in.

jenweber:
Card name is now separate from id, so you can use all kinds of characters. Lightweight FE validation is in #1136.

@habdelra do we have any server side validation for field ids in new-hub?

habdelra:
we have some real light validation, like the field cannot start with `cs`, and we also have the idea where the resulting field card needs to be valid JSONAPI (so i think that will catch special chars).

jenweber:
Ok, I think in that case, we should go ahead and close this, and come back to it if we need more powerful measures.

PR #1136, opened 2019-12-12
Field name validation

bagby:
Part of #1109 

This doesn't prevent users from saving a card if a field is invalid -- that work is not trivial and will be addressed in a separate PR.


jenweber:
Another thought, we should make sure this has a percy snapshot for the error state too, since the edges have some complex css.

PR #1140, opened 2019-12-14
Reenable interactive elements lint rule

### LINKED ISSUE #1139: Reenable interactive elements linting rule
In #1122 there was an update to ember-template lint that made adding a click to a div be an error instead of a warning.

We turned this off to get the PR merged, and so we need to fix this error and turn it back on.

jenweber:
Closes #1139 

We still use `div role="button"` on the draggable components. I didn't mess with those. I figure we can tackle that during "make drag & drop accessible" tickets. Switching them to buttons without doing that other work doesn't fix anything.

Added:
- some styles to get rid of button styling
- the lint rule that was removed in #1122 

Changed:
- clickable divs become buttons
- added aria-hidden to icons that are only for display
- some styles to account for these changes

### DIFF
--- a/packages/cardhost/app/styles/index.css
+++ b/packages/cardhost/app/styles/index.css
@@ -35,7 +35,8 @@
   box-shadow: none;
   border: none;
   cursor: pointer;
-  font-size: 1rem;
+  font: 400 13px/18px var(--ch-font-family);
+  letter-spacing: 0.015em;
 }
 
 .cardhost-welcome-left-edge--nav ul {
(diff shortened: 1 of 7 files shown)


packages/cardhost/app/styles/index.css
```diff
@@ -30,6 +30,14 @@
   background: url(/assets/images/cardstack-logo.svg) 30px 30px/24px 25px no-repeat;
 }
 
+.cardhost-welcome-left-edge--nav button {
```
burieberry: @jenweber Since `button`s bring their own font properties and the whole family to the party, I'd add the below. Everything else looks good.
```
    font: 400 13px/18px var(--ch-font-family);
    letter-spacing: 0.015em;
```

jenweber: Added, thank you!

Issue #1133, opened 2019-12-12
Bug: Fix monaco resize and HTML that is cut off in the Themer


jenweber:
To-do:

- [ ] see if it's still a problem - For some reason, when we insert HTML into the themer, you can't scroll far enough to see all of it. Possibly we're either doing something weird with the HTML string that truncates it, or we have messed up scrolling by setting height based on line numbers.
- [ ] fix ^^^^
- [ ] When you resize the editor, we need to update the monaco container size. There is a hook for onResizeEnd that we could use to determine when to redraw. In the future, we can use onResize/onResizeStart hook to grey out the code blocks during a resize, or even destroy and rerender

PR #1142, opened 2019-12-16
Themer focus 1083

### LINKED ISSUE #183: Add a mechanism to associate users into groups
As part of asserting grants, we should have a mechanism to associate users into groups that can be specified in grants. Probably need to handle the ability for users to be added to or removed from a group when user's attributes/relationships change.

jenweber:
Closes #0183 and #1133 

Added:
- background color animation between themer and editor pages
- bright blue border around CSS editor
- aria-label with Tab trap escape instructions
- enable responsive resize of the themer pane
- check for testing ENV before turning responsive resize checker on. Could refactor this in the future to only resize after the resize motion finishes, depending on the UX we want

Changed
- Moved themer-related styles into their own file, `themer.css`
- Added some height to ui-components code blocks so that fewer are cut off
- minor whitespace formatting

### DIFF
--- a/packages/cardhost/app/components/code-editor.js
+++ b/packages/cardhost/app/components/code-editor.js
@@ -32,7 +32,7 @@ export default class CodeEditor extends Component {
   // Sets default resize interval check in milliseconds.
   // This limits how often updatedDimensions is called.
   get resizeCheckIntervalMs() {
-    return this.args.resizeCheckIntervalMs || 2000;
+    return this.args.resizeCheckIntervalMs || 1000;
   }
 
   // readOnly defaults to true. Affects whether the editor lets you type in it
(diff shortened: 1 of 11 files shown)


jenweber:
## Before
[video](https://drive.google.com/drive/u/0/folders/1tnELDNaOC4HyPaDSnd8FxEcxEOJ1xFzm)
Screen Shot 2019-12-16 at 5 15 15 PM

## After
[video](https://drive.google.com/drive/u/0/folders/1tnELDNaOC4HyPaDSnd8FxEcxEOJ1xFzm)
Screen Shot 2019-12-16 at 5 15 24 PM

jenweber:
@burieberry @bagby I moved some of the CSS for themer out of app.css and into it's own file, themer.css. Just a heads up in case you see some merge conflicts on any work you are doing.

PR #1147, opened 2019-12-18
Refactor top edge buttons

### LINKED ISSUE #1145: Rework layout of top edge in the themer
TODO
- [x] change layout to match mockup
- [x] layout should be absolute/fixed. Should not move when you collapse the right pane.
- [x] keep in mind that we will add more elements in the future
- [x] also keep in mind that we will have a "more" button eventually, for handling mobile view

Should look like this:

Screen Shot 2019-12-17 at 4 45 35 PM

Currently looks like this:

Screen Shot 2019-12-16 at 5 15 24 PM

jenweber:
Changed:
- arrange themer buttons using CSS grid instead of flex
- create a container for top edge buttons
- use white for inactive state text/icons. The gray I was using was not in the palette and was insufficient for text contrast.

Added:
- text inside the full width and responsive buttons. Done with changes to button size and SVG viewbox. These buttons themselves use flexbox since I need nice text centering.

About the grid:
- There is one grid container that specifies columns of auto width
- There are two groups of buttons, those that run to the left and those that go to the right, using justify-self
- the size/position of the grid is meant to match the boundaries of the rest of the elements on the page

Closes #1145

### DIFF
--- a/packages/cardhost/app/styles/components/button.css
+++ b/packages/cardhost/app/styles/components/button.css
@@ -151,7 +151,8 @@
 .ch-button--icon.responsive {
   --icon-url: url('/assets/images/buttons/responsive-inactive.svg');
   --width: auto;
-  --height: 60px;
+  --height: auto;
+  padding-top: 25px;
   background-size: 25px 25px;
   background-position: top;
   justify-content: center;
@@ -171,7 +172,8 @@
 .ch-button--icon.full-width {
   --icon-url: url('/assets/images/buttons/full-width-inactive.svg');
   --width: auto;
-  --height: 40px;
+  --height: auto;
+  padding-top: 25px;
   background-size: 20px 20px;
   background-position: top;
   display: inline-flex;
--- a/packages/cardhost/app/styles/components/button.css
+++ b/packages/cardhost/app/styles/components/button.css
@@ -158,6 +158,7 @@
   justify-content: center;
   align-items: flex-end;
   box-shadow: none;
+  border-radius: 0;
 }
 
 .ch-button--icon.responsive.selected {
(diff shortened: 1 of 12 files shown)


jenweber:
Before
Screen Shot 2019-12-18 at 10 49 10 AM

After

Screen Shot 2019-12-18 at 10 57 00 AM

jenweber:
The percy diffs look somewhat odd in terms of button positioning. I am not sure why...

burieberry:
@jenweber the positioning of the buttons look different on mine:
Screen Shot 2019-12-18 at 11 45 00 AM

For those two buttons on the left, if we're using background images instead of inline SVG (if switching the highlighted icons because of the color change is causing any issues, we can use inline SVG), I'd say give a `padding-top: 25px;` to account for the space the images are taking. Those buttons can then have `--height: auto;`. If I'm looking at the right XD file, the space between them is `45px`. The font seems to now be `font-size: 11px` and `line-height: 15px;` in the XD, though might need to check with Gardella.

For the right side, I'm not sure why the Close button is hanging off. Making the inner container  (`themer--top-edge-pane-controls` class) `display: flex;` seems to alleviate the alignment issue. I should probably go over Cardhost button styling after you merge this to clean up the css there and make it more flexible.

Right now it looks like we're pretty much using the grid like a flexbox (which makes sense for this case), but grid can come handy down the line for more complex things.

jenweber:
Ah, it looks different/bad in Chrome, but ok in firefox. I'll keep working on it.

jenweber:
Updated to fix the Chrome/Firefox discrepancies.

The padding and height fixes worked perfectly. I had a bit of a battle to get the button aligned but `vertical-align: top;`  worked in the end.

PR #1156, opened 2019-12-18
Add shadow to scrolled view

jenweber:
Added:
- "scrolled" style that is optionally applied to the header
- code for detecting, debouncing, and cleaning up scroll events
- test to simulate scrolling
- CSS for the shadow

We'll probably want to use this elsewhere in any fixed containers with scrollbars.

The debounce timeout is really, really short, but anything more than that and there was a noticeable lag.

Before scrolling:

Screen Shot 2019-12-18 at 6 28 01 PM

After scrolling:
Screen Shot 2019-12-18 at 6 28 09 PM

### DIFF
--- a/packages/cardhost/app/styles/components/card-renderer.css
+++ b/packages/cardhost/app/styles/components/card-renderer.css
@@ -67,6 +67,12 @@
   background-color: var(--ch-highlight);
   color: var(--ch-dark);
 }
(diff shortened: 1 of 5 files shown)


jenweber:
Closed, will reopen a new PR adding the same functionality to other page elements

PR #1160, opened 2019-12-20
Fix button save state, closes #1154

### LINKED ISSUE #1154: Making changes to CSS doesn't trigger dirty state
As a result save button is never enabled on the layout page and css changes cannot be saved.

jenweber:
The problem was passing card and not model to the save button in the view route.

Changed:
- correct the undefined variable
- use better DDAU for setting CSS

Added
- test coverage for the bug
- test coverage for rendered styles


bagby:
LGTM 👍

PR #1181, opened 2019-12-27
Create image core type

### LINKED ISSUE #1064: Create an image core type
Similar to how we have fields for Text, Date, Number, we should create a core type for image. This can be an "old style" core type, similar to text and date, for this first pass. In the future, our core types will be cards but this is out of scope.

The image core type/card can be as lightweight as:

```handlebars
<div class="some-class {{additionalClasses}}">
  <img src={{someUrl}} alt={{altText}} />
  <div>{{caption}}</div>
</div>
```

The user will provide the image via their own URL (i.e. something on imgur)

In cases of decorative images, `alt=""` is what should be used, so we should make sure that property and value show up as a default, when no value is provided. [more info](https://www.w3.org/WAI/tutorials/images/) and [this](https://www.w3.org/WAI/tutorials/images/decision-tree/)

Helper test (if we are using it) can give the user a clue about how to fill in alt.

jenweber:
closes #1064 

Added
- create `decorative-image-viewer.hbs` which is an image tag
- create `decorative-image-editor.hbs` which is just the existing textfield editor
- create `decorative-image.js` which validates the entry is a string
- add core-types data to `mappings.js`
- add the new core type to the embroider config for cardhost in `ember-cli-build.js` so that it gets included in builds
- create `decorative-image-editor.js`
- add image icon to the left edge selector in schema mode
- style the image (just doing a hero image for now)

Changed
- rework the event card to use the image
- tests! Modified the existing event card tests

How to review:
Try running this locally. Also look to see if any more CSS needs to be removed/changed since this replaces the event-card functionality.


habdelra:
tests?

habdelra:
You should really add a test for this in this module https://github.com/cardstack/cardstack/blob/master/packages/cardhost/tests/acceptance/edit-card-test.js#L29

jenweber:
Thanks @habdelra, I've added a test there.

Screenshots of the work in this PR:

Screen Shot 2019-12-30 at 1 54 57 PM

Screen Shot 2019-12-30 at 1 55 06 PM

Screen Shot 2019-12-30 at 1 55 16 PM

Screen Shot 2019-12-30 at 1 53 22 PM

burieberry:
@jenweber this looks good. After you merge, I can update my branch for the css styling.

PR #1185, opened 2019-12-30
Fix moving catalog card label bug

### LINKED ISSUE #1146: On the launcher screen label is still moving on card hover
Tha label resizing was fixed in #1104, but it is still slightly moving - as if several conflicting animations are taking place at the same time - see attached video. 

https://www.loom.com/share/a01c11452669416aa438d54f2a16c3c4

jenweber:
Closes #1146 

This PR positions the label absolutely and then uses box-shadow to show the lighter gray tray. By using Box Shadow, it avoids shifting things around and simplifies the animation.

The downside is that we can't directly control the border radius, however when it is time to put tools in the tray, we'll need to refactor the CSS and markup anyway, and can make a dedicated tray container.

Screen Shot 2019-12-30 at 5 11 05 PM

### DIFF
--- a/packages/cardhost/app/styles/index.css
+++ b/packages/cardhost/app/styles/index.css
@@ -139,12 +139,13 @@
   transition: transform var(--ch-card-transition-duration) ease-in-out calc(var(--ch-card-transition-duration) / 3);
 }
 
-.card-catalog .embedded-card {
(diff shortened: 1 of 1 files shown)


jenweber:
@burieberry this is ready for another review

PR #1190, opened 2019-12-31
trigger redeploy to builder-prod

jenweber:
I merged in `master` and pushed directly to this branch, but it did not seem to trigger a deployment, so I am opening a PR to see if that has an effect. If it does, I'll edit the travis.yml config to deploy on direct pushes to this branch so we can easily keep it up to date.


jenweber:
This wasn't the problem - see #1191

PR #1194, opened 2020-01-02
Handle logged out state

### LINKED ISSUE #1099: User should not be able to edit or make layout and schema changes without logging in
When users are not logged in they should see only the "Login" button. Other buttons (e.g. "Save") should be hidden. Currently, Schema and Edit mode already hide them, but in Layout mode they are still displayed: 
image

The "Login" button itself will be moved to the left edge under the profile button context menu (#1014)

Eng notes:
- We should be cautious not to hack this
- needs design - what should it look like if you are not logged in

jenweber:
Changed:
- logged out users who visit `edit` or `schema` via direct URL are redirected to `view`
- clicking the logout button from `edit` or `schema` redirects to `view`
- Save controls and right edge are hidden in `view` if you are logged out

Added
- tests for the behavior above
- CSS to make the card wider when the edges are hidden
- CSS to move the login button to the right when logged out

Removed
- redundant auth checks

Closes #1099 

You can see a video of the expected behavior on Google Drive if you search for `login-logout-issue-1099.mov`

To review, please try to break it, and also look for any places that you think are overly complicated. Our login/logout behavior needs a larger refactor but I want to start from a place of things working, and then if it still works the same when I'm done, we're good.


jenweber:
I'm holding off on merging this until #1195 is in and I resolve any merge conflicts.

PR #1196, opened 2020-01-03
Continuing git refactor

alexspeller:
Make some more progress in isomorphic git refactor by abstraction, wanted to get this merged before any divergence happens


habdelra:
I'm wondering if we really should be merging this into the `new-hub` branch. the hub API's that you are using (like the indexer and writer) are changing drastically in the `new-hub` branch. I'm unsure how useful it is to try to make the new isomorphic git work on both the old and the new hub. If I had to pick, I'd suggest that you concentrate your efforts on the new-hub branch, as the old hub is going away.

/cc @ef4

habdelra:
I'm trying to merge master into the new-hub branch, and I think i'm going to have to just leave the git package out for now, as I'm not sure how to get your stuff working in the new-hub branch without having the firsthand knowledge that you have.

PR #1213, opened 2020-01-07
Fix the back button for the themer

### LINKED ISSUE #1199: Fix the back button for the themer
To-do:
- [x] Create a new route that has existing themer functionality
- [x] Remove themer logic from the layout route
- [x] Write an auth redirect for themer (unauthenticated users get sent to layout)
- [x] Make sure the back button works
- [x] Replace "Close Editor" with "Save" button
- [x] ~~Show the logout button???~~ Decided to skip
- [x] tests
- [x] make the close button be a "secondary" button
- [x] figure out if latest changes to master broke something under the hood (unexpected test failures)

jenweber:
Closes #1199

## Added
- themer logic and layout are copied into a new route

## Changed
- The close button for the themer is now "Save and Close"
- Save and Close button width
- Testing: Replaced monaco editor with a simple `<code>` element for speedier testing/renders
- Testing: renamed the test selector for the card json
- Testing: update existing tests to use the code selector
-  Testing: moved themer tests out of view tests and into editing-css-tests

## Removed 
- logic and layout for the themer from view
- reliance on the editingCss query param throughout the app

The close button is now "Save and Close" because, since the themer route is a sibling of view, the model hook reruns and wipes out the changes.

## How to review:
The diff on this PR is pretty big, but it's mostly copy & paste and tests. I am not sure why some older commits by Will show up on the log for this PR. Possibly I branched from something other than master by mistake.

- try navigating to and from the themer, using in-app links/buttons as well as the browser "back" button
- see that changes made to the CSS get saved successfully, and you can see the preview
- check for any code quality issues


packages/cardhost/app/controllers/cards.js
```diff
@@ -4,15 +4,20 @@ import { inject as service } from '@ember/service';
 export default class CardsController extends Controller {
   @service cssModeToggle;
   @service cardstackSession;
+  @service router;
 
   get themerClasses() {
-    let editing = this.cssModeToggle.editingCss;
-    if (editing && this.cssModeToggle.isResponsive) {
+    let onThemerRoute = this.router.currentRouteName.includes('themer');
```
bagby: you could use `this.router.currentRoute.localName === 'themer'`...not sure if that's better though 😄

bagby: also why not use the getter on line 20, i.e. `this.onThemerRoute`?

jenweber: I think your suggestion is an improvement! I'll fix these two things and merge.

PR #1229, opened 2020-01-15
Show newly created cards

### LINKED ISSUE #1002: Newly created cards are not added to the Recent Cards section of the launcher screen
I can see cards created in the GitHub repo, but they are still not appearing on the Recent Cards screen. 

image

image

jenweber:
Closes #1002 

## Added:
- `cardLocalStorage` service to persist the ids of recent cards to local storage (get, add, remove, clear)
- unit tests for the service
- acceptance tests

I named the local storage service and methods in such a way that we can add more types of things to local storage in the future.

## Changed
- the model hook returns recent cards, and does not use `allCardsInStore` anymore
- in deployment, templates such as "event-card" will not show up in the "Recent Cards" list. This is a good thing, because we don't want our users to edit those directly, only adopt from them.

## Not included:
- I couldn't actually refresh the page in an acceptance test, so the service unit tests will have to do.
- there's no query param for clearing the local storage, since it's easier to do via command line. `    localStorage.removeItem('recentCardIds');`

## How to test manually

Create a card. Create another card. Go to the catalog page. Do you see the cards? Refresh. Are they still there? Cool!


packages/core/addon/services/data.js
```diff
@@ -17,6 +17,7 @@ let store = {
 
 export default class DataService extends Service {
   @service cardstackSession;
+  @service cardLocalStorage;
```
bagby: is this necessary?

jenweber: It is not, and I have since removed it. I also moved the local storage service into cardhost altogether, so this PR introduces no changes to the core package.

PR #1233, opened 2020-01-15
Create a minimum view width for the Builder, closes #1167

### LINKED ISSUE #1167: Set a minimum screen width so that mobile view doesn't look broken
We aren't doing mobile styling yet, so in the meantime, all pages should have a minimum width, similar to what we do on the Catalog page:

Screen Shot 2019-12-24 at 11 33 56 AM

Here's an example of what should just be unfolded in a larger width. Set the minimums so that the buttons don't overlap.

Screen Shot 2019-12-24 at 11 33 48 AM

jenweber:
The goal of this PR is to make minimal changes to help the Builder be usable at smaller screen widths, and overall look not-broken during UX testing. It does not aim to be truly responsive.

## How to test
In each main page view, try viewing it in a half-size browser. Compare with what is currently deployed.

I'm open to any feedback about how to tackle this differently.

### DIFF
--- a/packages/cardhost/app/styles/app.css
+++ b/packages/cardhost/app/styles/app.css
@@ -18,6 +18,9 @@
 @import '/assets/components/themer.css';
 @import '/assets/components/dialog.css';
 
+/* Handling minimum screen sizes for desktop */
+@import '/assets/responsive.css';
+
 /* Cardhost styling for core package templates */
 @import '/assets/components/card-renderer.css';
 @import '/assets/components/field-renderer.css';
@@ -90,7 +93,7 @@ Until that work is done we are statically adding this css
 }
 
 body {
-  min-width: 100%;
+  min-width: 900px;
   min-height: 100%;
   width: 100%;
   height: auto;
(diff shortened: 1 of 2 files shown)


burieberry:
@jenweber  Under 900px, the page scrolls horizontally but the top edge and the right-edge seem to be cropped. If the goal is to make it scroll like the catalog homepage, I think the `right-edge`, `cardhost-top-edge`, and `cardhost-left-edge` classes need to be absolutely positioned instead of fixed.

right-edge-cropped

edges-cropped

jenweber:
Ok thanks! I'll fix that.

jenweber:
I've made some edits in the last commit to:
- add a background color that was missing in the panels that fell off the edge of the screen
- position more things absolutely
- set the breakpoint for the schema page a bit smaller so that it handles resize more gracefully

burieberry:
@jenweber  The changes look good. I realized there're a couple other places that need absolute positioning, too. `themer--top-edge-container` (css-editor view), `card-renderer--top-edge-buttons` (cropped on schema view),  and `cardhost-left-edge`, if you want that one to be part of the scrolling page too.

jenweber:
I'm going to let this sit until we get the new navigation in place, otherwise it will need to be redone.

jenweber:
I am closing this because navbar designs are still in flux, and they are different enough that we'll need to redo how the top edge works.

Issue #1209, opened 2020-01-07
Edges should slide in/out when summoned/dismissed


dnaidionov:
When edges are summoned or dismissed they should not just appear on screen, but slide in/out like in the following animation: 
https://www.loom.com/share/2f7e4496c76c40ac950896e994784a20
The transition speed should be synced with the other animation happening on screen. 

Currently, they are just appearing on screen on top of other animation that is happening:
https://www.loom.com/share/90bd9ab35aef4911bd9197a84e83782c

alexspeller:
I believe this is blocked by #1249 - or at least, #1249 mentions that animations are currently blocked

burieberry:
A little caviat, this _might_ become throwaway work _if_ we end up using the cardstack-edges package (exploratory work for that in #1236). Since then the sliding in and out comes built-in. But #1236 requires a ton of changes so I'm not sure when or if it would be ready to be merged.

dnaidionov:
we need this transition to be able to judge if this is the right way to handle edges during route transitions. so even if we later switch to 1236, i'd rather have this sooner so that we have a working transition meanwhile

PR #1236, opened 2020-01-16
Explore edges

burieberry:
@ef4 I'm wondering if there's a very obvious error here. Am I missing something when trying to use cardstack edges? (This work was just for exploration purposes, in the case that we do end up using the edges package.) 

1- Created service that registers `cards` template as top level component
2- Wrapped outlet with `CardstackEdges` component, which inserted the ember-toolbars components
3- Attempted to insert `cardhost-top-edge` component in cards component by using `{{cardstack-edge "top" (component "cardhost-top-edge")}}`. This is the error I got:

in-top-toolbar-error

4- Inserted the not-found `in-top-toolbar` component manually to get around that error. Now the `cardhost-top-edge` component is actually inserted in the top edge (`cst-top` div) in the HTML. But its visibility is still set to `hidden`. The page loads, but this error appears in the console:

liquid-versions-error

This error continued to appear after I updated ember versions to 3.15 and yarn linked ember-toolbars and liquid-fire packages.


burieberry:
Closing this, since it will be handled in #1209 using `ember-toolbars` addon. @alexspeller if you come across similar issues as I posted above, see Ed's comment about installing `jquery` and adding to the `packageRules`.

Issue #1014, opened 2019-11-18
Move Login/logout buttons to the context menu displayed when user clicks profile icon


dnaidionov:
When user is logged out profile icon should display cardstack logo. 
When user clicks on Cardstack logo, the left edge would open and context menu displayed in it. The context menu would have "Log In | Sign Up" items for social logins. The only active button would be "Log In with GitHub", all other are disabled and marked "Coming Soon" (see xd). 
When user is logged in profile icon should display user's profile picture. For GitHub login it would display user's GitHub avatar. When user clicks on it, left edge would open. Next to profile picture we would display user's name and GitHub username. When user clicks on them we would take user to profile page (#1200). 

XD: `login sequence.xd` https://drive.google.com/open?id=1RSxnpEKSHzm4SLQiYE2FbfS5zZ4fRK-T

jenweber:
Todo:
- [ ] replace profile icon with cardstack logo
- [ ] apply a background to the left edge
- [ ] Click to toggle left edge width and reveal more buttons
- [ ] animate the left edge reveal
- [ ] remove login/logout buttons from all pages
- [ ] add buttons to the revealed panel
- [ ] Check to make sure nav works for keyboard commands
- [ ] make sure it still looks ok in smaller screens
- [ ] tests

PR #1239, opened 2020-01-16
WIP Context menu #1014

jenweber:
Partially addresses #1014 

## TODO

- [x] Wait to merge until #1235 is merged in, and this PR has the latest commits
- [x] Tests need some work
- [x] Figure out why the if helper + tracked isn't working in test mode. Maybe fix it by upgrading Ember versions??? EDIT it's because I put the data-test on the wrong 

After merging
- [ ] the login button styling in #1233 should be removed, after that is merged.

## Added
- left edge can expand and collapse when you click the Cardstack logo
- menu automatically collapses when you click outside of the nav, or it loses focus
- Ability to navigate the menu by keyboard
- some aria labels and roles

## Changed
- moved login/logout buttons to the left edge
- Styled these buttons as text only, instead of a CTA
- The profile icon is now the Cardstack logo

## Removed
- login/logout from the top edge
- some login/logout button styling

## Will do in another PR
- More "pixel-perfect" styling
- placeholder elements for the social sign on
- Some way to prompt the user to sign in???

## How to test/feedback needed

To manually test, click on the cardstack logo from the edit/schema/layout pages

- What, if any, additional aria things need to be done?
- How to simplify/refactor/rename CSS?
- The sidebar looks kinda ugly with the current nav, but it may be better after Will's work is in
## Demo
See `login-logout-left-edge-2019-01-16.mov` in Google Drive for a short video of how it works.


packages/cardhost/app/components/cardhost-left-edge.js
```diff
@@ -0,0 +1,59 @@
...
+  }
+
+  /*
+    closeListener, registerCloseListener, and destroyCloseListener are the
+    actions responsible for closing the left edge when you click somewhere
+    outside of it. They are only registered when the left edge is open,
+    and torn down when it is closed via the did-insert and will-destroy
+    render modifiers.
+  */
+  @action
```
bagby: you could simplify this by putting a `{{did-update this.toggleListener this.isExpanded}}` on the `#cardhost-left-edge` element, and do something like
```
  @action
  closeListener(element) {
    return event => {
      if (!element.contains(event.target)) {
        this.closeLeftEdge();
      }
    };
  }

  @action
  toggleListener(el, [expanded]) {
    let listener = expanded ? 'removeEventListener' : 'addEventListener';
    ['click', 'focusIn'].forEach(evt => document.querySelector('body')[listener](evt, this.closeListener(el)));
  }
```

bagby: although I suppose the argument could be made to keep it more explicit 🤷‍♂

jenweber: I think we need to keep it as `will-destroy` because we need to ensure that we tear down this click handler when we leave the route as well.

PR #1244, opened 2020-01-17
Builder app - update to Ember 3.15 Octane

jenweber:
This applies version bump and blueprint changes only.

A followup PR will include linting fixes to allow us to use `extends: octane` in templatelint.


jenweber:
I'm not sure why there's a failing test. Working to replacate locally.

jenweber:
How interesting, we need sync Observers. I think we should fix the `"default-async-observers": false,` issue some other time.

PR #1251, opened 2020-01-17
Base card css updates; event-ticket card themed styles

burieberry:
default-styles
themed-event-card
themed-event-scrolled


jenweber:
This makes sense to me!

I see a gap for Firefox:
Screen Shot 2020-01-20 at 3 49 14 PM

jenweber:
You may also want to turn on the in-browser a11y testing (temporarily) in ember-cli-build.js to check for issues.

PR #1254, opened 2020-01-20
implement view mode animations

bagby:


packages/core/addon/templates/components/field-renderer.hbs
```diff
@@ -3,7 +3,7 @@
 
 {{!-- TODO: let's think of a more accessible solution here, consider making this a button--}}
 {{!-- TODO: also we should also turn on a11y linting --}}
-<section {{on "click" (fn this.selectField @field)}} {{on "dragstart" (action this.startDragging @field)}} {{on "dragend" (action this.finishDragging)}}
+<section {{on "click" (action this.selectField @field)}} {{on "dragstart" (action this.startDragging @field)}} {{on "dragend" (action this.finishDragging)}}
```
jenweber: If we use `fn` instead of `action`, will it work? Enabling the "octane" app linting would require us to remove uses of `action` from templates.

bagby: ahh good catch, I changed it while I was thrashing around trying to fix a thing. I'll change it back, it should work fine.

PR #1255, opened 2020-01-20
Create Link and CTA core types

### LINKED ISSUE #1240: Create a URL core type
The URL core type would be similar to case insensitive text, but will render as `href`. Links should open in a new tab. Before we have full support for card as fields we will have the value of the field serve as both display value and link URL. In the future, those two will be distinct.

### LINKED ISSUE #1241: Create a CTA core type
Initially, the CTA core type would be similar to case insensitive text but will render as a button. 
When rendered, field label would not be displayed, only the button itself. We will start with link buttons. A user would specify the button label and link URL. 
In the future (when we have support for cards as fields) in addition to links, we would have this type configurable to support HTTP request methods, custom actions via javascript, and cardstack action cards.

jenweber:
closes #1240 
closes #1241 

After merge, we need to rev the hub.

## Added
- Link core type
- CTA core type
- tests for CTA and Link
- Styling for the Link core type
- A new color to the palette (in ui-components and variables)

## Changed
- Switched some former urls to be actual URL types

## How to test
Create a card, drop in a CTA and Link field using the schema editor, fill in the URLs in edit mode, and see how it looks in layout


jenweber:
Sidenote, I think we should use these in one of the new cards and make sure we snapshot it.

PR #1257, opened 2020-01-21
testing - only take desktop screenshots with percy

jenweber:
So, the reason we had such high Percy totals is that we take 18 snapshots, but they are repeated for all widths (mobile, desktop) and browsers (Chrome, firefox). So that turns into 72 screenshots for every push to a PR. I can cut this in half by removing mobile, since we intend this as a desktop app for the time being.


jenweber:
Screen Shot 2020-01-21 at 12 33 18 PM

Issue #1249, opened 2020-01-17
Edit Mode Fields screen


dnaidionov:
## Top edge:
The top edge should display the following controls: 
- Edit mode indicator/close button
- Layout/Fields switching controls
- Temporarily (until we have autosave) display Save button on the right side (not reflected on screenshot)

![image](https://user-images.githubusercontent.com/1836666/72645880-afe70880-3942-11ea-8a8c-3bc8a5310592.png)

## Right edge: 
The right edge should display a CTA to switch to **Schema mode** - "Configure Schema" button

## Card header: 
The card header displays card type icon and card name on the left side, and "Done" and context (...) buttons on the right side. 
- Done button closes Edit mode and returns to the View mode with card header hidden (Exit edit mode button also closes Edit mode, but leaves card header visible in the View mode). 
- Context button is reserved for future use

![image](https://user-images.githubusercontent.com/1836666/72645697-3a7b3800-3942-11ea-85b0-1334872a0a05.png)

## Background
The screen background is the color from our palette closest to #B7B1BA. 
![image](https://user-images.githubusercontent.com/1836666/72645727-4d8e0800-3942-11ea-93e7-30e3a4b791f5.png)
The top edge should have a transparent background. 

## Transitions: 
- When entering this screen from View mode, right edge slides in from the right. Top edge slides in from the top. 
- When switching to this screen from the "**Edit Mode Layout screen**", right edge's content fades out/in and edge resizes if needed. On the top edge "screen size" and "preview" buttons are sliding in from the top of screen. 
- Switching between this and Schema screens are described in Schema screen ticket
- When exiting Edit mode to View mode, top edge retracts to the top, right edge - to the right off the screen.

jenweber:
The white text against the bright blue background in this mockup is not accessible. We should aim to use black instead or get new designs.

bagby:
@dnaidionov I believe all the items in the issue have been done *except* the animations, can you confirm and create a new issue just for the animations?

dnaidionov:
@bagby: 
- the exit edit mode button should be on the left side of the screen (inconsistency in the placement of "exit mode x" buttons is intentional)
- the items in the card header section are not implemented
- the background of the top edge transitions later than the page background. can we just make top edge background transparent?

PR #1258, opened 2020-01-21
Edit mode styles

bagby:
Addresses most of #1249, except for animations, which are blocked at the moment. Plus some misc. fixes to view mode animation and a few other style tweaks.


bagby:
I think something broke with my latest master merge, some weird `z-index` thing.

bagby:
@jenweber I make the claim that the visual diff for the cards is actually correct (see screenshot of diff below). We have an existing CSS var `--card-renderer-header-height: 50px` that we use to determine the height of the content area, ex:
```
.card-renderer-isolated--content {
  // ...
  min-height: calc(var(--card-renderer-height) - var(--card-renderer-header-height));
}
```
but we weren't actually using it as the height of the header. The actual height of the header is 56px, and I'm using the CSS var to set that explicitly. So the min-height of the content div is effectively 6px shorter.

Screen Shot 2020-01-23 at 4 29 12 PM

bagby:
@jenweber fixed

Screen Shot 2020-01-24 at 3 37 28 PM

bagby:
@jenweber ahhhh nevermind something else is broken on the right side of the page 😠

bagby:
ok now it's fixed :)

PR #1259, opened 2020-01-21
Fix custom/default theme switcher

### LINKED ISSUE #1252: Finish the "custom theme" button/dropdown

jenweber:
Closes #1252 
Now when you choose "default" from the theme dropdown, it resets your CSS to default. If you have made a custom theme, that is shown as the selected option in the dropdown.

## Added
- logic for checking if the card is using default styles (i.e. has no custom styles defined)
- tests
- selecting "default style" clears any custom CSS

## Removed
- cruft code from the card-viewer route

## How to test by hand

Create a card. Note that the Theme dropdown says "Default." Add some custom styling. Go back to layout, and note that the dropdown says "custom." Save. Select "default" from the dropdown and see the custom styles disappear. Save. Refresh the page and see that they are still gone, and css has been set back to `.cardstack_base-card-isolated { }`


jenweber:
@bagby it is lost to the ether that is git. Dmitry and I are going to meet to figure out how it should work in the future.

PR #1275, opened 2020-01-22
Auto-save card after 5 seconds

bagby:
Adds a `{{did-update}}` modifier to the Save button to hook into existing save functionality. Addresses parts of #1276, with the following behaviors:

 * Autosaves the card any time `card.isDirty` is `true`.
 * User can still save the card manually
 * Does *not* autosave if it is a new/adopted card (i.e. `card.isNew === true`), the user must manually save it first. This behavior will change with #1277
 * Autosave has a 5000ms (5s) debounce, this can be overridden globally in `config/environment.js` ex. `ENV.autosaveDebounce = 1000`
 * Autosaving is turned off in tests by default. For component tests, you can enable it with:
```
<SaveButton autosaveDisabled={{false}} ... />
```
 * In the app, you can disable it with:
```
<SaveButton autosaveDisabled={{true}} ... />
```
 * It is reusing the save button UI -- spinner when saving, checkmark and temporary green background when saved, etc


packages/cardhost/app/components/save-button.js
```diff
@@ -13,7 +18,16 @@ export default class SaveButton extends Component {
...
-  @task(function*() {
+  autosaveDebounce = autosaveDebounce || AUTOSAVE_DEBOUNCE;
+  autosaveDisabled = typeof this.args.autosaveDisabled === 'boolean' ? this.args.autosaveDisabled : !!autosaveDisabled;
+
+  get cardIsNew() {
+    let card = this.args.card;
+    return card.isNew;
+  }
+
+  @enqueueTask
```
jenweber: I think maybe we want `[@keepLatest`](http://ember-concurrency.com/api/TaskProperty.html) instead of `enqueue`. A GitHub roundtrip is 4 seconds, and so I think bursts of typing or whatever could rapidly spawn 20+ seconds of queued requests. Is that accurate?

bagby: @jenweber  the saves need to be enqueued so that we don't have `/meta/version` mismatches. We could mitigate the scenario you described by making the debounce longer (maybe 3-5 seconds?)

I'll put some comments in the code to reflect this.

jenweber:
Followup work is tracked in #1280. If you switch routes, saving is interrupted.

PR #1278, opened 2020-01-24
Upgrade embroider to get addon rebuilds

ef4:


ef4:
What's the deal with Percy failures?

PR #1284, opened 2020-01-28
Card name modal dialog when creating/adopting card

bagby:


packages/cardhost/app/templates/components/card-name-dialog.hbs
```diff
@@ -0,0 +1,16 @@
+<Dialog open={{true}} @title={{this.title}} @mode={{@mode}} @icon={{@icon}}>
+  <p class="dialog--message">
```
jenweber: All form element need labels. I think you could make this into the label and it would work well:

```html
<label class="dialog--message" for="card_name">
    Give your card a name:
</label>
<TextField id="card_name" />
```

packages/cardhost/app/templates/components/card-name-dialog.hbs
```diff
@@ -0,0 +1,16 @@
...
+    Give your card a meaningful name
+  </p>
+  <TextField
+    @id="card__name"
+    @placeholder="i.e. Millenial Puppies"
```
jenweber:
```suggestion
    @placeholder="i.e. My Event Name"
```

We should avoid 'Millennial Puppies' and Ember Meetup things for our examples, and take inspiration from the cards that are shown on the website now.

packages/cardhost/app/templates/components/card-name-dialog.hbs
```diff
@@ -0,0 +1,16 @@
...
+    @setValue={{action this.updateCardName}}
```
jenweber: Optional - If you can just pass the function here and it works, that's better. When we turn on Octane linting, we'll have to remove all uses of "action."

packages/cardhost/app/templates/components/cardhost-modal-target.hbs
```diff
@@ -0,0 +1,12 @@
+{{#from-elsewhere name="cardhost-modal" as |args outsideParams|}}
+  {{#if args}}
+    <div class="cardhost-modal-backdrop"></div>
+    <div
+      role="button"
```
jenweber: We should not use `<div role=button>`. We lose all the automatic keyboard navigation and focus management that the browser gives you. Can we make it a `<button>`?

packages/cardhost/tests/acceptance/create-card-test.js
```diff
@@ -151,11 +113,36 @@ module('Acceptance | card create', function(hooks) {
...
 
+  test('creating a card from the homepage', async function(assert) {
+    await login();
+    await visit('/');
+
+    assert.equal(currentURL(), '/');
+    await percySnapshot(assert);
+    await click('[data-test-new-blank-card-btn]');
+    await setCardName(card1Name);
+
```
jenweber: - I think we should add a percy snapshot that shows the open modal.
- You could also add the open modal view to the `accessibility-test.js`
- maybe we need a test that also clicks the close button and asserts that the modal is gone, unless that is already tested under the dialog component

packages/cardhost/app/controllers/index.js
```diff
@@ -19,4 +22,15 @@ export default class IndexController extends Controller {
...
+
+  @action
+  openCardNameDialog(title, model /*, evt*/) {
+    if (arguments.length > 2) {
+      this.cardModel = model;
+    }
+    if (arguments.length > 1) {
+      this.dialogTitle = title;
+    }
+    this.showDialog = true;
```
jenweber: Whenever you open a modal, it should receive focus. I am not sure if that is already handled in the dialog component or not.

See https://webaim.org/techniques/keyboard/

Screen Shot 2020-01-29 at 10 37 42 AM

This page is also worth a read. It would be good to identify which things are the responsibility of this PR vs what should be implemented in the dialog compomnent: http://web-accessibility.carnegiemuseums.org/code/dialogs/

jenweber: This is maybe the first resource to look at: https://melsumner.gitbook.io/ember-component-patterns/messaging/modals

jenweber: cc @alexspeller since you are also working on modals

packages/cardhost/app/templates/components/card-name-dialog.hbs
bagby: Good call, will do.

packages/cardhost/app/templates/components/card-name-dialog.hbs
bagby: 👍

packages/cardhost/app/templates/components/cardhost-modal-target.hbs
bagby: sure...?

jenweber:
The outside click to close doesn't seem to be working for me. I tested it in Chrome and Firefox.

bagby:
@jenweber outside click should be working now. I'm also auto-focusing the dialog when opened, and I added better tests with percy screenshots.

packages/cardhost/app/controllers/index.js
bagby: @jenweber I did add an `{{autofocus}}` modifier that automatically gives the dialog focus when inserted. Any other changes that need to be made to the `dialog` component should probably be done in a separate PR.

As an aside...using the `<dialog>` HTML element is already somewhat problematic since it's not supported in Firefox or Safari. There is a `dialog` role, i.e. `<div role="dialog">` but I'm not sure whether or not that's more widely supported.

jenweber:
I think this is good for us to merge now and do some ux testing.

PR #1285, opened 2020-01-29
Add warning when trying to leave with unsaved content

### LINKED ISSUE #1155: Display confirmation prompt when a user tries to leave the dirty page.
If a user has some unsaved changes and tries to switch to a different view  (layout/edit/schema), we should display a warning modal with options to proceed and return to the current page. 
Proceeding with route change will abandon changes. Returning to current card will allow the user to save them before navigating away.

Modal design - see  #1043
Icon: Warning
Title: "Unsaved changes"
Text: "You have some unsaved changes in this card. If you proceed to change the view or navigate away, they will be lost. If you want to save changes, please stay in this view and save them. Do you want to proceed, or stay?"
Buttons: "Proceed", "Stay"

alexspeller:
Adds a warning when trying to leave with unsaved content.

Closes #1155 

Todo:

- [ ] Tests
- [ ] Browser native confirm warning for when navigating outside of ember actions

### DIFF
--- a/packages/cardhost/app/controllers/cards/card.js
+++ b/packages/cardhost/app/controllers/cards/card.js
@@ -1,0 +1,16 @@
+import Controller from '@ember/controller';
+import { tracked } from '@glimmer/tracking';
+import { action } from '@ember/object';
+
+export default class CardsCardController extends Controller {
+  @tracked attemptedLeaveTransition;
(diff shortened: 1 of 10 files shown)


alexspeller:
No longer relevant with autosave

Issue #1283, opened 2020-01-28
Routing for creating and adopting cards


dnaidionov:
- Click on New Blank Card takes user to new card in Schema mode. 
- Adopting a card (e.g. clicking on template or selecting “adopt” action) takes user to new card in Edit mode.

dnaidionov:
Blocked by #1279. This ticket should be picked only after  1279 is done.

PR #1301, opened 2020-01-31
Create bottom edge

### LINKED ISSUE #1243: Turn launcher screen into bottom edge
Right now the launcher is the starting screen of the application, and it is the way to open existing cards and create new ones. Instead, we will move its functionality and design to the bottom edge, and create a new starting page that will have just featured (and recent - for logged in users) cards and different design (#1272).

image

- Bottom edge always opens to full screen. 
- Menu on the bottom edge should be the same as currently on the launcher screen. 
- Search bar should be present, but not functional. 
- Do not implement sorting/modes switching at this moment 
- Design and behavior of embedded cards stays current (see note below). 

Design of bottom edge (for reference only, keep content as current) can be viewed in https://drive.google.com/open?id=1xSnis7586pGMysyWycrjvGHM7GQzbLZm in the artboard group "Library" on the artboard "Without Guides". 

**Note:** Until we have search implemented in hub, we will just duplicate current launcher screen functionality as a bottom edge. When the search is ready, bottom edge will be revisited.

burieberry:
Closes #1243 

Index page now only displays featured cards:

index-page

If user is logged in, they can click on the library button on the left edge, which pulls up the bottom edge library. The library has the contents of the previous launcher page, including the navigation on the left:

bottom-edge-library
scrolled-down

Completed: 
Phase 1 (this PR): Currently, the bottom edge is only available from the index page. 

TODO: 
Phase 2: Make bottom edge available from other routes. Requires some discussion. We can make index page redirect to `/cards` and display the new homepage there, along with the left edge. This was the model with recent-cards and saved-templates arrays would be accessible from all `/cards/*` routes. This works, however, then the bottom-edge remains open after the route changes. Maybe need to create a service to signal when the edge should be expanded or not.


dnaidionov:
> * New card takes you to edit mode and it ought to go to schema. This is maybe being done in another PR though.

There's a ticket for that - #1283 

> * If you click Log In and then immediately click to open the bottom edge, you get into a weird state (screenshot below). I think we need to make the left edge close when the bottom edge is open.

Bottom edge should take over entire screen hiding all other edges.
Please refer to this diagram (will also pin the XD to the product channel)

image

> * The left edge needs a home button that is separate from the button that opens the bottom edge

We have updated navigation for edges/home/user profile coming. All that would be left edge concerns. I will create a separate ticket for that.

burieberry:
@jenweber 
Updated:
- Index page can now be scrolled vertically
- Regression on card scroll behavior fixed
- Added padding around card, so the themer pane editor doesn't overlap anymore
- Bottom edge is now above the left edge and covers the full view
- Added "Hide Editor" button --> had to place this on left-edge with a higher z-index, since left-edge is controlling the opening/closing of the bottom edge. But this causes the button to appear independently from the bottom edge. Any ideas? I could make it a separate component and have it slide-in from the top edge while library is sliding in from the bottom. Or is there a good way to place the button on the bottom edge? Not sure how I would close the toolbar it is on.

I'll update the tests to test for closing of the library.

dnaidionov:
> * Added "Hide Editor" button --> had to place this on left-edge with a higher z-index, since left-edge is controlling the opening/closing of the bottom edge. But this causes the button to appear independently from the bottom edge. Any ideas? I could make it a separate component and have it slide-in from the top edge while library is sliding in from the bottom. Or is there a good way to place the button on the bottom edge? Not sure how I would close the toolbar it is on.

#1315 (left edge) addresses that. In short - Library button belongs to both left end bottom edges, and it stays with the edge that is on top becoming open/close button. 
If the question is how to display that button while bottom edge is closed, we may assume that bottom edge is not completely hidden, but minimized, and while minimized it has a transparent background

jenweber:
The smallest media query size isn't scrollable, but the others seem good:

Screen Shot 2020-02-05 at 5 10 47 PM

jenweber:
I think the way you did the hide button is good for now. I was thinking that maybe we should keep Hide Library in the bottom left hand corner, and show an x in the top right corner. However now that you mention it, it should probably travel up with the pane. I'd do this by duplicating the button on the bottom edge, and make sure that the initial position of the duplicate matches the position of the "original" underneath. I'm not really sure though. It'd be good for us to try this in XD first.

PR #1308, opened 2020-02-03
bugfix narrow schema width, closes 1294

### LINKED ISSUE #1294: In Schema mode card is too narrow.
It used to be wider, then after some update became narrower. We need to restore the initial width.

jenweber:
closes #1294

This fixes very broken Firefox layout, and in Chrome, there is slightly more space around the trash can. Is this enough to fix the issue that things are too narrow, or should they be wider?

See the percy tests for smaller screen widths. The screenshots below are large monitor proportions.

Before - Firefox
before-firefox

After - Firefox
after-firefox

Before - Chrome
before-chrome

After - Chrome
after-chrome

### DIFF
--- a/packages/cardhost/app/styles/components/card-renderer.css
+++ b/packages/cardhost/app/styles/components/card-renderer.css
@@ -24,9 +24,7 @@
   letter-spacing: 0.015em;
 }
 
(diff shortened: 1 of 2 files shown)


dnaidionov:
lgtm

PR #1309, opened 2020-02-03
add return to editing button link, closes #1220

### LINKED ISSUE #1220: Update design of schema mode screen
In the schema mode implement the following changes: 
- [x] Remove view mode dropdown 
- [x] Replace all other top edge buttons with a single button "Schema Mode" with a close icon - this indicates current mode and allows exiting it. 
- [x] Clicking "Schema Mode" closing button brings users back to the "Fields" view. 
- [x] When the card is selected, the "Configure Schema" button in the Template section changes to the "Return to Editing" button, which performs the same function as the closing schema mode button above. #1309 

image

"State_Transition_V21" https://drive.google.com/open?id=1SN0rhwF1n_mSn9Rs1kArm4dggPRxkhcm
Artboards: "Schema Mode - Card", "Schema Mode - Field"
Note: At the moment we have only wireframes, high fidelity prototypes will come later. This illustrates changes in navigation. Reuse as much of current design elements as possible

jenweber:
Added:
Return to editing button

Changed:
I fixed a bug where you had to click on the words inside the button to change routes, when you should be able to click anywhere inside the button. Using a link inside a CTA button doesn't work great. I took the styles from the CTA button and used them on the link-to.

I opened a ticket in ui-components to create a link-to variation of our CTA: https://github.com/cardstack/ui-components/issues/155

Screen Shot 2020-02-03 at 12 41 25 PM

### DIFF
--- a/packages/cardhost/app/templates/components/right-edge.hbs
+++ b/packages/cardhost/app/templates/components/right-edge.hbs
@@ -38,16 +38,27 @@
                     {{#if (eq @mode "fields")}}
                       <div class="right-edge--items-container">
                         <div class="right-edge--item">
-                          <Cta @variant="secondary responsive" @theme="cs-dark">
-                            <LinkTo
-                              @route="cards.card.edit.fields.schema"
-                              @class="right-edge--configure-schema-link"
-                              @model={{@card}}
-                              data-test-configure-schema-btn
-                            >
-                              Configure Schema
-                            </LinkTo>
-                          </Cta>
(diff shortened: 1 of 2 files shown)


jenweber:
@bagby do you think I should add a test for this? On one hand, there's some logic, but on the other hand it's all hbs logic and if it broke, other things would break too. I'm wary of creating too many "this thing exists" tests when we are redoing navigation soon.

PR #1314, opened 2020-02-05
Ember animated drag drop

bagby:
Adding fields from the catalog via drag & drop uses Ember Animated instead of the HTML Drag & Drop APIs. This allows more control over the layout/UI. The intent of this PR is to achieve parity with the current functionality, new designs will be implemented in future PRs. 

Reordering fields within the card is *still* using native drag & drop APIs. There is a not insignificant amount of refactoring of the `field-renderer` that would need to happen to use Ember Animated, and I opted to leave that for a separate PR.

### Added
 * A `draggable` service for keeping track of dragged fields and dropzones, and triggering appropriate events when dropping
 * Adds an additional method for adding fields -- you can simply click on the catalog field, hover over a drop zone, and click to add the field. This was primarily added to make testing easier, but it could be useful on its own.

### Changed
 * Uses `ember-animated` for adding fields in schema mode via drag & drop instead of HTML Drag & Drop API


bagby:
FYI I had started to implement adding fields via keyboard, but didn't complete it, so there's a lot of copy-pasted code in here around keyboard events. I'm happy to take it out, or I can just implement it in this PR.

jenweber:
@bagby it's up to you. I think keyboard controls would be a reasonable followup PR.

PR #1329, opened 2020-02-10
Leave ghost element

### LINKED ISSUE #1265: Only part of the dropzone is active on a blank card
When adding fields to the blank card it shows the entire area of a card as an active dropzone, however, only top part is accepting the card - if you drop the card in the lower part of the dropzone it would be discarded. The entire dropzone should be accepting cards. 
Kapture 2020-01-21 at 16 01 30

bagby:
Addresses parts of the UI design for drag & drop

Fixes #1265 

image

### DIFF
--- a/packages/cardhost/app/templates/components/card-renderer.hbs
+++ b/packages/cardhost/app/templates/components/card-renderer.hbs
@@ -14,7 +14,7 @@ Ember Concurrency into the rendering of the card component here.
           {{safe-css-string @card.adoptedFromName}}
           {{safe-css-string @card.name}}
           {{if (and this.cardstackSession.isAuthenticated @cardSelected) "selected"}}
-          {{if this.fields.length "has-fields" "no-fields"}}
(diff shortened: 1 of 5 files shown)


packages/cardhost/app/templates/components/card-renderer.hbs
```diff
@@ -14,7 +14,7 @@ Ember Concurrency into the rendering of the card component here.
           {{safe-css-string @card.adoptedFromName}}
           {{safe-css-string @card.name}}
           {{if (and this.cardstackSession.isAuthenticated @card.isSelected) "selected"}}
-          {{if this.fields.length "has-fields" "no-fields"}}
+          {{concat "field-count-" this.fields.length}}
```
jenweber: This is a super nice improvement. Could be quite useful in the future.

jenweber:
Unfortunately the snap-back behavior still happens when you successfully drop a field, but your mouse is over another field. I'll send you a video of what I mean.

Issue #1330, opened 2020-02-11
Refactor `@card.isSelected` and `@cardSelected` state management


jenweber:
On hold until there's an ok time for making a bunch of changes.

We have two booleans that control "selected" card state, `@cardSelected` and `@card.isSelected`. They clash and are repetitive, making debugging and animations difficult.

The boolean has two jobs:
1. Influence styling, i.e. `{{if @cardsSelected "selected"}}` determines what is bright blue
2. Influence animation, i.e. whether the card header slides up on hover in the View mode
3. Influence state of what is displayed in the Right Edge on the Schema page. When you click on the card, the right edge shows card settings. When you click on the field, the right edge shows field config.

## What should happen instead

We should use just one boolean, and it should not be a property of the card itself. State should be changed via an action on the controller only.

### Recommended approach

Find all the places that depend on selected state. Note what they do. Design the controller API for managing that state. Starting with schema, comment out/remove existing state management and replace with the redesigned management from the top down. All other card modes are less complicated than schema.

Overall intended header states across modes:
![26D6F065-5392-4D06-8E51-0DE540A81927](https://user-images.githubusercontent.com/16627268/74258365-3255b480-4cc4-11ea-8871-fdcdf2467483.jpeg)

### Areas of possible improvement

There are also some aliased actions handing around like `@cardFocused` that make debugging state changes difficult.

We also use the `mut` helper to modify parent component state and card property state. We should replace this with an action on the controller. Changing this may break animations, so check them at every step; I found that turning mut helpers into actions made state changes be hidden from ember-animated components, which is very mysterious.

When `@card.isSelected` is removed, it may require changing `{{animated-value @card watch="isSelected"}}` to `{{animated-if @cardSelected}}` instead.

jenweber:
I am not sure if this still needs to be done. Step 1 is to see if this pattern still exists after the new hub merge. If it seems fine now, close this ticket.

PR #1331, opened 2020-02-11
Bugfix - header and field highlighting for schema

### LINKED ISSUE #1323: Bugfix - regression on schema card and field highlighting
When I merged #1317, it was with the knowledge that there was a regression on the highlighting for the Card Schema mode. When a field is selected, it should be blue and the header should be grey. When a field is not selected, the header should be blue. I think I missed a boolean somewhere. It's easier to fix this regression after the fact than keep chasing after merge conflict problems:

Screen Shot 2020-02-07 at 3 29 09 PM

Screen Shot 2020-02-07 at 3 29 00 PM

jenweber:
Closes #1323

Problems I ran into are described in #1330 for possible future refactor. However in this PR I changed as little as possible.

To review:
Run it locally and click around/hover in each mode to make sure header and field selection match the intended states.


jenweber:
Superseded by #1332

PR #1333, opened 2020-02-11
Set animation speed to be super fast in testing

jenweber:
must-go-faster

Changed:
Our CI runs were really long due to 250ms animations on card renderer and 1000ms on the index page. Will fixed this at one point, but it looks like we missed a few spots in subsequent work. This PR sets animation duration to 20ms everywhere that I could find it.


jenweber:
Somehow, this had almost no impact. 🤔

jenweber:
Unfortunately the bottom edge animations don't appear to have a configurable duration, via ember-toolbars.

Issue #1315, opened 2020-02-05
Left Edge


dnaidionov:
Left edge contains organization and meta navigation elements. 
- [x] Top button (cardstack logo) serves as "Home" button. It leads to home page (user's dashboard). 
- [x] Second button is user avatar. Clicking it opens left edge with log in/out controls. 
- [x] When user is not logged in it shows generic user icon. 
- [ ] When user is logged in it shows user's avatar

- [ ] **TBD**
>Next block is list of user's organizations. Clicking on organization opens left edge with inbox for that org and list of quick actions available. This block is optional, and may not be included on launch. 

- [x] At the bottom is the library button. It opens the bottom edge. 
- [ ] Second from the bottom is catalog button. It opens card catalog page.

![image](https://user-images.githubusercontent.com/1836666/73864210-4191b980-480f-11ea-9352-d01811a6453d.png)

- [x] Library button, while positioned at the bottom of the left edge, belongs to both bottom and left edges, so when the bottom edge is opened, library button retains its position on the bottom edge - it slides up with the bottom edge, covering the other buttons on the left edge, and with the open bottom edge it serves as "close library" button:

![image](https://user-images.githubusercontent.com/1836666/73308094-8739fb00-41ed-11ea-8dac-acfbae968517.png)

burieberry:
Moving this back to to-do, since we can't yet do the remaining 3 items (different users, organizations, catalog page)

PR #1335, opened 2020-02-11
Update left-edge controls; make bottom-edge available

### LINKED ISSUE #1318: Bottom-edge improvements
This picks up where #1301 left off. See the comments thread there.

- [x] Currently bottom edge is only accessible from the index page. Make bottom edge available on other routes (make sure the Recent Cards and Saved Templates cards are also available).
- [x] This means we lose the homepage button, but that should be handled in #1315.
- [x] The current behavior of the bottom edge when user clicks on a card in the library should be preserved.
- [x] When the bottom edge is opened/closed, the "Hide Library" button should travel up/down with the bottom edge, not just immediately appear on the page as it currently does.

### LINKED ISSUE #1357: Change the color of the Card Catalog on schema page
Color should be as close to #413E4E as accessibility features allow

burieberry:
Part of #1315 
Closes #1318 
Closes #1357 

Updates:
- MaybeAnimate component resolves the id conflicts issue
- Update recent-cards fetching so new cards appear when library is opened
- Updated `createCards` test helper to directly go to `/cards/new` route instead of waiting for the library animation

@bagby it looked fine to me but can you check to see if drag-and-drop is still working as expected?

---

- Compressed uncompressed svgs using image optim and general svg cleanup
- Added profile button and moved login there
- CS logo is the new home button
- Library button pulls up the library in all pages (not just the homepage)
- Hide Library button travels with the bottom edge
- Created library service
- Homepage now redirects to `/cards` so the model with bottom edge library cards is now available in other places too


jenweber:
@burieberry The first things I would look for with the bug you are seeing are:
- It kinda looks like the model hook isn't running when it should. How are we doing the LinkTos from the bottom edge? If you provide the entire record to a LinkTo, it can skip over the model hook, whereas if you provide only id, it will run.
- Is something missing `@tracked`?
- Is that container managed by Ember Animated?

jenweber:
To fix the issue with the new card behavior, we need to save the card before attempting the transition: `yield newCard.save();`

I don't quite understand how it ever worked without this, but maybe @bagby has some insights. If we do a save here, we would also need to handle the id conflicts (err 409) in some way.

```js
  @task(function*() {
    let adoptedFrom;

    if (this.args.model) {
      adoptedFrom = yield this.data.getCard(`local-hub::${this.args.model.name}`, 'isolated');
    }
    let newCard = this.data.createCard(`local-hub::${this.cardId}`, adoptedFrom);
    yield newCard.save();
    if (adoptedFrom) {
      this.router.transitionTo('cards.card.edit', newCard);
    } else {
      this.router.transitionTo('cards.card.edit.fields.schema', newCard);
    }
  })
  createCardTask;
```

habdelra:
@burieberry thanks for the heads up, this looks good. I know what I need to do to incorporate it into new-hub.

PR #1337, opened 2020-02-12
Reduce dev build times by 30% 

### LINKED ISSUE #1336: Try configuring monaco to speed up builds and reduce bundle size
## status quo for dev builds
Without config, it's 1.05 gzipped, 6.79 parsed. It takes approximately 1:26 to build.

## configurations for dev builds

With language configuration (ts, js, html, and css only), gzipped size of monaco deps is 1.04MB, parsed 6.64MB. There is no substantial change in build time.

Configuring features as well brings it down to 804kb gzipped, 4.95 parsed.

Build time is reduced to 58s.

## conclusion 

There is some variation in time to build, but improvements are roughly that build times are improved by 30% (86 seconds to 58). The greatest improvements came from reducing features.

## limitations

I wasn't able to update the webpack plugin or monaco. Possibly a webpack version update in Embroider is needed.

jenweber:
Closes #1336 

Also reduces monaco dependency bundle size by about a third.

To test - run it locally. Make changes in the code editor. Confirm that it works how you would expect.

### DIFF
--- a/packages/cardhost/app/components/code-editor.js
+++ b/packages/cardhost/app/components/code-editor.js
@@ -1,6 +1,6 @@
 import Component from '@glimmer/component';
 import requirejsPolyfill from '@cardstack/requirejs-monaco-ember-polyfill';
-import * as monaco from 'monaco-editor';
+import * as monaco from 'monaco-editor/esm/vs/editor/editor.api.js';
 import { action } from '@ember/object';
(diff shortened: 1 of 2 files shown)


bagby:
@jenweber I ran this locally and everything works as expected, *and* the build time *seemed* faster 👍

PR #1339, opened 2020-02-12
Field position ember animated

bagby:


jenweber:
Since this predates the v2 merge, I think it makes sense to close it.

PR #1344, opened 2020-02-13
occlusion refactor

ef4:
This cleans up several things I found hard to follow. It mostly keeps the same opinions about how rules should work, with a couple small exceptions.

### DIFF
--- a/packages/core/card-document.ts
+++ b/packages/core/card-document.ts
@@ -222,7 +222,7 @@ export class CardDocument {
       doc.data.attributes!.csRealm = this.csRealm;
     }
 
-    if (this.csOriginalRealm) {
+    if (this.csOriginalRealm && this.csOriginalRealm !== this.csRealm) {
       doc.data.attributes!.csOriginalRealm = this.csOriginalRealm;
     }
 
--- a/packages/core/card.ts
+++ b/packages/core/card.ts
@@ -15,17 +15,10 @@ import flatten from 'lodash/flatten';
 import uniqBy from 'lodash/uniqBy';
 import difference from 'lodash/difference';
 import intersection from 'lodash/intersection';
-import isObjectLike from 'lodash/isObjectLike';
+import isEqual from 'lodash/isEqual';
 import * as J from 'json-typescript';
 import { CARDSTACK_PUBLIC_REALM } from './realm';
-import {
-  OcclusionFieldSets,
-  InnerOcclusionRules,
-  assertOcclusionFieldSets,
-  OcclusionRulesOrDefaults,
-  InnerOcclusionRulesOrDefaults,
-  Format,
-} from './occlusion-rules';
+import { OcclusionFieldSets, assertOcclusionFieldSets, Format, OcclusionRules } from './occlusion-rules';
 import { Container } from './container';
 import { ModuleLoader } from './module-loader';
 import { CardReader } from './card-reader';
@@ -45,7 +38,7 @@ export const apiPrefix = '/api';
 export async function makeCollection(
   cards: AddressableCard[],
   meta: ResponseMeta,
-  rules?: OcclusionRulesOrDefaults
+  rules?: OcclusionRules
 ): Promise<CollectionResourceDoc> {
   let pristineDocs = await Promise.all(cards.map(card => card.serializeAsJsonAPIDoc(rules)));
   let doc: CollectionResourceDoc = {
@@ -277,6 +270,19 @@ export class Card {
     }
   }
 
+  @Memoize()
+  private async allFieldNames(): Promise<string[]> {
+    let fields = [];
+    for (let card of [this, ...(await this.adoptionChain())]) {
+      if (card.csFields) {
+        for (let name of Object.keys(card.csFields)) {
+          fields.push(name);
+        }
+      }
+    }
+    return fields;
+  }
+
   private fieldsWithData(): string[] {
     let names: string[];
     if (this.attributes && this.relationships) {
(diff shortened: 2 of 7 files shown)


packages/core/card-document.ts
```diff
@@ -222,7 +222,7 @@ export class CardDocument {
       doc.data.attributes!.csRealm = this.csRealm;
     }
 
-    if (this.csOriginalRealm) {
+    if (this.csOriginalRealm && this.csOriginalRealm !== this.csRealm) {
```
ef4: We generally don't include csOriginalRealm when it's redundant.

packages/core/card.ts
```diff
@@ -330,25 +336,29 @@ export class Card {
...
-          this.container
-        );
-      }
-    }
-    let parent = await this.adoptsFrom();
-    if (parent) {
-      return await parent.field(name);
+    let data;
+
+    if (name === 'csAdoptsFrom') {
```
ef4: If we're going to let people treat `csAdoptsFrom` the same as the user defined fields (and I think that's a good idea), it needs to have a FieldCard defining it.

packages/core/card.ts
```diff
@@ -412,13 +463,10 @@ export class Card {
...
   }
 
-  private async serialize(
-    rules: OcclusionRulesOrDefaults,
-    included: Map<string, OcclusionRulesOrDefaults[]>
-  ): Promise<ResourceObject> {
+  private async serialize(rules: OcclusionRules, included: Map<string, OcclusionRules[]>): Promise<ResourceObject> {
     let data = Object.create(null);
 
-    if (this instanceof AddressableCard) {
```
ef4: This doesn't really work, because sometimes we're an instance of FieldCard.

(Also we were sometimes an instance of Card, but I addressed that by adding a `patch` method to `FieldCard`so that patching a FieldCard gives another FieldCard and not a Card.)

packages/core/card.ts
```diff
@@ -508,147 +561,69 @@ export class Card {
...
+    included: Map<string, OcclusionRules[]>
   ): Promise<void> {
-    let includeFields = await this.includeFieldsForField(rules, field);
-    if (includeFields === 'everything' || includeFields === 'upstream') {
-      attributes[field] = cloneDeep(data);
-    } else if (rules) {
-      if (!includeFields.length) {
-        return;
-      }
-      if (includeFields.find(i => i.name === field) && !isObjectLike(data)) {
```
ef4: `isObjectLike` here was a problem. It's not a reliable way to test for "scalar" or "basic" fields, because cards are free to represent those as POJOs too. The reliable test is `this.hasFeature('field-deserialize')`

packages/core/card.ts
```diff
@@ -508,147 +561,69 @@ export class Card {
...
-        typeof includeFields !== 'string' ? includeFields.map(i => i.includeFields || []) : []
-      );
-      let interiorRule: OcclusionRulesOrDefaults =
-        typeof includeFields !== 'string' ? { includeFields: interiorIncludeFields } : includeFields;
-      setInteriorCardFieldSet(rules, interiorRule);
-
-      if (Array.isArray(data)) {
-        attributes[field] = await Promise.all(
-          data.map(async cardValue => {
-            let interiorCard = await this.container.instantiate(
```
ef4: Instantiation here is not always right. It's better to use `this.value()`, which will respect the custom deserializer.

packages/core/card.ts
```diff
@@ -1138,6 +1125,19 @@ export class FieldCard extends Card {
     }
     return await this.patch({ data: value });
   }
+
+  async patch(otherDoc: SingleResourceDoc): Promise<FieldCard> {
```
ef4: This specialized `patch` for `FieldCard`.

packages/core/card.ts
```diff
@@ -1262,28 +1262,18 @@ function relationshipToCardId(ref: RelationshipObject): CardId | CardId[] | unde
...
   }
-  if (rules.find(i => i === 'everything')) {
-    return 'everything';
-  }
-  throw new Error(`unimplemented: mergeRules`);
-}
-
-// need to run this by Ed, this assumes that all interior cards will return
-// embedded field sets if the outer card was requested with a fieldset.
-function setInteriorCardFieldSet(rules: OcclusionRulesOrDefaults, interiorCardRules: OcclusionRulesOrDefaults): void {
```
ef4: I think the policy is fine, but the place this is handled is now `rulesForField`.

packages/hub/node-tests/card-service-test.ts
```diff
@@ -2004,7 +2004,6 @@ describe('hub/card-service', function() {
           { type: 'cards', id: vanGogh.canonicalURL },
           { type: 'cards', id: mango.canonicalURL },
         ]);
-        expect(doc).to.not.have.property('included');
```
ef4: This was a policy change. 

OcclusionRules don't distinguish between value and reference. A field is either in or out, and that doesn't depend on whether it's a value or reference. But this means that if we would have wanted the field to be present in upstream attributes, it's necessarily also present in upstream included when it's a reference.

packages/hub/node-tests/card-service-test.ts
```diff
@@ -2258,6 +2257,7 @@ describe('hub/card-service', function() {
             {
               name: 'puppies',
               includeFields: [
+                'name',
```
ef4: This was another policy change. I think before the puppy's `name` field was being included because they would get their embedded field set in addition to the things explicitly requested in the rules.  But then there's no way to occlude the embedded field set.

The new policy is that you only get the embedded fields if you don't specify any rules of your own:

```js
let doc = await daddy.serializeAsJsonAPIDoc({
    includeFieldSet: 'isolated',
    includeFields: [ 'puppies' ]
});
```

Once you start controlling the rules for puppies, that's what you get, with nothing else added:

```js
let doc = await daddy.serializeAsJsonAPIDoc({
    includeFieldSet: 'isolated',
    includeFields: [
      {
        name: 'puppies',
        includeFields: ['name']
      }
    ]
});
```

packages/hub/node-tests/card-service-test.ts
ef4: An alternative that would let us control `included` separately is to distinguish between when to include the reference in `relationships` vs when to actually include the related resource in `included`. Today those are the same thing and the occlusion rules don't let you say one or the other.

packages/core/card.ts
```diff
@@ -484,10 +537,10 @@ export class Card {
     return data;
   }
 
-  private async serializeIncluded(included: Map<string, OcclusionRulesOrDefaults[]>): Promise<ResourceObject[]> {
+  private async serializeIncluded(included: Map<string, OcclusionRules[]>): Promise<ResourceObject[]> {
```
ef4: The `serializeIncluded` method is re-serializing the same cards repeatedly, and then spending effort uniquing them back out. I didn't get around to improving that yet.

PR #1360, opened 2020-02-18
Remove Save Button and autosave instead

### LINKED ISSUE #1280: Handle route changes during autosave
Following #1275, there is one condition where saving gets interrupted and does not resume.

To close this ticket:
- discuss and decide what we want the behavior to be (get input from Chris)
- implement it
- write a test

To reproduce:
- Make a change to a previously-saved card
- In less than 5 seconds, navigate to a new route (click layout, schema, etc)
- Note that the button continues to display "Save"
- Refresh (or check network requests) and you can see that the last changes before switching routes were not persisted

Some possible options
- save on willDestroy for the button, if the record is dirty
- change nothing
- somehow continue the save loop independent of route
- ????

### LINKED ISSUE #1313: Autosave status
Autosave status is displayed as the first item in the context popup menu. It is not actionable. 
image
If there is an error, context menu button changes border color as a secondary indicator of an error 
image

XD File: https://drive.google.com/open?id=1SrH4cxeUt9Jxg0AwHCrvJy-LMwb39e8z

jenweber:
Closes #1280 
Closes #1313 

## Added:

- Save state is reflected in the context menu, together with a timestamp
- Save state uses dialog icons
- All modes where the user can change the card now have a test added that enables autosave to make sure it's working.

## Changed

- Saving is handled by the card-renderer-header, not the save button
- Some tests to accommodate for the save button being gone. Instead, use `[data-test-card-is-dirty="no"]` to check saved state.

## Removed

- JavaScript for the `save-button` component. Clicking save won't do anything. It will be removed in a followup PR.
- Save Button

## How to review

Click through the app. Does the top edge look right on each page? Make changes to a card. Are the changes saved after about 5 seconds? Click on the context menu to see the saved state change.

## Notes

Whenever `isDirty` becomes true, the autosave feature waits 5 seconds and then saves. Similar to how the save button worked, saving is debounced so that additional typing won't send out many requests. A {{did-update}} on the `...` button in `card-renderer-header` is what does the saving, so this element _must_ be present or autosaving will not work.

Oddly enough, if you do `[data-test-something=false]` that causes the test selector to not be rendered, so we have to do `data-test-something={{if isTrue "yes" "no"}}`.

### Testing

By default, tests do _not_ use the autosave functionality unless they specifically look up the autosave service and set the `autosaveDisabled` property:

```js
this.owner.lookup('autosave').autosaveDisabled = false;
```

If it's important to save during a test, use the `saveCard` method directly within the test:

```
await saveCard();
```


packages/test-support/addon-test-support/card-ui-helpers.js
```diff
@@ -81,8 +80,10 @@ export async function createCards(args) {
 }
 
 export async function saveCard() {
-  await click(`[data-test-card-save-btn]`);
-  await waitFor(`[data-test-card-save-btn].saved`, { timeout });
+  getContext()
```
bagby: 👍

PR #1362, opened 2020-02-18
Drag drop bugs

bagby:
fixes a bug where the dragged field remains selected after dropping, cleans up some of the event handling stuff.


jenweber:
Tested locally, seems like an improvement!

PR #1363, opened 2020-02-18
Make the full mode label a link

### LINKED ISSUE #1355: Make mode exiting buttons a single button instead of segmented control
All mode exiting buttons are currently segmented controls. The active segment is apparent only on hover: 
image
Instead let's make the entire button a single active control. 
image

The images above are for reference only - the same applies to all mode buttons:
- Edit mode
- Schema mode
- Themer mode
- Preview

burieberry:
Closes #1355

@dnaidionov can you confirm this is what we want?

mode-label

On hover:
Color darkens and tooltip appears that says "Link to X mode"

mode-label-hover

Update:
It looks much better with save button gone:

mode-label


dnaidionov:
yes

PR #1364, opened 2020-02-19
prevent dupe field previews

bagby:
also fixes a bug where we weren't setting `this.draggable.dragging` to `false` when clicking on a dragged card.

### DIFF
--- a/packages/cardhost/app/components/card-manipulator.js
+++ b/packages/cardhost/app/components/card-manipulator.js
@@ -312,19 +312,14 @@ export default class CardManipulator extends Component {
 
       self.draggable.setDragging(true);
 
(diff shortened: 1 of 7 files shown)


jenweber:
I'm still seeing some odd behavior, though I don't see the duplicates problem anymore. I sent you a video of it to see if it is expected.

jenweber:
Here are the things I do for all drag & drop PR reviews:
- Try dropping in multiple fields
- Rearrange field order
- delete a field
- drag a field offscreen
- Grab a field and wave my mouse around wildly
- Try placing a field at the top and bottom
- try rearranging fields to put them at the top/bottom

PR #1365, opened 2020-02-19
click outside to close context menu #1361

### LINKED ISSUE #1361: When clicking outside of context menu it should close

jenweber:
Closes #1361 

This PR implements a `{{click-outside}}` modifier and uses it for the left edge and the `...` context menu. Use it for things where you want to trigger an action whenever someone clicks outside of that element. It takes an action/function and an `ignore` string selector. Clicking on something passed to `ignore` will be ignored by the click handlers.

```handlebars
<button class="show-nav" {{on "click" this.toggleNav}}>
  Open
</button>

{{#if this.showNav}}
  <nav {{click-outside this.closeNav ignore=".show-nav"}}>
    ....
  </nav>
{{/if}}

<div>
clicking here will close the menu
</div>
```

## Added:
- click-outside modifier
- docs for the new modifier (in the modifier file itself)
- tests!

## Changed:
- the left edge's "click outside to close" behavior is removed and handled via the modifier instead
- clicking outside the context menu closes it

## Needs feedback

Is there a better way to handle events here? The `ignore` property exists because without it, the menu would rapidly open and then close. It seems like the chain of events was this:

- user clicks the '...' button, which triggers an action to set a boolean
- menu is inserted into the DOM when boolean becomes true
- modifier initializes and the click handlers register
- the original `...` click event is caught by the modifier click handlers
- the menu closes

## How to test:
Click on the `...` to open the context menu. Click on something in the context menu and it should remain open (including the non-interactive headers like the last-saved timestamp). Click somewhere outside the menu and it closes.

Repeat for the left edge.


jenweber:
@bagby @ef4 I especially would like your review of the click handler business. It seems weird and maybe there's a better way to do this than via the `ignore` property. Thanks!

PR #1366, opened 2020-02-20
Add ability to press Enter to create card in dialog

bagby:


bagby:
@burieberry simplified it based on your feedback.

PR #1368, opened 2020-02-20
Bugfix - interrupt autosaving when changes are made

jenweber:
Without this, as you make changes to a Card, the autosave will complete and the changes get wiped out to match server state. This will help reduce that occurrence. Eventually we need to depend on something other than isDirty.


jenweber:
It is a mystery to me why this works, but @bagby this is perfect. Thanks!

PR #1372, opened 2020-02-24
Add percy token

alexspeller:
Added percy token to ember build and added to repo secrets so that it will run on github actions build


alexspeller:
@jenweber I think this is what you meant now :)

PR #1375, opened 2020-02-24
Run tests before deploying

### LINKED ISSUE #1371: Deploy workflow doesn't wait for CI tests
It looks like test and deploy are setup in github actions as completely separate workflows, so we don't wait for tests to succeed before deploying.

cc @jenweber @alexspeller

jenweber:
Closes #1371 

This ought to run the tests before deploying. 

It will unfortunately run tests for pushes to master twice, but the conditional syntaxes don't seem to work as documented.


alexspeller:
@jenweber it looks good but github is saying there's a yaml error - I can't see the error though

jenweber:
I tried many combinations using env, inputs, etc. and the action did not ever seem to recognize the custom vars. I went back to what I had earlier.

jenweber:
After a little more research, I realized that I probably can't do this as-written. Actions and workflow files are not interchangeable. Looks like we'll have to make a really long file.

jenweber:
@alexspeller would you have another look at this? I combined the test and deploy scripts into one file so that we could make use of the "needs" key. I added an "if" statement to the deploy steps to check if the branch is master.

Actions themselves seem to be way too granular for what we need.

I am a little fuzzy on whether each job needs to do its own installing. Any ideas?

PR #1378, opened 2020-02-24
Prevent overscroll, closes #1358

### LINKED ISSUE #1358: Body of the page background color should match card background on all pages
If you scroll up or down beyond the visible page, you can get a peek of page background color, and it's not always the same. Page background color should always be the same as card's background color. We should inject the background color of currently isolated card/mode into the page. 
e.g.:
Edit mode - backgrounds match
edit-mode-page-bg
View mode - page background is different from card background and is peeking when scrolling past the card (dark bars at the top/bottom)
view-mode-page-bg

jenweber:
Closes #1358 

Please see the ticket for the goal and the reason I chose this approach.

How to test:
Open the builder in multiple browsers. Using the touchpad, try scrolling too far down or up a page. You should not see exposed edges like in the gif Dmitry put in the ticket.

### DIFF
--- a/packages/cardhost/app/styles/app.css
+++ b/packages/cardhost/app/styles/app.css
@@ -117,6 +117,11 @@ body {
   line-height: 1.2;
 }
 
+/* Prevent overscroll bounce across all browsers. Should use `overscroll-behavior: none` once there is better cross-browser support. */
+html {
+  overflow: hidden;
+}
+
 a {
   color: inherit;
   text-decoration: none;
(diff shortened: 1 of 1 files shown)


burieberry:
@jenweber I tested this on a couple of browsers and it looks good. If there are benefits to using `overscroll-behavior`, we can add this to use it where it's supported:

```
@supports (overscroll-behavior: none) {
  html {
    overflow: initial;
  }

  body {
    overscroll-behavior: none;
  }
}
```

jenweber:
Cool, I will add this! Thanks!

PR #1379, opened 2020-02-24
Edit mode transitions and card transitions from mode to mode

### LINKED ISSUE #1348: Animate transitions when switching between layout and fields modes
- [x] When switching from Fields to Layout, on the top edge "screen size" and "preview" buttons are sliding up off screen.
- [x] When switching from Layout to Fields, on the top edge "screen size" and "preview" buttons are sliding in down from the top of the screen.
- [ ] When switching between Layout and Fields, animate Right Edge in a similar way to #992
- [x] When switching between Layout and Fields, animate card resize in a similar way to #992 (more advanced animations would come later)

burieberry:
Most of #1348 (except the right-edge, I'll push it in a separate PR)

**What to test:**
- On edit mode, when toggling between Fields and Layout, the transitions checked in #1348 happen smoothly
- When moving between different modes, the card transitions smoothly (had to do this here too because edit mode changes was messing up transitions other modes)
- Previous animations continue working (ie. card-view border animation, schema drag & drop)


habdelra:
@burieberry I think you'll need to redo this in new-hub. things are so different there that your animations are not working across the card async boundaries. sorry about that :-(

habdelra:
(specifically when we animate to the schema mode in new-hub, the `{{animated-value}}`'s watch isn't kicked off correctly and the ember animated inline styles hide all the fields.)

habdelra:
I'll play around with it a bit more, and I'll push my work to a branch if you want to explore. I think i'll need to give you a primer on now new-hub works first, though, as it might be a little disorienting without some guidance on the new API.

burieberry:
@habdelra no worries, in that case I can redo it. it would be great to get a primer on how new hub works

packages/cardhost/app/templates/components/card-renderer.hbs
```diff
@@ -8,78 +8,83 @@ Ember Concurrency into the rendering of the card component here.
...
+                @setContextMenu={{@setContextMenu}}
               />
-            </div>
-          </MaybeAnimate>
-        </div>
-      </section>
+            {{/animated-value}}
+          {{else}}
+            <AnimatedContainer>
+              {{#animated-value @card watch=this.mode use=this.headerAnimation group="header"}}
```
habdelra: @burieberry `watch=this.mode` does actually do the thing you think it does. you can only watch properties on the item being animated, `@card` in this case. the `watch` field actually indicates the property on `@card` that you want to use to trigger the animation. because you are evaluating `this.mode` the `@card`property that you will actually end up watching is `@card.view` or `@card.edit` or `@card.schema`, etc.. which are nonsensical. `this.mode` will not trigger the animation.

packages/cardhost/app/templates/components/card-renderer.hbs
```diff
@@ -8,78 +8,83 @@ Ember Concurrency into the rendering of the card component here.
...
+                  @buttonText="Done Editing"
+                  @mode={{this.mode}}
+                  @contextMenuOpen={{@contextMenuOpen}}
+                  @setContextMenu={{@setContextMenu}}
+                />
+              {{/animated-value}}
+            </AnimatedContainer>
+          {{/if}}
+          <AnimatedContainer class="card-renderer-isolated--content">
+            {{#animated-value @card watch=(if (eq this.mode "view") "isSelected" this.mode) use=this.cardTransition group="body"}}
```
habdelra: same as the above in terms of watching `this.mode`, you actually end up watching a property on `@card` that `this.mode` evaluates to: `@card.view` or `@card.edit` or `@card.schema`, etc

PR #1399, opened 2020-03-02
Update background colors for the Themer page

### LINKED ISSUE #1261: Themer mode screen
## Top edge:
See #1306

## Right edge: 
Right Edge is hidden

## Code Editor
The code editor occupies the right half of the screen. 

## Card header: 
The card header is not shown. Only the contents of the card as seen by the user in **view mode** are displayed. 

## Background
The background on which the card is displayed is the color from our palette closest to #5D5C6F. 
The background of the code editor side is the color from our palette closest to #000000 with 77% opacity. 

image

The top edge should have a transparent background. 

## Transitions: 
- For transitions when switching between this and Edit Layout screens see #1316 
- For transitions when switching between this and Preview screens see Preview screen ticket (#1262)

### LINKED ISSUE #1269: Show drag handle focus
Here, I temporarily removed the outline from a button. This is very bad for a11y, but it was totally misaligned and weird.

https://github.com/cardstack/cardstack/pull/1270/files#diff-c71bc13c08b22f8ff44b2c06f50d9aa1R87

To close this ticket, make the handle itself (the little dots) the button that gets the focus when you interact with it. Restore the default blue outline by removing `outline: none` hack.

jenweber:
Closes #1261 

## Changed
- lighter background that the card is on top of
- Darker editor panel
- Small fixes to some invalid aria attributes. I guess an empty string is not ok for `aria-expanded`. Neither is `true` or `false`. It needs to be a quoted string for `"true"` and `"false"`

## Needs work
- The color contrast of the buttons has problems after making these changes. See below.

@dnaidionov could you have a read through the checklist in #1261 and see if the substitutions I made are ok? Also, I would like some direction about what to do with the top edge controls. Now that I made the background lighter, the S-M-L buttons are too light. Should I make the inactive state white for all top edges? There are screenshots of the color contrast checks in the issue.


jenweber:
Sorry, had the wrong issue number initially. This goes with #1261

jenweber:
One more thing @dnaidionov, is this overlap effect intended for the transparent header? Should we do something else when the card is scrolled?

Screen Shot 2020-03-02 at 5 37 23 PM

dnaidionov:
> @dnaidionov could you have a read through the checklist in #1261 and see if the substitutions I made are ok?

all substitutions are ok.

> Also, I would like some direction about what to do with the top edge controls. Now that I made the background lighter, the S-M-L buttons are too light. Should I make the inactive state white for all top edges?

Now, this is a bit of a problem - white is reserved for hover state, so we need another color for inactive control. I'll ask Gardella to take a look and suggest color combinations

dnaidionov:
> One more thing @dnaidionov, is this overlap effect intended for the transparent header? Should we do something else when the card is scrolled?

hm... maybe completely transparent header is not such a good idea, and we would be better with a scrim. I'll ask Gardella to look at this too

dnaidionov:
@jenweber here's note from Gardella: 

> We can find a suitable, lighter replacement for inactive grey, but I don't think we need hover states for every button that can be interacted with
> Themer mode:  I don't think we need hover states for the size control (S,M,L) or the panel orientation control (dock left, dock bottom)
> Also we only need to pass the "Graphical Objects and User Interface Components" test since we are not displaying any text in this color, only icons.  I consider the size control letters to be icons, not text - user is not reading a sentences in this case
> Using the WebAIM contrast checker this color passes for the graphical objects and user interface components test: #B1B1B9
> This is ONLY for size icons and docking controls while in themer mode.

I agree about SML being GUI component, not text. In that case we don't need to worry about hover state at all since current tests are passing.

jenweber:
Awesome, thanks for looking into this!

jenweber:
I think I still need to know what to do about the transparent header. @dnaidionov do you think I should just keep it as-is (transparent) and merge? Then we see if we like it or not?

dnaidionov:
Keep it as-is for now, we will decide later

jenweber:
The color Gardella suggested is nearly identical to `--ch-light-op50` so I will leave it as-is.

PR #1401, opened 2020-03-02
Remove hub scaffolding and add logic to create meta realm and default realm (as well as seed realm cards in test env)

habdelra:
@ef4 this is ready for review now. thanks!

The node-tests continue to use the `first-ephemeral-realm`, which is now seeded by the test environment. The ember tests actually rely on a fully running hub, and as such they now use the new default realm. As well as, the meta realm is created in both the test env setup and as part of the server startup. As part of this we have removed the hub scaffolding completely now.


habdelra:
(I think we can safely ignore the failing percy test)

PR #1403, opened 2020-03-03
Add git to new hub

alexspeller:
Currently basic writer and indexer implemented.

Next step is to implement remote repo I think

### DIFF
--- a/cards/git-realm/lib/git-settings.ts
+++ b/cards/git-realm/lib/git-settings.ts
@@ -21,7 +21,6 @@ export function extractSettings(realmCard: AddressableCard): IndexerSettings {
   if (typeof attributes.repo == 'string') {
     repo = attributes.repo;
   } else {
-    console.log(realmCard);
     throw new CardstackError('You must provide a repo attribute when instantiating the git realm card');
   }
 
--- a/cards/git-realm/lib/git-settings.ts
+++ b/cards/git-realm/lib/git-settings.ts
@@ -24,8 +24,8 @@ export function extractSettings(realmCard: AddressableCard): IndexerSettings {
     throw new CardstackError('You must provide a repo attribute when instantiating the git realm card');
   }
 
-  let branchPrefix = (attributes.branch ?? '') as string;
-  let basePath = (attributes.basePath as string) ?? undefined;
+  let branchPrefix = await realmCard.value('branch');
+  let basePath = await realmCard.value('basePath');
 
   // TODO: handle remote config
   // if (typeof attributes.remote == 'object' && attributes.remote.url && attributes.remote.cacheDir) {
--- a/cards/git-realm/lib/git-settings.ts
+++ b/cards/git-realm/lib/git-settings.ts
@@ -9,23 +9,15 @@ import { RemoteConfig } from './git';
 import CardstackError from '@cardstack/core/error';
 import { AddressableCard } from '@cardstack/core/card';
 
-export function extractSettings(realmCard: AddressableCard): IndexerSettings {
-  let attributes = realmCard.attributes;
+export async function extractSettings(realmCard: AddressableCard): Promise<IndexerSettings> {
+  let repo = await realmCard.value('repo');
 
-  if (!attributes) {
-    throw new CardstackError('You must provide attributes to configure the git realm card');
-  }
-
-  let repo;
-
-  if (typeof attributes.repo == 'string') {
-    repo = await realmCard.value('repo');
-  } else {
+  if (typeof repo != 'string') {
     throw new CardstackError('You must provide a repo attribute when instantiating the git realm card');
   }
 
-  let branchPrefix = await realmCard.value('branch');
-  let basePath = await realmCard.value('basePath');
+  let branchPrefix = (await realmCard.value('branch')) ?? '';
+  let basePath = (await realmCard.value('basePath')) ?? undefined;
 
   // TODO: handle remote config
   // if (typeof attributes.remote == 'object' && attributes.remote.url && attributes.remote.cacheDir) {
--- a/cards/git-realm/lib/git-settings.ts
+++ b/cards/git-realm/lib/git-settings.ts
@@ -19,7 +19,7 @@ export function extractSettings(realmCard: AddressableCard): IndexerSettings {
   let repo;
 
   if (typeof attributes.repo == 'string') {
-    repo = attributes.repo;
+    repo = await realmCard.value('repo');
   } else {
     throw new CardstackError('You must provide a repo attribute when instantiating the git realm card');
   }
--- a/cards/git-realm/lib/git-settings.ts
+++ b/cards/git-realm/lib/git-settings.ts
@@ -1,0 +1,44 @@
+interface IndexerSettings {
+  repo: string;
+  basePath: string;
+  branchPrefix: string;
+  remote?: RemoteConfig;
+}
+import { RemoteConfig } from './git';
+
+import CardstackError from '@cardstack/core/error';
+import { AddressableCard } from '@cardstack/core/card';
+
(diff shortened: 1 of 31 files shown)


habdelra:
The tests are green, so I say go for it. 👍

cards/git-realm/lib/git-settings.ts
```diff
@@ -0,0 +1,43 @@
...
+  branchPrefix: string;
+  remote?: RemoteConfig;
+}
+import { RemoteConfig } from './git';
+
+import CardstackError from '@cardstack/core/error';
+import { AddressableCard } from '@cardstack/core/card';
+
+export function extractSettings(realmCard: AddressableCard): IndexerSettings {
+  let attributes = realmCard.attributes;
```
habdelra: @alexspeller AddressableCard.attributes is private. you'll need to use AddressableCard.value('repo'), etc to get values off the card.

cards/git-realm/lib/git-settings.ts
```diff
@@ -0,0 +1,43 @@
...
+
+  if (!attributes) {
+    throw new CardstackError('You must provide attributes to configure the git realm card');
+  }
+
+  let repo;
+
+  if (typeof attributes.repo == 'string') {
+    repo = attributes.repo;
```
habdelra:
```suggestion
    repo = await realmCard.value('repo');
```

cards/git-realm/lib/git-settings.ts
```diff
@@ -0,0 +1,43 @@
...
+  } else {
+    throw new CardstackError('You must provide a repo attribute when instantiating the git realm card');
+  }
+
+  let branchPrefix = (attributes.branch ?? '') as string;
+  let basePath = (attributes.basePath as string) ?? undefined;
```
habdelra:
```suggestion
  let branchPrefix = await realmCard.value('branch');
  let basePath = await realmCard.value('basePath');
```

cards/git-realm/writer.ts
```diff
@@ -0,0 +1,401 @@
...
+//     this.ephemeralStorage.store(null, id, this.realmCard.csId, version);
+//   }
+// }
+
+import stringify from 'fast-json-stable-stringify';
+
+const defaultBranch = 'master';
+
+function getType(document: UpstreamDocument) {
+  return document.jsonapi.data.type;
```
habdelra: you can hardcode this to 'cards'. the type will _always_ be `cards`.

cards/git-realm/writer.ts
```diff
@@ -0,0 +1,401 @@
...
+}
+
+function getId(document: UpstreamDocument) {
+  return document.jsonapi.data.id;
```
habdelra: let's use document.canonicalURL()--it's simpler logic wise and more straight forward/ consistent with what we are doing else where.
```suggestion
  return document.canonicalURL;
```

cards/git-realm/writer.ts
```diff
@@ -0,0 +1,401 @@
...
+  githereumConfig: todo;
+  githereum: todo;
+  _githereumPromise?: Promise<todo>;
+
+  constructor(private realmCard: AddressableCard) {
+    let hostname = os.hostname();
+    this.myName = `PID${process.pid} on ${hostname}`;
+    this.myEmail = `${os.userInfo().username}@${hostname}`;
+
+    let settings = extractSettings(realmCard);
```
habdelra: this will be async after you start using the public API on `AddressableCard` which means you wont be able to set this in the contructor. 

Our container actually uses async instantiation, so we do have an async intantiation hook for you to use for this kind of stuff: `async ready(){ ...} `

as part of the way the container will instantiate these features, it will await the `ready()` method in your writer immedately after instantiating the class (and providing injections as well into your instantiation). you put any async that you need as part of your writer there. So you'll need to have something like:
```ts
async ready(): Promise<void> {
  let settings = await extractSetting(realmCard);
  this.repoPath = settings.repo;
  this.basePath = settings.basePath;
  this.branchPrefix = settings.branchPrefix;
  this.remote = settings.remote;
}
```

cards/git-realm/writer.ts
```diff
@@ -0,0 +1,401 @@
...
+
+  async create(session: Session, document: UpstreamDocument, upstreamId: UpstreamIdentity | null) {
+    let createId = upstreamId || getId(document);
+    let id: string | undefined;
+
+    if (createId) {
+      id = upstreamIdToString(createId);
+    }
+
+    let type = getType(document);
```
habdelra: type is _always_ `"cards"`. let's just hardcode this.

packages/core/card.ts
```diff
@@ -95,7 +95,9 @@ export class Card {
   // value for the field).
   readonly sourceCard: Card | undefined;
 
-  private readonly attributes: SingleResourceDoc['data']['attributes'];
+  // TODO: should this remain private? I needed to access the attributes in the
+  // git writer but it was private before
+  readonly attributes: SingleResourceDoc['data']['attributes'];
```
habdelra: Noooooo. please do not do this! this is definitely not what we want (and also explains why you have no linting error.

habdelra: You absolute need to use the `Card.value()` method to get values out of your card--this is wrong for so many reasons.

packages/core/card.ts
```diff
@@ -249,6 +251,10 @@ export class Card {
     return cardDocumentFromJsonAPI(jsonapi);
   }
 
+  get cardDir(): string | undefined {
```
habdelra: Is this the directory there the module lives for the card? or is this something specific to just cards that come from the git realm. If this is purely a git realm need, then let's not do this here.

cards/git-realm/writer.ts
habdelra: Doh, I think I can you a bad suggestion here: I was mistakingly thinking `document` was an instance of `Card`. I see it's not. what you had before is ok.

habdelra:
i'm not sure if you saw the notifcation--my suggestion around the `canonicalURL` was incorrect. I was assuming you were working with an instance of `Card`, but I just now noticed it was an `UpstreamDocument`, in which case I don't think you have a `canonicalURL` getter on it. What you had before was ok. sorry!

packages/core/card.ts
alexspeller: > Is this the directory where the module lives for the card

Yes, this was at @ef4 's suggestion, so that cards are loaded from the local fs if they are present there, and the data is stored on the card - this allows using of npm modules in cards which was not working previously.

packages/hub/module-service.ts
```diff
@@ -20,10 +20,17 @@ export class ModuleService implements ModuleLoader {
...
     // collision resistance
-    let hash = createHash('md5');
-    hash.update(stringify(await toIdempotentDoc(card)));
-    let cardDir = join(cardFilesCache, hash.digest('hex'));
-    await this.cachedWriteCard(card, cardDir);
+
+    let cardDir;
+
+    if (card.cardDir) {
+      cardDir = card.cardDir;
```
ef4: This is the only legitimate place where anybody should access `card.cardDir`. It would be nice to limit it more clearly in the type system. A challenge is that this state needs to persist in the database and persist across card cloning, so it's not sufficient to use something like a WeakMap on the Card instances.

Given that many developers are likely to encounter the `Card` class in the future and none of them should ever really see or think about `cardDir`, it's probably worth hiding it better. We could just do a type-unsafe cast here at the one spot where we legitimately access it:

```js
cardDir = (card as any)?.meta?.cardDir
```

The downside is that this won't generate a compile error if `Card` changes to not have a private `meta` anymore. But as long as we cover it with a test I'm fine with that.

ef4: I filed a separate issue to track this since this PR is merged already https://github.com/cardstack/cardstack/issues/1404

PR #1406, opened 2020-03-04
Add remote writing and indexing with some initial tests

alexspeller:


habdelra:
Is the `remoteCacheDir` a local directory that the remote repo is cloned into?

alexspeller:
> Is the remoteCacheDir a local directory that the remote repo is cloned into?

Yes, that's correct, it's the same as the cacheDir config from the previous implementation

packages/hub/node-tests/git/writer-remote-test.ts
```diff
@@ -19,342 +26,355 @@
...
+import { dir as mkTmpDir, DirectoryResult } from 'tmp-promise';
+
+const repoRealm = `${myOrigin}/api/realms/test-git-repo`;
+
+async function resetRemote() {
+  let tmpDir = await mkTmpDir({ unsafeCleanup: true });
+  let root = tmpDir.path;
+
+  let tempRepo = await makeRepo(root, {
+    'contents/events/event-1.json': JSON.stringify(
```
habdelra: in general I highly suggest using `CardDocument` to create test cards--this way you can be sure the card json is legit--like it has cs-fields, schema definitions, and adoption specification, all of which you are missing in this test data. that was its original purpose (it's branched out since then). There are lots of examples in hub/node-tests/card-service-tests.ts. the simplest form of it will create the field schema for you:
```ts
let doc = cardDocument().withAutoAttributes({ title: 'This is a test event', 'published-date': '2018-09-25' });
let jsonapi = doc.jsonapi;
```

alexspeller: Yeah definitely meaning to update to that, I have a _lot_ of cleanup to do here from the old tests

Issue #1341, opened 2020-02-12
When selecting a field from card catalog that is below the fold, the catalog scrolls up and ghost image of the card is displayed in the wrong place


dnaidionov:
This behavior happens when selecting a field with a mouse click. 
See how the catalog scrolls up by itself when picking a field card that is below the fold. Moreover, the shadow image of the selected field is now displayed over the other scrolled fields, and remains there even after the field has been released over the drop zone. 
![new field bug 2](https://user-images.githubusercontent.com/1836666/74384255-e3d91080-4dbe-11ea-8840-242ac7561452.gif)

PR #1407, opened 2020-03-04
Fix ghost field when dragging from bottom of list

bagby:
This *mostly* addresses #1341, it attempts to mitigate the `mousedown` -> `mouseup` -> `click` sequence by introducing a short debounce for the main drag action (which is now a `restartableTask`).

It can still be "broken" by clicking for longer than the debounce time, i.e. holding down the mouse button for a period of time without moving the mouse, then letting go of the button. In reality, no one really clicks like that. but it's still a thing. That said I would be curious to know if anyone can think of a better approach.


bagby:
> What is causing that scroll-to-top behavior? Can we turn that off?

I believe it's because `ember-animated` is re-rendering the fields list when it shouldn't, it would require more investigation.

PR #1408, opened 2020-03-04
New cards show a loading animation while they save 1388

jenweber:
The goal of this PR is to show a "loading" screen in the time between clicking the button to create a blank card and being brought to the schema or edit view. It is implemented as a component used in `application.hbs` and a service because its animation behavior should not be cut off during route transitions. There also should not be multiple active overlays at any given time.

## Added
- An `overlays` service that controls which overlays/modals are visible.
- A `loading` component, referenced from `application.hbs`
- fade animation for the loading component
- tests

## Changed
- Some small testing fixes
- unrelated regression in the context menu positioning (misaligned in view mode)

## How to test

To see what this animation will look like in practice, wrap the `save` and transition in a `setTimeout` in `card-name-dialog.js`. Saving to GitHub takes about 3 seconds.

```js
    setTimeout(() => {
      this.overlays.reset();
      if (adoptedFrom) {
        this.router.transitionTo('cards.card.edit.fields', newCard);
      } else {
        this.router.transitionTo('cards.card.edit.fields.schema', newCard);
      }
    }, 3000);
```

The mockup I was working from:

Screen Shot 2020-03-04 at 2 56 48 PM


jenweber:
@burieberry @bagby this is ready for review now.

The tricky thing about this feature is that we will not see it in our future local development because saves are so fast, so please keep an eye out for regressions in Percy screenshots and such moving forward.

jenweber:
@habdelra want to have a look at this and see how it meshes with new-hub? I am happy to help make it work there too.

habdelra:
it looks mostly plainless. I can shepard this into new-hub after its merged.

habdelra:
👍 we've colocated all the templates in new-hub--it will definitely make the new-hub merges go easier to do that

jenweber:
Ok, I'll move those files!

jenweber:
I see, they were already in the co-located spot. I updated with `master` and I think this is good to go now.

PR #1424, opened 2020-03-10
bugfix - loading page never resolved for new card

jenweber:
Whenever you created a new card in a production environment, it would just hang with the loading icon spinning, even on a successful server response. I am still a little fuzzy on what `yield` should have done here/why my previous code din't work. My understanding was that for ember-concurrency, `yield` was like `await`, but perhaps my `.catch` caused a problem. So, I moved it into a `.then` which is more canonical anyway.

You can see this issue in our .space stack currently.

If we need to get it working immediately, we can comment out the `this.overlays` lines in this same file and it will have the old behavior.

### DIFF
--- a/packages/cardhost/app/components/card-name-dialog.js
+++ b/packages/cardhost/app/components/card-name-dialog.js
@@ -54,17 +54,21 @@ export default class CardNameDialog extends Component {
 
     if (environment !== 'test') {
       this.overlays.setOverlayState('showLoading', true);
-      yield newCard.save().catch(() => {
-        // if there's a problem saving, go back to showing the dialog
-        this.overlays.setOverlayState('showLoading', false);
-        return;
-      });
-    }
-    this.overlays.reset();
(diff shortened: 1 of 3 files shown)


dnaidionov:
@jenweber Here's more - if you refresh the page after that, and try to create a new card, loading screen doesn't come up at all - site reverts to the old behavior where there's no indication of what it is doing. It takes hard reset and clear cache to display the waiting indicator again (but then it spins forever)

jenweber:
That is super weird. I'm not doing anything with local browser storage for the loading screen. We are still saving the IDs for showing "recent cards" but that's unrelated.

jenweber:
This is super strange behavior. Will looked at it for a bit and discovered that properties weren't rerendering as they should. Also timeout doesn't work right in this task.

I pushed a commit that simplifies some things, and would like to give this a try in the .space stack, since the bug is only really present in a deployed context.

PR #1430, opened 2020-03-11
Adding files realm type

ef4:
This adds a files-realm card that lets you store cards in a comfortable-to-manually-edit format on disk. It provides file watching so your edits are reflected immediately in the hub.


ef4:
All CI passed except percy.

PR #1435, opened 2020-03-13
Added relative image reference field card

habdelra:
to use for URL's that are relative paths (and thus not serializable as `URL` objects)


habdelra:
TODO add test

PR #1465, opened 2020-03-19
add adopt to the card context menu #1461

### LINKED ISSUE #1461: Create an "adopt from this card" button on all cards
When I was making the video demo, there was no easy way to copy a featured card. I needed to modify the URL directly to go to the adopt endpoint.

We should add a button on each card that lets you do this. I think this is in a mockup somewhere.

jenweber:
Closes #1461 

This PR adds an "adopt" option to the context menu, plus tests.

@habdelra I tagged you for review just in case there is anything missing from what you would have done.

Screen Shot 2020-03-19 at 4 57 04 PM

### DIFF
--- a/packages/cardhost/tests/acceptance/card-mode-navigation-test.js
+++ b/packages/cardhost/tests/acceptance/card-mode-navigation-test.js
@@ -5,7 +5,13 @@ import Fixtures from '../helpers/fixtures';
 import { login } from '../helpers/login';
 import { percySnapshot } from 'ember-percy';
 import { cardDocument } from '@cardstack/core/card-document';
-import { waitForCardLoad, waitForSchemaViewToLoad, encodeColons, waitForTestsToEnd, setCardName } from '../helpers/card-ui-helpers';
+import {
+  waitForCardLoad,
+  waitForSchemaViewToLoad,
+  encodeColons,
+  waitForTestsToEnd,
+  setCardName,
+} from '../helpers/card-ui-helpers';
 
 const csRealm = `http://localhost:3000/api/realms/default`;
 const testCard = cardDocument().withAutoAttributes({
--- a/packages/cardhost/tests/acceptance/card-mode-navigation-test.js
+++ b/packages/cardhost/tests/acceptance/card-mode-navigation-test.js
@@ -5,7 +5,7 @@ import Fixtures from '../helpers/fixtures';
 import { login } from '../helpers/login';
 import { percySnapshot } from 'ember-percy';
 import { cardDocument } from '@cardstack/core/card-document';
-import { waitForCardLoad, waitForSchemaViewToLoad, encodeColons, waitForTestsToEnd } from '../helpers/card-ui-helpers';
+import { waitForCardLoad, waitForSchemaViewToLoad, encodeColons, waitForTestsToEnd, setCardName } from '../helpers/card-ui-helpers';
 
 const csRealm = `http://localhost:3000/api/realms/default`;
 const testCard = cardDocument().withAutoAttributes({
@@ -67,6 +67,21 @@ module('Acceptance | card mode navigation', function(hooks) {
     await waitFor('[data-test-context-menu-button]');
     await click('[data-test-context-menu-button]');
     assert.dom('[data-test-context-menu-items] .checked').hasText('View');
+  });
+
+  test('can use the context menu to adopt from a card', async function(assert) {
+    await visit(`/cards/${cardPath}/edit/fields`);
+    await waitForCardLoad();
+
+    assert.equal(encodeColons(currentURL()), `/cards/${cardPath}/edit/fields`);
+    await click('[data-test-context-menu-button]');
+
+    await click('[data-test-context-adopt]');
+    await waitFor('[data-test-card-name]');
+    let adoptedCardName = 'Adopted Millennial Puppies';
+    await setCardName(adoptedCardName);
+    assert.ok(/^\/cards\/.*\/edit\/fields$/.test(currentURL()), 'URL is correct');
+    assert.dom('.card-renderer-isolated--header-title').hasText(adoptedCardName);
   });
 
   test('clicking outside the context menu closes it', async function(assert) {
--- a/packages/cardhost/tests/acceptance/card-mode-navigation-test.js
+++ b/packages/cardhost/tests/acceptance/card-mode-navigation-test.js
@@ -5,13 +5,7 @@ import Fixtures from '../helpers/fixtures';
 import { login } from '../helpers/login';
 import { percySnapshot } from 'ember-percy';
 import { cardDocument } from '@cardstack/core/card-document';
-import {
-  waitForCardLoad,
-  waitForSchemaViewToLoad,
-  encodeColons,
-  waitForTestsToEnd,
-  setCardName,
-} from '../helpers/card-ui-helpers';
+import { waitForCardLoad, waitForSchemaViewToLoad, encodeColons, waitForTestsToEnd } from '../helpers/card-ui-helpers';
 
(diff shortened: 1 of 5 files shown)


jenweber:
@habdelra do you have any idea about why unrelated tests are failing? I tried restarting the job and it had the same result.

habdelra:
I don't know. everything looks like it is passing in master. when is the last time you have rebased? maybe you're missing the latest ver bump in your branch.

habdelra:
do you see these failures locally?

jenweber:
@habdelra yes, I do. But only when the whole test suite is run. The "card mode" tests themselves work fine isolation. Is there some destroy or teardown that I'm missing? Does adopting from a card prevent its cleanup?

habdelra:
Ah, it sounds like you are not tearing down the test correctly. I’m away from my computer right now, but when I get a chance I’ll show you how to do that.

jenweber:
It's ok, I think I got it. I don't understand when that's necessary vs not, but I was able to copy another test's patterns.

habdelra:
ok, so test test are divided into tests that modify the system, and test that introspect the system. you are adding a test that modifies the system (creates a card), into a test suite used for introspection of the system (navigating around the site). this is the difference between the moduleSetup and the testSetup. I'm still away from my computer, so when I get back home I'll provide better guidance and examples.

habdelra:
the reason we separate it out, is that we can run the test a lot faster this way, when you have a bunch of tests that can utilize the same setup instead of having to be setup up and torn down in between each test.

habdelra:
really it would be better for you to add this into the card adoption tests....

habdelra:
@jenweber Here's the promised write up (done from my phone, so please excuse all the inevitable typos :-)

The long explanation is that because we are doing full end-to-end testing, where we touch a live DB (most tests use something like mirage to mock the back end--and thus dont have the extra challenge) we need to be really careful about cleaning up after ourselves, otherwise state will bleed into other tests (as you have seen). This was the case for both the old system and the new system. anytime you see a test that runs successfully in isolation, but fails when run with other tests is in this scenario: where state is bleeding into other tests. The state may be data-based, or async based (promise that is not being awaited). In this case it looks like you have data-based state that is bleeding into other tests. 

The rule of thumb is that you need to make sure that your test returns the state of the system back to the state that was in when the test started. This means that, if within your test, a card is created, you need to make sure that it is cleaned up when the test is done. There should be no net change in the cumulative number of cards that exist the in DB in between each test, or that cards that do exist globally (like realm cards) remain unchanged after the test concludes. Adopting a card is a form of card creation, which means that you must take care of the card cleanup when the test concludes. To support this, we have a very unique mechanism in our test framework (because we are testing against a live DB) that allows us to control the cards in the DB between each test. This includes logic to create cards that your tests depend on, as well as to remove cards created within your tests. This is our `Fixtures` system.

In our tests you can create scenarios by instantiating a `Fixture` and then specifying the cards that should exist before the test is run, as well as cards that should be removed that were created in the course of running your test. Any cards that you tell the system that you need to be created before the test is run will automatically be removed for you--you don't need to say anything special to remove these cards. For tear down you only need to tell the Fixture about the cards created by your test itself (like a test that creates cards). 

To specify cards to create in your Fixture (that you'll rely upon within your tests), use the `create` property in the fixture configuration (Typescript will help you here). The `create` property accepts an array of `CardDocument` objects that specify the cards you want to create (and these will be torn down for you automatically).

To specify cards to destroy after the test is done running in your Fixture, use either the `destroy.cards` property in the Fixture constructor to specify an array of `CardId`'s to destroy, or use `destroy.cardTypes` to specify the an array of card parents whose adopted children you want to destroy. And again, what you are specifying here are not the cards you specified in the create hook--those are cleaned up automatically, its the cards that were created in the test itself that need to be cleared out.

The next thing to be cognizant of is _when_ the Fixtures are run: do you want them to run in between each test? or do you want them run once at the beginning of the test module, and once at the end of the test module. The reason we give you a choice is that the work to setup and teardown the test requires web request for each card to create and each card to destroy. depending on your tests, the cumulative amount of time it takes to issue these cause could be very large (lots of tests or lots of cards for the tests, or a combination of lots of tests using lots of cards). To help make the tests as efficient as possible we recommend that you group your tests into "read-only" tests and "mutating" tests. 

The read-only tests are tests that don't actually change any system data. They are tests where you are just navigating around the system and making sure the correct things appear in the various routes or card states. All the system configuration in terms of setting up the testing scenario occurs within the Fixture, and the test is basically just poking around looking at things, but not changing any card data, like the library tests, or card view tests. In this scenario you can greatly speed up the time it takes to run the tests by having all the tests just reuse the same cards without creating the cards from the fixtures or cleaning up the cards from the fixtures in between each test--since the promise is that the test wont alter that card data. In this case we use:
```js
scenario.setupModule(hooks);
```
to setup the Fixtures. `setupModule()` means that we only setup the fixture at the module boundary, not the test boundary. 

The mutating tests are tests that actually change card data--they may create cards, update cards, or delete cards--but the fundamental thing is that they are changing the state of the application within the tests themselves. These tests need to create the cards from the fixture and teardown the cards from the fixture (plus any other cards that were created in the tests themselves) after each individual test. This is the kind of test that you wrote, since it is creating an adopted card within the test itself. For these tests you can use:
```js
scenario.setupTest(hooks);
```
to setup the Fixtures. `setupTest()` means that we setup the fixture at the individual test boundary.

Some test actually layer these 2 ideas (like the adoption tests). for the `adopt-card-test.js` we never alter the parent card--that is held static throughout all the test. So we use a Fixture to govern the setup of the parent card specifically that is a module scoped Fixture (does't run in between each test). This prevents us from having to expend the extra time to create and destroy the parent card in between each test, since we never change that anyways. Then, since the adoption tests actually do create new cards within the tests themselves--but cards that specifically only ever extend from the parent card that we'd defined, we use a Fixture that deletes all the cards that adopt from the parent that is a test scoped Fixture (so it runs in between each individual test).

habdelra:
this write up should probably go into an MD file that we maintain for contributors....

habdelra:
it's also worth noting that we follow this exact same strategy for our node tests as well, using in mocha we use `before()` and `beforeEach()` (paired with `after()` and `afterEach()`) to setup "read-only" tests and to setup "mutating" tests. In general, it's a really nice testing strategy.

packages/cardhost/tests/acceptance/card-mode-navigation-test.js
```diff
@@ -14,8 +20,18 @@ const testCard = cardDocument().withAutoAttributes({
...
+  csId: 'gen-z-puppies',
+  title: 'The Gen Z Puppy',
+});
+const adoptableCardPath = encodeURIComponent(adoptableTestCard.canonicalURL);
 const scenario = new Fixtures({
-  create: [testCard],
+  create: [testCard, adoptableTestCard],
+  destroy: {
+    cardTypes: [adoptableTestCard],
+  },
```
habdelra: you actually have a test leak here. Please note that this Fixture is  module scoped (line 38):
```js
  scenario.setupModule(hooks);
```

This is basically a bug lying in wait--you are not returning the system back to a clean state because this Fixture is not being run after your test--it's being run after the module itself. This means that the card that was created in your test is leaking into all the other tests in this module.

This type of test is a "mutating" test, and as such, it cannot live in this module, because this module is setup for "read-only" tests. You can either create a new module in this file specifically for mutating tests like this that uses a Fixture that run at the test boundary. or, what might be easier, would be to move this test into the adoption test suite, which is already setup as "mutating" tests.

jenweber:
Ok, will do. Thanks Hassan! I can add your writeup to `MAINTAINERS.md` while I'm at it!

jenweber:
I think I have addressed the issues described, and this is ready for another look. No rush!

PR #1477, opened 2020-03-23
Improve left-edge transitions

burieberry:
- Use ember-toolbars for `cardhost-left-edge` component
- Resolve resulting z-index/visibility issues on schema mode when dragging fields
- Slide catalog in/out using `ember-animated` when transitioning to/from schema mode
*(not using ember-toolbars here because margins do not recalculate when toolbar width changes)*
- Only render the left-edge in `cards` template


habdelra:
This looks good to me, let's merge this.

Issue #1474, opened 2020-03-20
Missing field icons


burieberry:
I don't know what we decided to do about the icons, but we should either add them, or remove the gap before the field name. Both in the card catalog menu, and in the schema field header.

<img width="1282" alt="icons" src="https://user-images.githubusercontent.com/16160806/77203138-1e248480-6ac6-11ea-8ff7-c2a184d9522b.png">

dnaidionov:
Let's use the same icons we have now in old hub

jenweber:
These missing icons have some code comments about needing to add them to the API.
@habdelra can you describe what you would do here? I see a few paths:

1. add a `csIcon` property to all cards, and make it a path. Put the files for our icons in public.
2. add a `csIcon` property to all cards, and make it an SVG
3. Handle it 100% in cardhost until we determine the next steps (see #1483)
4. ???

habdelra:
@jenweber yeah when last we discussed with Chris, he said that he'd like to have a new cs field to categorize the card--so that there would be different types of card categorizations, and that the icon we display is based on the card categorization. I think, then, that means we are in a bit of a holding pattern in that we need to get some initial card categories, as well as, UI for how we want the user to assign/change a categorization on a card.

jenweber:
Ok, in that case maybe I'll clean up my front-end-only solution so that it is acceptable to be merged, and we'll use that for now. I will also open up a new ticket that summarizes the work to be done to handle this for real.

PR #1483, opened 2020-03-25
Restore missing field icons

jenweber:
Closes issue #1474

This PR adds back in the missing field icons. This is accomplished on the front end as a stopgap until icons become part of the Card API.

@burieberry the thing I could use your help with is figuring out how to visually balance the icons.

I also don't have an SVG for the input, and the text icon is different than what we had in the past. Do you have those assets anywhere?


burieberry:
@jenweber yes, there are two different text icons. The icons I used for fields are in `/images/field-types` folder. I pushed some suggestions to your branch. I think the text icon actually needs to be 12px x 15px for the field-renderer, though. Feel free to make any changes

burieberry:
@jenweber we can use `has-many-field-icon.svg` for the card field. It's not used anywhere else. We can rename it

jenweber:
Something still isn't right here for the Chrome snapshots in Percy. It looks fine in Chrome on my computer. I added animation waiters and it had no effect.

Screen Shot 2020-03-31 at 10 25 27 AM

burieberry:
@jenweber I see this issue visible on Safari and adding `max-width: 100%` to the `.ch-catalog-field--icon` class seems to work

PR #1488, opened 2020-03-27
update README

jenweber:


packages/cardhost/README.md
```diff
@@ -23,15 +23,32 @@ You will need the following things properly installed on your computer.
...
 * Visit your app at [http://localhost:4200](http://localhost:4200).
 * Visit your tests at [http://localhost:4200/tests](http://localhost:4200/tests).
 
-`INDEX_INTERVAL` determines how often the hub should reindex data, in minutes.
-It is recommended to set this number high for local development.
+### Common server configuration
+
+#### `DEV_DIR`
+
+If you would like your cards to be saved to a git repository on your hard drive, specify the path to an existing, separate git repository when you start the servers.
```
habdelra: it's more precise to say that we should specify a path the the directory that we want to write cards out to. In the git realm we write cards to `./cards`, so in this case you should point your file realm to the `./cards` of the git repository.

jenweber: added, thanks!

Issue #1489, opened 2020-03-30
Create a modal that informs users that cards are public


jenweber:
We should make sure that anyone who tries our app in deployment understands that their cards are public.

I was thinking that this should be an application-level query param, so that when people are doing demos or working locally, they don't have to contend with the modal. Also, it should probably only be displayed for production builds.

jenweber:
Closed by #1490

PR #1490, opened 2020-03-30
Add user-info modal #1489

burieberry:
modal

To test locally, set the `displayUserDialog` env var to be `true`.

Do the env variable for production look right? For tests, I only enabled the modal for the `user-info-modal-test`. How is the modal's appearance frequency? Right now it appears when a page first loads, after being dismissed, it wouldn't appear again unless there's a page full page refresh.


jenweber:
The prod env variables look fine. We can confirm with `yarn start-ember --environment=production` and see what the app looks like.

I think the main question is, does Chris want this to always show up whenever he demos? That's why I was thinking that we should store whether the modal has been shown in a query param, so he could visit `cards.example.com?confirmed=true` or something like that in order to skip the modal.

burieberry:
@jenweber oh I'll try that, good idea

burieberry:
@jenweber I did the queryParam part easily but spending time with tests. What do you think is the best way to handle the tests? If I hide the dialog in the tests using an env variable, then I still need to change some url checks to `?confirmed=false`. Or if the default is that the dialog is visible, then I'd either need to click out of it in tests or change urls to `?confirmed=true`

jenweber:
The way I have solved something similar for testing is to check the env via a getter:

```js
get showDialog() {
  if (this.showDialog !== undefined) {
    return this.showDialog
  } else {
     return ENV.showDialog
  }
}
```

This allows you to override showDialog in your acceptance and integration tests.

burieberry:
@jenweber  This is ready for another review. Using a getter didn't prevent the query param from being added, unless I was doing it wrong. After certain transitions (using the library, going on the themer page), the model hook runs again and the `?confirmed=false` gets added to the url, so I updated the tests accordingly. (I was hoping the queryParam wouldn't appear on the url if the value is `false` or `undefined`)

jenweber:
The trick here is to not assign an initial value to the query param, and then I think you'll get the behavior you want. I opened a PR to your branch so you can decide if you like it or want to make changes: https://github.com/cardstack/cardstack/pull/1492

PR #1491, opened 2020-04-01
Show recent cards only for a given device in deployed apps

### LINKED ISSUE #1487: Create mode that shows recent cards from local storage only
In hub v2, we can now show all recent cards. For the purposes of demoing, we should have an option to show only the cards recently created on that specific machine, in order to avoid cluttering the home page with all the cards that everyone has made.

This will build on https://github.com/cardstack/cardstack/pull/1229/files

jenweber:
Closes #1487 

The goal is that when we send people a link to demo this, they should not see other people's cards. Note that this has nothing to do with authorization. All cards in a demo are public-readable by anyone.

## Added:
- `csCreatedBy` string to the Card API
- `DEVICE_CARDS_ONLY` env var that is passed into the app.
- When ^^^ is true, the Library will fetch the recent cards associated with the given device, not all recent cards
- `cardLocalStorage` service, which sets a semi-random string for `Device` in the browser local storage
- The application controller sets up the cardLocalStorage
- Tests

## How to test
- in config/environment.js, set deviceCardsOnly to true
- run the app. Make a card. Open the library and see only your card there. Refresh and see it's still there. Clear local storage. See that the card doesn't show up in the library. Make a new one, and see it show up.


packages/cardhost/app/services/library.ts
```diff
@@ -16,6 +20,7 @@ const size = 100;
 
 export default class LibraryService extends Service {
   @service data!: DataService;
+  @service cardLocalStorage: any;
```
habdelra, line 23: you should really use TS for your service and then use that as the type definition

packages/hub/card-document.ts
```diff
@@ -44,6 +45,9 @@ export class CardDocument {
               this.csOriginalRealm = value;
             }
             break;
+          case 'csCreatedBy':
+            this.csCreatedBy = value;
+            break;
```
habdelra, line 50: this is incorrect, instead of this just add `csCreatedBy` to the `case` list below.

packages/cardhost/app/services/library.ts
jenweber: It was tough to turn CardLocalStorage into a ts file because the polyfill had no types. I removed the polyfill altogether, which means that only browsers with local storage can use this feature. However, I think that's ok given https://caniuse.com/#feat=mdn-api_window_localstorage

jenweber:
@habdelra This is ready for another review.

jenweber:
@habdelra test added.

PR #1498, opened 2020-04-06
Add card toString function

### LINKED ISSUE #1494: Don't show [object object] in schema view
just noticed an issue: in the schema view when we show a card that has a relationship to another card, we see [Object Object]. In the edit view we see the card ID of the related card. I think we need to show something more sensible here (like we do in the edit view).

image

alexspeller:
When a card should be represented as a string, show something more meaningful
than `[object object]`

Fixes #1494 (and any similar places where this would show up)

### DIFF
--- a/packages/hub/card.ts
+++ b/packages/hub/card.ts
@@ -238,6 +238,10 @@ export class Card {
     }
   }
 
+  toString() {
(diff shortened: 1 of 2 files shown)


habdelra:
looks good, merge when green

PR #1499, opened 2020-04-08
Improve how header and border show when hovering over a card

### LINKED ISSUE #1389: In view mode, only show header and border when hovering over card
Currently, the element that has the hover (mouseover) event takes up the full width of the screen, so hovering over any part of it triggers the header/border animation. That hoverable area should be confined to the card itself.

alexspeller:
Prevent header and border showing when hovering outside of the visible card
boundary.

Also clean up some monkey patching of the card api by making the property live
on the component instead of the card and changing what the animated-value component
watches

Fixes #1389


habdelra:
@alexspeller looks like you have some linting errors. I wonder if adjusting the aria role will resolve these...

alexspeller:
@habdelra yeah that was it, my editor is using a dated ember-template-lint for some reason

PR #1500, opened 2020-09-14
Add computed query example test and fix pgclient pool release async bug

alexspeller:
So the issue was the postgres pool filling up. I realised that it was due to a single line that was throwing an error outside of the finally block, preventing the client getting released in that case.

Moving the line inside the finally block to ensure client pool release prevents the tests hanging in this situation now


lukemelia:
Closing as outdated

PR #1504, opened 2020-10-08
Collection view start

burieberry:
- Add `isolated-collection` component
- Add ts file for changing collection view format and card selection
- CSS file currently has active styles for grid format
- Add `background-header` component that all cards can use to have a special header with background image
- Add `collection` route with 3 sample cards (note: the json data and css for the sample cards are in a separate repo)
- New svg icons

collection-grid-view

master-recording

participant

musical-work


ef4:
The `.gitignore` needs to be updated to exclude these `.d.ts`, `.js`, and `.js.map` files that are typescript build output. 

It's unfortunately pretty manual right now because of the mix of JS and TS in these directories.

ef4:
This PR still contained some of the files that should not have been committed.

(Once they're added to git, adding them to .gitignore won't automatically remove them from git.)

PR #1535, opened 2021-03-16
card-loader

mattmcmanus:


core/src/raw-base-cards.ts
```diff
@@ -1,41 +1,13 @@
...
-import DateSchemaJS from '!raw-loader!@cardstack/base-cards/date/schema.js';
-import DateEmbedded from '!raw-loader!@cardstack/base-cards/date/embedded.hbs';
-import TagSchemaJS from '!raw-loader!@cardstack/base-cards/tag/schema.js';
-import TagEmbedded from '!raw-loader!@cardstack/base-cards/tag/embedded.hbs';
-import CommentSchemaJS from '!raw-loader!@cardstack/base-cards/comment/schema.js';
-import CommentEmbedded from '!raw-loader!@cardstack/base-cards/comment/embedded.hbs';
+import StringCard from '!@cardstack/card-loader!@cardstack/base-cards/string/card.json';
+import DateCard from '!@cardstack/card-loader!@cardstack/base-cards/date/card.json';
+import TagCard from '!@cardstack/card-loader!@cardstack/base-cards/tag/card.json';
+import CommentCard from '!@cardstack/card-loader!@cardstack/base-cards/comment/card.json';
```
mattmcmanus, line 4: I would have loved to just do
```
import CommentCard from '@cardstack/base-cards/comment'
```
But it wouldn't work because: 
1. The config approach wasn't picking up the card-loader
2. It was yelling that nothing was there since it's a directory. I wasn't sure if we could simulate a card.json when not there, or something along those lines?

card-loader/src/index.ts
```diff
@@ -0,0 +1,27 @@
...
+  cardJson: any
+): Promise<string> {
+  let base = this.context;
+  let files = {};
+  // TODO: Get config based loader working with options passed in correctly working
+  // let options = getOptions(this);
+  let realm = "https://cardstack.com/base/models";
+  let name = basename(base);
+
+  let cardDir = await opendir(base);
```
ef4, line 18: If you use [walk-sync](https://www.npmjs.com/package/walk-sync) here instead we can support arbitrarily nested files.

core/src/raw-base-cards.ts
ef4: I think we want to go even further than that, like

```js
import everything from '!@cardstack/card-loader!@cardstack/base-cards/package.json'
```

ef4 (commented):
Combining these two suggestions, we could crawl the whole package of base cards at once as a single tree of JSON. We'll need to remember to pass options to walkSync to ignore node_modules and probably dotfiles.

ef4:
Oh, one more challenge though: I think any of these approaches that combines many files will escape webpack's rebuilding. Webpack might not know to watch all those files.

ef4:
One way around that is to have the loader emit code that imports all the other files via raw loader.

mattmcmanus:
Looks like we can declare those as dependencies within the loader: https://webpack.js.org/contribute/writing-a-loader/#loader-dependencies

PR #1537, opened 2021-03-18
Card views as native JS files

mattmcmanus:
- [x] "templates" are now JS files
- [x] What is responsible for module urls?
   - I've thrashed on a couple different approaches. I keep coming back to wanting the passed in defineModule to return the full path to the module and then we store that as part of the compiled card. There are places, like in the glimmer plugin, where we will also need to import the appropriate file and there isn't enough info there to do that correctly (especially in our dual world of requires and imports). I suppose this points to needed to create a `CardResolver` of some sort that can decide on the paths and know when to do imports vs requires?
- [x] Simple cards, like the StringCard can be automatically embedded if they require no changes in the components scope
- [x] Rewrite Card Templates that have cards that need imports
   - [x] Add the imports, properly named
   - [x] Add those imports to the component scope
   - [x] Rewrite component invocation to use imported component
- [x] TemplateComponents should be transformed so they can render in the ember app


base-cards/date/embedded.js
```diff
@@ -0,0 +1,9 @@
+import { setComponentTemplate } from "@ember/component";
+import { precompileTemplate } from "@ember/template-compilation";
+import templateOnlyComponent from "@ember/component/template-only";
+export default setComponentTemplate(
+  precompileTemplate("Date: {{@model}}", {
+    strictMode: true,
+  }),
+  templateOnlyComponent()
+);
```
ef4, line 9: This is just an aside, but: pzuraq (who made the ember-template-imports addon and the underlying glimmer features it uses) mentioned yesterday wanted to offer a shorthand for all of this. It seems like a good idea. It's a lot of verbosity.

cardhost/app/lib/builder.ts
```diff
@@ -11,12 +12,35 @@ import { setComponentTemplate } from '@ember/component';
...
+  constructor(params: {
+    modulePrefix?: string;
+    defineModule: (fullModulePath: string, source: unknown) => void;
+  }) {
     this.cache = new Map();
+    this.customDefine = params.defineModule;
+    this.modulePrefix = params.modulePrefix ?? '';
+  }
+
+  // TODO: is there a way to reuse the arguments of the defineModuleCallback type?
```
ef4, line 32: Yes:

```ts
async defineModule(...args: Parameters<defineModuleCallback>) {
```

`Parameters` is a type that ships with TypeScript itself. If you use VSCode to jump-to-definition on it, you can see a whole family of similar helpers, and their (very short) implementations. In this case the implementation:

```ts
/**
 * Obtain the parameters of a function type in a tuple
 */
type Parameters<T extends (...args: any) => any> = T extends (...args: infer P) => any ? P : never;
```

is using `infer`, a pretty neat feature that's explained [here](https://www.typescriptlang.org/docs/handbook/2/conditional-types.html#inferring-within-conditional-types). Background reading for that section if it's totally alien would be starting at [this chapter](https://www.typescriptlang.org/docs/handbook/2/types-from-types.html).

ef4 (commented):
Regarding "imports vs requires", I think we can force only imports, even on the server. Node has native ES module support now, and we make that work for us. 

I think it's helpful to try to list the build stages we've discussed and give them names:

 - author writes some code
 - **prebuild stage** can (optionally) convert their code to the standard ember low-level with `precompileTemplate`, `setComponentTemplate`, etc. It's optional because they could also just write the low-level by hand.
 - **card compiler stage** runs here and emits code that is still ES modules like it's input, and still uses `precompileTemplate` etc, but now all the card-to-card interactions have been resolved to real module imports & invocations. Imports of other cards are expressed as URLs.
 - **build stage** is where we do different things for 
   -  normal browser build (embroider does the work)
   -  dynamically-defined cards like our test suite or a card that is being live-edited (we need to do our own compilation and module registration to get the code to run)
   -  server-side build (we don't have this yet, but this would be how code like the `routeFor` runs in node. It will probably still need transpilation because we want to map the URL imports to real paths on disk, probably paths sitting in a card compiler cache so all the compiled cards can see each other.

It's premature for the card compiler to worry about converting the module URLs into a form that actually runs. That's the concern of the build stage. That's why I think the card compiler can just unilaterally pick URLs, based on the card's own URL and configuration.

cardhost/app/lib/builder.ts
```diff
@@ -49,13 +82,27 @@ export default class Builder implements BuilderInterface {
...
+    // );
+    // this.defineModule(compiledCard.modelModule, templateSource);
+
+    /*
+      TODO Build stage:
+        1. Require template module from it's path, eg: "http://mirage/cards/hello/isolated"
+        2. Babel transpile it
+        3. Do the createTemplateFactory song and dance?
+        4. Redefine the module? Make a new definintion?
+    */
```
mattmcmanus, line 102: @ef4 This is the part that I'm not stuck on. 

1. I'm not quite sure how we want to handle the module resolution and definition for these built modules
2. I'm not sure of the full scope of babel plugins we'll need and how to configure them. Do you think we'll just copy the logic from here? https://github.com/embroider-build/embroider/blob/master/packages/core/src/app.ts#L337
3. Apart from the babel config, will we also need to wrap the components in the templateFactory stuff?

cardhost/app/lib/builder.ts
```diff
@@ -3,38 +3,37 @@ import type {
...
-// import type { TemplateFactory } from 'htmlbars-inline-precompile';
-
-// // @ts-ignore
-// import { createTemplateFactory } from '@ember/template-factory';
-// function compileTemplate(source: string): TemplateFactory {
-//   return createTemplateFactory(JSON.parse(precompile(source)));
-// }
 
+// @ts-ignore
+import HTMLBarsCompiler from 'ember-template-compiler/vender/ember-template-compiler';
```
mattmcmanus, line 10: @ef4 Webpack doesn't like the main file of this package because it's not statically analyzable. What we ultimately want is the file under vendor (PS: Whats up with this?). Any tricks here?

ef4: I think this is the same as `import { precompile } from '@glimmer/compiler';` that we already use in `tests/helpers/template-compiler.ts`.

PR #1545, opened 2021-04-01
Start adding boxel dashboard components

burieberry:
dashboard

qr-code

Some notes:
- Need help with colocating css files with components without importing the stylesheet via the js file in `web-client/app/components/card-pay` dirs
- When adding css files in `web-client/app/styles` folder, additional css files imported to the app.css aren't being read, but maybe I didn't restart my server properly, or missing some config for those files to be read somewhere
- `connect` modal with the QR code will probably need to live outside of the card-pay route, since there'll also be a connect button on the cardstack landing page
- There's a ton of data in [cardstack landing page in boxel repo](https://github.com/cardstack/boxel/blob/main/tests/dummy/app/routes/cardstack.js). I don't know if that's how we want to handle the landing page info.


Aierie:
> There's a ton of data in cardstack landing page in boxel repo. I don't know if that's how we want to handle the landing page info.

Yep! I didn't know what format of data I would be receiving, so I just added some JSON for the time being. We can change that later.

PR #1550, opened 2021-04-05
Co-locate web-client app component css files

burieberry:


packages/web-client/ember-cli-build.js
```diff
@@ -1,6 +1,8 @@
 'use strict';
 
 const EmberApp = require('ember-cli/lib/broccoli/ember-app');
+const path = require('path');
+const concat = require('broccoli-concat');
```
burieberry, line 5: Here the linter is complaining `"broccoli-concat" is extraneous`, but when I remove this line, it complains that `concat` is not defined. What was your solution to this? @lukemelia

Aierie: https://github.com/mysticatea/eslint-plugin-node/blob/master/docs/rules/no-extraneous-require.md

Was curious about the rule and looked into this, it looks like this might be solved by installing `broccoli-concat` explicitly as a dependency/dev dependency.

lukemelia: This is correct @burieberry

PR #1551, opened 2021-04-06
Enhancement/migrate cardstack landing page

Aierie:
CS-547

Migrating the cardstack landing page. Currently, the org-card has colocated css that will need #1550 to work. Some issues:
1. Where do we want to place styles for routes?
2. Asset resolution is a bit weird because imports from a route happen after embroider has done some processing - I guess it's the same for the Boxel repo?

TODO:
- [x] html:
  - [x] document content
  - [x] button as link-to (needs fix in Boxel for disabled state - for some reason `@disabledClass` is not applied when Boxel is used as an addon)
- [x] tests


packages/web-client/app/routes/index.ts
```diff
@@ -0,0 +1,110 @@
+import Route from '@ember/routing/route';
+
+import CardstackLogo from '../images/icons/cardstack-logo-white-text.svg';
```
lukemelia, line 3: Is this an example of the surprising pathing that you need to use? Would full paths work, and would that be better? i.e. `@cardstack/web-client/images/icons/cardstack-logo-white-text.svg`

Aierie: That seems better! Thanks @lukemelia.

Aierie: Will also apply to Boxel later

packages/web-client/app/services/edges.ts
```diff
@@ -0,0 +1,25 @@
...
+
+  @action
+  updateDisplayLeftEdge(isDisplayed: boolean) {
+    if (isDisplayed) {
+      this.displayLeftEdge = true;
+    } else {
+      this.displayLeftEdge = false;
+    }
+  }
+}
```
burieberry, line 25: @Aierie  Let's not install bring this over yet. There's no left-edge in Card Pay, and it's static on the home page (you can use the named block for left-edge that's in the dashboard component). We can figure out a good way to implement this when we start working on Card Space.

packages/web-client/app/templates/index.hbs
```diff
@@ -1,36 +1,50 @@
-<main class="cardstack-container" data-test-cardstack>
-  <header class="cardstack-section-header">
-    <h1 class="cardstack-section-header__title" data-test-cardstack-header>
+<Boxel::Dashboard
+  {{did-insert (fn this.updateEdges)}}
```
burieberry, line 2: Remove this line

packages/web-client/app/templates/index.hbs
```diff
@@ -1,36 +1,50 @@
...
-
-    <div class="cardstack-landing-hero">
-      <img src={{@model.logo}} alt="" role="presentation" class="cardstack-landing-hero__logo">
-      <p class="cardstack-landing-hero__description">The world’s first cooperative SaaS platform.</p>
-      <p class="cardstack-landing-hero__description">Own the platform. Choose the tools. Work your way.</p>
-      <LinkTo @route="index" class="cardstack-landing-hero__cta">
-        Learn more {{svg-jar "caret-thin-right" class="cardstack-landing-hero__cta-icon"}}
-      </LinkTo>
+    </Boxel::Button>
+    {{!-- {{/unless}} --}}
```
burieberry, line 21: This will be same as the Layer 2 connect button. Please remove references to any cardstack session. I'm not sure which design we'll use, so remove this button for now.

burieberry (changes requested):
Left a couple comments

Aierie:
Oh right, will remove all the edges. Thanks!

packages/web-client/app/templates/index.hbs
Aierie: Hmm, on this particular page isn't this button correct?

Aierie:
Will have to examine why the `@disabledClass` argument is not passed to the `LinkTo` in `Boxel::Button`, but using a placeholder style to overwrite for now.

Aierie:
Also, now that we're working on the Cardstack repo - @burieberry what are your thoughts on utility classes? Would like to use them if possible.

PR #1555, opened 2021-04-07
Implement first version of Layer 2 Connect modal

lukemelia:
CS-532

See Linear card for checklist


lukemelia:
Screenshot 2021-04-08 at 12 51 15 PM

lukemelia:
Note that I decided to create separate Linear cards for the video, "text it to me" and alternate xDai wallet features.

Aierie (commented):
I'm having trouble building for some reason (even on main) so I'll have to look at the built product more closely after I debug that. Overall this looks fine to me! 

Minor detail: If we're focusing on using `rem` units, we might want to get our `img` tags and `canvas` to use them so that proportions are maintained if users change their root font sizes - but then how designs change with font size seems like a gray area, so I think it's fine if some parts aren't `rem`-ified for now.

lukemelia:
> Minor detail: If we're focusing on using rem units, we might want to get our img tags and canvas to use them so that proportions are maintained if users change their root font sizes - but then how designs change with font size seems like a gray area, so I think it's fine if some parts aren't rem-ified for now.

@Aierie img tags can likely be styled with rems, although you will end up with subpixel rendering for bitmap images, so that's a tradeoff. Canvas is a bitmap concept, so AFAIK there is no direct way to use rems to size elements on a canvas, or the canvas itself.

Aierie:
> > Minor detail: If we're focusing on using rem units, we might want to get our img tags and canvas to use them so that proportions are maintained if users change their root font sizes - but then how designs change with font size seems like a gray area, so I think it's fine if some parts aren't rem-ified for now.
> 
> @Aierie img tags can likely be styled with rems, although you will end up with subpixel rendering for bitmap images, so that's a tradeoff. Canvas is a bitmap concept, so AFAIK there is no direct way to use rems to size elements on a canvas, or the canvas itself.

@lukemelia Ah right bitmap is an issue. Canvas can be styled with css with rems, and the contents will resize to fit - but again bitmap will be an issue.

PR #1560, opened 2021-04-12
Add conditional logic to deposit workflow and update layer 2 wallet connect uI

lukemelia:
CS-559
CS-446

Also, there is a commit here with the results of running yarn-deduplicate -- this promise to speed up yarn installs substantially.


burieberry (approved):
👍

PR #1563, opened 2021-04-14
Unit test coverage for Workflow/Milestone/WorkflowPostable/etc

lukemelia:
CS-589

### DIFF
--- a/packages/web-client/app/models/workflow/milestone.ts
+++ b/packages/web-client/app/models/workflow/milestone.ts
@@ -17,11 +17,7 @@ export class Milestone {
     this.postables.invoke('setWorkflow', wf);
   }
   get isComplete() {
-    return this.postables
-      .filter((p) => {
-        return !this.#excludedPostables.has(p);
-      })
-      .isEvery('isComplete', true);
+    return this.visiblePostables.isEvery('isComplete', true);
   }
   completedDetail;
 
--- a/packages/web-client/app/models/workflow/milestone.ts
+++ b/packages/web-client/app/models/workflow/milestone.ts
@@ -9,15 +9,19 @@ interface MilestoneOptions {
 export class Milestone {
   title: string;
   postables: WorkflowPostable[] = [];
-  includedPostables = new Set<WorkflowPostable>();
-  excludedPostables = new Set<WorkflowPostable>();
+  #includedPostables = new Set<WorkflowPostable>();
+  #excludedPostables = new Set<WorkflowPostable>();
   workflow: Workflow | undefined;
   setWorkflow(wf: Workflow) {
     this.workflow = wf;
     this.postables.invoke('setWorkflow', wf);
   }
   get isComplete() {
-    return this.postables.isEvery('isComplete', true);
+    return this.postables
+      .filter((p) => {
+        return !this.#excludedPostables.has(p);
+      })
+      .isEvery('isComplete', true);
   }
   completedDetail;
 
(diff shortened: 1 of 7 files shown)


packages/web-client/app/models/workflow/milestone.ts
```diff
@@ -9,15 +9,19 @@ interface MilestoneOptions {
...
-  includedPostables = new Set<WorkflowPostable>();
-  excludedPostables = new Set<WorkflowPostable>();
+  #includedPostables = new Set<WorkflowPostable>();
+  #excludedPostables = new Set<WorkflowPostable>();
   workflow: Workflow | undefined;
   setWorkflow(wf: Workflow) {
     this.workflow = wf;
     this.postables.invoke('setWorkflow', wf);
   }
   get isComplete() {
```
Aierie, line 19: Would changing this to `this.visiblePostables.isEvery('isComplete', true)` make sense? To make the `isComplete` property less dependent on `visiblePostables` having recently been accessed.

lukemelia: Hmm, not sure. visiblePostables has side effects -- see my comment here: https://github.com/cardstack/cardstack/pull/1564/files#diff-217fd8162cfba6de1caa5800dc8cd7a75e9ff7df25dffad73d3c55ae72dde876R30-R34

Aierie: Hmm, looks a bit tricky. If we don't change to `this.visiblePostables.isEvery('isComplete', true)`, then we can't reliably check milestone completion status before `Milestone.visiblePostables` is accessed right? I guess this would be relevant to something like this where we have a quick view of different workflows' completion statuses without rendering the whole thread:

image

But if we do, then we might add an incorrect timestamp to the postables for that milestone - or running code that adds the postables to `Milestone.#excludedPostables` because it's called before the milestone is 'ready'.

Maybe this is something we can wait to change, but make a comment on?

Aierie (approved):

packages/web-client/app/models/workflow/milestone.ts
Aierie: Sorry, github looked like it doubled up my comment.

Can we leave a comment about not accessing `Milestone.isComplete` without accesssing `Milestone.visiblePostables`? It might be relevant for things like overviews of workflows:

image

lukemelia: This is a really good point. I guess since the `visiblePostables` getter is idempotent, it would be OK to use from `isComplete` as well as the template. The only real issue maybe is the timestamp side effect, but maybe even that is OK.

lukemelia: This will all change once we have workflow persistence, too.

packages/web-client/app/models/workflow/milestone.ts
```diff
@@ -9,15 +9,19 @@ interface MilestoneOptions {
 export class Milestone {
   title: string;
   postables: WorkflowPostable[] = [];
-  includedPostables = new Set<WorkflowPostable>();
-  excludedPostables = new Set<WorkflowPostable>();
+  #includedPostables = new Set<WorkflowPostable>();
+  #excludedPostables = new Set<WorkflowPostable>();
```
burieberry, line 13: What does the # in the property name signify?

burieberry (approved):

packages/web-client/app/models/workflow/milestone.ts
Aierie: Private class fields I think!

https://www.typescriptlang.org/play?ts=3.8.3#example/private-class-fields

burieberry: Ah is it just to mean number? At first I thought it was some new ember or typescript thing lol.

burieberry: Ok so it was for typescript. I've only seen _ used before.

lukemelia: "Private"! It's a newish JS feature now in Stage 3: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_class_fields

lukemelia: "_" is just a convention that communicates "this is private". "#" is part of Javascript that enforces privacy.

PR #1565, opened 2021-04-15
Render post-workflow postables once workflow is complete

lukemelia:
- extract PostableCollection to support Milestone and use for Workflow#epilogue as well

CS-594

### DIFF
--- a/packages/web-client/tests/integration/components/workflow-thread-test.ts
+++ b/packages/web-client/tests/integration/components/workflow-thread-test.ts
@@ -1,10 +1,12 @@
 import { module, test } from 'qunit';
 import { setupRenderingTest } from 'ember-qunit';
-import { render } from '@ember/test-helpers';
+import { render, waitFor } from '@ember/test-helpers';
 import hbs from 'htmlbars-inline-precompile';
 import { Workflow } from '@cardstack/web-client/models/workflow';
 import { Milestone } from '@cardstack/web-client/models/workflow/milestone';
 import { WorkflowMessage } from '@cardstack/web-client/models/workflow/workflow-message';
+import { WorkflowPostable } from '@cardstack/web-client/models/workflow/workflow-postable';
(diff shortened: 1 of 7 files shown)


packages/web-client/tests/integration/components/workflow-thread-test.ts
```diff
@@ -26,46 +28,74 @@ module('Integration | Component | workflow-thread', function (hooks) {
...
+      }),
+    ]);
+    this.set('workflow', workflow);
+    workflow.attachWorkflow();
+    await render(hbs`
+      <WorkflowThread @workflow={{this.workflow}} />
+    `);
+    assert.dom('[data-test-epilogue][data-test-postable]').doesNotExist();
+    postable1.isComplete = true;
+    await settled();
```
Aierie, line 96: This is interesting. The tests fail if I remove this, but I can't see what asynchronous work is going on that would require settled - is it simply the DOM changes?

Aierie (approved):

packages/web-client/tests/integration/components/workflow-thread-test.ts
lukemelia: @Aierie `settled` waits for any "test waiters" to finish, including any queued tasks on the Ember run loop. An alternative that might be clearer would be to wait for the DOM state that we expect.

Aierie: I see, thanks. So these waiting items might block ember from rerendering?

lukemelia: No, the `settled` test waiter "waits" for Ember to complete rendering (among other things). There are a bunch of layers here, but you can have a look at https://github.com/emberjs/ember-test-helpers/blob/master/addon-test-support/%40ember/test-helpers/settled.ts as a starting point.

That being said, this conversation has convinced me I should use an approach that is more intention-revealing, so I'm going to make that change.

PR #1567, opened 2021-04-16
Card Pay styling updates

burieberry:
A bunch of small updates including:
- Update header connected button styling, remove layer labels, adjust network names to be two words
- Add link on the homepage and broken styles fix
- Workflow messages broken styles fix
- Edit prose styles
button-login-state


lukemelia (approved):
:shipit:

lukemelia:
Rebased. Also, I notice that the `card-pay-prose` is not class is no longer used. If that is intentional, should we also remove the CSS?

burieberry:
> Rebased. Also, I notice that the `card-pay-prose` is not class is no longer used. If that is intentional, should we also remove the CSS?

Ok, removed now because they don't really apply to the expandable banner and the thread message. We might bring them back as app-wide styles if we find ourselves writing the same styles repeatedly for various html tags.

PR #1568, opened 2021-04-19
Feature/layer one wallet selection

Aierie:
CS-539

Adds a radio with several wallets to connect to. Code for wallet connections to follow in a separate PR.

Also modifies an existing test - we should show this to people who are not connected:
image

Existing test tests for:
image


lukemelia (approved):
Do you expect `CardPay::LayerOneWalletSelection` to be used from anywhere else besides `CardPay::DepositWorkflow::ConnectLayerOne`? If not, I suggest nesting it underneath that component. If so, it is good where it is.

Aierie:
> Do you expect `CardPay::LayerOneWalletSelection` to be used from anywhere else besides `CardPay::DepositWorkflow::ConnectLayerOne`? If not, I suggest nesting it underneath that component. If so, it is good where it is.

Might be used in the main CardPay dashboard, when the user clicks the button on the left:

image

Will leave it like this for now then.

PR #1569, opened 2021-04-19
Add a11y testing and fix some errors

burieberry:
- Update boxel version
- Add accessibility tests
- Related fixes
- Make modal more responsive
- Focus modal and allow scroll and keyboard scroll

Visual changes:
modal-small-focus
Scroll bottom:
modal-small-scroll

### DIFF
--- a/packages/web-client/app/components/card-pay/layer-one-connect-modal/index.css
+++ b/packages/web-client/app/components/card-pay/layer-one-connect-modal/index.css
@@ -1,4 +1,22 @@
+.layer-one-connect-modal__scroll-wrapper {
+  height: 100%;
+  margin: 0 var(--boxel-sp-sm);
+  padding: var(--boxel-sp-lg) var(--boxel-sp-sm);
+  overflow-y: auto;
+}
+
+.layer-one-connect-modal__card:focus {
+  outline-color: transparent;
+  outline-offset: -5px;
+  box-shadow: 0 0 0 2px var(--boxel-blue);
+}
+
+.layer-one-connect-modal__card:focus:not(:focus-visible) {
+  box-shadow: none;
+}
+
 .layer-one-connect-modal__card {
   height: auto;
 }
 
+/* TODO: Close button */
--- a/packages/web-client/app/components/card-pay/layer-one-connect-modal/index.hbs
+++ b/packages/web-client/app/components/card-pay/layer-one-connect-modal/index.hbs
@@ -1,13 +1,18 @@
 <Boxel::Modal
+  @isOpen={{@isOpen}}
   @onClose={{@onClose}}
-  style={{css-var boxel-modal-max-width="45rem"}}
+  style={{css-var boxel-modal-max-width="45rem" boxel-modal-offset-top="0px"}}
   class="layer-one-connect-modal"
   data-test-layer-one-connect-modal
 >
-  <CardPay::LayerOneConnectCard
-    class="layer-one-connect-modal__card"
-    @onComplete={{@onClose}}
-    @suppressHeader={{true}}
-    @suppressConnectingCard={{true}}
-  />
+  <div class="layer-one-connect-modal__scroll-wrapper">
+    <CardPay::LayerOneConnectCard
+      class="layer-one-connect-modal__card"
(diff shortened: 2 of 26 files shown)


packages/web-client/app/templates/card-pay/balances.hbs
```diff
@@ -8,8 +8,9 @@
   </Boxel::Button>
 </section>
 
-{{#if this.isShowingDepositWorkflow}}
-  <Boxel::Modal @onClose={{set this "isShowingDepositWorkflow" false}}>
-    <CardPay::DepositWorkflow />
-  </Boxel::Modal>
-{{/if}}
\ No newline at end of file
+<Boxel::Modal
```
Aierie, line 11: I think we should check with design if they would prefer closing on Escape or not, but we can leave that for when the design team starts playing with the UI since this seems like a fairly straightforward change. Will add this to CS-627.

packages/web-client/app/styles/app.css
```diff
@@ -0,0 +1,32 @@
...
+img {
+  max-width: 100%;
+  height: auto;
+}
+
+a {
+  display: inline-block;
+}
+
+ul {
```
Aierie, line 28: Do we want to overwrite `ol` too?

packages/web-client/app/components/card-pay/layer-one-connect-modal/index.css
```diff
@@ -1,4 +1,26 @@
...
+
+/* remove */
 .layer-one-connect-modal__card {
-  max-height: 48rem;
+  min-height: 40rem;
+  padding: var(--boxel-sp-lg);
 }
 
+
+/* TODO: Close button */
```
lukemelia, line 26: Should this comment be left in? The design didn't have a close button AFAIK.

packages/web-client/app/components/card-pay/layer-one-connect-modal/index.hbs
```diff
@@ -1,10 +1,17 @@
...
-  </Boxel::CardContainer>
+  <div class="layer-one-connect-modal__scroll-wrapper">
+    <Boxel::CardContainer
+      class="layer-one-connect-modal__card"
+      tabindex="0"
+      {{did-insert this.focus}}
+    >
+      Connect...
+    </Boxel::CardContainer>
+  </div>
```
lukemelia, line 16: Would we prefer to put a scroll wrapper around a CardContainer, or a way to make a CardContainer scrollable?

packages/web-client/app/components/card-pay/layer-two-connect-modal/index.hbs
```diff
@@ -1,12 +1,17 @@
...
-    @suppressHeader={{true}}
-    @includeExplanations={{true}}
-  />
+  <div class="layer-two-connect-modal__scroll-wrapper">
+    <CardPay::LayerTwoConnectCard
+      class="layer-two-connect-modal__card"
+      @suppressHeader={{true}}
+      @includeExplanations={{true}}
+      tabindex="0"
+      {{did-insert this.focus}}
```
lukemelia, line 14: In my experience, if we don't have a test around this focus, we will very likely break it when we introduce animations.

lukemelia (approved):

packages/web-client/app/components/card-pay/layer-one-connect-modal/index.css
burieberry: There is an X in the top right in the design that's labelled "Modal Version". Might be a later version of the design.

packages/web-client/app/templates/card-pay/balances.hbs
burieberry: I think the escape key for modal is standard for accessibility. (Focus trap is also standard, but we haven't implemented it yet as you mentioned.) https://www.w3.org/TR/wai-aria-practices/#dialog_modal

packages/web-client/app/components/card-pay/layer-two-connect-modal/index.hbs
burieberry: Good experience. 👍  I'll add

PR #1570, opened 2021-04-20
Acceptance tests include layer 1 token balances

lukemelia:
Fixes #CS-539


lukemelia (commented):
@Aierie a couple of comments

PR #1574, opened 2021-04-22
Disconnect wallet

Aierie:


packages/web-client/app/utils/web3-strategies/kovan.ts
```diff
@@ -79,6 +81,26 @@ export default class KovanWeb3Strategy implements Layer1Web3Strategy {
     }
   }
 
+  async disconnect(): Promise<void> {
+    if (this.currentProviderId === 'metamask') {
```
Aierie, line 85: @lukemelia Metamask doesn't seem to have a good way to disconnect from the Dapp side. Wondering if we should show a specific message to tell the user to disconnect from the Dapp?

lukemelia: Ideally, yes, but as long as we remove the connection from the dapp side, it's OK for now I think.

Aierie: Will leave it like this for now, and bring this up with the design team in the next meeting, just in case there are specific requirements/user expectations.

lukemelia (changes requested):
What if any test coverage should we have of this?

Aierie:
> What if any test coverage should we have of this?

This is related to:
- showing/hiding cards based on connection status
- dashboard button/address

Showing/hiding cards based on connection status is yet to be implemented. Dashboard button/address can be worked on now.

lukemelia:
@Aierie  ready for you to take another look and then review. Per discussion in design-internal, the handling for disconnections in the deposit flow will likely change, so this is "for now"

packages/web-client/app/models/workflow/postable-collection.ts
```diff
@@ -49,6 +49,13 @@ export default class PostableCollection {
       }
 
       if (!post.isComplete) {
+        // clear out remaining posts from #includedPostables and #excludedPostables,
+        // in case we have a situation where we went backward
+        for (let post of this.postables.slice(i, -1)) {
```
Aierie, line 54: `Array.slice(i, -1)` always leaves out the last item of an array. This means that a milestone that has one `WorkflowCard`, A, as its last item, and another, B, before it, will have A automatically marked as completed if B's `onIncomplete` method is called and then B is completed. For this workflow, I think this doesn't affect anything, but implementation might need to change for the 'issue card' workflow.

Aierie: Example here: https://github.com/cardstack/cardstack/tree/example/incompletion-bug

lukemelia: Good catch. I pushed an update with an improved approach.

Aierie: Thanks @lukemelia, it works nicely! I've added a test for the Workflow calling Milestone `resetFrom` with the right arguments. Let's merge if [this commit](https://github.com/cardstack/cardstack/pull/1574/commits/1f6e78a384942beeca73161c2ea269b49bdf5c67) looks ok?

Aierie: added some detail to the epilogue test.

PR #1576, opened 2021-04-22
Show layer 2 balance

lukemelia:
Fixes #CS-631

### DIFF
--- a/packages/web-client/app/utils/web3-strategies/sokol.ts
+++ b/packages/web-client/app/utils/web3-strategies/sokol.ts
@@ -16,7 +16,7 @@ export default class SokolWeb3Strategy implements Layer2Web3Strategy {
   provider = new WalletConnectProvider({
     chainId: this.chainId,
     rpc: {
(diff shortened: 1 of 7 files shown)


Aierie (approved):
Seems good!

packages/web-client/app/utils/web3-strategies/sokol.ts
```diff
@@ -16,8 +16,9 @@ export default class SokolWeb3Strategy implements Layer2Web3Strategy {
   provider = new WalletConnectProvider({
     chainId: this.chainId,
     rpc: {
-      77: 'https://sokol.poa.network',
+      77: 'https://sokol.stack.cards',
     },
+    infuraId: undefined,
```
Aierie, line 21: What does `infuraId: undefined` do? I don't seem to get errors when I remove this

lukemelia: Good catch. I added that while troubleshooting. Will remove.

PR #1578, opened 2021-04-26
Introduce WorkflowSession to allow a workflow session to have state

lukemelia:
- supports the deposit source/destination card passing along info to the deposit amount card, etc.


Aierie (approved):
Works for me!

PR #1579, opened 2021-04-26
Repurpose layer one connect card for modal context launched from header nav Connect button

lukemelia:


packages/web-client/app/components/card-pay/layer-one-connect-modal/index.css
```diff
@@ -1,4 +1,4 @@
 .layer-one-connect-modal__card {
-  max-height: 48rem;
+  max-height: 22.8rem;
```
lukemelia, line 2: @Aierie this is a fragile solution to the problem of sizing the modal height properly. Do you have any suggestions on how to tackle this?

Aierie: You could replace it with `height:auto`. This happens because of `height:100%` on `Boxel::CardContainer`.

Aierie (approved):

packages/web-client/app/components/card-pay/layer-one-connect-modal/index.css
Aierie: Would also add `top: 50%; transform: translateY(-50%);` for centering.

PR #1582, opened 2021-04-28
Initialize layer 1

Aierie:
CS-655

Store wallet provider and try to reconnect when page is reloaded.


lukemelia (approved):
Looks good. I'd prefer to see this tested, but we haven't come up with a good way to test one of these strategies yet.

PR #1583, opened 2021-04-28
Token Bridge Step 1 functionality

lukemelia:
- share deposit transaction receipt via WorkflowSession
- Update Boxel


Aierie (approved):
This works!

PR #1584, opened 2021-04-28
Deposit amount card

Aierie:
CS-541

Deposit amount card that currently mainly supports the happy path. Will aim to add error handling and input validation in a future PR. Note that there is a missing line separating the two CTAs - this is because I missed out splattributes on the `Boxel::CtaBlock` component - this should be updated in a future release of Boxel.

image


packages/web-client/app/components/card-pay/deposit-workflow/transaction-amount/index.hbs
```diff
@@ -32,8 +32,13 @@
       >
       <div class="card-pay-transaction-amount__amount-field-input-container">
         {{svg-jar this.currentTokenDetails.icon class="card-pay-transaction-amount__amount-field-currency-icon" role="presentation"}}
+        {{!-- 
+          .boxel-input is a hack to get the boxel styles.  
+          depending on how we plan to handle validation of this input,
+          we may need to change Boxel's input to use a more DDAU pattern
+        --}}
         <input
```
Aierie, line 40: I'm not sure how we want to handle the input validation/filtering, but it doesn't seem easy to use Boxel's input component to do input filtering since Ember's `<Input>` component auto-updates its `@value`.

packages/web-client/app/components/card-pay/deposit-workflow/transaction-amount/index.ts
```diff
@@ -39,7 +40,7 @@ const tokenDetails: {
 };
 
 class CardPayDepositWorkflowTransactionAmountComponent extends Component<CardPayDepositWorkflowTransactionAmountComponentArgs> {
-  @tracked amount = 0;
+  @tracked amount = '0';
```
Aierie, line 43: We're storing the amount as a string and validating it before we update it when an `input` event comes in, to prevent floating point issues. We convert it to a `FixedNumber` when we want to do any operations on it.

packages/web-client/app/utils/token.ts
```diff
@@ -1,6 +1,15 @@
-import { BigNumber } from '@ethersproject/bignumber/lib/bignumber';
+// ideally we would want to just use parseEther from @ethersproject/units
+// but there seems to be a typescript incompatibility between BigNumber produced by that
+// and BigNumber imported from @ethersproject/bignumber
+import { BigNumber, FixedNumber } from '@ethersproject/bignumber';
 
-export function amountInWei(amount: number, decimals: number = 18) {
-  let dec = BigNumber.from(10).pow(BigNumber.from(decimals));
-  return BigNumber.from(amount).mul(dec);
+export function amountInWei(amount: FixedNumber, decimals: number = 18) {
```
Aierie, line 6: `amountInWei` seems to only be used in a few places that receive the value to relay. Though we're storing this value as a string, I didn't feel good passing a string to this function. Passing a `FixedNumber` felt like a guarantee of more structure by the time it got here - though I'm not very sure if this is true.

packages/web-client/app/components/card-pay/deposit-workflow/transaction-amount/index.hbs
Aierie: This uses the `boxel-input` class directly to get styles from Boxel's input.

packages/web-client/app/components/card-pay/deposit-workflow/transaction-amount/index.ts
Aierie: (Will definitely need more validation)

burieberry (approved):
Looks good. There are some design changes in the xd, but we can merge this version, and create another ticket for those (and/or to discuss if they're necessary).

PR #1586, opened 2021-04-29
Style memorialized layer 2 wallet card

lukemelia:
image


burieberry (commented):
In the screenshot there's an * on the "Balance" label, which makes me think there should be a footnote. We can use the semantic html for creating footnotes, if there's supposed to be one. Otherwise, looks good.

PR #1588, opened 2021-04-29
Render bridge explorer button in token bridge progress card

lukemelia:
Just function, not yet designed.


lukemelia:
@burieberry You mentioned you were considering what to work on next. Once this is merged, there is a task you could pick up which is the layout and styling of this card.

burieberry (approved):

Aierie (approved):

PR #1589, opened 2021-04-29
Add cardpay-sdk and cardpay-cli

habdelra:


habdelra:
Until we integrate the cardpay protocol and gnosis services into the mono repo, I'm using the launch configs in VSCode to confirm that the SDK is working--and all the calls are operating correctly right now

mattmcmanus:
@habdelra this recommendation is probably a bit late, but if there plans to continue to invest in the CLI it might be worth exploring OCLIF: https://oclif.io/docs/introduction. It's the framework that the heroku cli is built on and it's impressive

habdelra:
@mattmcmanus LOL, I just spent the day ripping out oclif. it is specifically the lib that does not play nice with mono repos which required Dominic to make it its own repo--which I'm now decommissioning.

mattmcmanus:
Ha! Oh boy.

PR #1591, opened 2021-04-30
Update deposit (transaction-setup) card UI

burieberry:
Changes:
- Brought over the ActionContainer component to this repo (It's easier to make changes, as needed. We can also use the one in boxel, I'm fine with it either way.)
- Deposit (Transaction Setup) card template and styling

"Continue" button is disabled until a selection is made:
deposit-data-entry

Selection made:
deposit-selected

View-only mode after pressing "Continue":
deposit-view

### DIFF
--- a/packages/web-client/app/components/card-pay/deposit-workflow/transaction-setup/token-option/index.css
+++ b/packages/web-client/app/components/card-pay/deposit-workflow/transaction-setup/token-option/index.css
@@ -1,0 +1,82 @@
+.token-option {
+  display: grid;
+  grid-template-columns: auto 1fr 1fr;
+  align-items: center;
+  gap: var(--boxel-sp);
+  padding: var(--boxel-sp);
+  border-radius: var(--boxel-border-radius);
+  box-shadow: 0 0 0 1px var(--boxel-light-400);
+  transition: box-shadow var(--boxel-transition);
+}
+
+.token-option:hover {
+  box-shadow: 0 0 0 1px var(--boxel-dark);
+}
+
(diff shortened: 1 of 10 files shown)


Aierie:
This needs the token balance in USD. I'll make a PR for that. I think we should raise the frequency of getting exchange rates in USD with the design team to see if there are any requirements or implications.

packages/web-client/app/components/card-pay/deposit-workflow/transaction-setup/token-option/index.css
```diff
@@ -0,0 +1,82 @@
...
+.token-option:hover {
+  box-shadow: 0 0 0 1px var(--boxel-dark);
+}
+
+.token-option--checked,
+.token-option:focus,
+.token-option:focus-within {
+  box-shadow: 0 0 0 2px var(--boxel-highlight);
+  outline-color: transparent;
+  outline-width: 0px;
```
Aierie, line 21: We can leave the outline-width as is for Windows High Contrast Mode, which ignores box-shadow.

https://sarahmhigley.com/writing/whcm-quick-tips/

lukemelia:
We can merge without the USD conversions, and fill them in with the helper after.

burieberry:
> We can merge without the USD conversions, and fill them in with the helper after.

I agree. Also merging without the Depot address, there's a linear ticket for that [CS-662](https://linear.app/cardstack/issue/CS-662/fetch-existing-depot-to-show-on-deposit-sourcetarget-selection-card)

PR #1592, opened 2021-04-30
Refactored the web-client to use the cardpay-sdk

habdelra:
I tested this out using the new cardpay-sdk and was able to successfully bridge tokens into the latest v0.1.2 card protocol deployment (here are the bridge tokens from this update https://alm-test-amb.herokuapp.com/42/0xaf433952d57088f2bede1a26084f0714b04fbc60a678bff451de6a6a9ccdd859). @lukemelia note the 3rd commit "Refactored to use cardpay-sdk" was specifically the changes that I made to the web-client. most of the rest of the work was around wrangling ES6 to play nice with CJS so we can use the sdk from node and from the browser as well as making a few updates to make things more pleasant.

The important thing to note is that all network specific addresses, constants (like the block explorer URL), as well as ABI's now come from the SDK--in addition to the actual functionality itself. This should hopefully make creating a new prepaid card from the tokens you bridged a piece of cake. As well as you can always double check your work using the CLI, which will run thru the same code that your web client uses when interacting with the card protocol.

Also:
> +149 −1,237 Lines of code 🎉 🌮 🎉


lukemelia (approved):
👏 👏 👏  This is terrific!

PR #1593, opened 2021-05-03
assign provider on pre-connected layer one card

Aierie:
CS-653

- Adds `currentProviderId` property to the `Layer1WebStrategy` type
- Assigns provider for layer one connection card by checking `layer1Network.strategy.currentProviderId` if there is a connected layer 1 account.

### DIFF
--- a/packages/web-client/app/components/card-pay/layer-one-connect-card/index.ts
+++ b/packages/web-client/app/components/card-pay/layer-one-connect-card/index.ts
@@ -27,7 +27,7 @@ class CardPayDepositWorkflowConnectLayer1Component extends Component<CardPayDepo
   @service declare layer1Network: Layer1Network;
   @reads('layer1Network.hasAccount') declare hasAccount: boolean;
   @tracked isWaitingForConnection = false;
-  @tracked currentWalletProviderId = '';
+  @tracked radioWalletProviderId = '';
 
   constructor(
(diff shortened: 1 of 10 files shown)


lukemelia (changes requested):
Let's add a test.

packages/web-client/app/components/card-pay/layer-one-connect-card/index.ts
```diff
@@ -36,24 +36,23 @@ class CardPayDepositWorkflowConnectLayer1Component extends Component<CardPayDepo
...
     if (this.hasAccount) {
       next(this, () => {
-        this.currentWalletProviderId =
-          this.layer1Network.strategy.currentProviderId || '';
         this.args.onComplete?.();
       });
     }
   }
-  get currentWalletProvider(): WalletProvider | undefined {
+  get connectedWalletProvider(): WalletProvider | undefined {
```
Aierie, line 43: When writing the test for this I realised that the `LayerOneConnectCard` component  could get its provider from the network's strategy directly, so it's now separated from the selection state.

lukemelia: Tangential benefit of writing tests!

lukemelia (approved):
:shipit: !

PR #1595, opened 2021-05-03
Remove 'about' tab; refactor connection modal and fix outside click

burieberry:
- Remove "About" tab from Card Pay header
- Redirect "card-pay" route to "card-pay.balances" route
- Refactor "Connection" modal so we don't have to write the same styles twice
- Improve outside click for the modals
- Center the modal container and adjust sizing
- Mobile styles not implemented on the dashboard yet, but allow everything to wrap and stack, since currently there's some overflowing content which makes the page unusable on mobile


packages/web-client/app/components/card-pay/layer-connect-modal/index.ts
```diff
@@ -0,0 +1,35 @@
...
+  }
+  @task *closeOnConnectedTask() {
+    if (this.args.name === 'layer1') {
+      yield this.layer1Network.waitForAccount;
+    } else {
+      yield this.layer2Network.waitForAccount;
+    }
+    this.args.onClose();
+  }
+  @action focus(element: any) {
```
lukemelia, line 30: Seems like `element` should be typed as `HTMLElement`

packages/web-client/app/templates/card-pay.hbs
```diff
@@ -14,7 +14,6 @@
     />
     <CardPay::HeaderNav
       @tabs={{array
-        (hash title="About Card Pay" route="card-pay.index" icon="home")
```
lukemelia, line 17: We should make sure that visiting `/card-pay` results in a redirect to `card-pay/balances` now.

lukemelia (changes requested):
I think we should address the routing issue I flagged before merging. Also, on a meta note: the work in this PR is not reflected as in-progress in Linear. You have discretion to take on the work that you judge needs to be done, but you should create visibility for the rest of the team by reflecting that in Linear.

lukemelia:
FYI, I rebased on main and force pushed to fix an issue where I couldn't successfully build this branch.

Aierie (approved):
Looks good, would feel comfortable merging after the changes Luke requested! 

Some of the modal changes would be nice in Boxel's Modal, like centering and focus.

burieberry:
> I think we should address the routing issue I flagged before merging. Also, on a meta note: the work in this PR is not reflected as in-progress in Linear. You have discretion to take on the work that you judge needs to be done, but you should create visibility for the rest of the team by reflecting that in Linear.

Yeah, that's true. This was work that started with "let me fix this one very little thing" (removing the about tab) and then turned into a full digression, which is usually what happens. I'll try to stick to the task at hand, and create future tickets on Linear for the stuff that come up...

PR #1596, opened 2021-05-03
Use action container component

burieberry:
- Uses the ActionContainer component created in #451 across the relevant Card Pay cards. 
- Incorporates ThreadMessage updates from latest Boxel release

See the styling differences for the cards below for different card modes:

Data-entry mode:
data-entry-view

Memorialized view:
memorialized-view

### DIFF
--- a/packages/web-client/app/components/card-pay/deposit-workflow/supplier-badge/index.hbs
+++ b/packages/web-client/app/components/card-pay/deposit-workflow/supplier-badge/index.hbs
@@ -1,3 +1,6 @@
-<Boxel::CardContainer>
+<ActionCardContainer
+  @header="Badge"
+  @isComplete={{true}}
+>
   Supplier Badge
-</Boxel::CardContainer>
+</ActionCardContainer>
--- a/packages/web-client/app/components/card-pay/deposit-workflow/supplier-badge/index.hbs
+++ b/packages/web-client/app/components/card-pay/deposit-workflow/supplier-badge/index.hbs
@@ -1,6 +1,6 @@
 <ActionCardContainer
   @header="Badge"
-  @isComplete={{true}}
+  @isComplete={{@isComplete}}
 >
   Supplier Badge
 </ActionCardContainer>
--- a/packages/web-client/app/components/workflow-thread/index.css
+++ b/packages/web-client/app/components/workflow-thread/index.css
@@ -1,5 +1,5 @@
 .workflow-thread {
-  --inner-max-width: 36.6rem; /* ~585px */
+  --inner-max-width: 36.25rem; /* 580px */
   --outer-max-width: 43.75rem; /* 700px; */
 }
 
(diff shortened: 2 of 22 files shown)


packages/web-client/app/components/card-pay/deposit-workflow/supplier-badge/index.hbs
```diff
@@ -1,3 +1,6 @@
-<Boxel::CardContainer>
+<ActionCardContainer
+  @header="Badge"
+  @isComplete={{true}}
```
Aierie, line 3: I would go with `@isComplete` like the other `ActionCardContainers` in this PR rather than hardcoding `true`, in case the behaviour/purpose of this card changes.

packages/web-client/app/components/workflow-thread/index.css
```diff
@@ -1,11 +1,29 @@
...
   text-align: center;
 }
 
+.workflow-thread__card > .action-card-container {
+  max-width: var(--outer-max-width);
+  box-shadow: 0 15px 30px rgba(0, 0, 0, 0.15);
+  transition: all var(--boxel-transition);
+}
+
+.workflow-thread__card--view-only > .action-card-container {
```
Aierie, line 19: `view-only` isn't accurate here. You can often go back and edit things in a workflow. How about `is-complete`, same as the argument that controls whether this class is applied?

Aierie: Would also replace `.workflow-thread__card > .action-card-container` and `.workflow-thread__card--view-only > .action-card-container` with `workflow-thread__card` and `workflow-thread__card--view-only`, and remove `box-shadow`. Box shadow styling is repeated here and in action container, and we want the margins to apply to the containing element to prevent the card exceeding the bounds of the thread:

image

Aierie: There's also an issue with the margins that I think we can iron out later - the right margin is not really present once the viewport shrinks a bit.

image

Aierie (approved):
I think this works, there are some issues around slightly smaller viewports related to the margins of a card, but we can probably fix those when polishihng.

packages/web-client/app/components/workflow-thread/index.css
burieberry: I couldn't replicate this, but I think this is actually closer to how it's supposed to look in smaller viewports. We can check while doing the mobile version

burieberry: > There's also an issue with the margins that I think we can iron out later - the right margin is not really present once the viewport shrinks a bit.
> 
> image

I couldn't replicate this, but I think this is actually closer to how it's supposed to look in smaller viewports. We can check while doing the mobile version

burieberry: > Would also replace `.workflow-thread__card > .action-card-container` and `.workflow-thread__card--view-only > .action-card-container` with `workflow-thread__card` and `workflow-thread__card--view-only`, and remove `box-shadow`. Box shadow styling is repeated here and in `ActionCardContainer`, and we want the margins to apply to the containing element to prevent the card exceeding the bounds of the thread:
> 
> image

The problem is, when the margin is only on the containing element, the latency causes the switch from one mode to the next to be janky. So I can add a note here to find a better way to implement the transition. Ember-animated would probably resolve this issue, but we haven't added it in yet.

burieberry: I'll push a different solution to this ^

lukemelia (approved):
No issues from me -- I'll leave it to you to hash out the styling questions.

packages/web-client/app/components/workflow-thread/index.css
burieberry: > Would also replace `.workflow-thread__card > .action-card-container` and `.workflow-thread__card--view-only > .action-card-container` with `workflow-thread__card` and `workflow-thread__card--view-only`, and remove `box-shadow`. Box shadow styling is repeated here and in `ActionCardContainer`, and we want the margins to apply to the containing element to prevent the card exceeding the bounds of the thread:
> 
> image

This crop issue should be resolved now. Let me know if it's not.

PR #1597, opened 2021-05-04
mock token to usd

Aierie:
We need a way to convert tokens to USD for #1591 and #1584. This attempts to do that using a helper + a service that manages exchange rates.

### DIFF
--- a/packages/web-client/app/helpers/format-token-amount.ts
+++ b/packages/web-client/app/helpers/format-token-amount.ts
@@ -1,13 +1,10 @@
 import Helper from '@ember/component/helper';
-import { inject as service } from '@ember/service';
-import TextFormattingService from '../services/text-formatting';
 import { BigNumber } from '@ethersproject/bignumber';
 import { formatUnits } from '@ethersproject/units';
 
 type FormatTokenAmountHelperParams = [BigNumber];
 
 class FormatTokenAmountHelper extends Helper {
-  @service declare textFormatting: TextFormattingService;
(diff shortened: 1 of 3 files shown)


Aierie:
As this is only a mock prior to @habdelra finishing the API for getting exchange rates, will hardcode static values for this implementation.

habdelra:
sure, ultimately you'll query the blockchain instead of a web service for the rates, as CARD token rates will only be available to us via the blockchain. and as such I'll be adding a new API in the cardpay-sdk to obtain these.

habdelra:
This does imply that you have to connect to your wallet before any kind of rate information is available to the UI (since the wallet provides the gateway to make queries to the blockchain). I'm not sure if this causes you grief. it does mean that if you want to use an Ember service you'll need to consider 2 states since ember services have a lifetime that begins when the app boots: 
1. before the wallet is connected
2. after the wallet is connected

Aierie:
> This does imply that you have to connect to your wallet before any kind of rate information is available to the UI (since the wallet provides the gateway to make queries to the blockchain). I'm not sure if this causes you grief. it does mean that if you want to use an Ember service you'll need to consider 2 states since ember services have a lifetime that begins when the app boots:
> 
> 1. before the wallet is connected
> 2. after the wallet is connected

Good to know, thanks @habdelra!

Aierie:
Closing temporarily, since API PR is up. Will reopen after it's merged.

PR #1598, opened 2021-05-04
SDK: wait for bridging

lukemelia:
Also adds a CLI command and wires the dapp to use this

### DIFF
--- a/packages/cardpay-cli/README.md
+++ b/packages/cardpay-cli/README.md
@@ -6,27 +6,43 @@ CLI tool for basic actions in Cardpay
 
 # Commands
 - [Commands](#commands)
-  - [`yarn cardpay bridge <TOKEN_ADDRESS> <AMOUNT> --network=NETWORK [--mnemonic=MNEMONIC]`](#yarn-cardpay-bridge-token_address-amount---networknetwork---mnemonicmnemonic)
+  - [`yarn cardpay bridge <AMOUNT> <TOKEN_ADDRESS> --receiver=<RECEIVER> --network=NETWORK [--mnemonic=MNEMONIC]`](#yarn-cardpay-bridge-amount-token_address---receiverreceiver---networknetwork---mnemonicmnemonic)
+  - [`yarn cardpay await-bridged <FROM_BLOCK> [RECIPIENT] --network=NETWORK [--mnemonic=MNEMONIC]`](#yarn-cardpay-await-bridged-from_block-recipient---networknetwork---mnemonicmnemonic)
   - [`yarn cardpay prepaidcard-create <SAFE_ADDRESS> <TOKEN_ADDRESS> <amounts..> --network=NETWORK [--mnemonic=MNEMONIC]`](#yarn-cardpay-prepaidcard-create-safe_address-token_address-amounts---networknetwork---mnemonicmnemonic)
   - [`yarn cardpay safes-view [ADDRESS] --network=NETWORK [--mnemonic=MNEMONIC]`](#yarn-cardpay-safes-view-address---networknetwork---mnemonicmnemonic)
 
 
 
-## `yarn cardpay bridge <TOKEN_ADDRESS> <AMOUNT> --network=NETWORK [--mnemonic=MNEMONIC]`
+## `yarn cardpay bridge <AMOUNT> <TOKEN_ADDRESS> --receiver=RECEIVER --network=NETWORK [--mnemonic=MNEMONIC]`
 
 Bridge tokens from L1 wallet to L2 safe
 
 ```
 USAGE
-  $ yarn cardpay bridge <tokenAddress> <amount> --network=NETWORK [--mnemonic=MNEMONIC]
+  $ yarn cardpay bridge <amount> <tokenAddress> --receiver=RECEIVER --network=NETWORK [--mnemonic=MNEMONIC]
 
 ARGUMENTS
-  TOKEN_ADDRESS   The token address of the token to bridge
   AMOUNT          Amount in ether you would like bridged
-  NETWORK         The network to use ("kovan" or "mainnet")
+  TOKEN_ADDRESS   The token address of the token to bridge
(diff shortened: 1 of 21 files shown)


habdelra:
I'm fine with the change to specify the receiver address in the token bridge SDK in order to bridge to a different wallet in L2.

lukemelia:
@habdelra I've got the new command to the point that I'm getting an error `The current provider doesn't support subscriptions: HDWalletProvider` As a Web3 Provider Specialist, :) do you have a suggested course of action?

habdelra:
So, AFAIK, the HDWalletProvider is only being used in the CLI. This wallet provider is a mnemonic wallet which we are not using in the web-client. The HDWalletProvider in the CLI I setup to use an HTTPProvider for web3 calls. Like I was mentioning in our chat, the HTTPProvider does not work with subscriptions. Can you confirm that you are seeing this when using the CLI? I wasn't really planning on extending the CLI to use a web socket provider since I wasn't too sure how that would even work from the CLI (waiting for arbitrary events). If you are getting this error from the web-client, then I have many more questions for you :-)

habdelra:
Also, dumb question, the URL you are providing to your provider here, https://github.com/cardstack/cardstack/blob/main/packages/web-client/app/utils/web3-strategies/sokol.ts#L39,  does it start with "https://" or does it start with "wss://"? I added a new constant called "rpcWssNode" to get the wss provider URL (some clients like the cli need the http URL, and others like the web-client will want the wss URL)

packages/cardpay-cli/await-bridged.ts
```diff
@@ -0,0 +1,28 @@
...
+  fromBlock: number,
+  recipient?: string
+): Promise<void> {
+  let web3 = new Web3(
+    new HDWalletProvider({
+      chainId: networkIds[network],
+      mnemonic: {
+        phrase: mnemonic,
+      },
+      providerOrUrl: new HttpProvider(await getConstant('rpcNode', network)),
```
habdelra, line 17: This is why you are getting the errors around the subscription not working--HttpProvider doe snot support subscriptions. This HttpProvider I had to do a bunch of patching to get it to work with the HDWalletProvider. We could try doing something similar with a web socket provider--but it will be like a day or 2 of time to get that working I would imagine if its anything like the HttpProvider...

packages/cardpay-cli/README.md
```diff
@@ -6,27 +6,43 @@ CLI tool for basic actions in Cardpay
...
 
-## `yarn cardpay bridge <TOKEN_ADDRESS> <AMOUNT> --network=NETWORK [--mnemonic=MNEMONIC]`
+## `yarn cardpay bridge <AMOUNT> <TOKEN_ADDRESS> --receiver=RECEIVER --network=NETWORK [--mnemonic=MNEMONIC]`
 
 Bridge tokens from L1 wallet to L2 safe
 
 ```
 USAGE
-  $ yarn cardpay bridge <tokenAddress> <amount> --network=NETWORK [--mnemonic=MNEMONIC]
+  $ yarn cardpay bridge <amount> <tokenAddress> --receiver=RECEIVER --network=NETWORK [--mnemonic=MNEMONIC]
```
habdelra, line 22: I'd recommend a positional arg within a sub command like we are doing for `amount` and `tokenAddress`

packages/cardpay-cli/index.ts
```diff
@@ -2,41 +2,72 @@
...
+    });
     yargs.positional('tokenAddress', {
       type: 'string',
       default: '0x4F96Fe3b7A6Cf9725f59d353F723c1bDb64CA6Aa', // Kovan DAI
       description: 'The token address (defaults to Kovan DAI)',
     });
-    yargs.positional('amount', {
-      type: 'number',
-      description: 'Amount of tokens you would like bridged (*not* in units of wei)',
+    yargs.option({
```
habdelra, line 51: parhaps we make this positional to match the other args? I left the overarching args for all the commands as options.

packages/cardpay-cli/await-bridged.ts
lukemelia: I tried it with the WebsocketProvider. It required similar patches to the HttpProvider and then worked for basic operations, but HDWalletProvider itself doesn't support subscriptions, so that path was a dead end. Instead, I added a polling fallback for when the provider doesn't support subscriptions.

packages/cardpay-cli/README.md
lukemelia: Done ✅

packages/cardpay-cli/await-bridged.ts
habdelra: Hmmm, yeah I’ve seen that some providers make assumptions around downstream providers (it stacks like middleware), which can mess things up. That is very unfortunate to hear. However I don’t thing you should give up on it. The HDWallet provider is just for the CLI. For the web client you’ll be using the wallet connect provider and I think that one will work with subscriptions. It’s unfortunate that the selection of provider by the client can prevent us from using specific SDK methods, but maybe we can use polling as a fall back when subscriptions aren’t  supported

packages/cardpay-sdk/README.md
```diff
@@ -22,22 +24,23 @@ This is a package that provides an SDK to use the Cardpay protocol.
...
   - [`RewardPool.withdraw` (TBD)](#rewardpoolwithdraw-tbd)
+- [`getBlockHeight`](#getBlockHeight)
 - [`getAddress`](#getaddress)
 - [`getConstant`](#getconstant)
 - [`networkIds`](#networkids)
 - [ABI's](#abis)
 
-## `TokenBridge`
-The `TokenBridge` API is used to bridge tokens into the layer 2 network in which the Card Protocol runs. The `TokenBridge` class should be instantiated with your `Web3` instance that is configured to operate on a layer 1 network (like Ethereum Mainnet or Kovan).
+## `TokenBridgeForeignSide`
```
habdelra, line 33: good call!

habdelra (approved):
👍  LGTM

PR #1600, opened 2021-05-05
Add Exchange Rates API

habdelra:
This introduces exchange rates API's. Because the math can be very sensitive when dealing with ethereum numbers, especially when you have floating points involved, this API will do the math for you instead of just returning a rate. These rates are a snapshot of the oracles from today (our partners didn't deploy any oracles in sokol, so we have manual feeds that emulate the structure of our partner's contracts--Chainlink and DIA). I did a test of all the rates and they look totally sane. (DAI is just a hair more valuable than USD--you'll notice when you start getting the value of more than 100 DAI)

### DIFF
--- a/packages/cardpay-cli/README.md
+++ b/packages/cardpay-cli/README.md
@@ -10,9 +10,9 @@ CLI tool for basic actions in Cardpay
   - [`yarn cardpay await-bridged <FROM_BLOCK> [RECIPIENT] --network=_NETWORK [--mnemonic=MNEMONIC]`](#yarn-cardpay-await-bridged-from_block-recipient---network_network---mnemonicmnemonic)
   - [`yarn cardpay prepaidcard-create <SAFE_ADDRESS> <TOKEN_ADDRESS> <amounts..> --network=NETWORK [--mnemonic=MNEMONIC]`](#yarn-cardpay-prepaidcard-create-safe_address-token_address-amounts---networknetwork---mnemonicmnemonic)
   - [`yarn cardpay safes-view [ADDRESS] --network=NETWORK [--mnemonic=MNEMONIC]`](#yarn-cardpay-safes-view-address---networknetwork---mnemonicmnemonic)
-  - [`yarn cardpay usd-price <TOKEN> <AMOUNT> --network=NETWORK [--mnemonic=MNEMONIC]`](#yarn-cardpay-usd-price-token-amount---networknetwork---mnemonicmnemonic)
-  - [`yarn cardpay eth-price <TOKEN> <AMOUNT> --network=NETWORK [--mnemonic=MNEMONIC]`](#yarn-cardpay-eth-price-token-amount---networknetwork---mnemonicmnemonic)
-  - [`yarn cardpay oracle-updated-at <TOKEN> --network=NETWORK [--mnemonic=MNEMONIC]`](#yarn-cardpay-oracle-updated-at-token---networknetwork---mnemonicmnemonic)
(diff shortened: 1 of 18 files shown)


packages/cardpay-cli/README.md
```diff
@@ -59,3 +62,42 @@ ARGUMENTS
...
+
+ARGUMENTS
+  TOKEN     The token symbol (without the .CPXD suffix)
+  AMOUNT    The amount of the specified token (not in units of wei)
+  NETWORK   The network to use ("sokol" or "xdai")
+  MNEMONIC  (Optional) Phrase for mnemonic wallet. Also can be pulled from env using MNEMONIC_PHRASE
+```
+
+## `yarn cardpay oracle-updated-at <TOKEN> --network=NETWORK [--mnemonic=mnmonic]`
+This returns the date that the oracle was last updated for the specified token.
```
lukemelia, line 93: Is this the only oracle we are likely to be concerned with? If not, perhaps `exchange-oracle-updated-at` or `price-oracle-updated-at` would be clearer

packages/cardpay-cli/index.ts
```diff
@@ -68,6 +77,47 @@ const { network, mnemonic = process.env.MNEMONIC_PHRASE, tokenAddress, amount, a
...
+    'Get the USD value for the USD value for the specified token in the specified amount',
+    (yargs) => {
+      yargs.positional('token', {
+        type: 'string',
+        description: 'The token symbol (without the .CPXD suffix)',
+      });
+      yargs.positional('amount', {
+        type: 'string',
+        description: 'The amount of the specified token (*not* in units of wei)',
+      });
```
lukemelia, line 91: It might be convenient to have `amount` default to `1`

packages/cardpay-cli/utils.ts
```diff
@@ -0,0 +1,15 @@
+import HDWalletProvider from 'parity-hdwallet-provider';
+import Web3 from 'web3';
+import { HttpProvider, getConstant, networkIds } from '@cardstack/cardpay-sdk';
+
+export async function getWeb3(network: string, mnemonic: string): Promise<Web3> {
```
lukemelia, line 5: Nice!

packages/cardpay-sdk/sdk/exchange-rate.ts
```diff
@@ -0,0 +1,49 @@
...
+}
+
+// because BN does not handle floating point, and the numbers from ethereum
+// might be too large for JS to handle, we'll use string manipulation to move
+// the decimal point. After this operation, the number should be safely in JS's
+// territory.
+function safeFloatConvert(rawAmount: BN, decimals: number): number {
+  let amountStr = rawAmount.toString();
+  return Number(`${amountStr.slice(0, -1 * decimals)}.${amountStr.slice(-1 * decimals)}`);
+}
```
lukemelia, line 49: @Aierie you may find this interesting based on our conversation yesterday

lukemelia (approved):
Looks good. I left a few ideas -- feel free to apply or ignore!

habdelra:
These are all great suggestions, thx!

packages/cardpay-cli/index.ts
habdelra: Actually we'd want to default to 10<sup>18</sup>, since everything is in units of wei. but yeah, that is a good suggestion

habdelra: or wait--duh, from teh CLI we use non-wei, it's the SDK that uses wei--nevermind

PR #1601, opened 2021-05-05
Transaction status steps UI

burieberry:


packages/web-client/app/components/card-pay/deposit-workflow/transaction-status/index.css
```diff
@@ -0,0 +1,47 @@
...
+  line-height: calc(24/18);
+}
+
+.transaction-status__steps {
+  color: var(--boxel-dark);
+  font: 600 var(--boxel-font);
+  letter-spacing: var(--boxel-lsp-xs);
+}
+
+.transaction-status__steps > li > span {
```
Aierie, line 22: Made a task in the backlog to update Boxel to have more flexible height for progress steps: 

https://linear.app/cardstack/issue/CS-746/progress-steps-height-should-adjust-to-its-content

packages/web-client/app/components/card-pay/deposit-workflow/transaction-status/index.hbs
```diff
@@ -2,36 +2,46 @@
...
+      <span class="transaction-status__step-status">
+        {{#if (eq progressStep.index 0)}}
+          <Boxel::Button @as="anchor" @size="extra-small" @href={{this.depositTxnViewerUrl}} target="_blank" data-test-etherscan-button>
+            View on Etherscan
+          </Boxel::Button>
+        {{/if}}
+        {{#if (eq progressStep.index 1)}}
+          <Boxel::Button @as="anchor" @size="extra-small" @href={{this.bridgeExplorerUrl}} target="_blank" data-test-bridge-explorer-button>
+            View on Bridge Explorer
+          </Boxel::Button>
```
Aierie, line 33: not sure if this is necessary these days - but do we need to add `rel="noopener"` to links with `target="_blank"`?

Aierie (approved):

packages/web-client/app/components/card-pay/deposit-workflow/transaction-status/index.hbs
Aierie: https://mathiasbynens.github.io/rel-noopener/

Seems like most browsers we plan to support have decided that `rel="noopener"` is implicit when `target="_blank"` this year.

Aierie: Not sure about their previous versions though, or if that matters

lukemelia (approved):
👍

PR #1602, opened 2021-05-06
Use BN.js instead of @ethersproject/bignumber

lukemelia:
Hassan sez:

```
The ethereum community has standardized around BN
it's been around a lot longer, and its actually embedded into web3
generally you use web3 or ethers--but not generally both
using truffle greatly pushes us in the web3 direction
if we want to consolidate our contracts into the mono repo, using web3 will make life way easier
(I used bn.js instead of pulling it from web3 because the types in TS worked out more nicely that way--it looks like the typings for web3 are not totally up to date)
```


Aierie (approved):
Thanks! It's nice having an official go-to utility.

Is `web3-utils` necessary if we have `web3` as a dependency?

lukemelia:
>Is web3-utils necessary if we have web3 as a dependency?

Hmm, maybe not. I'll take a look.

habdelra (approved):
LGTM

habdelra:
I dived a bit deeper, and in 2 specific cases web3 does actually depend on ethers-project. 
1. it looks like web3 uses a function from ethers project to parse raw transactions https://github.com/ChainSafe/web3.js/blob/1.x/packages/web3-core-method/src/index.js#L447
2. to encode and decode ABI's here https://github.com/ChainSafe/web3.js/blob/1.x/packages/web3-eth-abi/src/index.js
I still feel good about our choice though, since web3 is the library from the ethereum foundation itself.

PR #1604, opened 2021-05-06
Fix/layer one connection bugs

Aierie:
CS-734 

CS-707 + CS-733 fixed in https://github.com/cardstack/cardstack/commit/0f617c42bdb7fa27b4b5062ed68dd9c3e69e9246
CS-732 fixed in https://github.com/cardstack/cardstack/commit/de17084f2d53bab127e33b98bee71f725b391bfb

Batches some fixes and tests for the layer one connection card.


Aierie:
There is an accessibility issue (contrast of text and background of CTA block) on CI, but not on local. I'll update Boxel for this.

burieberry (approved):

PR #1607, opened 2021-05-07
Feature/confirmation card

Aierie:
Confirmation card for CardPay's deposit workflow.

localhost_4200_card-pay_balances (1)

Styling without logic at https://github.com/cardstack/cardstack/pull/1607/commits/e2e24b4ceeb4e850f18eb6a3de613ab1a88b1def


packages/web-client/app/components/card-pay/deposit-workflow/transaction-setup/index.ts
```diff
@@ -19,18 +20,19 @@ interface token {
...
-    icon: 'dai-token',
-  },
-  {
-    symbol: 'CARD',
-    description: 'ERC-20 Cardstack token',
-    icon: 'card-token',
-  },
-];
+
+const DAI_TOKEN = {
```
lukemelia, line 24: It could be nice to build on the `Token` type in `app/utils/web3-strategies/types.ts` for this.

packages/web-client/app/components/card-pay/deposit-workflow/transaction-amount/index.ts
```diff
@@ -171,6 +171,10 @@ class CardPayDepositWorkflowTransactionAmountComponent extends Component<CardPay
           'relayTokensTxnReceipt',
           transactionReceipt
         );
+        this.args.workflowSession.update(
+          'depositedAmount',
+          this.amountAsBigNumber.toString() // BN is mutable
```
lukemelia, line 176: I don't understand the comment. What is the significance of BN being mutable?

packages/web-client/app/components/card-pay/deposit-workflow/index.ts
```diff
@@ -144,7 +145,7 @@ class DepositWorkflow extends Workflow {
       completedDetail: 'Tokens received on xDai',
     }),
   ];
-  epiloguePostables = [
+  epilogue = new PostableCollection([
```
lukemelia, line 148: Whoops, good catch

lukemelia (changes requested):

packages/web-client/app/components/card-pay/deposit-workflow/transaction-setup/index.ts
Aierie: Yep, was thinking of doing that as a follow-up PR.

packages/web-client/app/components/card-pay/deposit-workflow/transaction-amount/index.ts
Aierie: Since BN is mutable, I was thinking it would be better to add the amount as a string to the shared state of the workflow. We don't mutate it right now, but this is a 'just in case'. Not sure if necesssary.

PR #1610, opened 2021-05-10
Feature/token to usd

Aierie:
CS-757 - A helper to convert DAI and CARD to USD.

### DIFF
--- a/packages/cardpay-sdk/sdk/exchange-rate.ts
+++ b/packages/cardpay-sdk/sdk/exchange-rate.ts
@@ -11,6 +11,17 @@ const ten = new BN('10');
 export default class ExchangeRate {
   constructor(private layer2Web3: Web3) {}
 
+  async getUSDConverter(token: string): Promise<(amountInWei: string) => number> {
+    const oracle = await this.getOracleContract(token);
+    const usdRawRate = new BN((await oracle.methods.usdPrice().call()).price);
+    const oracleDecimals = Number(await oracle.methods.decimals().call());
+
+    return function convertToUsd(amountInWei: string): number {
+      let rawAmount = usdRawRate.mul(new BN(amountInWei)).div(ten.pow(tokenDecimals));
+      return safeFloatConvert(rawAmount, oracleDecimals);
+    };
+  }
+
   async getUSDPrice(token: string, amount: string): Promise<number> {
     let oracle = await this.getOracleContract(token);
     let usdRawRate = new BN((await oracle.methods.usdPrice().call()).price);
--- a/packages/cardpay-sdk/sdk/exchange-rate.ts
+++ b/packages/cardpay-sdk/sdk/exchange-rate.ts
@@ -16,7 +16,7 @@ export default class ExchangeRate {
     const usdRawRate = new BN((await oracle.methods.usdPrice().call()).price);
     const oracleDecimals = Number(await oracle.methods.decimals().call());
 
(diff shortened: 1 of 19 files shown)


packages/cardpay-sdk/sdk/exchange-rate.ts
```diff
@@ -11,6 +11,17 @@ const ten = new BN('10');
 export default class ExchangeRate {
   constructor(private layer2Web3: Web3) {}
 
+  async getUSDConverter(token: string): Promise<(amountInWei: string) => number> {
+    const oracle = await this.getOracleContract(token);
+    const usdRawRate = new BN((await oracle.methods.usdPrice().call()).price);
+    const oracleDecimals = Number(await oracle.methods.decimals().call());
+
+    return function convertToUsd(amountInWei: string): number {
```
Aierie, line 19: Wondering if this should be something like an instance of a class that can internally keep track of when this exchange rate was fetched.

habdelra: I like the approach. although naming it is probably unnecessary, since the name is lost. probably more concise to say:
```ts
  return (amountInWei: string): number => {
    ...
  }
```

habdelra: knowing when to check on chain for the latest rate is a bit tricky. You really want to stay in sync with the contract, otherwise you are going to get weird results if you get out of sync with what is happening on chain. Perhaps there an be an internal polling mechanism that queries the contract every 60 seconds. if it sees that the price is different it can update this function via closure (in which case the `=>` will become ver important).

habdelra: albeit you also want to be cause in that you only ever have one poller at a time (using this 2 should not result in 2 pollers--rather one poller that can update all the instances where it is being used)

habdelra: Also we should definitely bounce this idea above off @lukemelia to see if he has any insight around good approaches here.

habdelra: Completely different approach could be using ethereum events to know when the rate changes. The quirk here is that the addresses that you'll need to listen to the events will originate from contracts we don't control (but the SDK can hide the address details). When the rates change ethereum events are emitted (the annoyance is that it is two totally different events because we have to different exchange rate providers). I think, though, that you'd still need some kind of polling mechanism here, as luke was encountering issues with the `subscribe()` API from web3. In this scenario the timing would be very precise, as you'll have ethereum events that are emitted at exactly when the rate changes.

packages/web-client/app/helpers/token-to-usd.ts
```diff
@@ -0,0 +1,32 @@
...
+  amount: BN | undefined;
+
+  constructor() {
+    super(...arguments);
+    this.tokenToUsdService.register(this);
+  }
+
+  willDestroy() {
+    this.tokenToUsdService.unregister(this);
+  }
```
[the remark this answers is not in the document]
lukemelia: I think @habdelra may be thinking of https://api.emberjs.com/ember/3.25/functions/@ember%2Fdestroyable/registerDestructor, which would be a minor but worthwhile improvement to our helper code here.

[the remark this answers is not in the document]
habdelra: Confirm, that was what I was thinking about

PR #1616, opened 2021-05-11
Specify 'latest' for toBlock when waiting for chain event

lukemelia:
- This fixes an issue where deposit flow hung on TokenBridge transaction status card. Odd because 'latest' is the default for `toBlock` according to the docs


habdelra (approved):
👍

PR #1617, opened 2021-05-11
Fix meta showing on messages w same time

Aierie:
CS-624

Not sure about the interval length for comparison, but I set it to 1 minute because Boxel formats time down to minutes.


lukemelia (approved):
One minute seems reasonable to me 👍

PR #1619, opened 2021-05-12
static next steps component

Aierie:
CS-718 

Scaffolding + styling for the suggested next steps at the end of the deposit workflow.

### DIFF
--- a/packages/web-client/app/components/card-pay/deposit-workflow/next-steps/index.hbs
+++ b/packages/web-client/app/components/card-pay/deposit-workflow/next-steps/index.hbs
@@ -1,1 +1,20 @@
-<Boxel::NextStepsBox />
+<Boxel::NextStepsBox>
+  <div class="card-pay-deposit-button-container">
+    <Boxel::Button 
+      @size="base" 
+      @kind="primary" 
+      {{on "click" this.openNewDepositWorkflow}} 
+      data-test-deposit-next-step="new-deposit"
+    >
+      Make Another Deposit
+    </Boxel::Button>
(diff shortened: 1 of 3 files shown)


packages/web-client/app/components/card-pay/deposit-workflow/next-steps/index.hbs
```diff
@@ -1 +1,20 @@
-<Boxel::NextStepsBox />
\ No newline at end of file
+<Boxel::NextStepsBox>
+  <div class="card-pay-deposit-button-container">
```
Aierie, line 2: @burieberry Had a think about the `<:item>` named block for the `Boxel::NextSteps` component. Considering Safari 14 supports flex gap, it feels better to me to just use the `default` block especially if it's just 2 items. We aren't launching immediately, and Safari 13 has a small market share based on initial search results. We can define our browser support later and add a polyfill if necessary (maybe we won't even need one). So here, I'm just using a flex gap container in the default block.

Aierie: CS-788 is a task to remove the  `item` named block from the component. If you're in agreement, let me know and I will pick it up when I'm back.

lukemelia (approved):

packages/web-client/app/components/card-pay/deposit-workflow/next-steps/index.hbs
burieberry: Sure, works for me

burieberry (approved):

PR #1622, opened 2021-05-13
Use updated card chin with connecting state

burieberry:
CS-709, CS-719

- Update boxel version
- Use chin
- Update layer-1 connecting state chin
- Fix alignment regression on transaction setup card

**Note:** Cancel button is TBD. Need to check with Gardella when he returns. Leaving the button as is for now.

connecting-chin


lukemelia (approved):
LGTM! 🚢

PR #1624, opened 2021-05-14
Transaction-amount card styling update

burieberry:
CS-712

I removed any other refactor, here are the tx-amount-card-only styling changes:

unlock
unlock-expanded
unlocking
memorialized
memorialized-expanded

### DIFF
--- a/packages/web-client/app/components/card-pay/deposit-workflow/transaction-amount/balance-view/index.css
+++ b/packages/web-client/app/components/card-pay/deposit-workflow/transaction-amount/balance-view/index.css
@@ -1,70 +1,138 @@
-.card-pay-token-balance-view {
-  display: flex;
-  align-items: start;
-  padding: var(--boxel-sp);
-  box-shadow: 0 0 0 2px var(--boxel-light-400);
+/* details - summary container styles */
+.balance-view {
+  --details-summary-height: 3.125rem; /* 50px */
+  --details-open-height: 11rem; /* ~180px */
+
+  position: relative;
+  border: 1px solid var(--boxel-light-400);
+  box-shadow: 0 1px 1px rgba(0, 0, 0, 0.35);
   border-radius: var(--boxel-border-radius);
+  height: var(--details-summary-height);
+  overflow: hidden;
+  transition: height var(--boxel-transition);
 }
 
-.card-pay-token-balance-view__currency-details, .card-pay-token-balance-view__balance {
-  flex-shrink: 0;
+.balance-view[open] {
+  height: var(--details-open-height);
+  overflow: auto;
+}
+
+.balance-view__summary {
+  cursor: pointer;
   display: flex;
-  flex-direction: column;
+  align-items: center;
+  gap: 0 var(--boxel-sp);
+  height: var(--details-summary-height);
+  padding: var(--boxel-sp-xxs) calc(var(--boxel-sp) + 3rem) var(--boxel-sp-xxs) var(--boxel-sp);
+  font: 600 var(--boxel-font-sm);
+  letter-spacing: var(--boxel-lsp-xs);
+  transition: padding var(--boxel-transition),
+    font var(--boxel-transition);
 }
 
-.card-pay-token-balance-view__currency-icon {
+.balance-view__summary:focus {
+  outline-color: var(--boxel-highlight);
+  outline-offset: -5px;
+}
+
+.balance-view__summary--memorialized {
+  flex-direction: column;
+  align-items: flex-start;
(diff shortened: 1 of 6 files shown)


packages/web-client/app/components/card-pay/deposit-workflow/transaction-amount/balance-view/index.css
```diff
@@ -1,70 +1,127 @@
...
+  overflow: auto;
+}
+
+.balance-view__summary {
+  cursor: pointer;
   display: flex;
+  align-items: center;
+  gap: 0 var(--boxel-sp);
+  height: var(--details-summary-height);
+  padding: var(--boxel-sp-xxs) var(--boxel-sp);
```
Aierie, line 26: Would do something like `var(--boxel-sp-xxs) calc(var(--boxel-sp) + 3rem) var(--boxel-sp-xxs) var(--boxel-sp);` to keep the Show/Hide part from overlapping with the contents as space gets less.

packages/web-client/app/components/card-pay/deposit-workflow/transaction-amount/balance-view/index.css
```diff
@@ -1,70 +1,127 @@
...
+  font: 600 var(--boxel-font-sm);
+  line-height: calc(20/13);
+  transition: padding var(--boxel-transition),
```
Aierie, line 29: The padding transition feels like it's lagging behind the rest of the content. I tried tinkering with it for a bit, `transition: font 0.05s;`, skipping padding, feels the most comfortable for me.

packages/web-client/app/components/card-pay/deposit-workflow/transaction-amount/index.hbs
```diff
@@ -48,13 +60,10 @@
           </div>
         </div>
         </Boxel::Field>
+        <Boxel::Field @label="Approximate value">
+          ≈ $ {{token-to-usd this.currentTokenSymbol this.amountAsBigNumber}} USD
```
Aierie, line 64: This field shouldn't have the ≈

packages/web-client/app/components/card-pay/deposit-workflow/transaction-amount/index.hbs
```diff
@@ -1,60 +1,69 @@
...
+          @label="Amount to deposit"
+          @fieldMode="edit"
+          class="transaction-amount__field"
+        >
+        <div class="transaction-amount__input-container transaction-amount__input-token-balance">
+          {{svg-jar this.currentTokenDetails.icon class="transaction-amount__currency-icon" role="presentation"}}
+          <Boxel::Input
+            class="transaction-amount__input"
+            id="deposit-amount-input"
+            value={{this.amount}}
```
Aierie, line 54: Not using `@value` with this input feels wrong, because Boxel's input uses Ember's `<Input>`, which says in its documentation that we should not use the html attribute.

https://api.emberjs.com/ember/release/classes/Ember.Templates.components/methods/Input?anchor=Input

This does bypass 2 way binding, but I don't think we should be doing this. I was wondering if we should change Boxel's input not to use Ember's input but use the plain HTML input internally - we need more control than a magically working component gives us.

Aierie (changes requested):

packages/web-client/app/components/card-pay/deposit-workflow/transaction-amount/index.hbs
lukemelia: +1 to using plain input rather than Ember's <Input> in most cases.

packages/web-client/app/components/card-pay/deposit-workflow/transaction-amount/index.hbs
```diff
@@ -1,60 +1,70 @@
...
+          @balance={{format-token-amount this.currentTokenBalance}}
+          @balanceInUsd={{token-to-usd this.currentTokenSymbol this.currentTokenBalance}}
+          @isComplete={{@isComplete}}
         />
-        <div class="card-pay-transaction-amount__amount-field-currency-text">
-          {{this.currentTokenSymbol}}
-        </div>
-      </div>
       </Boxel::Field>
+      {{#if (or this.isUnlocking this.isUnlocked)}}
```
lukemelia, line 26: Since we're using this `or` more than once in the template, consider using a getter with an intention-revealing name instead.

lukemelia (approved):
When you merge, please either use Squash & Merge this or clean up the commit history first to avoid commits that are just reverts and fixups.

PR #1627, opened 2021-05-17
add infura id for staging

Aierie:
CS-736

Adds the infura id for staging, with only staging site url on its allowlist.


lukemelia (approved):
👍

PR #1628, opened 2021-05-18
Enhancement/refactor web client tokens

Aierie:
CS-765

- Refactors tokens to use a set of utils and types defined centrally
- Separates responsibility for presentational information (`TokenDisplayInfo`) and contract-related information (`TokenContractInfo`) for tokens


lukemelia (approved):
Good cleanup. Did you consider unifying TokenContractInfo and TokenDisplayInfo?

Aierie:
> Good cleanup. Did you consider unifying TokenContractInfo and TokenDisplayInfo?

Thanks @lukemelia! I did. But decided to leave them separate because: 

1. right now, it looks like display info is not used where contract info is used and vice versa
2. the addresses for contracts are network-specific, and I couldn't think of a great way to introduce addresses as optional that wouldn't reduce the Typescript benefits for the contract side.
3. Tokens like eth that we have some display information for are not used the same way as bridgeable tokens.

Did you have a structure in mind that would combine the two nicely?

Aierie:
Possibly a `Token` class that has a `contractFor` method that returns a `TokenContractInfo` instance for a specific network, but otherwise behaves like a `TokenDisplayInfo` class?

Aierie:
Merging first. May follow up if can think of a good way to combine contract and display info for tokens.

PR #1629, opened 2021-05-18
Upgrade Boxel to 0.7.2

Aierie:
Fixes issue with static loading indicator on buttons, and allows use of NextStepsBox without having to add a containing element that applies a cluster layout.


lukemelia (approved):
👏

PR #1630, opened 2021-05-18
Fix/misc web client UI bugs

Aierie:
CS-825
CS-833

Small ui fixes for the deposit workflow.


Aierie:
Some color contrast bugs should be resolved first - the expandable banner small text does not have enough contrast with its background

Aierie:
> Some color contrast bugs should be resolved first - the expandable banner small text does not have enough contrast with its background

Should be solved with latest commit.

burieberry (approved):

PR #1631, opened 2021-05-18
Transfer safe tokens

habdelra:
This SDK and CLI command is used to transfer tokens from a safe to an arbitrary address. We will use this to load CARD.CPXD into the relay gas payer so that this account can air drop CARD.CPXD on new prepaid cards. (and perhaps other administrative functions)

### DIFF
--- a/packages/cardpay-sdk/sdk/utils.ts
+++ b/packages/cardpay-sdk/sdk/utils.ts
@@ -2,6 +2,7 @@ import Web3 from 'web3';
 import { TransactionReceipt } from 'web3-core';
 import { networks } from './constants';
 import { Contract, EventData, PastEventOptions } from 'web3-eth-contract';
+import { getConstant } from './constants.js';
 
 export async function networkName(web3: Web3): Promise<string> {
   let id = await web3.eth.net.getId();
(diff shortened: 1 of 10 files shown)


habdelra:
@lukemelia just an FYI, I needed to split out the signature stuff into their own libs for reasons other than what you are working on, but I think this is definitely something you will want in the SDK.

packages/cardpay-sdk/sdk/utils.ts
```diff
@@ -62,3 +117,178 @@ export function waitForEvent(contract: Contract, eventName: string, opts: PastEv
...
+      },
+      (err, response) => {
+        if (err) {
+          return reject(err);
+        }
+        resolve(response?.result);
+      }
+    );
+  });
+  const sig = result.replace('0x', '');
```
lukemelia, line 279: @habdelra OK if we extract this RSV conversion to a separate util method, so signTypedData returns the hex string? For the auth stuff, I'd like to send to the server as the hex string.

habdelra: Totally fine by me

PR #1632, opened 2021-05-18
[CS-803] Add formatting utilities to SDK

DominicSherman:
These are being pulled directly from what is used in Rainbow wallet. I'm not sure yet what exactly the business logic should be, so I figured this would be a good place to start. This will give us consistency across the mobile app and dapp and we can make changes to the logic as needed


habdelra:
Instead of using bignumber.js lets use bn.js. We've standardized everywhere else in our web-client and in the SDK with bn.js. bn.js should provide all the same capabilities as bignumber.js

habdelra:
Let's limit this API to just the formatting of numbers and leave out all the arithmetic functions.

habdelra (changes requested):
- use bn.js instead of bignumber.js
- limit the API to just functions related to formatting. We already use bn.js for big number arithmetic.

habdelra:
@lukemelia @Aierie @burieberry  I wanted to bring you guys attention to this PR. One thing you'll notice is that we are using bignumber.js instead of bn.js (and instead of ethers/BigNumber). @DominicSherman and I chatted about it, and the reason we are not using bn.js, is that bn.js is not meant for floating point (because ethereum has no floating point). bn.js is strictly for big *integers*. For the currency formatting, floating point becomes very important to handle, so for that reason we are using bignumber.js here (as well as providing arithmetic capabilities for large floating point numbers). This was actually the module that rainbow used for handling floating point math and formatting. So as soon as you guys need to drop down into floating point for number formatting, you should use this module.

habdelra (approved):

lukemelia:
@habdelra Would you suggest we move to bignumber.js for all bn.js usage in the dapp? Or just use it when we are in formatting territory?

habdelra:
We should stick with BN for interacting with contracts. Passing in a floating point into a solidity function is a recipe for disaster. We should only use bignumber.js for formatting and and downstream activity from interacting with a contract.

PR #1636, opened 2021-05-19
Enhancement/wallet connection close button

Aierie:
CS-714

Added a close button to the wallet connection modal(s) by somewhat hackily wrapping it in yet another card container. 

When the close button was added, realised that there was a problem with focus trapping. We were programmatically focusing the modal when someone opened it, which made sense because otherwise someone could accidentally change something in the background of the page without realising. However, it's still possible for someone to tab into the background without realising.

Added a focus trap library to handle this behaviour and prevent focus from leaving the modal via tabbing. There are caveats to this:
- Programmatic focus may not show the focus ring because it does not match the  `:focus-visible` selector we use to only show keyboard-triggered focus and ignore mouse-triggered focus. When a modal opens, you generally want the first actionable element to be selected, and have its focus styles visible, but it's not immediately clear how to do this programmatically with `focus-visible`. See discussions on focus behaviour in:
        - https://github.com/WICG/focus-visible/issues/88
        - https://github.com/w3c/csswg-drafts/issues/5885
- The library itself does not seem to like an unselected radio as the last focusable element within a trap. This might have to do with roving tabindex - the library (tabbable) will try to select only a checked radio. To avoid this issue, I set the starting selected value of the radio to metamask, providing a checked radio, and an enabled button after it.

### DIFF
--- a/packages/web-client/app/components/card-pay/layer-connect-modal/index.css
+++ b/packages/web-client/app/components/card-pay/layer-connect-modal/index.css
@@ -33,28 +33,28 @@
   }
 }
 
-.layer-connect-modal__close-button-container {
-  display: flex; 
-  justify-content: flex-end; 
-  padding-top: var(--boxel-sp-sm);
-  padding-right: var(--boxel-sp-sm);
+.layer-connect-modal__card {
+  position: relative;
+  padding-top: calc(var(--boxel-sp-sm) + 0.75rem);
+  outline-color: transparent;
+  outline-style: solid;
 }
 
 .layer-connect-modal__close-button  {
-  padding: 0;
+  position: absolute;
+  right: 0;
+  top: 0;
+  z-index: 1;
+  padding: var(--boxel-sp-sm);
+  height: calc(var(--boxel-sp-sm) * 2 + 0.75rem);
+  width: calc(var(--boxel-sp-sm) * 2 + 0.75rem);
+  line-height: 0;
+
   background: transparent;
   border: none;
-  height: 0.75rem;
-  width: 0.75rem;
-  line-height: 0;
   cursor: pointer;
 
   --icon-color: var(--boxel-dark);
-}
-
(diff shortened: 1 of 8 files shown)


Aierie:
As a next step, instead of wrapping a card in another card in order to tack on a close button, we can create a generic modal version of a card that has the close behaviour built in.

Aierie:
> Does the vertical scrolling experience hold together with this change? i.e. when the browser height is less than the modal content height.

Yep, it does.

burieberry (changes requested):
The focus trap looks great.

In terms of the container resizing and scrolling, I've noticed some regressions.

My recommendation is not to touch the containers at all (since that took so much effort to get just right), but adding an approx. 40x40 px button by absolutely positioning it top-right on a relatively positioned container. The X icon can be centered inside the button. This will ensure that the button is large enough to click on.

Aierie:
> The focus trap looks great.
> 
> In terms of the container resizing and scrolling, I've noticed some regressions.
> 
> My recommendation is not to touch the containers at all (since that took so much effort to get just right), but adding an approx. 40x40 px button by absolutely positioning it top-right on a relatively positioned container. The X icon can be centered inside the button. This will ensure that the button is large enough to click on.

What regressions are you noticing?

burieberry:
> > The focus trap looks great.
> > In terms of the container resizing and scrolling, I've noticed some regressions.
> > My recommendation is not to touch the containers at all (since that took so much effort to get just right), but adding an approx. 40x40 px button by absolutely positioning it top-right on a relatively positioned container. The X icon can be centered inside the button. This will ensure that the button is large enough to click on.
> 
> What regressions are you noticing?

Hmm weird. The card wasn't resizing and the scroll was off, but now I tried again and it seems fine. In that case I'd just say make the clickable area larger for the button.

burieberry (approved):

PR #1639, opened 2021-05-20
Authentication to hub based on mobile wallet signature

lukemelia:
CS-831

I will follow up this PR to add nonce usage tracking in Postgres. I'd like to review and merge now though.


packages/cardpay-sdk/sdk/utils.ts
```diff
@@ -285,7 +289,7 @@ async function signTypedData(web3: Web3, account: string, data: any): Promise<Si
   // For ethereum valid V is 27 or 28
   // In case V = 0 or 01 we add it to 27 and then add 4
   // Adding 4 is required to make signature valid for safe contracts:
-  // https://gnosis-safe.readthedocs.io/en/latest/contracts/signatures.html#eth-sign-signature
+  // https://docs.gnosis.io/safe/docs/contracts_signatures/
```
lukemelia, line 292: @habdelra the comments don't seem to match the code, unless I am missing something. Can you clarify? (Also, I updated the URL to what I think is the correct -- the URL there was a dead link.)

habdelra: This was copied from Tim's team when they did the POC. I never really paid too much attention this this since everything was working. Looking at the docs though, we don't use `eth_sign` for the signing this like the docs mention (which means adding `4`, rather we use `eth_signTypedData`. I'm guessing you don't need to alter `v` if you are signing with `eth_signTypedData` since this code already works for gnosis safe signatures.

habdelra: Are you encountering issues recovering the address when you use this logic?

habdelra: (for reference, this was the logic that Tim's team used in their POC: https://github.com/cardstack/card-pay/blob/refactor-code/app/services/xdai-network.js#L147-L174)

habdelra (commented):
This looks great, but also do we include in the readme the process for starting the server for a developer? This should probably include how to generate the SERVER_SECRET env var as well.

PR #1641, opened 2021-05-20
Workflow copy/text changes

burieberry:
CS-843

- Text changes
- Make each link in the workflow open in new tab

Note: I've made more places where the chain name is used dynamic, but we're still inconsistent about when we use the dynamic name and when we use the hardcoded names.

### DIFF
--- a/packages/web-client/app/components/card-pay/deposit-workflow/confirmation/index.hbs
+++ b/packages/web-client/app/components/card-pay/deposit-workflow/confirmation/index.hbs
@@ -10,7 +10,7 @@
     <Boxel::CardContainer @displayBoundaries={{true}} class="card-pay-confirmation__transfer-info">
       <div class="card-pay-confirmation__transfer-network">
         <span class="card-pay-confirmation__transfer-network-name">{{this.layer1Network.chainName}}</span>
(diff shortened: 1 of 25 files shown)


lukemelia (approved):

packages/web-client/app/components/card-pay/deposit-workflow/confirmation/index.hbs
```diff
@@ -39,8 +39,8 @@
   <div class="card-pay-confirmation__bridge-container">
     <div class="card-pay-confirmation__bridge-inner">
       Bridge
-      <Boxel::Button @as="anchor" @size="extra-small" href={{this.bridgeExplorerUrl}} class="card-pay-confirmation__bridge-inner-button">
-        View in Bridge Explorer
+      <Boxel::Button @as="anchor" @size="extra-small" href={{this.bridgeExplorerUrl}} target="_blank" rel="noopener" class="card-pay-confirmation__bridge-inner-button">
+        View AMB receipt
```
Aierie, line 43: This is actually changed to "View in Bridge Explorer" in the script. Both of these lead to the same link, and no one seems to know why we use different text.

burieberry: Good catch!

PR #1642, opened 2021-05-20
Add action-card section component for consistent layout and headings

burieberry:
Adds `ActionCardContainer::Section` component, which we can use for the action card inner layout going forward. Also takes a `@title` argument, which renders the section title. Sections can be nested or stacked, such as in the transaction-setup and wallet layer-2 connection cards. 

CS-841


packages/web-client/app/components/card-pay/deposit-workflow/confirmation/index.css
```diff
@@ -109,62 +97,49 @@
...
-  padding: var(--boxel-border-radius) calc(var(--boxel-sp-lg) / 2);
+  padding: 0 var(--boxel-sp-lg);
 
   background-color: var(--boxel-light-300);
 }
 
 .card-pay-confirmation__bridge-inner {
   width: 100%;
-  padding: var(--boxel-sp-xl) calc(var(--boxel-sp-lg) / 2);
+  padding: var(--boxel-sp) var(--boxel-sp-sm);
```
Aierie, line 135: Changing the padding breaks alignment:

image

packages/web-client/app/components/card-pay/layer-one-connect-card/index.hbs
```diff
@@ -6,12 +6,9 @@
...
 >
-  <section class="layer-one-connect-card__body">
-    {{#if this.connectedWalletProvider}}
-      <header aria-label="Layer 1 Connected Header" class="layer-one-connect-card__body-header" data-test-mainnet-connection-header>
-        {{this.connectedWalletProvider.name}}
-        <img class="layer-one-connect-card__body-header-logo" src={{this.connectedWalletProvider.logo}} alt="" role="presentation">
-      </header>
+  {{#if this.connectedWalletProvider}}
+    <ActionCardContainer::Section @title={{this.connectedWalletProvider.name}} class="layer-one-connect-card__body">
+      <img class="layer-one-connect-card__logo" src={{this.connectedWalletProvider.logo}} alt="MetaMask" width="60" height="60">
```
Aierie, line 11: I think there isn't much value in adding alt text to these logos. It is meant to be an accompaniment to the text, which already should have the same value. (this shouldn't be hardcoded).

packages/web-client/app/components/card-pay/layer-two-connect-card/index.hbs
```diff
@@ -2,16 +2,12 @@
...
   ...attributes
 >
   {{#if (eq this.cardState 'memorialized')}}
-    <section class="layer-two-connect-card__body">
-      <header aria-label="Layer 2 Connected Header" class="layer-two-connect-card__body-header">
-        Cardstack
-        <img class="layer-two-connect-card__body-header-logo" src={{this.cardstackLogo}} alt="" role="presentation">
-      </header>
+    <ActionCardContainer::Section @title="Cardstack" class="layer-two-connect-card__body">
+      <img class="layer-two-connect-card__logo" src={{this.cardstackLogo}} alt="Cardstack" width="60" height="60">
```
Aierie, line 10: Similar to the layer one connect card, I think this image shouldn't have alt/be detected by screenreaders.

packages/web-client/app/components/card-pay/deposit-workflow/transaction-status/index.hbs
```diff
@@ -2,11 +2,7 @@
   @header="Token Bridge"
   @isComplete={{@isComplete}}
 >
-  <div class="transaction-status">
-    <p class="transaction-status__description">
-      Bridging tokens to xDai blockchain
-    </p>
-
+  <ActionCardContainer::Section @title="Bridging tokens to xDai blockchain">
```
Aierie, line 5: We've been missing an icon for this header!

packages/web-client/app/components/action-card-container/section/index.hbs
```diff
@@ -0,0 +1,9 @@
+<section class="action-card-section" ...attributes>
+  {{#if @title}}
+    <header class="action-card-section__header">
+      {{@title}}
+    </header>
```
Aierie, line 5: I feel like this section should have logo and icon arguments that are laid out nicely with the header, this feels more comfortable than absolutely positioning logos to align with this header.

image

image

Aierie (changes requested):
I like this change. There is a lot of duplicated code that can be removed. But I feel like we need a bit more flexibility in the API.

packages/web-client/app/components/action-card-container/section/index.hbs
burieberry: @Aierie I made the header change to include icons and added the missing icon. Made the logo a background-image since it only serves a presentation purpose as you've said and we don't need to dynamically manipulate its colors etc like we do with svg icons. I made the ActionCardContainer::Title as a separate component from ActionCardContainer::Section, so Title can be used separately if needed with custom content, but just passing in the same arguments to Section also work. I named it "Title" and not "Header" because we already have a Boxel::Header component.

packages/web-client/app/components/card-pay/deposit-workflow/confirmation/index.css
burieberry: Made a small change here to align it better, but if you want to go back to using the older version let me know. The design seems to have been updated since you worked on this, that's why I made the changes. @Aierie

Aierie: This change seems good! Thanks!

Aierie (approved):

PR #1644, opened 2021-05-24
Enhancement/upgrade boxel

Aierie:
CS-848

- Removes temporary hack to prevent radios from causing sudden scrolls when focused
- Changes final milestone status to "Workflow completed" as per design feedback


lukemelia (approved):

burieberry (approved):
Looks good!

PR #1645, opened 2021-05-25
Enhancement/disconnect event and handling

Aierie:
### CS-860: Adds a way to trigger changes in the app when the Dapp is disconnected
- Added an event listener utility class `DappEvents` that allows consumers to listen to and emit events (modified, simpler [nanoevents](https://github.com/ai/nanoevents))
- Used the `DappEvents` class to emit a 'disconnect' event when disconnection is triggered from the Dapp or the Wallet. Upon disconnection, strategies emit a 'disconnect' event. Layer 1 and 2 network services listen for the 'disconnect' event from strategies, and emit a 'disconnect' event themselves. The `disconnect` event from the services can be listened for with the `Listener` component.
- Added a component `Listener`, which allows the user to specify event listeners. Including this component in your templates will allow you to trigger actions when an object that has the `Emitter` interface defined in `packages/web-client/utils/events.ts` emits them. This component also manages cleanup for event listeners using `registerDestructor` from `@ember/destroyable`.
- Used the `Listener` component in layer one and two connect cards to trigger `@onIncomplete` and hence rollback the thread whenever a disconnection happens, whether from the current tab in the Dapp or outside.

### CS-869: Ensures that a disconnection in Layer 1 is communicated to other tabs via a network-specific BroadcastChannel
- Sends a broadcast channel message with data 'disconnected' on a channel with the same name as the network. On receiving this message, tabs with local connected state will disconnect (clear local connected state and trigger disconnect event listeners).
- Also added a polyfill for `BroadcastChannel`, since Safari does not support it at the moment.


packages/web-client/app/components/card-pay/layer-one-connect-card/index.hbs
```diff
@@ -7,6 +7,11 @@
   ...attributes
 >
   {{#if this.connectedWalletProvider}}
+    <Listener
+      @emitter={{this.layer1Network}}
+      @event="disconnect"
+      @action={{optional @onIncomplete}}
+    />
```
Aierie, line 14: An example of use of the `Listener` component:
- `@emitter` is the service/object to listen to
- `@event` is the name of the event
- `@action` is the triggered callback

packages/web-client/app/utils/events.ts
```diff
@@ -0,0 +1,20 @@
+export type UnbindEventListener = () => void;
+
+export interface Emitter {
+  on(event: string, cb: Function): UnbindEventListener; // eslint-disable-line no-unused-vars
+}
+
+export class DappEvents {
```
Aierie, line 7: @lukemelia I renamed this to avoid the accidental imports of node's event emitter.

lukemelia: I'm a fan of avoiding the accidental imports but don't love the name. How about `SimpleEmitter` or `MinimalEmitter` instead?

packages/web-client/app/utils/web3-strategies/kovan.ts
```diff
@@ -19,13 +19,19 @@ import {
...
+} from '@cardstack/web-client/utils/events';
+import { registerDestructor } from '@ember/destroyable';
 
 const WALLET_CONNECT_BRIDGE = 'https://safe-walletconnect.gnosis.io/';
 
 let cardToken = new TokenContractInfo('CARD', 'kovan');
 
 let daiToken = new TokenContractInfo('DAI', 'kovan');
 
+// emits event
```
lukemelia, line 34: Remove this comment

lukemelia (changes requested):

lukemelia (approved):

PR #1647, opened 2021-05-25
Add merchant registration SDK API and CLI command

habdelra:


habdelra:
Hold off on merging this until the sokol node comes back so that I can verify this works correctly

habdelra:
Ok, this is tested in the latest cardpay-protocol and working 🎉

PR #1648, opened 2021-05-25
Tweak meta display logic

burieberry:
CS-852
Do not hide metadata above a post if a message post is following a card post (if same author and within the same minute)

### DIFF
--- a/packages/web-client/app/components/workflow-thread/postable/index.hbs
+++ b/packages/web-client/app/components/workflow-thread/postable/index.hbs
@@ -11,7 +11,7 @@
     @imgURL={{@postable.author.imgURL}}
     @name={{@postable.author.name}}
     @hideName={{true}}
-    @hideMeta={{postable-meta-identical @postable (object-at @postables (dec @index))}}
+    @hideMeta={{postable-meta-hidden @postable previous=(object-at @postables (dec @index))}}
     data-test-postable={{@index}}
     ...attributes
   >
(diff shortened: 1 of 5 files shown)


Aierie (commented):
I would lean towards creating a separate helper to see if post a has a card. If we keep everything in the same helper, we should probably rename it because it no longer just checks if meta is identical, but if we want to show meta at all. Something like `should-show-meta`? 

Argument order matters for the helper that checks if one post is a card. If we keep it in the same helper, we should not test for the helper being able to work regardless of the order of arguments. It might be good to find ways to enforce each argument's role - which one is first and which one is second, as named arguments/hashes, maybe.

Also sorry! I suggested checking `componentName` earlier, but `postA instanceof WorkflowCard` is probably a more robust way of checking if a post is a card - what do you think?

burieberry:
> I would lean towards creating a separate helper to see if post a has a card. If we keep everything in the same helper, we should probably rename it because it no longer just checks if meta is identical, but if we want to show meta at all. Something like `should-show-meta`?
> 
> Argument order matters for the helper that checks if one post is a card. If we keep it in the same helper, we should not test for the helper being able to work regardless of the order of arguments. It might be good to find ways to enforce each argument's role - which one is first and which one is second, as named arguments/hashes, maybe.
> 
> Also sorry! I suggested checking `componentName` earlier, but `postA instanceof WorkflowCard` is probably a more robust way of checking if a post is a card - what do you think?

Yeah I initially created this as `postable-meta-hidden` helper, but then realized I'm repeating a lot of the code in `postable-meta-identical` and decided to combine them. Now I'm thinking maybe I'll create the separate helper which can maybe reference the other helper or more likely extend it, and update the tests accordingly.

Checking for the class instance is a good idea.

lukemelia (approved):
👍

Aierie (approved):

PR #1649, opened 2021-05-27
Fixes #CS-692 version our API

habdelra:
Fixes #CS-692

@lukemelia I'd like to get your feedback on this. The idea is that I'm comparing the available SDK API versions that we have created against the on-chain protocol version. I'm using a semver Tilde Range https://www.npmjs.com/package/semver to comapre the specific on-chain version with a `~app_api_version` range. For instance if we have versions `0.2.0` and `0.3.0` of the API available, and the on-chain version is `0.2.5`, then we should choose version `0.2.0` of the API to use. The way I approach that is to generate a list in descending order of all the API versions available, and compare the on-chain version to the "~" app semver range, so like:
```js
satisfies("0.2.5", "~0.3.0"); // does not satisfy
satisfies("0.2.5", "~0.2.0"); // does satisfiy
``` 
(in that order)

That means that we need to maintain at the very least an API version for the minor versions that we care about (we should ideally not have more than 2 API versions to choose from at a time, like feature flags, as soon as we have landed the new version in our clients, we should remove the older API version from our SDK at our next opportunity). It also turns out that currently only 3 of our API's are tied to our cardpay protocol contracts. The API's that do not rely on the cardpay protocol contracts I left as-is.

One critique I have is that I think we should obtain all the API's in a consistent way, even the ones that are not versioned. so that means moving to an async getter function that returns the API instance instead of just directly instantiating the API instance.

I can't really test this out yet, though until our sokol node is back up--which is why this is a draft PR and not an actual PR.


lukemelia:
@habdelra I think the semver logic makes sense. I took a swing at an API that has us obtaining all the APIs in a consistent way here: https://github.com/cardstack/cardstack/pull/1650. See how you like it.

habdelra:
I left the both version 0.2.0 and 0.3.0 so we can see what it looks like for multiple versions. going forward I'll remove the n - 1 versions as part of ongoing cleanup

habdelra:
This has been tested out with the live cardpay-protocol in v0.2.0 and v0.3.0 and is working fine

PR #1651, opened 2021-05-27
only show nonzero token balances

Aierie:
CS-849

Omit tokens with zero balance. Show "None" if there are no nonzero balances.


lukemelia (approved):
LGTM

PR #1652, opened 2021-05-27
Enhancement/workflow cancelations

Aierie:
Attempt to implement CS-859. Adds the `cancelation` and `isCanceled` property to workflows.

- `cancelation` is a `PostableCollection` that will be shown at the end of the current thread if `isCanceled` is true.
- A workflow being complete (`isComplete` is true) precludes it from being canceled
- Workflows are expected to be cancelled by their `WorkflowComponent`s - eg. `DepositWorkflowComponent` and `IssuePrepaidCardWorkflowComponent`
- Removed rolling back of thread in response to layer 1 and layer 2 disconnection

Also made some changes to the way we pass data to postables in order to support comparisons between postables in different `PostableCollection`s.

After implementing cancelation, will implement a way to freeze a workflow to prevent inputs/changes after cancelation in this PR (issue in CS-867).

Looking for some feedback on whether this way of cancelling a workflow seems ok.

### DIFF
--- a/packages/web-client/app/components/card-pay/deposit-workflow/index.ts
+++ b/packages/web-client/app/components/card-pay/deposit-workflow/index.ts
@@ -6,6 +6,10 @@ import { Workflow, cardbot } from '@cardstack/web-client/models/workflow';
 import { Milestone } from '@cardstack/web-client/models/workflow/milestone';
 import { WorkflowCard } from '@cardstack/web-client/models/workflow/workflow-card';
 import PostableCollection from '@cardstack/web-client/models/workflow/postable-collection';
+import Layer1Network from '@cardstack/web-client/services/layer1-network';
+import Layer2Network from '@cardstack/web-client/services/layer2-network';
+import { inject as service } from '@ember/service';
+import { action } from '@ember/object';
 
 class DepositWorkflow extends Workflow {
   name = 'Reserve Pool Deposit';
@@ -146,6 +150,40 @@ class DepositWorkflow extends Workflow {
       componentName: 'card-pay/deposit-workflow/next-steps',
     }),
   ]);
+  cancelationMessages = new PostableCollection([
+    new NetworkAwareWorkflowMessage({
+      author: cardbot,
+      message:
+        'It looks like your wallets have been disconnected. If you still want to deposit funds, please start again by connecting your wallets.',
+      includeIf() {
+        let message = this as NetworkAwareWorkflowMessage;
+        return !message.hasLayer1Account && !message.hasLayer2Account;
+      },
+    }),
+    new NetworkAwareWorkflowMessage({
+      author: cardbot,
+      message:
+        'It looks like your mainnet wallet has been disconnected. If you still want to deposit funds, please start again by connecting your wallet.',
+      includeIf() {
+        let message = this as NetworkAwareWorkflowMessage;
+        return !message.hasLayer1Account && message.hasLayer2Account;
+      },
+    }),
+    new NetworkAwareWorkflowMessage({
+      author: cardbot,
+      message:
+        'It looks like your xDai chain wallet has been disconnected. If you still want to deposit funds, please start again by connecting your wallet.',
+      includeIf() {
+        let message = this as NetworkAwareWorkflowMessage;
+        return message.hasLayer1Account && !message.hasLayer2Account;
+      },
+    }),
+    new WorkflowCard({
+      author: cardbot,
+      componentName: 'card-pay/deposit-workflow/workflow-canceled-cta',
+    }),
+  ]);
+
   constructor(owner: unknown) {
     super(owner);
     this.attachWorkflow();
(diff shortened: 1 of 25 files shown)


Aierie:
Use cases for cancelation currently include:
- Disconnection from layer 1 and/or 2 in the deposit workflow
- Disconnection from layer 2 in the prepaid card issuance workflow
- Possibly, not having enough to create a prepaid card in the prepaid card issuance workflow?

packages/web-client/app/components/card-pay/deposit-workflow/index.ts
```diff
@@ -153,11 +158,54 @@ class DepositWorkflow extends Workflow {
...
+        message =
+          'It looks like your xDai chain wallet has been disconnected. If you still want to deposit funds, please start again by connecting your wallet.';
+      }
+
+      let result = new PostableCollection([
+        new WorkflowMessage({
+          author: cardbot,
+          message,
+        }),
+      ]);
```
lukemelia, line 194: Why imperatively add this to the workflow when all the other workflow posts are declared in the definition with conditional logic about whether they should be shown?

packages/web-client/app/models/workflow.ts
```diff
@@ -15,6 +16,7 @@ export abstract class Workflow {
   name!: string;
   milestones: Milestone[] = [];
   epilogue: PostableCollection = new PostableCollection();
+  @tracked cancelation: PostableCollection | undefined;
```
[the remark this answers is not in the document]
Aierie: If `epilogue` is replaced, we will be showing `epilogue` both after a workflow is canceled or completed. This is nice because some of the behaviour from cancelling, we do want from completion too - eg. freezing a thread to prevent changes. I'm up for trying this out!

Aierie:
Will be changing this to define `cancelation` (pending rename) declaratively within the workflow, and controlling whether messages show with `includeIf`. This is because declarative models fit better with persistence, and the ability to edit a workflow, than dynamically-added workflow messages.

packages/web-client/app/components/card-pay/deposit-workflow/index.ts
```diff
@@ -146,18 +150,65 @@ class DepositWorkflow extends Workflow {
...
+  cancelationMessages = new PostableCollection([
+    new NetworkAwareWorkflowMessage({
+      author: cardbot,
+      message:
+        'It looks like your wallets have been disconnected. If you still want to deposit funds, please start again by connecting your wallets.',
+      includeIf() {
+        return (
+          !(this as NetworkAwareWorkflowMessage).hasLayer2Account &&
+          !(this as NetworkAwareWorkflowMessage).hasLayer1Account
+        );
```
lukemelia, line 162:
```suggestion
        let message = this as NetworkAwareWorkflowMessage;
        return !message.hasLayer1Account && !message.hasLayer2Account;
```

packages/web-client/app/components/card-pay/deposit-workflow/index.ts
```diff
@@ -146,18 +150,65 @@ class DepositWorkflow extends Workflow {
...
+    }),
+    new NetworkAwareWorkflowMessage({
+      author: cardbot,
+      message:
+        'It looks like your mainnet wallet has been disconnected. If you still want to deposit funds, please start again by connecting your wallet.',
+      includeIf() {
+        return (
+          !(this as NetworkAwareWorkflowMessage).hasLayer1Account &&
+          (this as NetworkAwareWorkflowMessage).hasLayer2Account
+        );
```
lukemelia, line 173:
```suggestion
        let message = this as NetworkAwareWorkflowMessage;
        return !message.hasLayer1Account && message.hasLayer2Account;
```

packages/web-client/app/components/card-pay/deposit-workflow/index.ts
```diff
@@ -146,18 +150,65 @@ class DepositWorkflow extends Workflow {
...
+    }),
+    new NetworkAwareWorkflowMessage({
+      author: cardbot,
+      message:
+        'It looks like your xDai chain wallet has been disconnected. If you still want to deposit funds, please start again by connecting your wallet.',
+      includeIf() {
+        return (
+          !(this as NetworkAwareWorkflowMessage).hasLayer2Account &&
+          (this as NetworkAwareWorkflowMessage).hasLayer1Account
+        );
```
lukemelia, line 184:
```suggestion
        let message = this as NetworkAwareWorkflowMessage;
        return message.hasLayer1Account && !message.hasLayer2Account;
```

packages/web-client/app/components/card-pay/layer-one-connect-card/index.ts
```diff
@@ -113,7 +114,7 @@ class CardPayDepositWorkflowConnectLayer1Component extends Component<CardPayDepo
 
   @action onDisconnect() {
     this.args.onDisconnect?.();
-    this.args.onIncomplete?.();
+    // this.args.onIncomplete?.();
```
lukemelia, line 117: Should this comment be removed?

packages/web-client/app/components/workflow-thread/index.hbs
```diff
@@ -52,7 +65,7 @@
           />
         </div>
         <div class="workflow-thread__status">
-          {{@workflow.progressStatus}}
+          {{if @workflow.isCanceled "Workflow canceled" @workflow.progressStatus}}
```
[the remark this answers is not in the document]
lukemelia: This comment still applies

lukemelia (changes requested):

PR #1656, opened 2021-05-28
Fix reroute to balances tab and workflow button locations

burieberry:
CS-910

- Fix reroute to balances tab (only reroute from /card-pay to /card-pay/balances)
- Swap workflow buttons ("Deposit Tokens" is on Token Suppliers tab, and "Issue a Prepaid Card" is on Balances tab)


lukemelia (approved):
👍

PR #1658, opened 2021-05-31
disable completed or canceled workflows via a wrapping fieldset

Aierie:
Disable the workflow via a wrapping fieldset. This is a quick solution that disables all input elements within cards. This relies on us using proper html inputs and button elements, however it provides an opportunity to apply the appropriate styling and behaviour to all contained inputs if they use the `:disabled` pseudo-class to apply disabled styles. 

Some issues:
1. Things that require special handling will include non-semantic inputs and interactable items, plus QRCodes.
2. It's easy enough to remove the `disabled` html attribute and continue interacting with the input elements.  Since we don't know if this allows abuses, we should use this together with #1659.
3. This messes up screen readers. 
4. Radio inputs lack good disabled states at the moment.

### DIFF
--- a/packages/web-client/app/components/workflow-thread/index.hbs
+++ b/packages/web-client/app/components/workflow-thread/index.hbs
@@ -6,25 +6,27 @@
     {{#if (has-block 'before-content')}}
       {{yield to='before-content'}}
     {{/if}}
-    {{#each @workflow.visibleMilestones as |milestone i|}}
-      {{#each milestone.visiblePostables as |postable j|}}
-        <WorkflowThread::Postable
-          @postable={{postable}}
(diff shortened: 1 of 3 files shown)


lukemelia:
>It's easy enough to remove the disabled html attribute and continue interacting with the input elements. Since we don't know if this allows abuses, we should use this together with #1659.

There is no abuse vector here because the dapp only allows activities that are either a) acting on blockchain resources the user controls via wallet-approved transactions, and b) API calls to our hub that are authenticated and validated. As a general rule, we must never be reliant on browser HTML, CSS or Javascript for security of resources outside the browser.

Aierie:
Closing in favour of #1658

PR #1659, opened 2021-05-31
disable via action chin

Aierie:
Disables a workflow by controlling the action chins' buttons.

### DIFF
--- a/packages/web-client/app/components/action-chin/index.hbs
+++ b/packages/web-client/app/components/action-chin/index.hbs
@@ -1,0 +1,26 @@
+<Boxel::ActionChin @state={{@state}} @stepNumber={{@stepNumber}} ...attributes>
+    <:default as |a|>
+        {{yield (hash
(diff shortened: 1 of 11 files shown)


packages/web-client/app/components/action-chin/index.hbs
```diff
@@ -0,0 +1,26 @@
+<Boxel::ActionChin @state={{@state}} @stepNumber={{@stepNumber}} ...attributes>
+    <:default as |a|>
+        {{yield (hash
+          ActionButton=(if @frozen (component a.ActionButton disabled=true) a.ActionButton)
```
lukemelia, line 4: Would it be functionally equivalent and more succinct to write:

`ActionButton=(component a.ActionButton disabled=@frozen)`

Aierie: Seems like it! Thanks for pointing it out.

PR #1660, opened 2021-06-01
Add wrappers around workflow models to gradually reveal postables instead of all at once

lukemelia:
https://www.loom.com/share/3c2dddb2777142219267027b647bf270

### DIFF
--- a/packages/web-client/app/models/animated-workflow.ts
+++ b/packages/web-client/app/models/animated-workflow.ts
@@ -6,7 +6,7 @@ import { WorkflowPostable } from './workflow/workflow-postable';
 import { tracked } from '@glimmer/tracking';
 import { taskFor } from 'ember-concurrency-ts';
 import { task } from 'ember-concurrency-decorators';
-import { timeout } from 'ember-concurrency';
+import { rawTimeout } from 'ember-concurrency';
 import { A } from '@ember/array';
 import config from '@cardstack/web-client/config/environment';
 import { buildWaiter } from '@ember/test-waiters';
@@ -100,6 +100,7 @@ export default class AnimatedWorkflow {
 
   constructor(model: Workflow) {
     this.model = model;
+    this.model.animatedWrapper = this;
     this.epilogue = new AnimatedPostableCollection(model.epilogue);
     this.cancelationMessages = new AnimatedPostableCollection(
       model.cancelationMessages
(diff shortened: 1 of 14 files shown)


Aierie:
Edge cases:
- Rolling back the thread when users edit a card (`onIncomplete`)
  - We need to mark messages as incomplete in our animated classes
- The initial loading of messages if the wallet connection steps are completed before starting the workflow
  - Currently this scrolls each time a message is shown and can be a bit janky if we show messages quickly.

Aierie:
This PR does not include CSS changes. I think CSS animations are feasible and can be included in another PR.

packages/web-client/app/models/animated-workflow.ts
```diff
@@ -0,0 +1,244 @@
...
+      model.cancelationMessages
+    );
+    this.milestones = model.milestones.map((m) => new AnimatedMilestone(m));
+    taskFor(this.tickerTask).perform();
+  }
+  @reads('model.name') declare name: string;
+
+  @task
+  *tickerTask() {
+    this.startTestWaiter();
```
lukemelia, line 133: I don't understand why we are starting the waiter once, and stopping it many times.

packages/web-client/app/models/workflow/workflow-card.ts
```diff
@@ -25,8 +26,10 @@ export class WorkflowCard extends WorkflowPostable {
 
   @action onComplete() {
     this.isComplete = true;
+    this.workflow?.animatedWrapper?.startTestWaiter();
```
lukemelia, line 29: This class should not know about animation or test waiters. This should be completely encapsulated in the AnimatedWorkflow. So this line should be something like `this.workflow?.onPostableComplete(this)` and then that should in turn do what needs to be done.

packages/web-client/app/models/workflow/workflow-card.ts
```diff
@@ -25,8 +26,10 @@ export class WorkflowCard extends WorkflowPostable {
...
   }
   @action onIncomplete() {
+    this.workflow?.animatedWrapper?.startTestWaiter();
```
lukemelia, line 32: Same as my comment for complete handling.

packages/web-client/tests/acceptance/deposit-test.ts
```diff
@@ -81,7 +81,8 @@ module('Acceptance | deposit', function (hooks) {
     assert.dom(`${post} [data-test-balance="DAI"]`).containsText('250.50');
     assert.dom(`${post} [data-test-balance="CARD"]`).containsText('10000.00');
 
-    await waitFor(milestoneCompletedSel(0));
```
lukemelia, line 84: Why would this need to change? Same for all similar test changes.

.vscode/settings.json
```diff
@@ -7,7 +7,7 @@
   },
   "[typescript]": {
     "editor.formatOnSave": true,
-    "editor.defaultFormatter": "esbenp.prettier-vscode"
+    "editor.defaultFormatter": "dbaeumer.vscode-eslint"
```
lukemelia, line 10: This probably doesn't belong?

Aierie: Oops. Yep!

packages/web-client/app/models/animated-workflow.ts
```diff
@@ -0,0 +1,287 @@
...
+    // short timeout to prevent moving revealpointer within the same runloop
+    yield rawTimeout(10);
+
+    while (true) {
+      let result = this.revealNext();
+
+      // the last card in these things is not completed so isComplete will never
+      // be true. We check for completion by making sure all the things we need to
+      // show are shown
```
lukemelia, line 169: > the last card in these things is not completed
Why is this true? Does it need to be?

Aierie: I think it doesn't have to be. We could hardcode the "Next steps" CTA to have `isComplete` be true.

Alternatively, we could also subclass `PostableCollection` and have it determine `isComplete` based on whether all necessary postables are visible.

lukemelia: > I think it doesn't have to be. We could hardcode the "Next steps" CTA to have isComplete be true.

This! 👍

Aierie: Prior to this, will be changing the way that styling is applied to workflow cards to allow cards to override default width assignment/allocation by the thread. 

After making "Next steps" completed, will replace this check with `PostableCollection.isComplete`.

These changes will be made in future PRs.

PR #1670, opened 2021-06-04
Add prepaid-card issue funding card UI and display depot balances

burieberry:
- Card UI
- **Fetch depot balances** --> this can probably be moved to layer2Chain web strategy. currently in this card's index file.
- Dropdown --> thinking about adding this in boxel and then moving the css there
- Created a fake depot implementation for the tests
- AccountDisplay component --> this can be used in other places in the app.

Note: This PR is happy path only (assuming user already has a depot and funds in it) 

funding-data-entry
funding-dropdown-open
funding-memorialized

### DIFF
--- a/packages/web-client/tests/acceptance/deposit-test.ts
+++ b/packages/web-client/tests/acceptance/deposit-test.ts
@@ -127,7 +127,9 @@ module('Acceptance | deposit', function (hooks) {
         '[data-test-card-pay-layer-2-connect] [data-test-card-pay-connect-button]'
       )
       .hasText('0x1826...6E44');
-    await waitFor(milestoneCompletedSel(1));
+
+    await settled();
+
     assert
       .dom(milestoneCompletedSel(1))
       .containsText('xDai chain wallet connected');
@@ -428,7 +430,7 @@ module('Acceptance | deposit', function (hooks) {
         '[data-test-card-pay-layer-2-connect] [data-test-card-pay-connect-button]'
       )
       .hasText('0x1826...6E44');
-    await waitFor(milestoneCompletedSel(1));
+    await settled();
     assert
       .dom(milestoneCompletedSel(1))
       .containsText('xDai chain wallet connected');
--- a/packages/web-client/tests/acceptance/issue-prepaid-card-test.ts
+++ b/packages/web-client/tests/acceptance/issue-prepaid-card-test.ts
@@ -96,7 +96,9 @@ module('Acceptance | issue prepaid card', function (hooks) {
         '[data-test-card-pay-layer-2-connect] [data-test-card-pay-connect-button]'
       )
       .hasText('0x1826...6E44');
-    await waitFor(milestoneCompletedSel(0));
+
+    await settled();
+
     assert
       .dom(milestoneCompletedSel(0))
       .containsText('xDai chain wallet connected');


packages/web-client/app/components/card-pay/account-display/index.hbs
```diff
@@ -0,0 +1,66 @@
...
+      {{svg-jar @icon class="account-display__icon" role="presentation"}}
+      <div>
+        <div class="account-display__title-lg" data-test-account-title>
+          {{format-token-amount @balance}} {{uppercase @symbol}}.CPXD
+        </div>
+      </div>
+    {{else}}
+      {{svg-jar @icon class="account-display__icon-sm" role="presentation"}}
+      <div>
+        <div class="account-display__title-sm" data-test-account-title>{{uppercase @symbol}}.CPXD</div>
```
lukemelia, line 13: If this is intended as generic AccountDisplay, perhaps it should not always have the `.CPXD` suffix.

packages/web-client/app/components/card-pay/account-display/index.hbs
```diff
@@ -0,0 +1,66 @@
...
+        <div class="account-display__text-sm" data-test-account-balance>
+          {{format-token-amount @balance}} {{uppercase @symbol}}.CPXD
```
lukemelia, line 15: Same question about suffix.

packages/web-client/app/components/card-pay/account-display/index.hbs
```diff
@@ -0,0 +1,66 @@
...
+          <div class="account-display__text account-display__address-wrapped" data-test-account-address>
+            {{@address}}
+          </div>
+        </div>
+      {{else}}
+        {{svg-jar @icon class="account-display__icon-sm" role="presentation"}}
+        <div>
+          <div class="account-display__title-sm" data-test-account-title>{{@title}}</div>
+          <div class="account-display__text-sm account-display__address account-display__text--outer" data-test-account-address>
+            {{truncate-middle @address 7 4}}
```
lukemelia, line 34: Are `7` and `4` the standard way we use `truncate-middle`? If so, perhaps we should make those the default arguments.

packages/web-client/app/components/card-pay/issue-prepaid-card-workflow/funding-source/index.ts
```diff
@@ -0,0 +1,91 @@
...
+      .perform()
+      .then((depot: { address: string; tokens: [] }) => {
+        if (depot) {
+          this.args.workflowSession.update('depotAddress', depot.address);
+          this.args.workflowSession.update('depotTokens', depot.tokens);
+        }
+      })
+      .then(() => {
+        if (!this.selectedToken) this.chooseSource(this.tokens[0]);
+      });
```
lukemelia, line 51: Maybe we should move the code within these `then` blocks into `fetchDepotTask`?

lukemelia (changes requested):
Left a few comments. Overall, great work here except that it probably should have been two different PRs instead of one larger one.

packages/web-client/app/components/card-pay/issue-prepaid-card-workflow/funding-source/index.ts
Aierie: These should probably be managed within the `Layer2Chain` class, since our balances there should come from the depot too.

packages/web-client/app/components/card-pay/issue-prepaid-card-workflow/funding-source/index.hbs
```diff
@@ -1,19 +1,102 @@
...
+              </div>
+            {{/if}}
+          </:inner>
+        </CardPay::NestedItems>
+      </CardPay::LabeledValue>
+    </ActionCardContainer::Section>
+  </ActionCardContainer::Section>
   <Boxel::ActionChin
     @state={{if @isComplete "memorialized" "default"}}
     @disabled={{@frozen}}
```
Aierie, line 86:
```suggestion
    @disabled={{or @frozen (not this.selectedToken)}}
```

Aierie: Otherwise, `not this.selectedToken` can override `@frozen` after completion/cancelation of the workflow

Aierie: Alternatively, apply `@frozen` to the Edit button?

Aierie: Similar case in https://github.com/cardstack/cardstack/commit/a1da448ece3bb536b49a1155be79568c331497e4

packages/web-client/app/components/card-pay/account-display/index.hbs
burieberry: Removed these, also because it's not decided yet if we want to show CPXD everywhere or in some places or not

packages/web-client/app/components/card-pay/account-display/index.hbs
burieberry: It's actually 6. I don't know where 7 came from. Updated the helper to use default values if none are given

packages/web-client/app/components/card-pay/issue-prepaid-card-workflow/funding-source/index.ts
burieberry: I only had to do it here because it was easier to fetch the real values than create fake ones that didn't error for the UI. Should I do the layer2chain strategy stuff in this PR? I was thinking of refactoring separately since it's kinda hard to ship this UI without any data. I also saw that you created a similar ticket, hope that's not the same thing

packages/web-client/app/components/card-pay/issue-prepaid-card-workflow/funding-source/index.hbs
burieberry: Removed the other disabled condition since there's already a preselected token. Haven't handled the edge cases though

packages/web-client/app/components/card-pay/issue-prepaid-card-workflow/funding-source/index.ts
Aierie: I think a separate PR is good. I'll work on fetching depot balances for the layer two connect card, and make the necessary changes to the use of depots here and in the other cards that use it.

burieberry: if I move this line to fetchDepot, then the dropdown displays 0 balance. I moved it up to the previous block, that works

Aierie: Draft implemented in #1686. Some changes to come + perhaps restructuring, but:

We fetch the depot and store in layer 2 web3 strategy, because we need the depot in order to get the balances. The depot balances are the same as the layer 2 chain balances, so you can get them directly from `this.layer2Network`. As a result, all uses of `fetchDepotTask` should be refactored out in #1686.

PR #1675, opened 2021-06-08
Use a more suitable message for disconnections

Aierie:
As per design feedback, uses a more generic way to tell users that a workflow is canceled because of disconnection.

### Problem
Cancelation messages may not provide an accurate picture of events. Since the messages shown are selected via `includeIf`, they never change after the first event that causes a workflow cancelation. Future changes are not reflected.

### Example
We have three different messages for different scenarios:
- Message A: Layer 1 disconnected
- Message B: Layer 2 disconnected
- Message C: Both layer 1 and layer 2 disconnected

If a user disconnects from Layer 1, they are shown message A. If the user disconnects from layer 2 after this, the message shown is still Message A, not message C.

### Edit
A second, related problem is we base the messages on whether wallets are disconnected, not whether there was a disconnect event. This means that if you start the workflow with only Layer 1 connected, then disconnect Layer 1, we show you the message for having both Layer 1 and Layer 2 disconnected.

This is solvable. We can keep track of disconnect events (as opposed to connected state) in the workflow session and choose messages based on that. However, it seems like it is a lot of trouble for little return.

### DIFF
--- a/packages/web-client/app/components/card-pay/deposit-workflow/index.ts
+++ b/packages/web-client/app/components/card-pay/deposit-workflow/index.ts
@@ -154,33 +154,19 @@ class DepositWorkflow extends Workflow {
     new NetworkAwareWorkflowMessage({
       author: cardbot,
       message:
-        'It looks like your wallets have been disconnected. If you still want to deposit funds, please start again by connecting your wallets.',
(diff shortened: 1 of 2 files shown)


lukemelia (approved):
This feels unfortunate, but I don't think it is a meaningful enough shortcoming to make a lot of changes for.

PR #1677, opened 2021-06-08
Webpack 5 + Break build with test waiters

Aierie:
This should break the build, not for merging.


Aierie:
Closing as @ef4's commit has demonstrated how to fix the issue with using `@ember/test-waiters` in a monorepo with embroider - make sure that there are no duplicate copies of `@ember/test-waiters`.

PR #1681, opened 2021-06-09
Add prepaid card issued confirmation card

backspace:
This fills out the placeholder prepaid card confirmation card.

image

One unconventional aspect is the use of `:where(.prepaid-card)` in CSS to provide zero-specificity styles for the prepaid card so they can be easily overridden, which was needed to include the deeper box shadow from the design.


lukemelia:
> should there be a test selector to assert on the card issuer name? currently asserts that entire postable contains issuer name

Yes. 👍 

> is it preferable to not access card details via @workflowSession.state.* in the template

I think this is acceptable.

> used ActionCardContainer with no action 🤔 but Boxel::Card doesn‘t have ::Section

@burieberry or @Aierie will likely have a good perspective on this

burieberry (approved):
Looks great! 

Regarding your question, I think it's ok to use the ActionCardContainer here. While the naming might suggest that the user should be taking some action, it actually has two states, determined via the `@isComplete` arg. The default state is the "data-entry" state. "Memorialized" state is for the presentation of some action taken or the results of a previous action, which is the case here. After a certain point in the workflow, the user can no longer change the actions taken, which is when we disable the action chin (the card footer with CTA) or we don't have it at all.

lukemelia (approved):

PR #1686, opened 2021-06-10
Fix/layer2 balance from depot

Aierie:
CS-793

- Fetches layer 2 depot and balances within the layer 2 chain class
- Updates the way depot address and balances are accessed
- Refreshes layer 2 balances after bridging
- Adds some "loading" UI when fetching layer 2 balances since it has more latency than layer 1

### DIFF
--- a/packages/web-client/app/utils/web3-strategies/layer2-chain.ts
+++ b/packages/web-client/app/utils/web3-strategies/layer2-chain.ts
@@ -32,7 +32,7 @@ export default abstract class Layer2ChainWeb3Strategy
 
   simpleEmitter = new SimpleEmitter();
 
-  web3!: Web3;
+  web3: Web3 = new Web3();
   #exchangeRateApi!: IExchangeRate;
   #safesApi!: ISafes;
   @tracked depotSafe: DepotSafe | null = null;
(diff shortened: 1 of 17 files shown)


Aierie:
Because there is latency, the balance for layer 2 shows a loading indicator while the depot is being fetched, instead of showing "None".

https://user-images.githubusercontent.com/39579264/121679607-1b5ae880-caeb-11eb-81b2-53c6d7fce293.mov

Aierie:
Fetching depot balances after bridging updates non-zero token balances. This means you have to have some amount of token in your depot in order for refetching balances to show the updated amount.

Edge cases as a result: 
- If a depot is newly created, balance displayed will remain as "None" after refetch
- If the token that was deposited previously had an amount of 0 and hence wasn't displayed, it still won't be displayed after refetch

packages/web-client/app/utils/web3-strategies/layer2-chain.ts
```diff
@@ -124,15 +135,15 @@ export default abstract class Layer2ChainWeb3Strategy
     this.updateWalletInfo([], this.chainId);
   }
 
-  private async refreshBalances() {
-    let raw = await this.getDefaultTokenBalance();
-    this.defaultTokenBalance = new BN(String(raw ?? 0));
+  async refreshBalances() {
```
Aierie, line 138: @lukemelia currently we're just redoing the `fetchDepotTask` for this. We could also use the same `getErc20Balance` method that's used in layer 1 to fetch token balances, but it felt better to only use one way of fetching balances.

lukemelia (approved):

PR #1688, opened 2021-06-10
Add stylelint for web-client

backspace:
I‘m opening this for discussion after pairing with @Aierie today, there were some extraneous whitespace diffs I encountered as part of #1681. It mirrors the stylelint setup from Boxel. I ran the autofix and then manually addressed what it missed; nothing that significant, see commit messages for more.

It doesn’t add linting to the GitHub Actions workflow as I wasn’t sure about best practices for that in this monorepo.


lukemelia:
I'm +1. It should ideally run as part of `yarn lint` and `yarn lint:css` in `packages/web-client` as well as `yarn lint` and `yarn lint:css:web-client` in the root of the monorepo. That will make it get picked up correctly as part of CI as well.

lukemelia (commented):
I'm +1. It should ideally run as part of yarn lint and yarn lint:css in packages/web-client as well as yarn lint and yarn lint:css:web-client in the root of the monorepo. That will make it get picked up correctly as part of CI as well.

backspace:
I’ve added a `lint:css:web-client` script to lint CSS to the monorepo’s `package.json` and a call to it from the Actions workflow. Since it follows the `lint:*` pattern, it’ll be run as part of `yarn lint` calls 🎉 the same applies to `web-client`’s `package.json`

You can see an example CI failure from d4e87cf [here](https://github.com/cardstack/cardstack/runs/2806476566#step:8:15), which I’ve now removed.

lukemelia (approved):

backspace:
oops, thanks, I misread the email notification I see 🙃

PR #1690, opened 2021-06-11
Add fast-check to did-resolver and fix a bug it discovered

lukemelia:
Introduced fast-check, a library for property-based testing and refactor some did-resolver tests to use it. It discovered a boundary bug when passing version == 0 and this PR includes a fix.


habdelra (approved):
looks good 👍

PR #1697, opened 2021-06-14
Enhancement/thread animation improvement

Aierie:
CS-1029

- CSS animations for thread items
- Respects `prefers-reduced-motion` settings
- Adds an environment variable to set thread animation interval during development

### DIFF
--- a/packages/web-client/app/components/workflow-thread/animated-workflow.css
+++ b/packages/web-client/app/components/workflow-thread/animated-workflow.css
@@ -1,0 +1,54 @@
+/*
+  If the user has expressed their preference for
+  reduced motion, then don't use animations.
+*/
+@media (prefers-reduced-motion: reduce) {
+  .workflow-postable-animated,
+  .milestone-banner-animated,
+  .workflow-date-divider-animated {
+    animation: none;
+  }
+}
+
+/*
+  If the browser understands the media query and the user
+  explicitly hasn't set a preference, then use animations.
+*/
+@media (prefers-reduced-motion: no-preference) {
+  .workflow-postable-animated,
+  .milestone-banner-animated,
+  .workflow-date-divider-animated {
+    animation: thread-item-revealed ease-out
+      var(--thread-item-reveal-duration, 200ms) backwards;
+  }
+
+  .workflow-postable-animated, .workflow-date-divider-animated {
+    --thread-item-reveal-duration: 200ms;
+  }
+
+  .workflow-postable-animated--has-card {
+    --thread-item-reveal-duration: 400ms;
+  }
+
+  .milestone-banner-animated {
+    /*
+    this is 1s in XD.
+    However, setting to 800ms so we don't look like we have 2 items being added at once
+    when there is a completed milestone and a message coming right after it
+    Alternatively we can make the interval > 1s and set this to 1s.
+    */
+    --thread-item-reveal-duration: 800ms;
+  }
+}
+
+@keyframes thread-item-revealed {
+  0% {
+    transform: translateY(30px);
+    opacity: 0;
+  }
+
+  100% {
+    transform: translateY(0);
+    opacity: 1;
+  }
+}
--- a/packages/web-client/app/components/workflow-thread/animated-workflow.css
+++ b/packages/web-client/app/components/workflow-thread/animated-workflow.css
@@ -1,8 +1,9 @@
-:root {
+.workflow-thread-animated {
   /*
-    this is also set in the constructor in animated-workflow.ts
+    this is also set in workflow-thread.ts
+    based on prefers-reduced-motion
   */
-  --thread-animation-interval: 1000ms;
+  --thread-animation-interval: 0ms;
 }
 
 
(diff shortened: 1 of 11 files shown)


Aierie:
A bit concerned about exposing both `Workflow` and `AnimatedWorkflow` as possible models to render workflow threads from.

packages/web-client/tests/acceptance/deposit-test.ts
```diff
@@ -60,6 +60,9 @@ module('Acceptance | deposit', function (hooks) {
 
     assert.dom(post).containsText('Connect your L1 test chain wallet');
 
+    let finishedCssAnimation = false;
+    setTimeout(() => (finishedCssAnimation = true), 1000);
```
lukemelia, line 64: Can we make the animation zero duration in test to avoid needing to introduce this delay and noise?

packages/web-client/app/components/workflow-thread/animated-workflow.css
```diff
@@ -0,0 +1,63 @@
...
+    --thread-item-reveal-duration: 400ms;
+  }
+
+  .workflow-milestone-banner-animated {
+    /*
+    this is 1s in XD.
+    However, setting to 800ms so we don't look like we have 2 items being added at once
+    when there is a completed milestone and a message coming right after it
+    Alternatively we can make the interval > 1s and set this to 1s.
+    */
```
lukemelia, line 48: Pick what you think looks good and we can adjust based on design review

lukemelia:
> A bit concerned about exposing both Workflow and AnimatedWorkflow as possible models to render workflow threads from, even though they are both supposed to have the same "interface". Wonder if this might make it harder to track down bugs in the future, though perhaps better testing can prevent that.

Me too. I don't think we should do this. How about if we let prefer-reduced-motion clients still have items that appear just without the CSS animation?

Aierie:
> > A bit concerned about exposing both Workflow and AnimatedWorkflow as possible models to render workflow threads from, even though they are both supposed to have the same "interface". Wonder if this might make it harder to track down bugs in the future, though perhaps better testing can prevent that.
> 
> Me too. I don't think we should do this. How about if we let prefer-reduced-motion clients still have items that appear just without the CSS animation?

I think that works, if we turn off autoscrolling for `prefers-reduced-motion`. This would help with regular thread use as well.

burieberry (approved):
Good work! The reveal animations look good, and tested the `prefers-reduced-motion` and the environment variable for setting interval.

Aierie:
> Good work! The reveal animations look good, and tested the `prefers-reduced-motion` and the environment variable for setting interval.

Thanks! @burieberry are you on board with turning off autoscroll for `prefers-reduced-motion`?

packages/web-client/app/models/animated-workflow.ts
```diff
@@ -17,7 +17,15 @@ let waiter = buildWaiter('thread-animation');
...
-let interval = config.environment === 'test' ? 100 : 1000;
+let interval = 1000;
+if (config.environment === 'test') {
+  interval = 0;
+} else if (
+  config.environment === 'development' &&
+  config.threadAnimationInterval
+) {
+  interval = config.threadAnimationInterval;
+}
```
backspace, line 28: I wonder if it might be preferable to move as much of this as possible into `config/environment.js`? Since it already has blocks that override defaults for `development` and `test`, it might be nice to concentrate things there.

Also, a possible development productivity improvement would be to allow a query parameter override so one can change `threadAnimationInterval` without needing to stop the server and update `THREAD_ANIMATION_INTERVAL`. I’ve done this a few times, like:
* `?faker-seed=1234` to [change the randomisation seed at runtime](https://github.com/hashicorp/nomad/blob/0af476263b5c6f12a43a069070400e1762ad17c2/ui/mirage/faker.js#L4-L20) instead of build time
* `?mirage-logging=true` to [enable Mirage logging](https://github.com/hashicorp/nomad/blob/a9ec864a1ab47030761788ac11bd8d85ebd43145/ui/mirage/config.js#L27) without needing to restart

I haven’t been working with the animations enough yet to know whether this is something that would be tweaked enough to make a query parameter override useful, so this could be something to revisit later with more experience. Maybe it’s overkill, I just have a distaste for stopping and starting the server to adjust things haha. If this does seem worthy, despite what I said above I believe the override would have to happen here, at consumption, vs in `config/environment`, which doesn’t live-update.

packages/web-client/app/models/animated-workflow.ts
```diff
@@ -137,6 +145,13 @@ export default class AnimatedWorkflow {
 
   constructor(model: Workflow) {
     this.model = model;
+    // set this so css animations shorten their duration to at most the
+    // rate of release
+    // this is purely for devx
+    document.documentElement.style.setProperty(
+      '--thread-animation-interval',
+      `${interval}ms`
+    );
```
backspace, line 154: love it, I haven’t been using CSS variables enough and didn’t know this was possible, seems like an excellent way to inject dynamism 🤩

backspace (approved):
I made some suggestions but nothing blocking, looks good 😀

packages/web-client/app/models/animated-workflow.ts
Aierie: Very interesting, hadn't considered query params as a way to control this. Thanks for the suggestions, might need some time to digest and act on them!

PR #1704, opened 2021-06-14
Add prepaid card preview card

backspace:
While the design includes an xDai address, based on Discord conversation, it seems that wouldn’t exist until after this step, so I’ll be leaving it out.

image

or maybe it should stay with the `0x1234…` placeholder? 🤔

image

todos:
* [x] add memorialised state
  * that should actually have a card address I think
  * design has “prepaid card funding” header but should it be “preview”? image
  * action chin should instead be “Confirmed”
* [x] spinner in action chin as confirmation request is pending
* [x] fix indentation in preview state
* [x] add required parameters to issue card
* [x] define interface for face value placeholder
* [x] update `issuePrepaidCard` to `issuePrepaidCards`?
* [x] include customisation/DID options in call to create card?


Aierie:
> design has “prepaid card funding” header but should it be “preview”? 

Yep! It should keep the same header as the non-memorialized state since they are technically the same card.

PR #1706, opened 2021-06-15
Enhancement/prepaid card UI copy

Aierie:
CS-1048

- Adds ability to specify header text and pattern colors
- Fixes for copy and background positioning

### DIFF
--- a/packages/web-client/app/components/card-pay/prepaid-card/index.css
+++ b/packages/web-client/app/components/card-pay/prepaid-card/index.css
@@ -136,7 +136,7 @@
   top: 0;
   right: 0;
   left: 0;
-  border-radius: var(--boxel-border-radius) var(--boxel-border-radius) 0 0;
+  border-radius: var(--boxel-border-radius);
   overflow: hidden;
   z-index: 0;
   height: 14.375rem;
--- a/packages/web-client/app/components/card-pay/prepaid-card/index.css
+++ b/packages/web-client/app/components/card-pay/prepaid-card/index.css
@@ -142,28 +142,17 @@
   right: 0;
   left: 0;
   overflow: hidden;
+  height: 100%;
+  border-radius: var(--boxel-border-radius);
   z-index: 0;
 }
 
 .prepaid-card__background--color {
   background: var(--header-background);
-
-  /*
-    this height is shorter because we want the gradient change to be more obvious
-  */
-  height: var(--prepaid-card-header-height);
-  border-radius: var(--boxel-border-radius) var(--boxel-border-radius) 0 0;
 }
 
 .prepaid-card__background--pattern {
   background-color: var(--header-pattern-color);
-
-  /*
-    this height is the full size of the prepaid card because
-    the image is meant to be centered on the card.
-  */
-  height: 100%;
-  border-radius: var(--boxel-border-radius);
   -webkit-mask-size: cover;
   -webkit-mask-image: var(--header-theme);
   -webkit-mask-position: center;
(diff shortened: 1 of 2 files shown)


packages/web-client/app/components/card-pay/prepaid-card/index.css
```diff
@@ -129,10 +137,19 @@
...
+.prepaid-card__background--color {
+  background: var(--header-background);
+}
+
+.prepaid-card__background--pattern {
+  background-color: var(--header-pattern-color);
+  -webkit-mask-size: contain;
+  -webkit-mask-image: var(--header-theme);
+  mask-size: contain;
+  mask-image: var(--header-theme);
```
backspace, line 154: Has there been discussion of any kind of preprocessing or the like so both the prefixed and non-prefixed versions don’t need to be present? Or is there a guideline on when to use both?

I’m not sure but I suspect some tweaking here would help as the header background is poking out behind the bottom corners of the card for me:

image

Maybe the border is desired though?

image

I tried a few browsers and they all had the poking 😯

backspace (changes requested):
I noticed an undesirable rendering situation as discussed in a comment, but apart from that, just to make sure I understand, this is setting things up so the text colour can be configured, but not actually configuring it?

image

Sometimes a before and after screenshot can help illustrate changes and ease review, especially when there’s a lot of CSS wrangling 🤓

packages/web-client/app/components/card-pay/prepaid-card/index.css
Aierie: Ah! Thanks. I didn't spot the bottom poking out earlier. Will fix this. 

I don't think we've discussed prefixing, though I think it would be helpful. The reason the prefixing is present for this one is because `mask-image` is non-standard and only supported in Firefox, we aim to support Chrome, Safari, and Firefox.

Aierie:
> I noticed an undesirable rendering situation as discussed in a comment, but apart from that, just to make sure I understand, this is setting things up so the text colour can be configured, but not actually configuring it?
> 
> image
> 
> Sometimes a before and after screenshot can help illustrate changes and ease review, especially when there’s a lot of CSS wrangling 🤓

Yep, this is just setting up the prepaid card component for changes to the layout customization card - in particular assigning different text and pattern colors for each background. These can be white or black, and will be chosen based on contrast against each background. See: https://linear.app/cardstack/issue/CS-909#comment-36ac2a34.

re: before/after screenshots - will do so in the future, probably would have helped with checking that the visual output was the same.

Aierie:
Changed the gradient behaviour as design team prefers backgrounds with more contrast.

image

preferred over

image

though second version is more similar to the designs in XD.

backspace:
ah if there wasn’t meant to be visual changes, then before/after is less useful… but visual diffs might have helped with the corner thing 😯

burieberry (approved):

Aierie:
> Changed the gradient behaviour as design team prefers backgrounds with more contrast.
> 
> image
> 
> preferred over
> 
> image
> 
> though second version is more similar to the designs in XD.

This has been reversed because matching the height of the header and height of the curved "platter" below is fragile, and might also be hard to replicate on mobile.

Aierie:
The best place to review this is probably on the `prototype/use-real-prepaid-card-customizations` branch, which has real patterns and colors and is based on this branch.

Aierie:
image

Design feedback is this makes sense

backspace (approved):
I ran `prototype/use-real-prepaid-card-customizations` locally as suggested, looks good!
image

PR #1709, opened 2021-06-15
Use auto-bind to simplify routing declarations

lukemelia:
CS-1063


backspace (approved):
quality of life improvement! 🎉

PR #1712, opened 2021-06-15
Hub file organizations

lukemelia:
Move dependency injection implementation to a `di/` directory, services to a `services/` directory, and move Error until `utils/`


backspace (approved):
👍🏻 this looks like some sensible cleanup and introduction of directory structure convention

PR #1714, opened 2021-06-15
Clean up event subscriptions when server shuts down

lukemelia:
-avoids leaking event listeners during test runs


backspace (approved):
it reminds me of Ember’s [`no-incorrect-calls-with-inline-anonymous-functions`](https://github.com/ember-cli/eslint-plugin-ember/blob/master/docs/rules/no-incorrect-calls-with-inline-anonymous-functions.md) which makes me wonder if it could be caught by linting, but I couldn’t find anything like that 🤔

PR #1717, opened 2021-06-16
Prototype/use real prepaid card customizations

Aierie:
This is to demonstrate the use of real prepaid card patterns and colors in customization.

## Data structure
See https://github.com/cardstack/cardstack/blob/3a2e8f7ce8c25e1dffcfd6657be6d3580b20de27/packages/web-client/app/utils/prepaid-card-customization-options.ts for the data structure that patterns and color options use.

## Readability
I feel like readability is an issue with colors that have a high contrast with both black and white, because the patterns' shapes distract from the text itself:

image

This is improved by dropping opacity to 0.5:

image

I think any solution be applied from the design side as the mobile team will need to render these cards as well, and also applying opacity on patterns from the tech side imposes constraints that seem tricky for future designs to work around.


Aierie:
Closing as this is merely a demo branch. Will keep branch alive until #1721 is merged, as a reference.

PR #1719, opened 2021-06-16
Use refactored AccountDisplay component

burieberry:
Use the `AccountDisplay` and `BalanceDisplay` components

Here are some examples:

## BalanceDisplay component:
With `size="large"`, `icon`, `balance`, and `symbol` args:
lg-icon-balance-symbol

With `size="large"`, `icon`, `balance`, `symbol` and `displayUsdBalance=true` args:
lg-icon-balance-usdbalance

With `size="small"`, `icon`, `name`, `balance`, `symbol`:
sm-icon-title-balance-symbol

## AccountDisplay component:
With `size="large"`, `icon`, `name` and `text` args:
lg-title-text

With `size="large"`, `icon`, `name`, and `address` args:
lg-icon-title-address

With `icon`, `name`, and `text` args:
def-title-text-depot

With `name` and `address` args:
def-title-address

With `icon`, `name`, and `address` args:
def-icon-title-address

With `size="small"`, `name`, `address`, and `wrapped`:
sm-title-address-wrapped

With `size="small"`, `icon`, `name`, `address`:
sm-icon-title-address

### DIFF
--- a/packages/web-client/app/components/card-pay/balance-display/index.hbs
+++ b/packages/web-client/app/components/card-pay/balance-display/index.hbs
@@ -21,13 +21,13 @@
           balance-display__text-lg=isLarge
           balance-display__text-sm=isSmall
         }} data-test-balance-display-amount>
-          {{@balance}} {{@symbol}}
+          {{format-token-amount @balance}} {{@symbol}}
         </div>
-      {{/if}}
-      {{#if @usdBalance}}
-        <div class="balance-display__text-sm" data-test-balance-display-usd-amount>
-          ${{@usdBalance}} USD
-        </div>
+        {{#if @displayUsdBalance}}
+          <div class="balance-display__text-sm" data-test-balance-display-usd-amount>
+            ${{token-to-usd @symbol @balance}} USD
+          </div>
+        {{/if}}
       {{/if}}
       {{#if @text}}
         <div class="balance-display__text-sm" data-test-balance-display-text>
--- a/packages/web-client/app/components/card-pay/balance-display/index.hbs
+++ b/packages/web-client/app/components/card-pay/balance-display/index.hbs
@@ -1,15 +1,12 @@
 {{#let (eq @size "small") (eq @size "large") as |isSmall isLarge|}}
   <div class="balance-display" ...attributes>
     {{#if @icon}}
-      {{svg-jar @icon class=(if isLarge "balance-display__icon-lg" "balance-display__icon") role="presentation"}}
+      {{svg-jar @icon class=(cn "balance-display__icon" balance-display__icon--lg=isLarge) role="presentation"}}
     {{/if}}
     <div>
-      {{#if @label}}
-        <div class="balance-display__text-sm">{{@label}}</div>
-      {{/if}}
       {{#if @name}}
         <div
-          class={{if isSmall "balance-display__name-sm" "balance-display__name"}}
+          class={{cn "balance-display__name" balance-display__name--sm=isSmall}}
           data-test-balance-display-name
         >
           {{@name}}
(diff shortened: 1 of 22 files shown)


lukemelia (commented):
Overall looks good in terms of implementation. Conceptually, I'm not a fan of using a single component for two distinct purposes: 1) showing an account address, and 2) showing a token balance. What do you think about making these two separate components?

lukemelia (approved):
👍

packages/web-client/app/components/card-pay/balance-display/index.hbs
```diff
@@ -0,0 +1,39 @@
...
+          class={{if isSmall "balance-display__name-sm" "balance-display__name"}}
+          data-test-balance-display-name
+        >
+          {{@name}}
+        </div>
+      {{/if}}
+      {{#if @balance}}
+        <div class={{cn
+          balance-display__text=(not (or isSmall isLarge))
+          balance-display__text-lg=isLarge
```
Aierie, line 21: Are lg and sm modifiers, and if so should they use the `--lg`/`--sm` syntax?

packages/web-client/app/components/card-pay/deposit-workflow/transaction-setup/token-option/index.hbs
```diff
@@ -1,33 +1,26 @@
...
+    @name={{@token.symbol}}
+    @text={{@token.description}}
+  />
+  <CardPay::BalanceDisplay
+    @label="Balance"
+    @balance={{@balance}}
+    @symbol={{@token.symbol}}
+    @displayUsdBalance={{true}}
+    data-test-balance={{@token.symbol}}
+  />
```
Aierie, line 25: Using the component this way seems to tie us to a certain implementation of a balance display. I think keeping the original implementation is better here because it works, and we don't need to think about it, and come back to change this component if/when we decide we want a new implementation of balance display.

packages/web-client/app/components/card-pay/deposit-workflow/transaction-amount/index.css
```diff
@@ -83,35 +83,18 @@
...
+  align-items: center;
 }
 
-.transaction-amount__usd-balance {
-  color: var(--boxel-purple-500);
-  font: var(--boxel-font-sm);
-  letter-spacing: var(--boxel-lsp-xs);
-  margin-top: var(--boxel-sp-xs);
+.transaction-amount__token > div > :last-child {
+  text-align: center;
```
Aierie, line 99: Can we live without this styling? It being a `div > :last-child` seems a bit too generic to me.

with:
image


without:

image

Aierie: The approximate sign on the USD is also removed, it might be a more important detail. Perhaps this component should also retain the original implementation instead?

Aierie (approved):

packages/web-client/app/components/card-pay/deposit-workflow/transaction-amount/index.css
Aierie: image

Aierie: Left comments about the approximate sign on this. There actually isn't an approximate sign on this component and the original implementation's one was extraneous.

packages/web-client/app/components/card-pay/balance-display/index.hbs
```diff
@@ -0,0 +1,39 @@
...
+          balance-display__text-sm=isSmall
+        }} data-test-balance-display-amount>
+          {{format-token-amount @balance}} {{@symbol}}
+        </div>
+        {{#if @displayUsdBalance}}
```
Aierie, line 26: Many uses of usd balance use an approximate sign, which might be important:

image

Aierie (changes requested):

packages/web-client/app/components/card-pay/balance-display/index.hbs
burieberry: Chris said to remove all approximate signs during one of the meetings. We can add it back if needed

packages/web-client/app/components/card-pay/deposit-workflow/transaction-setup/token-option/index.hbs
burieberry: Yeah I was undecided about this one, so your opinion helps.

Aierie (approved):

PR #1721, opened 2021-06-17
Enhancement/customization options service

Aierie:
CS-1048

This fetches data from the hub to use in the customization form + other uses of prepaid cards.


Aierie:
@lukemelia @burieberry @backspace does this setup seem reasonable? Tests and development should be working (development requires that you start the hub and are able to GET `http://localhost:3000/api/prepaid-card-color-schemes` and `http://localhost:3000/api/prepaid-card-patterns`). 

I'm currently working on converting the mirage stuff to typescript, however some parts seem to be a bit tricky - not sure how much value there is in defining types for Mirage models? Ember OSF does this quite extensively, but they seem to use mirage throughout their app and we don't use it as much in ours:

https://github.com/CenterForOpenScience/ember-osf-web/tree/develop/types/ember-cli-mirage
https://github.com/CenterForOpenScience/ember-osf-web/search?q=MirageModelRegistry

Thinking of just using their `ember-cli-mirage` types for now, and noting that this is a possible way to add on to mirage in the future.

packages/web-client/app/services/card-customization-options.ts
```diff
@@ -0,0 +1,91 @@
...
+  @tracked loaded = false;
+  @tracked patternOptions: PatternCustomizationOption[] | null = null;
+  @tracked colorOptions: ColorCustomizationOption[] | null = null;
+
+  constructor(props: any) {
+    super(props);
+    this.fetchCustomizationOptions();
+  }
+
+  async fetchCustomizationOptions() {
```
lukemelia, line 59: Two thoughts here:

1) Consider making this an ember-concurrency task. The reason is that it will join the asynchrony with the lifecycle of the service so that in tests where the service is instantiated but end quickly, we don't need to worry about promise resolutions trying to set properties on a service instance or doing pointless parsing.
2) Consider initiating this operation in response to a method call rather than instantiation. It seems cleaner to me to do this rather than couple creation of the service and initiation of this operation.

I feel more strongly about 1 than 2. Both are judgement calls.

packages/web-client/app/services/card-customization-options.ts
```diff
@@ -0,0 +1,91 @@
...
+
+let convertToPatternCustomizationOption = (
+  o: PatternCustomizationResponseObject
+): PatternCustomizationOption => {
+  return {
+    patternUrl: o.attributes['pattern-url'],
+    description: o.attributes.description,
+    id: o.id,
+  };
+};
```
lukemelia, line 47: It would be common in Ember development, when faced with writing this code, to reach for a data library, usually ember-data or orbit.js. It avoids needing this kind of boilerplate deserialization (at the cost of a complex dependency). We can refactor to it later, just want to flag this.

Aierie:
Dropped typescript because of difficulties with embroider + mirage, tracking in CS-1081

packages/web-client/tests/acceptance/issue-prepaid-card-test.ts
```diff
@@ -27,8 +30,14 @@ function milestoneCompletedSel(milestoneIndex: number): string {
 
 module('Acceptance | issue prepaid card', function (hooks) {
   setupApplicationTest(hooks);
+  setupMirage(hooks);
 
   test('Initiating workflow without wallet connections', async function (assert) {
+    this.server.db.loadData({
+      prepaidCardColorSchemes: prepaidCardColorSchemes,
+      prepaidCardPatterns,
+    });
```
lukemelia, line 39: Doesn't matter since we have only one test in this module currently, but this should probably be in a `beforeEach`

packages/web-client/tests/acceptance/issue-prepaid-card-test.ts
```diff
@@ -27,8 +30,14 @@ function milestoneCompletedSel(milestoneIndex: number): string {
 
 module('Acceptance | issue prepaid card', function (hooks) {
   setupApplicationTest(hooks);
+  setupMirage(hooks);
 
   test('Initiating workflow without wallet connections', async function (assert) {
+    this.server.db.loadData({
+      prepaidCardColorSchemes: prepaidCardColorSchemes,
```
lukemelia, line 37:
```suggestion
      prepaidCardColorSchemes,
```

packages/web-client/mirage/config.js
```diff
@@ -0,0 +1,18 @@
...
+
+  this.get('/prepaid-card-patterns', (schema) => {
+    return schema.prepaidCardPatterns.all();
+  });
+
+  this.passthrough((request) => {
+    return (
+      !request.url.includes('/api/prepaid-card-color-schemes') &&
+      !request.url.includes('/api/prepaid-card-patterns')
+    );
```
lukemelia, line 16: Perhaps this should passthrough anything that is not under `/api/...`

lukemelia (changes requested):
Looks good overall. Workaround for fixture bug seems acceptable.

Aierie:
@lukemelia Sorry, deleted the `.ts` files because of the difficulties with embroider + ts. I do follow your points though, will respond in a bit.

packages/web-client/app/services/card-customization-options.ts
Aierie: Does 2. mean that any component that uses this service will need to call the method to fetch the customizations?

Aierie: Situation where this might be relevant is displaying single prepaid cards on the dashboard/in arbitrary threads. Originally was going to handle this by providing a map for ids to styles in this service.

Aierie: My understanding of initiating in response to a method call is in https://github.com/cardstack/cardstack/pull/1732

Aierie: update after discussion w @lukemelia - we are fetching details about styles directly, and we don't need the customization options service to be able to convert ids to styles. This means the only consumer of this service is the form to edit customization options, for now. Will move the customization fetching to a method.

Aierie: This is done!

packages/web-client/app/services/card-customization-options.ts
```diff
@@ -0,0 +1,123 @@
...
+    try {
+      let [_colorOptions, _patternOptions] = yield Promise.all([
+        fetch(`${config.hubURL}/api/prepaid-card-color-schemes`, {
+          method: 'GET',
+          headers: {
+            Accept: 'application/vnd.api+json',
+          },
+        })
+          .then((v) => v.json())
+          .then((v) => v.data.map(convertToColorCustomizationOption)),
```
lukemelia, line 75: Extract fetch and response processing to a separate e-concurrency task so that async promise resolution is managed.

packages/web-client/app/services/card-customization-options.ts
```diff
@@ -0,0 +1,123 @@
...
+
+        fetch(`${config.hubURL}/api/prepaid-card-patterns`, {
+          method: 'GET',
+          headers: {
+            Accept: 'application/vnd.api+json',
+          },
+        })
+          .then((v) => v.json())
+          .then((v) => v.data.map(convertToPatternCustomizationOption)),
```
lukemelia, line 84: Extract this one to separate task as well.

packages/web-client/app/services/card-customization-options.ts
```diff
@@ -0,0 +1,123 @@
...
+
+      let blankIndex = _patternOptions.findIndex(
+        (a: PatternCustomizationOption) => !a.patternUrl
+      );
+      if (blankIndex !== -1) {
+        let blank = _patternOptions.splice(blankIndex, 1)[0];
+        patternOptions = [blank, ..._patternOptions];
+      } else {
+        patternOptions = _patternOptions;
+      }
```
lukemelia, line 106: Consider extracting this to an intention-revealing function.

packages/web-client/app/services/card-customization-options.ts
```diff
@@ -0,0 +1,123 @@
...
+      this.patternOptions = patternOptions;
+      this.colorOptions = colorOptions;
+      this.loaded = true;
+    } catch (e) {
```
lukemelia, line 110: Probably should log this as well.

packages/web-client/app/services/card-customization-options.ts
```diff
@@ -0,0 +1,123 @@
...
+  @tracked patternOptions: PatternCustomizationOption[] | null = [];
+  @tracked colorOptions: ColorCustomizationOption[] | null = [];
+
+  async ensureCustomizationOptionsLoaded() {
+    if (!this.loaded) {
+      return taskFor(this.fetchCustomizationOptions).perform();
+    }
+  }
+
+  @task *fetchCustomizationOptions(): any {
```
lukemelia, line 63: Consider configuring task to drop other invocations while running.

lukemelia (approved):
Nice work! ✨

PR #1723, opened 2021-06-17
Migrations, seeding, APIs for prepaid card customizations

lukemelia:
Note: this does not yet persist to S3. Saving that for a separate PR.

### DIFF
--- a/packages/did-resolver/index.ts
+++ b/packages/did-resolver/index.ts
@@ -3,6 +3,7 @@ import { shake_128 } from 'js-sha3';
 import shortUuid from 'short-uuid';
 import invert from 'lodash/invert';
 import kebabCase from 'lodash/kebabCase';
+import * as uuidv4 from 'uuid';
 
 const CURRENT_VERSION = 1;
 
@@ -73,7 +74,7 @@ class CardstackIdentifier {
   constructor(version: number, type: CardstackIdentifierType, uniqueId: string) {
     this.version = version;
     this.type = type;
-    this.uniqueId = uniqueId;
+    this.uniqueId = normalizeUniqueId(uniqueId);
   }
 
   toDID() {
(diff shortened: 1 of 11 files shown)


packages/did-resolver/index.ts
```diff
@@ -84,6 +85,21 @@ class CardstackIdentifier {
   }
 }
 
+function normalizeUniqueId(candidate: string) {
```
Aierie, line 88: Are we using a short unique id for the sake of storage space?

packages/hub/node-tests/routes/prepaid-card-customizations-test.ts
```diff
@@ -0,0 +1,277 @@
...
+    return stubTimestamp;
+  }
+
+  validateAuthToken(encryptedAuthToken: string) {
+    return handleValidateAuthToken(encryptedAuthToken);
+  }
+}
+
+let stubUserAddress = '0x2f58630CA445Ab1a6DE2Bb9892AA2e1d60876C13';
+handleValidateAuthToken = function (encryptedString: string) {
```
Aierie, line 34: Can this be assigned at the declaration of `handleValidateAuthToken`, since we use the same definition throughout?

packages/hub/node-tests/routes/prepaid-card-customizations-test.ts
```diff
@@ -0,0 +1,277 @@
...
+          },
+          pattern: {
+            data: {
+              type: 'prepaid-card-patterns',
+              id: 'ab70b8d5-95f5-4c20-997c-4db9013b347c',
+            },
+          },
+        },
+      },
+    };
```
Aierie, line 111: Is the structure/boilerplate to account for the possibility of using different sources for color schemes and patterns (eg. tables specific to external users) in the future? 

Wondering what the effect of a minimal request structure like below would be:
```
{
  type: 'prepaid-card-customizations'
  attributes: {
     'issuer-name': 'Satoshi Nakamoto',
     'color-scheme': '5058b874-ce21-4fc4-958c-b6641e1dc175',
     'pattern': 'ab70b8d5-95f5-4c20-997c-4db9013b347c'
   },
}
```

Aierie (approved):
Seems good! There are a couple of things I'm curious about, wrt API/data structure design decisions, left comments below.

packages/did-resolver/index.ts
lukemelia: To keep the did more compact / less storage space in on-chain scenarios / easier to copy/paste

packages/hub/node-tests/routes/prepaid-card-customizations-test.ts
lukemelia: Yes. Copy/paste oversight.

packages/hub/node-tests/routes/prepaid-card-customizations-test.ts
lukemelia: This format follows the [JSON-API](https://jsonapi.org/) spec, which is the default serialization format for ember-data and orbit.js and what the card compiler API will use.

packages/hub/node-tests/routes/prepaid-card-customizations-test.ts
lukemelia: Fixed ✅

PR #1733, opened 2021-06-21
Enhancement/hub seeds with real values

Aierie:
CS-1048

image

`ember build --environment="production"` generates this. It appears that we don't have fingerprinting for the prepaid card patterns.

After updating the db, the correct values are reflected when running the hub + prod ember(`HUB_URL=http://localhost:3000 yarn start:web-client --environment=production`).


Aierie:
The pattern image files are dynamically loaded, so they will not be fingerprinted. If fingerprinting is necessary, we might need to do some configuration of embroider?

lukemelia:
Ok if they are not fingerprinted for now then. We can version them by hand
if/when needed.

Aierie:
@burieberry The image files have gone through SVGO and ImageOptim (aggressive settings), but are still pretty huge:

image

Flagging this in case we need to make changes for mobile, including changing to raster to reduce size.

lukemelia:
Wow, 423KB seems like an unreasonable price to pay for an image of this purpose.

packages/hub/db/seeds.ts
```diff
@@ -21,41 +21,60 @@ async function loadPrepaidCardColorSchemes(db: Client) {
     description = excluded.description;
 `;
   let colorSchemeRows = [
-    ['7CEE903B-E342-4BD7-A3DA-EB3AB2DC078D', '#00ebe5', 'white', 'black', 'Solid Light Blue'],
-    ['232B7E38-1F9A-4599-A870-8F7A9BCBA6E4', '#37eb77', 'white', 'black', 'Solid Lime Green'],
-    ['5ED643F8-A8BB-4748-BCD6-98E850FBA505', '#ac00ff', 'white', 'white', 'Solid Purple'],
-    ['6997F87C-D1EE-45F8-8A49-D19A9D5FAD73', '#efefef', 'white', 'black', 'Solid Light Grey'],
-    ['16B83804-1FF9-4AA2-B670-472919EFF8C3', '#393642', 'white', 'white', 'Solid Dark'],
-    ['7B05C4EA-6BAA-471F-9B49-7AC10ADF2F9A', '#c3fc33', 'white', 'black', 'Solid Yellow Green'],
+    ['4bb07815-7d4b-4312-8224-275d3c0d1af2', '#FFD800', 'white', 'black', ''],
```
lukemelia, line 24: Did you remove the descriptions on purpose?

Aierie: Nope, it is kind of weird that there is no description for some of them. I think it's worth keeping the description field, but since it's a plus rather than a must I'm leaving adding the descriptions till later. Will probably use this tool to get them manually:

https://www.color-blindness.com/color-name-hue/

Aierie: Would probably like to get some alignment from design on the names before adding them, will create a table for the patterns and the colors and let Gardella have a look.

Aierie:
> Wow, 423KB seems like an unreasonable price to pay for an image of this purpose.

Yep, allowing loss of precision during optimization of the svg with SVGO drops it to 100KBish, but results in distortion, most other settings leave it at 200-300KB.

Aierie:
Descriptions are okayed by design. 

Images might be further optimized by using raster instead of vector for image 1 and 2 - eg. a transparent PNG - there are many paths and curves in the svgs, which leads to the large size. Might be possible, since we have a somewhat consistent range for prepaid card size?

lukemelia:
Let's merge as is and make a card to revisit.

PR #1734, opened 2021-06-21
exclude web client .env from git clean

Aierie:
We are using some environment variables in the web client via `.env`. This prevents them from getting removed via `yarn clean`.


backspace (approved):
🥳

PR #1739, opened 2021-06-22
Update DI system to allow factory classes to return the service instance asynchronously

lukemelia:
I plan to use this to instantiate graphile-worker's WorkerUtils class


habdelra (approved):
👍  LGTM

PR #1741, opened 2021-06-22
Add prepaid card face-value selection card

burieberry:
- Added FaceValue selection card UI
- Added `spend-to-usd` helper
- Disabled options where balance is lower than the face value option
- Displaying only the approximate Dai amount
- Added `workflow-config.ts` file for storing common values
- Safari CSS fixes

Follow up PR: 
- Maybe spend to token conversion. Have a version that needs some cleanup.

face-value-card
face-value-card-expanded
face-value-card-selected
face-value-card-memorialized


packages/web-client/app/components/card-pay/issue-prepaid-card-workflow/face-value/balance-view/index.css
```diff
@@ -0,0 +1,87 @@
...
+
+/* Details */
+.balance-view-banner__details {
+  padding: 0 var(--details-padding-right) 0 var(--boxel-sp);
+}
+
+/* Specific to this card */
+.balance-view-banner[open] .balance-view-banner__summary-content {
+  display: none;
+}
```
Aierie, line 82:
```suggestion
/* Specific to this card */
.balance-view-banner__summary-content {
  height: 100%;
}

.balance-view-banner[open] .balance-view-banner__summary-content {
  display: none;
}
```

Fixes bug in Safari 14.1.1

image

packages/web-client/app/components/card-pay/issue-prepaid-card-workflow/face-value/option/index.css
```diff
@@ -0,0 +1,59 @@
...
+
+.face-value-option--checked,
+.face-value-option:focus,
+.face-value-option:focus-within {
+  box-shadow: 0 0 0 2px var(--boxel-highlight);
+  outline: 1px solid transparent;
+}
+
+.face-value-option__input {
+  appearance: none;
```
Aierie, line 25: Safari needs this prefixed with `-webkit-`: 

https://developer.mozilla.org/en-US/docs/Web/CSS/appearance

image

packages/web-client/app/utils/token.ts
```diff
@@ -98,24 +103,33 @@ const _tokenDisplayInfoMap: Record<TokenSymbol, DisplayInfo> = {
     description: '',
     icon: 'dai-token',
   },
+  SPEND: {
+    name: 'Spend',
+    symbol: 'SPEND',
+    description: '',
+    icon: 'spend',
+    sign: '§',
+  },
```
Aierie, line 112: Would prefer SPEND config be separate from the tokens. It's not technically a token (I think?), and doesn't come up where the other tokens do (eg. in a list of balances where we need to display different info/visuals for each token, or in a dropdown).

packages/web-client/app/components/card-pay/issue-prepaid-card-workflow/face-value/balance-view/index.hbs
```diff
@@ -0,0 +1,43 @@
...
+      <CardPay::BalanceDisplay
+        class="balance-view-banner__details-token"
+        @size="large"
+        @icon={{@token.icon}}
+        @balance={{@balance}}
+        @symbol={{@token.symbol}}
+        data-test-balance-view-token-amount
+      />
+    </:inner>
+  </CardPay::NestedItems>
```
Aierie, line 42: NestedItems kind of breaks down at 3 levels. If 3 levels of nesting is necessary and comes up often (which it seems to?), may be good to update the component to accommodate this (as a separate issue).

Aierie (changes requested):
I think overall this looks good. Just had some browser-specific (SAFARI!) issues.

Aierie:
> maybe a helper for spend-usd conversions (overkill?) or update token-to-usd helper to include spend
Would go for a spend-specific helper for this.

> better way to store face value options?
I'm also wondering about this, since the smallest possible prepaid card value affects #1742. Should there be a conventional place to store certain workflow-specific configuration values? Eg. `issue-prepaid-card-workflow/workflow-config.ts`

packages/web-client/app/components/card-pay/issue-prepaid-card-workflow/face-value/index.hbs
```diff
@@ -1,19 +1,76 @@
...
+            <CardPay::IssuePrepaidCardWorkflow::FaceValue::Option
+              @checked={{eq amount this.selectedFaceValue}}
+              @amount={{amount}}
+              @token={{this.faceValueToken}}
+              @amountInUSD={{mult amount this.spendToUSDRate}}
+              @amountInFundingToken={{@amountInFundingToken}}
+              @fundingTokenSymbol={{this.fundingToken.symbol}}
+              @onInput={{fn this.chooseFaceValue amount}}
+            />
+          {{/each}}
```
backspace, line 49: This is more a larger design question; I found issue 847 related to having commas for larger values, should that maybe be applied globally? I think it can help make it easier to recognise the values. Here it’s helped because the bottom values are an order of magnitude larger than the top ones, but I feel like in general it’s preferable.

image

backspace (changes requested):
I noticed in the ticket that it says

> Disable options that they don't have enough balance for

but when I ran it locally with a too-low balance, I was able to choose every value. Maybe this could be added in a followup PR, though.

image

I’m looking forward to using the chosen value in the confirmation step! @Aierie’s suggestion toward defining known values to be chosen through a workflow sounds promising to me, right now the sort of grab bag/arbitrary name seems to work in this nascent stage, but having something formalised might help in the future.

packages/web-client/app/components/card-pay/issue-prepaid-card-workflow/face-value/balance-view/index.hbs
burieberry: What breaks about it?

packages/web-client/app/components/card-pay/issue-prepaid-card-workflow/face-value/index.hbs
burieberry: Yes, we want the commas wherever there's a large enough value. I thought that ticket was already done, but I'm not seeing the commas anywhere. Maybe some sort of regression happened, and it's not being applied to the helpers.

no-commas

packages/web-client/app/components/card-pay/issue-prepaid-card-workflow/workflow-config.ts
```diff
@@ -0,0 +1,14 @@
...
+export const spendToUsdRate = 0.01;
+
+export interface Token {
+  address?: string;
+  balance?: BN;
+  icon: string;
+  name: string;
+  description?: string;
+  symbol: TokenSymbol;
+}
```
burieberry, line 14: Not sure where this should go

packages/web-client/app/components/card-pay/issue-prepaid-card-workflow/face-value/balance-view/index.hbs
Aierie: Nothing in the styles directly.

I think having to nest NestedItems is probably not great, or adding extra styles to support 3 levels each time either. It would be nice to have the component support 3 levels directly I think? Though this is not a must and definitely not blocking for this PR

packages/web-client/app/components/card-pay/issue-prepaid-card-workflow/face-value/index.hbs
```diff
@@ -25,19 +25,24 @@
...
         <div class="face-value-card__options">
-          {{#each this.amountOptions as |amount|}}
+          {{#each this.faceValueOptions as |amount|}}
             <CardPay::IssuePrepaidCardWorkflow::FaceValue::Option
+              @disabled={{lt this.balanceFromBN (mult amount this.spendToUsdRate)}}
               @checked={{eq amount this.selectedFaceValue}}
-              @spendAmount={{amount}}
+              @spendFaceValue={{amount}}
+              @usdAmount={{mult amount this.spendToUsdRate}}
+              @approxDaiAmount={{mult amount this.spendToUsdRate}}
```
Aierie: Approximate DAI amount can probably use the `convertFromSpend` method from the SDK?

packages/web-client/app/components/card-pay/issue-prepaid-card-workflow/workflow-config.ts
Aierie: Do we need this type, if we store the token's symbol and balance separately?

packages/web-client/app/components/card-pay/issue-prepaid-card-workflow/face-value/index.hbs
```diff
@@ -1,19 +1,75 @@
...
+        </CardPay::LabeledValue>
+      {{else}}
+        <div class="face-value-card__options">
+          {{#each this.faceValueOptions as |amount|}}
+            <CardPay::IssuePrepaidCardWorkflow::FaceValue::Option
+              @disabled={{lt this.balanceFromBN (spend-to-usd amount)}}
+              @checked={{eq amount this.selectedFaceValue}}
+              @spendFaceValue={{amount}}
+              @usdAmount={{spend-to-usd amount}}
+              @approxDaiAmount={{spend-to-usd amount}}
```
Aierie, line 45: Approximate DAI amount can probably use the ExchangeRate.convertFromSpend method from the SDK?

burieberry: Yes, I got that to work in a future commit but requires some clean up

packages/web-client/app/components/card-pay/issue-prepaid-card-workflow/workflow-config.ts
burieberry: I liked the convenience of having the address and balance be readily available. But now I'm wondering if there's the risk of not having the balances be updated. I think I'll give up on the convenience.

PR #1742, opened 2021-06-22
Enhancement/prepaid card workflow cancelations

Aierie:
Competing implementation (vs #1758, #1764) of prepaid card workflow cancelation when there isn't enough DAI to create a prepaid card.

- Postables pass their workflow's `cancel` method to their components in order to provide them with the ability to cancel workflows
- We wrap layer 2 in a component that checks that balances are enough to create a prepaid card and cancels if they are not, via the passed in `cancel` method.


Aierie:
Temporarily closing in favour of #1764

PR #1749, opened 2021-06-23
Internal/docs and qol for hub

Aierie:


Aierie:
Will convert API docs to https://gist.github.com/iros/3426278

lukemelia (approved):

PR #1751, opened 2021-06-23
Add “older messages” docked footer

backspace:
This has a way to go and will probably have a followup PR to address the “15 new messages” use case.

One notable TODO is that I was unable to get `ember-in-viewport`’s `in-viewport` modifier working, specifically the `onExit` was never firing. This is using `did-insert` and the service instead, which is less elegant, hopefully I can get the modifier working somehow.


burieberry:
I checked discord web, and they have this component as a `button` element. I think that makes sense since the whole thing should be clickable.

backspace:
hmm interesting that the whole thing is a button! With the design I figured just the “Jump to latest” and arrow would be the thing to click. Thanks for finding this precedent 😀

backspace:
We’ve put this aside for now, but it’ll be accomplished within Boxel regardless.

PR #1754, opened 2021-06-24
Introduce a Postgresql-based background task queue to the hub

lukemelia:


Aierie (approved):
This works nicely for me! Graphile recommends having one `WorkerUtils` instance for the whole app though, should we use a static property for workerUtils, and/or prevent reassignment?

https://github.com/graphile/worker#makeworkerutilsoptions-workerutilsoptions-promiseworkerutils

Not sure if the DI takes care of having only one WorkerClient already.

lukemelia:
> This works nicely for me! Graphile recommends having one `WorkerUtils` instance for the whole app though, should we use a static property for workerUtils, and/or prevent reassignment?

The semantics of the hub's dependency injection container is that one and only one instance of each service is created in the container. And our worker-client creates one and only one WorkerUtils instance.

PR #1755, opened 2021-06-24
S3 json task

lukemelia:
- [x] First merge #1754


Aierie (commented):
This seems good to me, though I've just gone through the s3 client docs and haven't tested it yet (will give it a try tomorrow if/when I can figure out my own aws account). Should the hello task be removed?

lukemelia:
> Should the hello task be removed?

Yes. Removed!

I tested it manually and saw it write to an S3 bucket. Merging...

PR #1758, opened 2021-06-25
Enhancement/workflow cancelation on render

Aierie:
Competing implementation (vs #1742, #1764) of prepaid card workflow cancelation when there isn't enough DAI to create a prepaid card.

- Adds an optional `onRender` callback to postables
- `onRender` hook is called by the layer 2 connect card to cancel the workflow if funds are insufficient

### DIFF
--- a/packages/web-client/app/components/card-pay/issue-prepaid-card-workflow/index.ts
+++ b/packages/web-client/app/components/card-pay/issue-prepaid-card-workflow/index.ts
@@ -10,6 +10,8 @@ import NetworkAwareWorkflowMessage from '@cardstack/web-client/components/workfl
 import Layer2Network from '@cardstack/web-client/services/layer2-network';
 import { action } from '@ember/object';
 import { WorkflowPostable } from '@cardstack/web-client/models/workflow/workflow-postable';
+import { BN } from 'bn.js';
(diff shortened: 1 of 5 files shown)


packages/web-client/app/components/card-pay/issue-prepaid-card-workflow/index.ts
```diff
@@ -50,6 +52,18 @@ class IssuePrepaidCardWorkflow extends Workflow {
         new WorkflowCard({
           author: cardbot,
           componentName: 'card-pay/layer-two-connect-card',
+          onRender: () => {
+            let layer2Network = this.owner.lookup(
+              'service:layer2-network'
+            ) as Layer2Network;
+
+            layer2Network.waitForAccount.then(() => {
+              let MIN_DAI_AMOUNT = new BN(toWei('50'));
```
lukemelia, line 61: Consider extracting this to a modue-scoped const.

Aierie: Yep, will get this from a shared configuration with the face value card.

Aierie:
Temporarily closing in favour of #1764

PR #1759, opened 2021-06-25
Issue prepaid card workflow cancelations

Aierie:
CS-898, CS-1062

## Some refactoring
- Cancelation messages are selectively shown based on a `cancelationReason` property on the workflow, that is assigned when a workflow is canceled via its `cancel` method.
- A workflow cannot be canceled twice
- `workflow-canceled-cta` is now `disconnection-cta`

## New stuff
- An optional `check` option for the `WorkflowCard` constructor, which is an async function that allows you to specify whether to cancel the workflow and for what reason. This reason should be used to decided what cancelation messages to show. This will be called in the `onComplete` method of the `WorkflowCard`. In a future PR, should be able to tack on a new property to specify the context of failure for more informative cancelation messages.

## Features
- The prepaid card workflow will be canceled when a user does not have enough balance, after the layer two connect card is loaded
- The prepaid card workflow will be canceled upon disconnection of layer 2

## Other
- Opted to reduce non-golden/happy path tests' assertions to a minimum, so that they are more robust and readable. Also changed the address assertions to check for the address in our layer 2 wallet connection cards, since these are within the workflow, instead of the button outside the workflow.


packages/web-client/app/components/card-pay/deposit-workflow/index.ts
```diff
@@ -186,8 +192,8 @@ class DepositWorkflowComponent extends Component {
     this.workflow = new DepositWorkflow(getOwner(this));
   }
 
-  @action cancelWorkflow() {
-    this.workflow.cancel();
+  @action onDisconnect() {
+    this.workflow.cancel('DISCONNECTED');
```
lukemelia, line 196: Consider making 'DISCONNECTED' a constant

packages/web-client/app/components/card-pay/deposit-workflow/index.ts
```diff
@@ -151,21 +152,26 @@ class DepositWorkflow extends Workflow {
...
+    // currently we only cancel because of disconnections from either wallet
+    new WorkflowMessage({
       author: cardbot,
       message:
         'It looks like your wallet(s) got disconnected. If you still want to deposit funds, please start again by connecting your wallet(s).',
       includeIf() {
-        let message = this as NetworkAwareWorkflowMessage;
-        return !message.hasLayer1Account || !message.hasLayer2Account;
+        return (
+          (this as WorkflowPostable).workflow?.cancelationReason ===
```
lukemelia, line 162: There might be a way to make `this` typed more elegantly. https://www.typescriptlang.org/docs/handbook/2/functions.html#declaring-this-in-a-function

packages/web-client/app/components/card-pay/issue-prepaid-card-workflow/index.ts
```diff
@@ -46,6 +52,31 @@ class IssuePrepaidCardWorkflow extends Workflow {
...
+            );
+
+            if (sufficientFunds) {
+              return {
+                success: true,
+              };
+            } else {
+              return {
+                success: false,
+                reason: 'INSUFFICIENT_FUNDS',
```
lukemelia, line 76: Maybe this is part of an enum that includes 'DISCONNECTED'?

packages/web-client/app/models/workflow/workflow-card.ts
```diff
@@ -25,14 +37,26 @@ export class WorkflowCard extends WorkflowPostable {
...
     return this.workflow?.session;
   }
 
   @action onComplete() {
-    this.workflow?.emit('visible-postables-changed');
-    this.isComplete = true;
+    if (this.check) {
+      this.check().then((checkResult) => {
+        if (checkResult.success) {
+          this.workflow?.emit('visible-postables-changed');
```
lukemelia, line 50: I know this is not *new* in your PR, but.... why do we have to do this? Isn't setting isComplete enough state for us to react to?

packages/web-client/tests/acceptance/issue-prepaid-card-test.ts
```diff
@@ -76,7 +93,7 @@ module('Acceptance | issue prepaid card', function (hooks) {
     let layer2AccountAddress = '0x182619c6Ea074C053eF3f1e1eF81Ec8De6Eb6E44';
     layer2Service.test__simulateAccountsChanged([layer2AccountAddress]);
     layer2Service.test__simulateBalances({
-      defaultToken: toBN('250000000000000000000'),
+      defaultToken: SLIGHTLY_LESS_THAN_MAX_VALUE,
```
lukemelia, line 96: 👏

lukemelia (changes requested):
We think the failure reason is under-typed -Luke & Buck

packages/web-client/app/models/workflow/workflow-card.ts
Aierie: 'visible-postables-changed' is for test waiters/ the animated workflow.

packages/web-client/app/components/card-pay/deposit-workflow/index.ts
Aierie: The ts handbook is pretty helpful, I'm finding. Thanks for pointing this out.

packages/web-client/app/components/card-pay/deposit-workflow/index.ts
Aierie: Will make `DISCONNECTED` and `INSUFFICIENT_FUNDS` constants in workflow-config.

packages/web-client/app/components/card-pay/deposit-workflow/index.ts
Aierie: It's possible to type it within the options with the expected `this`, added this in https://github.com/cardstack/cardstack/pull/1759/commits/730f1995537e753f4846706a8eb1f185f9e2c192

packages/web-client/app/components/card-pay/deposit-workflow/index.ts
Aierie: Added these constants as entries within a typescript const type:

https://www.typescriptlang.org/docs/handbook/enums.html#objects-vs-enums

packages/web-client/app/models/workflow/workflow-card.ts
Aierie: Renamed to `visible-postables-will-change` and added a comment about why we are emitting this.

PR #1768, opened 2021-06-28
Fix/hardcoded network names

Aierie:
CS-1143

Replaces hardcoded network names with the correct ones identified at build time. For `*.ts` files, this is done by importing an object with references to layer 1 and layer 2 names. For `*.hbs` files, this is done using a helper.

Will wait for pending PRs for prepaid card issuance flow to be merged, then update this before merging it.


packages/web-client/app/components/card-pay/deposit-workflow/index.ts
```diff
@@ -10,12 +10,14 @@ import Layer1Network from '@cardstack/web-client/services/layer1-network';
...
 import { action } from '@ember/object';
+import { currentNetworkDisplayInfo as c } from '@cardstack/web-client/utils/web3-strategies/network-display-info';
+import { capitalize } from '@ember/string';
 
 class DepositWorkflow extends Workflow {
   name = 'Reserve Pool Deposit';
   milestones = [
     new Milestone({
-      title: 'Connect mainnet wallet',
+      title: `Connect ${c.layer1.networkType} wallet`,
```
Aierie, line 20: Unsure about replacing mainnet with testnet.

packages/web-client/app/components/card-pay/deposit-workflow/index.ts
```diff
@@ -52,22 +54,22 @@ class DepositWorkflow extends Workflow {
           componentName: 'card-pay/layer-one-connect-card',
         }),
       ],
-      completedDetail: 'Mainnet wallet connected',
+      completedDetail: `${capitalize(c.layer1.networkType)} wallet connected`,
```
Aierie, line 57: Wondering whether this should be the responsibility of the `NetworkCopywriting` type - capitalized vs non-capitalized stuff?

packages/web-client/app/utils/web3-strategies/ethereum.ts
```diff
@@ -1,7 +1,8 @@
 import Layer1ChainWeb3Strategy from './layer1-chain';
+import { networkDisplayInfo } from './network-display-info';
 
 export default class EthereumWeb3Strategy extends Layer1ChainWeb3Strategy {
   constructor() {
-    super('mainnet', 'Ethereum mainnet');
+    super('mainnet', networkDisplayInfo.mainnet.fullName);
```
backspace, line 6: Now that there’s a `networkDisplayInfo` to collect these programmatic and UI names, maybe this second argument to the constructor is unnecessary and can be derived in the constructors for the strategies? 🤔

packages/web-client/app/utils/web3-strategies/types.ts
```diff
@@ -54,3 +54,4 @@ export interface Layer2Web3Strategy extends Web3Strategy {
 
 export type TransactionHash = string;
 export type ChainAddress = string;
+export type NetworkSymbol = 'kovan' | 'sokol' | 'mainnet' | 'xdai';
```
backspace, line 57: I’m happy to see this graduate into this file haha

packages/web-client/app/components/card-pay/deposit-workflow/index.ts
backspace: Maybe… semi-relatedly, I anticipate some difficulty around things like “an Ethereum wallet” vs “a Kovan wallet” but I feel like it’s not worth doing anything in that direction until it‘s truly needed, it doesn’t seem like this implementation will hamper updates toward that end if a need does emerge.

packages/web-client/app/components/card-pay/deposit-workflow/confirmation/index.hbs
```diff
@@ -25,7 +25,7 @@
         <CardPay::AccountDisplay
           class="card-pay-confirmation__transaction-wallet"
           @size="small"
-          @name={{concat this.layer2Network.chainName " wallet"}}
+          @name={{concat this.layer1Network.chainName " wallet"}}
```
backspace, line 28: 😯

backspace (approved):
This looks suitable to me, I made an optional suggestion about deriving one of the chain strategy constructor arguments but definitely not blocking. I’m looking forward to seeing the proper strings in the different environments, I think it’ll help reduce my confusion as I still learn 🤓

packages/web-client/app/utils/web3-strategies/ethereum.ts
Aierie: This seems good, it reduces the number of files that need to know about `networkDisplayInfo`.

Aierie: Will change

packages/web-client/app/components/card-pay/deposit-workflow/confirmation/index.hbs
Aierie: Yep, this is the top half for:
image

packages/web-client/app/components/card-pay/deposit-workflow/index.ts
lukemelia: Yeah, it is not clear, because both sokol and kovan are testnets.

Aierie: Will check with design about whether we can replace 'mainnet' with the full testnet name.

backspace:
#1704 has “on xDai chain” hardcoded in for now in the prepaid card preview step, if/when it’s merged, could update that here? 🤞🏻

Aierie:
> #1704 has “on xDai chain” hardcoded in for now in the prepaid card preview step, if/when it’s merged, could update that here? 🤞🏻

Yep! Planning to merge this only after your PR.

packages/web-client/app/components/card-pay/deposit-workflow/index.ts
Aierie: Kovan's name will read `Kovan testnet`.

PR #1769, opened 2021-06-28
Display token conversions of spend amounts on FaceValue selection card

burieberry:
Fixes cs-1178

- Convert spend to token via layer2-chain strategy, which has access to the exchange rate sdk
- Adjust the face value options on the template to display approximate token amounts
- Disable option if user balance is less than the actual converted token amount

### DIFF
--- a/packages/web-client/app/components/card-pay/issue-prepaid-card-workflow/face-value/index.hbs
+++ b/packages/web-client/app/components/card-pay/issue-prepaid-card-workflow/face-value/index.hbs
@@ -30,7 +30,7 @@
             @symbol="SPEND"
             @balance={{this.selectedFaceValue.spendAmount}}
             @usdBalance={{spend-to-usd this.selectedFaceValue.spendAmount}}
-            @text={{concat this.selectedFaceValue.tokenAmount " " this.fundingTokenSymbol "*"}}
+            @text={{concat "≈ " this.selectedFaceValue.approxTokenAmount " " this.fundingTokenSymbol "*"}}
             data-test-face-value-display
           />
         </CardPay::LabeledValue>
@@ -38,11 +38,11 @@
         <div class="face-value-card__options">
           {{#each this.options as |option|}}
             <CardPay::IssuePrepaidCardWorkflow::FaceValue::Option
-              @disabled={{lt this.balanceFromBN (spend-to-usd option.spendAmount)}}
+              @disabled={{option.isOptionDisabled}}
               @checked={{eq option.spendAmount this.selectedFaceValue.spendAmount}}
               @spendFaceValue={{option.spendAmount}}
               @usdAmount={{spend-to-usd option.spendAmount}}
-              @tokenAmount={{option.tokenAmount}}
+              @approxTokenAmount={{option.approxTokenAmount}}
               @tokenSymbol={{this.fundingTokenSymbol}}
               @onInput={{fn this.chooseFaceValue option}}
             />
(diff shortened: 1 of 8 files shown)


lukemelia (approved):

packages/web-client/app/components/card-pay/issue-prepaid-card-workflow/face-value/index.hbs
```diff
@@ -28,22 +28,23 @@
...
-          {{#each this.faceValueOptions as |amount|}}
+          {{#each this.options as |option|}}
             <CardPay::IssuePrepaidCardWorkflow::FaceValue::Option
-              @disabled={{lt this.balanceFromBN (spend-to-usd amount)}}
-              @checked={{eq amount this.selectedFaceValue}}
-              @spendFaceValue={{amount}}
-              @usdAmount={{spend-to-usd amount}}
-              @approxDaiAmount={{spend-to-usd amount}}
-              @onInput={{fn this.chooseFaceValue amount}}
+              @disabled={{lt this.balanceFromBN (spend-to-usd option.spendAmount)}}
```
Aierie, line 41: Just realised that `balanceFromBN` is not accurate conversion of the token balance to USD. If you want to compare in USD, I think using the `token-to-usd` helper or service and converting to a number is better.

Comparing USD might not be good because we are losing decimal precision. It might be better to compare the raw token amounts fetched in `getTokenAmounts` vs the balance as `BN`.

packages/web-client/app/components/card-pay/issue-prepaid-card-workflow/face-value/index.ts
```diff
@@ -5,22 +5,51 @@ import { reads } from 'macro-decorators';
...
+      );
+      let tokenAmount = parseFloat(fromWei(result)).toFixed(2);
+      this.options = [
+        ...this.options,
+        {
+          spendAmount,
+          tokenAmount,
+        },
+      ];
+    });
```
Aierie, line 51: This `map` can be a `forEach`, since we aren't returning any values.

Aierie (changes requested):
Overall this works, main concern is the code we're using to compare token values can lead to confusing states when conversion of DAI to USD is not 1-1. Image below shows the 5000 spend option disabled when I appear to have enough DAI

image

packages/web-client/app/components/card-pay/issue-prepaid-card-workflow/face-value/index.ts
Aierie: Another concern with this map is, there is no guaranteed resolution order of requests to the sdk, so assignment to `this.options` can happen out of order. Code below is an alternative that will always return the options in the same order as `spendArr`.

```
this.options = await Promise.all(
      spendArr.map(async (spendAmount) => {
        let result = await this.layer2Network.convertFromSpend(
          symbol,
          spendAmount
        );
        let tokenAmount = parseFloat(fromWei(result)).toFixed(2);
        return {
          tokenAmount,
          spendAmount,
        };
      })
    );
```

Aierie: Determining whether there is enough token in the assignment of the `this.options` - caveat being this does not update when the balance updates, though converting `this.options` to a getter should change that.
```
       {
          spendAmount,
          canPurchase: this.layer2Network.defaultTokenBalance?.gte(
            toBN(result)
          ),
          tokenAmount,
        },
```

burieberry: Updated my PR to be something similar to your suggestions. Decided to add the approximate symbol back because otherwise we need to display 18 digits. None of these convert to round numbers. However, the `isOptionDisabled` takes into account the real token conversion, so the option will be available to select if the user actually has enough tokens.

Aierie (approved):

PR #1771, opened 2021-06-29
L2 can connect after clicking 'cancel' in wallet

Aierie:
CS-1174

Fixes the issue @habdelra was facing with being unable to connect after canceling. This seems like a potentially very confusing issue for users. 

The core issue is not having a new URI generated, and if generated, not updated in the QR code, so when WalletConnect sends us a 'disconnect' event (odd that it's a disconnect event that happens when users click 'cancel' before connection, but it works!), we:
- generate a new uri
- make qr code rebuild in response to uri change

And users should be able to use the QR code to connect.


lukemelia (changes requested):
Good catch! Let's use a {{did-update}} helper or modifier instead of observers to implement.

Aierie:
Changed to use `{{did-update}}`!

lukemelia (approved):

PR #1772, opened 2021-06-29
Enhancement/check and cancel workflows via workflow card check method

Aierie:
This aims to fix issues identified in #1764. Adds `check` and `failureReason` to `WorkflowCard`. `check` is a callback that should return `Promise<boolean>`, `failureReason` is a string that becomes the workflow's `cancelationReason` if `check` returns `false`. `check` is called within the `WorkflowCard`'s `onComplete` method, and cancels the workflow using the `failureReason` provided if the check fails. 

Rationale:
- Adding `check` to a `WorkflowCard` means we can control cancelations to occur at set points in the workflow more easily (as opposed to adding checking to the workflow, which would mean we need to keep track of the visibility of different postables within the workflow)
- We make the `check` callback in `onComplete` because all of the actions required to be taken in the card are assumed to be done (unlikely to be racing async calls).
- We prevent the `WorkflowCard` from completing if the check fails because completing a card starts showing subsequent postables.
- `check` is asynchronous because we don't know whether the workflow will have all information for a check ready at the point of the check. This can be made synchronous if we assume that all information required for a check will be made available in local state at the time of a check, however I feel like this is hard to guarantee.

Issues:
- as @backspace has mentioned, there probably needs to be a more dynamic way to provide the context of failure. This context would need to be accessed and used for informative cancelation messages

Other todos if this method works out:
- Cherry pick / extract and adjust way of getting previous postable and accompanying changes in #1764 to new PR

### DIFF
--- a/packages/web-client/app/components/card-pay/issue-prepaid-card-workflow/index.ts
+++ b/packages/web-client/app/components/card-pay/issue-prepaid-card-workflow/index.ts
@@ -11,7 +11,6 @@ import Layer2Network from '@cardstack/web-client/services/layer2-network';
 import { action } from '@ember/object';
 import { WorkflowPostable } from '@cardstack/web-client/models/workflow/workflow-postable';
 import { BN } from 'bn.js';
-import { toWei } from 'web3-utils';
 import { faceValueOptions } from './workflow-config';
 
 class IssuePrepaidCardWorkflow extends Workflow {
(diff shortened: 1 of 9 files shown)


lukemelia (changes requested):
Instead of:

```
async check(): Promise<boolean> {...}
failureReason: string = ...
```

I would suggest:

```ts
class CheckResult {
  constructor(readonly success:boolean, readonly reason?:string = undefined){}
}

async check(): Promise<CheckResult> { ... }
```

This avoids splitting a responsibility between the method and the getter, and opens things up for a given card to have more than one reason for a failed check.

Aierie:
Updated, opted for type over class since `CheckResult` defines data but not behaviour.

Aierie:
Merging first, since the base for this is a draft PR at the moment.

PR #1773, opened 2021-06-29
use wallet connect to sign txns in the CLI

habdelra:
This is an enhancement to the CLI that allows us to use the card wallet mobile app to sign txn in the CLI. This facilitates an alternate way to perform cardpay SDK commands from the web-client to help narrow down issues as well as to allow the card wallet team to work ahead of the web-client team in terms of their signing work.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -95,7 +95,7 @@
         "safes-view",
         "--network", "sokol",
         // Hassan's testing mnemonic feel free to use your own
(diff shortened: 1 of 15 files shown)


lukemelia (approved):
~26 hours from "Probably not possible" to PR -- nice work!

habdelra:
I feel bad taking credit, the real hard work was from the WalletConnect folks

PR #1776, opened 2021-06-30
Fix/walletconnect transaction never resolves

Aierie:
CS-1177

Workaround for `web3@1.3.x` not receiving 'confirmation' and 'receipt' events when used with WalletConnect. 

- Wrap contract method calls in TokenBridgeForeignSide in a Promise
- Makes existing polling for receipts timeout after a set duration, defaults to 30 seconds
- Resolves the Promise in TokenBridgeForeignSide if a receipt is successfully retrieved within the duration
- Throws an error if polling is not successful after the duration

### DIFF
--- a/packages/cardpay-sdk/sdk/utils/general-utils.ts
+++ b/packages/cardpay-sdk/sdk/utils/general-utils.ts
@@ -14,7 +14,9 @@ export async function networkName(web3: Web3): Promise<string> {
   return name;
 }
 
-export function waitUntilTransactionMined(web3: Web3, txnHash: string): Promise<TransactionReceipt> {
+export function waitUntilTransactionMined(web3: Web3, txnHash: string, duration = 30000): Promise<TransactionReceipt> {
+  let endTime = Number(new Date()) + duration;
+
   let transactionReceiptAsync = async function (
     txnHash: string,
     resolve: (value: TransactionReceipt | Promise<TransactionReceipt>) => void,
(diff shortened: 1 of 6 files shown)


Aierie:
@habdelra Wondering about suitable timeout durations for each place that waits for the transaction to be mined:
- `RevenuePool.getMerchantSafeFromTxn`
- `PrepaidCard.loadGasIntoPrepaidCard`
- `TokenBridgeHomeSide.waitForBridgingCompleted`
- `TokenBridgeForeignSide.unlockTokens`
- `TokenBridgeForeignSide.relayTokens`

habdelra:
I think a 5 minute timeout across the board seems pretty suitable. If we run into issues with that being too long (or too short) we'll adjust. I think they way you have it where if the caller has an opinion about the timeout they can specify one, otherwise we default to 5 minutes is very nice.

packages/cardpay-sdk/sdk/utils/general-utils.ts
```diff
@@ -14,20 +14,24 @@ export async function networkName(web3: Web3): Promise<string> {
   return name;
 }
 
-export function waitUntilTransactionMined(web3: Web3, txnHash: string): Promise<TransactionReceipt> {
+export function waitUntilTransactionMined(web3: Web3, txnHash: string, duration = 30000): Promise<TransactionReceipt> {
```
habdelra, line 17: this is more of a style thing, but lots of times its nice to show the math. so instead of `30000` you can say `1000 * 60 * 5`, then it's a little more clear we mean 5 minutes.

habdelra (approved):
This looks great 👍

habdelra:
where you able to confirm that this fixes the problem?

Aierie:
> where you able to confirm that this fixes the problem?

@habdelra yep, both WalletConnect and the MetaMask browser extension work with this. Will change the timeout to 5 minutes by default, and make the style change you requested. Thanks!

Aierie:
Merging for now, will add UI later; tracking in CS-1186

PR #1781, opened 2021-07-01
explicitly reject errors when polling for receipt

Aierie:
The `transactionHash` event callback is called in a `setTimeout` (probably because web3 is also polling?) so `Promise.catch` does not work without explicitly rejecting the promise, and using `TokenBridgeForeignSide.unlockTokens` and `TokenBridgeForeignSide.relayTokens` will result in uncaught errors in Promises, if any happen. This is fixed by explicitly catching then rejecting the error.


habdelra (approved):
👍

lukemelia (approved):

PR #1782, opened 2021-07-01
Prepaid-card creation including authentication and API call

lukemelia:
- [x] Passing tests
- [x] layer2-chain implementation of authentication and prepaid card issuing
- [x] should not result in a crash of cardwallet
- [x] /api/prepaid-card-customizations POST should not return 422 when payload is valid 
- [x] Coordinate with @backspace to get preview-issued-prepaid-card-cs-727 and this branch merged


lukemelia:
Took care of this in #1704

PR #1785, opened 2021-07-02
Enhancement/unlock and deposit error UI

Aierie:
CS-1186

Basic error ui to direct users to try again when something goes wrong, or to ask for help:

image

### DIFF
--- a/packages/web-client/app/components/card-pay/deposit-workflow/transaction-amount/index.ts
+++ b/packages/web-client/app/components/card-pay/deposit-workflow/transaction-amount/index.ts
@@ -119,54 +119,54 @@ class CardPayDepositWorkflowTransactionAmountComponent extends Component<Workflo
     }
   }
 
-  @action unlock() {
+  @action async unlock() {
     this.errorMessage = '';
-    this.isUnlocking = true;
-    taskFor(this.layer1Network.approve)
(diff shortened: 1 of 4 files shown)


Aierie:
@burieberry added CS-1199 to check how users will get help, will update this when that's resolved; leaving the mailto link for now.

packages/web-client/app/components/card-pay/deposit-workflow/transaction-amount/index.ts
```diff
@@ -116,46 +119,54 @@ class CardPayDepositWorkflowTransactionAmountComponent extends Component<Workflo
...
+    this.errorMessage = '';
+
+    try {
+      this.isUnlocking = true;
+      let transactionReceipt = await taskFor(
+        this.layer1Network.approve
+      ).perform(this.amountAsBigNumber, this.currentTokenSymbol);
+      this.isUnlocked = true;
+      this.unlockTxnReceipt = transactionReceipt;
+    } catch (e) {
```
lukemelia, line 132: Probably should also log the error, to help with debugging.

Aierie: Makes sense!

PR #1789, opened 2021-07-05
Add sdk functions to query balance of reward tokens

tintinthong:
Addition to cli:
- `rewardTokenBalances`: You can query for a payees reward balance for ALL tokens that tally has rewarded (`payee`==`owner of prepaid card`). 

Adding `RewardPool` class to sdk (connect to tally & reward pool):
- `getProofs`: This is a generic function that is used to query proof objects directly from tally. User can query ALL proofs by payee and token addresses. Proofs are needed to view balance and withdraw from the reward pool.
- `rewardTokensAvailable`: This is a query that returns array of all reward tokens by address that tally rewarded to payee. Currently, the reward pool doesn't tell you about so tally db keeps an index of this.
- `rewardTokenBalance`: This is the **main function** which queries token balance by payee and token address. I anticipate frontend devs would use this sdk function the most.
- `rewardTokenBalances`: `rewardTokenBalances` is a catch-all for `rewardTokenBalance`.

You can inspect the tally api [here](https://tally-service-staging.stack.cards/).

If you want to test tally rewards, you can pay 50 spend with a prepaid card on sokol; you should get 10 dai to your owner address of the prepaid card. The rewards might take some time -- maybe ~5 mins.

Note: 
- `/reward-balance` **on tally** is intentionally excluded because it requires tally to make a web3 connection to query the reward pool for balances. You can see that this sdk handles that.
- `withdraw` user flow is intentionally excluded because it requires changes to reward pool contract.

Demo:
https://www.loom.com/share/7ad853fef1be45e596775247812bf961


habdelra (approved):
This looks great! good job 👍

PR #1792, opened 2021-07-05
Fix/wallet connection states

Aierie:
Extract providers and related event handling for layer 1 into their own class, and expose an API with callbacks centered around connection events we care about and that the Dapp's logic deals with - account changes (connection, disconnection), errors, and connecting to an incorrect chain. This PR as is should fix CS-1179 and sets up for CS-1088.

The API is intentionally very specific at the moment, the only callbacks that are in use at the moment (`onConnect`, `onDisconnect`, `onIncorrectChain`) are the events that we are concerned about and want to handle in the Dapp logic. There is a "standard" for Ethereum events (https://github.com/ethereum/EIPs/issues/2319), but I'm unsure about adoption, and WalletConnect and MetaMask seem to be using these events in ways fairly different from each other. I felt it was better to use specific callbacks to ensure we are handling the events we want, and keep the complexity out of `layer1-chain`.

### DIFF
--- a/packages/web-client/app/utils/web3-strategies/layer-1-connection-manager.ts
+++ b/packages/web-client/app/utils/web3-strategies/layer-1-connection-manager.ts
@@ -14,73 +14,7 @@ const GET_PROVIDER_STORAGE_KEY = (chainId: string | number) =>
   `cardstack-chain-${chainId}-provider`;
 const WALLET_CONNECT_BRIDGE = 'https://safe-walletconnect.gnosis.io/';
 
-export async function createConnectionManagerFromLocalStorage(
-  web3: Web3,
-  options: ConnectionManagerOptions
-): Promise<ConnectionManager | void> {
-  try {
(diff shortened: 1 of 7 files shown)


packages/web-client/app/utils/web3-strategies/layer-1-connection-manager.ts
```diff
@@ -0,0 +1,313 @@
...
+ * This class does not, at the moment, store any state used directly by the UI besides providerId.
+ */
+export abstract class ConnectionManager
+  implements Emitter<ConnectionManagerEvents> {
+  networkSymbol: NetworkSymbol;
+  chainId: number;
+  simpleEmitter: SimpleEmitter;
+  broadcastChannel: BroadcastChannel;
+  protected provider: any;
+  protected connected = false;
```
Aierie, line 53: Don't really like adding this here since the Layer1Chain has its own understanding of connected state - having a wallet address, but I wanted to make sure that we didn't end up having multiple onDisconnect callbacks.

PR #1797, opened 2021-07-06
Fix test date parsing strings in Safari

backspace:
The test was failing in Safari but succeeds with this change.

2021-07-06 13-30-10-date-parse
2021-07-06 13-32-03-date-parse

### DIFF
--- a/packages/web-client/tests/integration/components/workflow-thread-test.ts
+++ b/packages/web-client/tests/integration/components/workflow-thread-test.ts
@@ -60,10 +60,13 @@ module('Integration | Component | workflow-thread', function (hooks) {
       }),
     ];
     workflow.epilogue.postables = [postable4];
-    postable1.timestamp = new Date(Date.parse('2021-05-20T00:00:00'));
-    postable2.timestamp = new Date(Date.parse('2021-05-20T00:00:00'));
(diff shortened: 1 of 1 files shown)


Aierie (approved):
This does seem to fix it! 

However MDN doesn't recommend usage of `Date.parse` in general due to browser differences in implementation (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/parse). Could consider:

`new Date(2021, 4, 20, 0, 0, 0) // year, monthIndex, day, hour, minute, second`

Though that is less easy to read (especially `monthIndex`). For convenience, if you decide to use this:

```
postable1.timestamp = new Date(2021, 4, 20, 0, 0, 0);
postable2.timestamp = new Date(2021, 4, 20, 0, 0, 0);
postable3.timestamp = new Date(2021, 4, 21, 0, 0, 0);
postable4.timestamp = new Date(2021, 4, 21, 0, 0, 0);
```

backspace:
ya I saw that on MDN too… I feel like maybe it’s less important in test-only code, thoughts??

Aierie:
I think it's less likely to be an issue in test-related code too. Raised mainly because there is a tendency for people to mimic/copy existing code for the sake of convention and convenience. Writing about this more, I think it makes sense to remove `Date.parse`.

PR #1799, opened 2021-07-06
Add amount card for withdrawal workflow

backspace:
This fills out the amount-setting card for the withdrawal workflow:

screencast 2021-07-08 14-50-23

It’s mostly a copy of `CardPay::DepositWorkflow::TransactionAmount`, but it uses a `Layer2BalanceView` component renamed from `IssuePrepaidCardWorkflow::FaceValue::BalanceView`. There’s an opportunity for future extraction here as the currency input field and accompanying validation could be shared.

One significant caveat: I’ve only exercised this with DAI.CPXD and not CARD.CPXD. See CS 1224 for more.

### DIFF
--- a/packages/web-client/app/components/card-pay/withdrawal-workflow/transaction-amount/index.css
+++ b/packages/web-client/app/components/card-pay/withdrawal-workflow/transaction-amount/index.css
@@ -1,0 +1,98 @@
+.withdrawal-transaction-amount {
+  --tx-amount-banner-height: 3.125rem; /* 50px */
+  --tx-amount-input-height: 3.75rem; /* 60px */
+}
+
+.withdrawal-transaction-amount__input-section > * + * {
+  margin-top: var(--boxel-sp-lg);
+}
+
+.withdrawal-transaction-amount__input-container {
+  --input-currency-icon-size: 2.5rem;
+  --input-currency-text-width: 5ch;
+  --input-currency-icon-space: var(--boxel-sp-xs);
+  --input-currency-text-space: var(--boxel-sp-sm);
+
+  position: relative;
+}
+
+.withdrawal-transaction-amount__input-token-balance {
+  font-family: var(--boxel-font-family);
+  font-size: 1.25rem;
+  line-height: calc(27 / 20);
+  letter-spacing: var(--boxel-lsp-xs);
+}
+
+.withdrawal-transaction-amount__input {
+  height: var(--tx-amount-input-height);
+  padding-left: calc(var(--input-currency-icon-size) + var(--input-currency-icon-space) * 2);
+  padding-right: calc(var(--input-currency-text-width) + var(--input-currency-text-space) * 2);
+  font: inherit;
+  letter-spacing: inherit;
+}
+
+.withdrawal-transaction-amount__currency-icon {
+  position: absolute;
+  width: var(--input-currency-icon-size);
+  height: var(--tx-amount-input-height);
+  left: var(--input-currency-icon-space);
+  top: 0;
+  user-select: none;
+}
+
+.withdrawal-transaction-amount__currency-text {
+  position: absolute;
+  height: var(--tx-amount-input-height);
+  display: flex;
+  align-items: center;
+  right: var(--input-currency-text-space);
+  top: 0;
+  user-select: none;
+}
+
+.withdrawal-transaction-amount__banner-field-label {
+  align-self: flex-start;
+  display: flex;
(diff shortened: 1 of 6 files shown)


packages/web-client/app/components/card-pay/transaction-amount/index.ts
```diff
@@ -48,6 +78,10 @@ class CardPayDepositWorkflowTransactionAmountComponent extends Component<Workflo
       balance = this.layer1Network.daiBalance;
     } else if (this.currentTokenSymbol === 'CARD') {
       balance = this.layer1Network.cardBalance;
+    } else if (this.currentTokenSymbol === 'DAI.CPXD') {
+      balance = this.layer2Network.defaultTokenBalance;
+    } else if (this.currentTokenSymbol === 'CARD.CPXD') {
+      balance = this.layer2Network.cardBalance;
```
backspace, line 84: This extends what was already present, but maybe it could be encapsulated in the service somehow? And is `defaultTokenBalance` even correct?

burieberry: `defaultTokenBalance` in layer2 refers to the DAI.CPXD balance

lukemelia (changes requested):
I think that the conditionals in the extracted component introduce an unacceptable level of complexity to this component and that we should explore other options, even at the cost of some duplication. This is arguable, and I'm OK with the proposed path if everyone else thinks it is OK. To me, one hint that this is not a good solution is asking "Would be happy to reuse this in a 3rd workflow with the same approach?" The answer is a clear no.

packages/web-client/app/components/card-pay/balance-display/index.hbs
```diff
@@ -22,7 +22,7 @@
         </div>
         {{#if @usdBalance}}
           <div class="balance-display__text" data-test-balance-display-usd-amount>
-            ${{@usdBalance}} USD
+            ${{@usdBalance}} USD{{if @usdFootnote "*"}}
```
Aierie, line 25: Afaik the disclaimer is always true. USD balance is an estimate. I think there was a designwide decision on how this is presented - with/without the *, using the estimate (≈) symbol or not. @burieberry  do you remember what it is, if there was one?

packages/web-client/app/components/card-pay/transaction-amount/index.hbs
```diff
@@ -0,0 +1,158 @@
...
+>
+  <ActionCardContainer::Section @title={{this.title}} @dataTestId="transaction-amount">
+    <div class="transaction-amount__input-section">
+      <Boxel::Field
+        class="transaction-amount__field"
+        @label="Funding From:"
+        @labelClass="transaction-amount__banner-field-label"
+      >
+        <CardPay::TransactionAmount::BalanceView
+        @wallet={{network-display-info "layer1" "fullName"}}
```
Aierie, line 14: This should also change to `layer2` for the withdrawal workflow, I believe.

Aierie:
> I think that the conditionals in the extracted component introduce an unacceptable level of complexity to this component and that we should explore other options, even at the cost of some duplication. This is arguable, and I'm OK with the proposed path if everyone else thinks it is OK. To me, one hint that this is not a good solution is asking "Would be happy to reuse this in a 3rd workflow with the same approach?" The answer is a clear no.

I agree, the conditionals do make this hard to update this component for new use cases/ to update it if contents change.

Perhaps removing duplication at a lower level might be good? A possible target is the token amount input, which seems like a reusable component, contributes a fair amount of logic that would be otherwise duplicated, and needs some improvements that are good for both the deposit and withdraw workflows (CS-781).

packages/web-client/app/components/card-pay/balance-display/index.hbs
burieberry: I agree with the comments about the complexity of this card.

In terms of the approximate sign (≈), in a design meeting we said that we won't use it. (I had to use it in one place though (the face value selection card in the issue card workflow) because the DAI amounts were very rounded.) The footnotes remain as is.

backspace:
Thanks for the feedback, all. @Aierie and I paired on this and started fresh without attempting to generalise the component. We concluded that the prepaid card workflow’s balance view component was more suitable for withdrawal, so we promoted it outside its workflow, but it didn’t need any changes otherwise.

I didn’t extract the currency input component, but if someone wants to pair on that to include it in this PR, let me know!

PR #1801, opened 2021-07-07
catch and reject web3/json-rpc errors in l1 -> l2 bridge

Aierie:
Directly reject errors that originate from `Contract.send` calls in `TokenBridgeForeignSide`. From web3's [docs](https://web3js.readthedocs.io/en/v1.2.11/web3-eth-contract.html#methods-mymethod-send) about the error event:

> "error" returns error: Error: Fired if an error occurs during sending. If the transaction was rejected by the network with a receipt, the receipt will be available as a property on the error object.

Necessary to catch user rejections of transaction requests and other errors to display helpful messages on the unlock and deposit UI.


habdelra (commented):
👍

habdelra:
These "promi-events" that web3 uses are definitely tricky to work with when you need to deal with errors. thanks for digging into this.

PR #1803, opened 2021-07-08
Add withdrawal tx confirmation card

burieberry:
- Add transaction approval card
- Extract `CardPay::BalanceViewBanner` component
- Update NestedItems component to include a third level
- Style fixes in expanded view of the banner component
- Update tests

We will need to revisit this component:
- When the withdrawal amount set in the previous card is available. We need to replace the current withdrawal amount with the user selected amount.
- Need to request confirmation from Card Wallet app

confirm-default
confirm-expanded
confirm-memorialized

### DIFF
--- a/packages/web-client/app/components/card-pay/issue-prepaid-card-workflow/face-value/index.hbs
+++ b/packages/web-client/app/components/card-pay/issue-prepaid-card-workflow/face-value/index.hbs
@@ -13,7 +13,7 @@
         </span>
       </p>
       <CardPay::LabeledValue @vertical={{true}} @label="Funding From:">
-        <CardPay::IssuePrepaidCardWorkflow::FaceValue::BalanceView
+        <CardPay::BalanceViewBanner
           @walletAddress={{this.layer2Network.walletInfo.firstAddress}}
           @depotAddress={{this.layer2Network.depotSafe.address}}
           @token={{this.fundingToken}}
--- a/packages/web-client/app/components/card-pay/withdrawal-workflow/transaction-approval/index.hbs
+++ b/packages/web-client/app/components/card-pay/withdrawal-workflow/transaction-approval/index.hbs
@@ -1,19 +1,61 @@
 <ActionCardContainer
-  @header="Withdrawal"
+  @header="Confirmation"
   @isComplete={{@isComplete}}
   data-test-withdrawal-transaction-approval-is-complete={{@isComplete}}
 >
-  <ActionCardContainer::Section>
+  <ActionCardContainer::Section @title="Confirm the withdrawal details">
+    <CardPay::LabeledValue @vertical={{true}}  @label="Funding From:">
+      <CardPay::BalanceViewBanner
+        @walletAddress={{truncate-middle this.layer2Network.walletInfo.firstAddress}}
+        @depotAddress={{truncate-middle this.depotAddress}}
+        @token={{this.withdrawalToken}}
+        @balance={{format-token-amount this.tokenBalance}}
+        data-test-withdrawal-tx-approval-balance
+      />
+    </CardPay::LabeledValue>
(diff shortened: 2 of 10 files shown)


lukemelia (approved):
Looks good! Please add subtasks in Linear for the remaining TODOs you identified.

Aierie (approved):
Looks good!

packages/web-client/app/components/card-pay/withdrawal-workflow/transaction-approval/index.ts
```diff
@@ -1,17 +1,84 @@
...
+
+  get depotAddress() {
+    return this.layer2Network.depotSafe?.address || undefined;
+  }
+
+  get isDisabled() {
+    return (
+      !this.depotAddress ||
+      !this.withdrawalAmount ||
+      this.withdrawalAmount.isZero() ||
```
backspace, line 56: Maybe it doesn’t matter but it should be impossible to get to this step with a zero withdrawal amount because of validation in the amount card 🤔

packages/web-client/app/components/card-pay/issue-prepaid-card-workflow/face-value/index.hbs
```diff
@@ -13,7 +13,7 @@
         </span>
       </p>
       <CardPay::LabeledValue @vertical={{true}} @label="Funding From:">
-        <CardPay::IssuePrepaidCardWorkflow::FaceValue::BalanceView
+        <CardPay::BalanceViewBanner
```
backspace, line 16: In our pairing, @Aierie and I decided to rename this `Layer2BalanceView` (as contrasted with the `BalanceView` within the deposit workflow that’s about layer 1), what do you think of that?

packages/web-client/app/components/card-pay/withdrawal-workflow/transaction-approval/index.hbs
```diff
@@ -1,19 +1,61 @@
...
+      />
+    </CardPay::LabeledValue>
+    <CardPay::LabeledValue @vertical={{true}} @label="Amount to Withdraw:">
+      <CardPay::BalanceDisplay
+        class="withdrawal-transaction-approval__amount-display"
+        @size="large"
+        @icon={{this.withdrawalToken.icon}}
+        @symbol={{this.tokenSymbol}}
+        @balance={{format-token-amount this.withdrawalAmount}}
+        @text={{concat "$" (token-to-usd this.convertibleSymbol this.withdrawalAmount) " USD*"}}
```
backspace, line 23: I didn’t realise this was an option vs `@usdBalance`, I’ll adopt this in #1799 vs the unsightly `@usdFootnote` option I added, thanks.

backspace (approved):
I had a question about naming but otherwise this looks good, once this is merged I can update #1799 with the selected amount.

packages/web-client/app/components/card-pay/issue-prepaid-card-workflow/face-value/index.hbs
burieberry: I was thinking of using this component for that purpose too, but if it ends up being too complex, we can have two different ones that are named more specifically.

packages/web-client/app/components/card-pay/withdrawal-workflow/transaction-approval/index.hbs
burieberry: > I didn’t realise this was an option vs `@usdBalance`, I’ll adopt this in #1799 vs the unsightly `@usdFootnote` option I added, thanks.

I'm fine with that approach, too, if we want to follow that approach. Another idea is to extract a new component for the amount display with the border around it, since it's used in a bunch of places.

PR #1804, opened 2021-07-08
Add Withdraw sdk (bridge from L2 -> L1) to SDK

habdelra:
This completes the final part of the withdraw support in the SDK. Albeit, it's almost impossible to test this with the current ethereum node issues we are encountering


lukemelia (approved):
👍  Looks good. +1 for the added callbacks too!

PR #1805, opened 2021-07-08
Fix issue with persisting card token choice in balance chooser component

lukemelia:


packages/web-client/tests/integration/components/card-pay/withdrawal-workflow/choose-balance-test.ts
```diff
@@ -0,0 +1,88 @@
...
+import { toBN } from 'web3-utils';
+import { DepotSafe } from '@cardstack/cardpay-sdk/sdk/safes';
+import WorkflowSession from '@cardstack/web-client/models/workflow/workflow-session';
+
+module(
+  'Integration | Component | card-pay/withdrawal-workflow/choose-balance',
+  function (hooks) {
+    setupRenderingTest(hooks);
+
+    test('It should show a selection UI if layer 1 is not connected', async function (assert) {
```
backspace, line 15:
```suggestion
    test('It should allow a layer 2 balance to be chosen', async function (assert) {
```

backspace (approved):
I noticed a possible copypaste error but otherwise this looks good!

PR #1807, opened 2021-07-09
Fix/clear hub auth

Aierie:
CS-1239

Remove the auth token from local storage when a user disconnects, or when the prepaid card customization saving fails due to an invalid auth token. I think using customized errors which hold codes and communicative messages could be useful, do people have any thoughts about how to do this?

Edit: Noticed that the case where we have multiple errors is returned is for invalid prepaid card customization attributes, seems straightforward enough to pass the error information up to the calling component so it can show error messages as a first step. To follow up with UI in [task 1240](https://linear.app/cardstack/issue/CS-1240/add-ui-for-errors-to-prepaid-card-preview-card)

### DIFF
--- a/packages/web-client/app/services/card-customization.ts
+++ b/packages/web-client/app/services/card-customization.ts
@@ -190,6 +190,27 @@ export default class CardCustomization extends Service {
       }
     );
     let customization = yield response.json();
+    if (customization.errors) {
+      if (
+        customization.errors.length === 1 &&
+        Number(customization.errors[0].status) === 401 &&
+        customization.errors[0].title === 'No valid auth token'
+      ) {
+        console.error(
+          'Failed to store prepaid card customization due to invalid auth token'
+        );
+        this.hubAuthentication.authToken = null;
+        throw new Error('No valid auth token');
+      } else {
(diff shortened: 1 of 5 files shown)


packages/web-client/app/services/card-customization.ts
```diff
@@ -190,6 +190,27 @@ export default class CardCustomization extends Service {
...
+      } else {
+        // TODO: this should be changed to a form that communicates the errors
+        console.error(
+          'Failed to store prepaid card customization, got errors:',
+          customization.errors
+        );
+        throw new Error('Failed to store prepaid card customizations');
+      }
+    }
+
```
lukemelia, line 213: Consider breaking this task body up by extracting a couple of methods.

lukemelia (approved):

packages/web-client/app/services/card-customization.ts
Aierie: It is pretty huge. Will do in CS-1240, where the errors that need to be handled might be clearer.

backspace:
re customised errors, would this be as part of an event? Maybe some simple event wrapper would work so it’s not just a bare string but something that can have an error attached…? 🤔

PR #1809, opened 2021-07-09
Add withdrawal status card

backspace:
This adds a new withdrawal transaction status card, similar to the deposit workflow’s status card:

screencast 2021-07-14 17-21-44-two-step

It assumes these properties exist in the workflow session, and for now has inserted placeholder values in the `TransactionApproval` component (and the component test’s workflow session):
* `layer1BlockHeightBeforeBridging`
* `relayTokensTxnReceipt`

It adds `completedLayer1TransactionReceipt` to the workflow session. I see now that #1810 is expecting a `completedLayer2TransactionReceipt` property, it may be that I’m mixed up on this 🤔

### DIFF
--- a/packages/web-client/app/components/card-pay/withdrawal-workflow/transaction-status/index.hbs
+++ b/packages/web-client/app/components/card-pay/withdrawal-workflow/transaction-status/index.hbs
@@ -5,6 +5,7 @@
 >
   <ActionCardContainer::Section
     @title={{concat "Bridging tokens to " (network-display-info "layer1" "fullName")}}
+    @icon={{if this.isInProgress "clock" "success-bordered"}}
     @dataTestId="withdrawal-confirmed"
   >
     <Boxel::ProgressSteps
--- a/packages/web-client/app/components/card-pay/withdrawal-workflow/transaction-status/index.hbs
+++ b/packages/web-client/app/components/card-pay/withdrawal-workflow/transaction-status/index.hbs
@@ -44,15 +44,4 @@
       </span>
     </Boxel::ProgressSteps>
   </ActionCardContainer::Section>
-  <Boxel::ActionChin
-    @state={{if @isComplete "memorialized" "default"}}
-    data-test-withdrawal-transaction-confirmed
-  >
-    <:default as |d|>
-      <d.ActionButton {{on "click" this.toggleComplete}}>
-        Confirm
-      </d.ActionButton>
-    </:default>
-  </Boxel::ActionChin>
-
 </ActionCardContainer>
--- a/packages/web-client/app/components/card-pay/withdrawal-workflow/transaction-status/index.hbs
+++ b/packages/web-client/app/components/card-pay/withdrawal-workflow/transaction-status/index.hbs
@@ -28,14 +28,9 @@
           </Boxel::Button>
         {{/if}}
         {{#if (eq progressStep.index 1)}}
-          <Boxel::Button @as="anchor" @size="extra-small" @href={{this.bridgeExplorerUrl}} target="_blank" rel="noopener" data-test-bridge-explorer-button>
-            View in Bridge Explorer
-          </Boxel::Button>
-        {{/if}}
-        {{#if (eq progressStep.index 2)}}
           {{#if progressStep.completed}}
-            <Boxel::Button @as="anchor" @size="extra-small" @href={{this.withdrawalTxnViewerUrl}} target="_blank" rel="noopener" data-test-etherscan-button>
-              View on Etherscan
+            <Boxel::Button @as="anchor" @size="extra-small" @href={{this.bridgeExplorerUrl}} target="_blank" rel="noopener" data-test-bridge-explorer-button>
+              View in Bridge Explorer
             </Boxel::Button>
           {{else}}
             Processing...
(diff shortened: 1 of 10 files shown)


packages/web-client/app/components/card-pay/withdrawal-workflow/transaction-status/index.ts
```diff
@@ -0,0 +1,70 @@
...
+  get progressSteps() {
+    return [
+      {
+        title: `Withdraw tokens from ${c.layer2.fullName}`,
+      },
+      {
+        title: `Bridge tokens from ${c.layer2.fullName} to ${c.layer1.fullName}`,
+      },
+      {
+        title: `Release tokens on ${c.layer1.conversationalName}: ${c.layer2.shortName}`, // FIXME script says "DAI", not "xDai" 🤔
```
backspace, line 45: The script specifies

> Release tokens on mainnet: DAI

while the current configuration in `network-display-info` would only produce this

> Release tokens on mainnet: xDai

if this departure from the script seems okay, I can remove this note!

lukemelia: The tokens being released on mainnet would be DAI and not xDai, so I don't think the departure is OK.

packages/web-client/app/components/card-pay/withdrawal-workflow/transaction-status/index.hbs
```diff
@@ -0,0 +1,47 @@
...
+            View on Blockscout
+          </Boxel::Button>
+        {{/if}}
+        {{#if (eq progressStep.index 1)}}
+          <Boxel::Button @as="anchor" @size="extra-small" @href={{this.bridgeExplorerUrl}} target="_blank" rel="noopener" data-test-bridge-explorer-button>
+            View in Bridge Explorer
+          </Boxel::Button>
+        {{/if}}
+        {{#if (eq progressStep.index 2)}}
+          {{#if progressStep.completed}}
```
lukemelia, line 36: I think we ideally want to show this etherscan button as soon as we have `withdrawalTxnViewerUrl`, even if the operation is not yet complete.

lukemelia (approved):
Looks great overall, assuming we fix the xdai/dai issue and get the buttons showing up as early as possible. FWIW, I will share that I have a mostly negative relationship with sinon. In my experience, it adds test implementation complexity vs hand-rolled mocks. I would not seek to impose my sinon opinion on you or the team, just something to consider as we use it: "What would this look like without sinon? Would it be easier or harder to understand?"

packages/web-client/app/components/card-pay/withdrawal-workflow/transaction-status/index.ts
Aierie: Yep, DAI here is referring to the token's symbol, not the network's name. We should use DAI.

Aierie: This should be obtained from the component's workflow session's `withdrawalToken` property.

packages/web-client/app/components/card-pay/withdrawal-workflow/transaction-status/index.hbs
Aierie: Had a look, there is a [task in Linear for claiming tokens](https://linear.app/cardstack/issue/CS-1234/withdrawal-claim-card). This task has a proposal for the addition of a claim card that replaces step 3. There is also a note in the original task about this.

Aierie (changes requested):
Based on reviewing the tasks in the Linear issue, I think there might be some UI changes we need for the claim card, though will wait on @burieberry to confirm, since these changes seem to be proposed but not confirmed.

Aierie:
If we do remove step 3 (Release tokens on xxx chain: TOKEN), I think step 2 will need some adjustment to appropriately convey a "waiting" state, a responsibility that was held by step 3 before this (Processing...); Might also want to check with design whether this change would need to be applied to the deposit flow.

packages/web-client/app/components/card-pay/withdrawal-workflow/transaction-status/index.ts
backspace: oh yes, of course, thank you both. I can use the `getUnbridgedSymbol` function I added for the withdrawal amount card 🎉

backspace: fixed in 01a9c66b8

backspace:
I should have read everything again before making the change for the third message, not realising it was meant for removal!

regarding the waiting state, what about something like this?

image

I recognise the chin is meant for actions, but I wondered whether it might be okay to stretch it this way, somewhat akin to waiting for the user to do something in Card Wallet 🤔

Otherwise, any suggestions about how I can proceed with this?

burieberry (approved):

burieberry:
> I should have read everything again before making the change for the third message, not realising it was meant for removal!
> 
> regarding the waiting state, what about something like this?
> 
> image
> 
> I recognise the chin is meant for actions, but I wondered whether it might be okay to stretch it this way, somewhat akin to waiting for the user to do something in Card Wallet 🤔
> 
> Otherwise, any suggestions about how I can proceed with this?

I thought we were displaying `Processing...` instead of the Bridge Explorer link button until we have a link. There isn't supposed to be a chin for this card.

As for the upcoming changes regarding the claim card, we have a general idea but we don't know the details until we actually implement the bridging and claiming functionality. So this card can be merged as is, until we know more, and then there'll probably be some changes.

backspace:
👍🏻  okay thanks, I’ve removed the third step and the button won’t appear until `awaitBridged` has completed. I also noticed that the title icon was missing so I’ve added it now:

screencast 2021-07-14 17-21-44-two-step

PR #1810, opened 2021-07-09
Add withdrawal summary card

burieberry:
To be added once we have the real withdrawal: links for Blockscout, Bridge Explorer, Etherscan

with-actions


packages/web-client/tests/acceptance/withdrawal-test.ts
```diff
@@ -294,6 +294,42 @@ module('Acceptance | withdrawal', function (hooks) {
...
+    assert
+      .dom(
+        '[data-test-withdrawal-transaction-confirmed-to] [data-test-bridge-item-amount]'
+      )
+      .containsText('200.00 DAI');
+    assert
+      .dom(
+        '[data-test-withdrawal-transaction-confirmed-to] [data-test-bridge-item-wallet]'
+      )
+      .containsText(layer1AccountAddress);
```
backspace, line 331: I’m at EOD so I won’t do a thorough review today, but one comment would be that in discussions with L&M, the acceptance tests are starting to feel enormous, what do you think about moving these assertions into a component test? I took a step toward that [here](https://github.com/cardstack/cardstack/pull/1809/commits/37d07cd17705e83f39572aa34f8c8abe62d34214) for example, now the acceptance test for that card only has a single assertion. That card in particular doesn’t add anything to `workflowSession`, but I think it could be good to test that user actions result in values being stored under particular variables in the session state 🤔

I retain a desire for acceptance tests to assert on the flow of choices through the entire workflow, so maybe the user-entered `200` would be good to check for at least. Still figuring all this out! 💞

lukemelia (approved):
Nice refactor -- the components are very expressive. I agree with @backspace's comments about acceptance tests and integration tests. Let's make sure we have linear tasks to update tests accordingly and to update the links currently missing in this PR.

Aierie (approved):
Seems good. Are we omitting `Bridge::Action` for the withdrawal component?

burieberry:
> Seems good. Are we omitting `Bridge::Action` for the withdrawal component?

Ah you're right. We do have it here, too.

PR #1814, opened 2021-07-12
Fix/metamask incorrect chain

Aierie:
Focus of this PR is to not allow MetaMask to connect with the wrong chain undetected upon initial connection. As dApp-side disconnection from MetaMask is a faux-disconnection, we need to manually check for this within the connect and reconnect methods, the `accountsChanged` event listener is not a reliable place to do this.

Also made disconnections more reliable by removing a check to prevent emitting doubled-up disconnect events (WalletConnect emits one for its local disconnection and another for the wallet-side disconnection). Responsibility to decide whether or not a disconnection is relevant to communicate to the UI rests completely with layer 1 chain as of https://github.com/cardstack/cardstack/pull/1814/commits/f9419c492b774b78bbd513ea2fea6148c5be789a, we always clear connection state upon disconnect.

### DIFF
--- a/packages/web-client/app/utils/web3-strategies/layer-1-connection-manager.ts
+++ b/packages/web-client/app/utils/web3-strategies/layer-1-connection-manager.ts
@@ -159,10 +159,6 @@ class MetaMaskConnectionManager extends ConnectionManager {
       return;
     }
 
-    // remove all listeners that previous instances of metamask connections have added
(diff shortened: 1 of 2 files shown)


lukemelia (approved):
Looks good. All this logic is making me wonder if we should be modeling this as a state machine.

Aierie:
> Looks good. All this logic is making me wonder if we should be modeling this as a state machine.

I do like that idea, but not very clear on how to start it.

PR #1816, opened 2021-07-12
Add logic for catching errors and providing useful error messages in prepaid card creation

Aierie:


packages/web-client/app/components/card-pay/issue-prepaid-card-workflow/preview/index.ts
```diff
@@ -32,23 +33,52 @@ export default class CardPayDepositWorkflowPreviewComponent extends Component<Ca
...
+      // - Other arbitrary errors from request to save prepaid card customizations: try again, reconnect, restart, cardstack support
+
+      yield taskFor(this.layer2Network.issuePrepaidCard)
+        .perform(this.faceValue, customization.did)
+        .then((address: string) => {
+          this.args.workflowSession.update('prepaidCardAddress', address);
+          this.args.onComplete();
+        });
+      // Possible errors:
+      // - Not enough to create a card: top up. should cancel the workflow w INSUFFICIENT_FUNDS, and include the necessary amount. cancelation needs some updates to allow showing the necessary amount
```
Aierie, line 65: This will be done in a separate PR.

packages/web-client/app/components/card-pay/issue-prepaid-card-workflow/preview/index.ts
```diff
@@ -29,26 +29,49 @@ export default class CardPayDepositWorkflowPreviewComponent extends Component<Ca
...
+        // We probably want to cancel the workflow at this point
+        // And tell the user to go deposit funds
+        this.error = 'INSUFFICIENT_FUNDS';
+      } else if (tookTooLong) {
+        this.error = 'TIMEOUT';
+      } else {
+        // Basically, for pretty much everything we want to make the user retry or seek support
+        this.error = 'DEFAULT';
+      }
+    }
```
Aierie, line 74: @habdelra currently we're catching two errors from the sdk's prepaid card creation and displaying specific error messages for these cases:

When the transaction times out: We tell the user to check with Cardstack support about your transaction (we should give them the transaction hash for this)
When there aren't enough funds: We tell the user to add funds via the deposit workflow
Do you think there are any other cases to catch, or does this cover it? Also, wondering whether you have any thoughts on more structured/specialized errors that we can identify by type/by a shared constant rather than string matching? Having errors like this would make it easy to get the minimum amount required for the prepaid card the user wants, so we can tell the user in the UI.

packages/web-client/app/components/card-pay/issue-prepaid-card-workflow/preview/index.ts
```diff
@@ -29,26 +29,49 @@ export default class CardPayDepositWorkflowPreviewComponent extends Component<Ca
...
+
+      yield taskFor(this.layer2Network.issuePrepaidCard)
+        .perform(this.faceValue, customization.did)
+        .then((address: string) => {
+          this.args.workflowSession.update('prepaidCardAddress', address);
+          this.args.onComplete();
+        });
+    } catch (e) {
+      let insufficientFunds = e.message.startsWith(
+        'Safe not have enough balance to make prepaid card(s).'
```
habdelra, line 59: my bad, can we fix this error message so that the grammar is correct? "Safe _does_ not have enough balance to make prepaid card(s)"

Aierie: Will do!

habdelra:
I have slightly better logic for testing the safe balance is enough, I made the change in the Safes.transferTokens() API, but I forgot to make it in the PRepaidCards.create()/ Specifically we should also test that the safe has enough tokens for the gas as well. Maybe you can fix this? Here is the logic https://github.com/cardstack/cardstack/blob/main/packages/cardpay-sdk/sdk/safes/base.ts#L200-L207. The idea is that after we get the estimate from the server, we also test that the safe can cover the estimated gas charges too.

packages/web-client/app/components/card-pay/issue-prepaid-card-workflow/preview/index.ts
habdelra: Another kind of error that is potentially possible is creating a prepaid card with unapproved tokens--but that would be impossible to do from the UI, so I think it's ok to let that blow up naturally.

You could also get an error if you attempted to sign the txn from a card wallet that had a different private key than the owner of the safe. I'm not too sure how to replicate this in the real world. Perhaps that is also ok to let blow up naturally.

Yeah error codes would not be a bad idea--where we could throw our own type of error from the SDK that has a `code` property` on it that you could test for instead. I'm totally down for this.

Aierie:
Thanks @habdelra , will make the changes you suggested, and think about the error codes!

Aierie:
> I have slightly better logic for testing the safe balance is enough, I made the change in the Safes.transferTokens() API, but I forgot to make it in the PRepaidCards.create(). Specifically we should also test that the safe has enough tokens for the gas as well. Maybe you can fix this? Here is the logic https://github.com/cardstack/cardstack/blob/main/packages/cardpay-sdk/sdk/safes/base.ts#L200-L207. The idea is that after we get the estimate from the server, we also test that the safe can cover the estimated gas charges too.

Will make this change in a separate PR, have made the grammar change.

PR #1817, opened 2021-07-12
Deploy green boxel main branch builds to boxel-preview bucket

lukemelia:


Aierie:
Bulk of current accessibility test failures should be fixed when axe fixes this bug:

https://github.com/dequelabs/axe-core/issues/3082

Aierie:
In the meantime will see if there is a way to turn off the rule

Aierie:
Tracking the bug by subscribing to the GitHub issue, and also in an issue with a due date of 1 week:

https://linear.app/cardstack/issue/CS-1265/turn-aria-allowed-rule-back-on-in-boxel-after-axe-core-and-ember-a11y

PR #1820, opened 2021-07-13
Use strings when referring to token amount in cli

Aierie:
Noticed when testing things out with the cli just now that yargs was converting amount inputs to numbers and token amounts sometimes lost precision. To prevent this, changed the type of `amount` inputs from 'number' to 'string'. Also changed an existing `amount` argument (in `pay-merchant`) that referred to spend to `spendAmount`.


habdelra (changes requested):
the spend amount is actually the amount of USD cents. it is absolutely safe to use this as a number in javascript since it is a decimal "2" value--it would be impossible to max this out in javascript--that would represent more USD that actually exist in circulation. And we are definitely not making prepaid cards that big 😉 . Given that I would like to keep these as numbers in the CLI. We only need to use strings for numbers when we are dealing with token amounts in `wei`. Since that is not the case for SPEND we should be just fine.

Aierie:
> the spend amount is actually the amount of USD cents. it is absolutely safe to use this as a number in javascript since it is a decimal "2" value--it would be impossible to max this out in javascript--that would represent more USD that actually exist in circulation. And we are definitely not making prepaid cards that big 😉 . Given that I would like to keep these as numbers in the CLI. We only need to use strings for numbers when we are dealing with token amounts in `wei`. Since that is not the case for SPEND we should be just fine.

Yep, spend is kept as a number; only token amounts are changed to strings, they were previously getting casted as numbers.

habdelra:
oh my bad! I misunderstood your PR. (the renaming of the SPEND_AMOUNT in the arguments threw me off)

habdelra (approved):

PR #1824, opened 2021-07-13
Updated to support "available to spend" balance for prepaid card

habdelra:
Fixes #CS-819
In this PR we are using a new `faceValue` function for a prepaid card. After a prepaid card is used this returns the most conservative SPEND balance for a prepaid card based on the rate drift (since a rate lock must be supplied when using a prepaid card).


habdelra:
I'm holding off on merging this until I can test the subgraph (my local graph node is busy indexing erc-20 in xdai right now)

PR #1826, opened 2021-07-14
Cardstack hosted rpc

habdelra:
switching back to cardstack hosted RPC. the difference between relay server and app RPC is too severe, we should keep them consistent


lukemelia (approved):
I had troubles getting the websocket endpoints to work last time I tried. Are they working for you?

habdelra:
That why I turned them off, but the mobile app team is going to use these to try and solve the crashing issue. Many people report that using ws fixes the specific error that they are running into.

PR #1833, opened 2021-07-15
Enhancement/incorrect chain handling logic

Aierie:
To follow #1832 . This adds a network correction service that is responsible for determining:
1. Whether we need to block user actions and show a message that the user needs to reload
2. When we should automatically reload


Aierie:
Current behaviour:

Layer 2 has an incorrect chain - Disconnect and show a dismissable modal telling the user why they are disconnected. Open workflows will receive the disconnect event as usual and be canceled. The user is free to reconnect and start workflows again.

Layer 1 has an incorrect chain - Show a non-dismissable modal telling the user that they are on the wrong chain, and they need to switch back to the correct chain; We will reload the page when they disconnect or connect to the right chain. This happens regardless of whether the user is connected to MetaMask or WalletConnect.

PR #1836, opened 2021-07-15
check whether account is null as part of checking for bridging success

Aierie:
Account may be null when polling if it is the user's first time bridging.

Your switching to the subgraph seems to fix CS-1094 (tokens not showing up after deposit in new depot if refetching balances happens too soon) as well, just noticed while working on this. Thanks!


habdelra:
thanks!

PR #1840, opened 2021-07-16
Add more deposit/withdrawal amount validation

backspace:
This adds a mostly-styling `TokenAmountInput` and uses it in the deposit and withdrawal workflows, as well as some validation functions that the consuming components use to determine whether the entered amount is present, has 18 or fewer decimal points, is lower than the balance, etc.

screencast 2021-07-20 14-55-07-submitting

While we’d prefer to show the USD equivalent even when a too-high or too-many-decimals number has been entered, I’ve stopped short of including that here.


lukemelia (approved):

packages/web-client/app/utils/validation.ts
```diff
@@ -0,0 +1,66 @@
+import BN from 'bn.js';
+import { toBN, toWei } from 'web3-utils';
+
+// token input validations that assume string inputs are in wei
```
Aierie, line 4: I just realised this comment is incorrect - it should assume string inputs are in ether.
```suggestion
// token input validations that assume string inputs are in ether
```

packages/web-client/app/components/card-pay/withdrawal-workflow/transaction-amount/index.hbs
```diff
@@ -34,31 +34,15 @@
...
-            {{this.currentTokenSymbol}}
-          </div>
-        </div>
-        </Boxel::Field>
+          @amount={{this.amount}}
+          @tokenSymbol={{this.currentTokenSymbol}}
+          @tokenBalance={{this.currentTokenBalance}}
+          @onInputAmount={{this.onInputAmount}}
+          @invalid={{this.isInvalid}}
+          @errorMessage={{this.errorMessage}}
```
Aierie, line 44: We could use a blur event to trigger validation if a user clicks on this and then leaves.

backspace: Makes sense, done in 13caeed, thanks!

PR #1844, opened 2021-07-19
Update face value options for prepaid cards

lukemelia:
- so that we don't have to give people so much DAI to bridge in beta


lukemelia:
Screenshot 2021-07-19 at 2 13 35 PM

tintinthong (approved):
LGTM

PR #1848, opened 2021-07-19
Add layers to Boxel's modal

Aierie:
## Intention
Prepare for implementation of urgent modals:
image

## Changes
- add two different z-indices to boxel's variables
- use in boxel's modal as a "layer" argument. Intentionally used "layer" instead of "z-index" to make it harder to add new z-indices ad hoc.

## Pics
1 layer:
image

2 layers. I don't like that the background where modal overlays overlap is totally dark, but since this is off the happy path, and we can come back to fix this later, not changing the overlay opacity for now:
image

### DIFF
--- a/packages/boxel/addon/components/boxel/modal/index.hbs
+++ b/packages/boxel/addon/components/boxel/modal/index.hbs
@@ -1,7 +1,11 @@
 {{#if @isOpen}}
   {{set-body-class "has-modal"}}
   {{on-key "Escape" @onClose event="keydown"}}
-
+<div
+  style={{css-var
+    boxel-modal-z-index=(if (eq @layer "urgent" ) "var(--boxel-layer-modal-urgent)" "var(--boxel-layer-modal-default)")
+  }}
+>
   <button
     type="button"
     {{on "click" @onClose}}
@@ -21,4 +25,5 @@
       {{yield}}
     </div>
   </dialog>
+</div>
 {{/if}}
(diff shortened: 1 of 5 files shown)


lukemelia (approved):
It is with a held nose that I approve this PR. 👃 ☁️

packages/boxel/addon/styles/variables.css
```diff
@@ -119,6 +119,10 @@
 
   /* Fix: Other colors used in app, potentially not in color palette: */
 
+  /* z-index layers */
+  --boxel-layer-modal-default: 15;
+  --boxel-layer-modal-urgent: 20;
+
```
backspace, line 125: Should this go below the following colours, so the comment isn’t mixed up?

packages/boxel/addon/components/boxel/modal/index.hbs
```diff
@@ -21,4 +25,5 @@
       {{yield}}
     </div>
   </dialog>
+</div>
```
backspace, line 28: maybe indentation comments are annoying but all of this could be indented 🤓

backspace (approved):
This seems like a sensible way to restrict the interface so it doesn’t become infinite modals 🎉

packages/boxel/addon/styles/variables.css
Aierie: I think keeping the variables visible is better than adding it below the colours to keep consistency, it feels like someone might scroll along and see the "less-used colors" and not realize there are z-indices around.

Aierie: Will move it up. Thanks!

PR #1852, opened 2021-07-19
Extract amount input validation functions

backspace:
This is an extension of #1840 as suggested by @Aierie, as a start toward common validation functions that can be used by other cards.

screencast 2021-07-20 09-01-44-true-function

At least one refinement would be nice: having the USD conversion amount still show even when the number exceeds the balance (or has too many decimals).


backspace:
I’ve added these commits to to #1840 and will continue there 🤓

PR #1854, opened 2021-07-20
Make address arg for getAddressByNetwork type safe

DominicSherman:


habdelra:
This looks good, I need to fix how our subgraph reads this  info though to match your changes. I'll make a commit in this branch for that

PR #1855, opened 2021-07-20
Add card listing to balances tab

backspace:
This adds a listing of prepaid cards to the balances route:

image

Implementation notes:
* there’s a new `CardPay::PrepaidCardSafe` component that wraps `CardPay::PrepaidCard` to render `PrepaidCardSafe`s as returned by the SDK, including a task to resolve the DID and fetch customisation options
* as a result, the indicators on the card about reloadability and transferability now reflect the SDK’s response instead of being hardcoded
* when the account sidebar is added, the `prepaidCards` getter should likely be promoted upward as it will need to be used elsewhere for the card count
* this replaces the mock `POST /prepaid-card-customizations` handler with an implementation that makes more use of Mirage models, relationships, and serialisers; the assertions on what is sent with the `POST` are now accomplished by examining the request via Pretender’s `handledRequests`
* the `viewSafe` ended up being unnecessary because of #1878 but I figured I might as well leave it


packages/web-client/app/css/card-pay/balances.css
```diff
@@ -1,3 +1,49 @@
...
+}
+
+.card-pay-dashboard__balances-count {
+  padding-left: var(--boxel-sp-xxs);
+  color: var(--boxel-cyan);
+}
+
+.card-pay-dashboard__balances-cards {
+  display: grid;
+  gap: var(--boxel-sp); /* FIXME cards seem to have an extra right-side gap? */
```
backspace, line 42: The gap is dependent on the container width; you can see in this GIF that the column and row gaps are the same at the beginning but the column gaps grow as the container gets wider. I couldn’t find a way around this, is it okay?

grid-gap-screencast 2021-07-27 12-39-47

packages/web-client/tests/acceptance/issue-prepaid-card-test.ts
```diff
@@ -14,18 +14,18 @@ import { toWei } from 'web3-utils';
...
-import { Response as MirageResponse } from 'ember-cli-mirage';
 import { timeout } from 'ember-concurrency';
 import { currentNetworkDisplayInfo as c } from '@cardstack/web-client/utils/web3-strategies/network-display-info';
 import { faceValueOptions } from '@cardstack/web-client/components/card-pay/issue-prepaid-card-workflow/workflow-config';
 
 import { MirageTestContext } from 'ember-cli-mirage/test-support';
 
-interface Context extends MirageTestContext {}
+interface Context extends MirageTestContext {
+  normalizedRequestAttrs(): any;
```
Aierie, line 27: I haven't needed this property in local tests, does anything here use normalizedRequestAttrs?

Nice to know that this convenience method exists though!

packages/web-client/app/components/card-pay/prepaid-card-safe/index.ts
```diff
@@ -0,0 +1,31 @@
...
+    if (this.args.safe.customizationDID) {
+      this.fetchTask = taskFor(
+        this.cardCustomization.fetchCardCustomization
+      ).perform(this.args.safe.customizationDID);
+    }
+  }
+
+  get customization() {
+    return this.fetchTask?.value;
+  }
```
Aierie, line 30: Do we want a fallback customization here? Would the first selectable color scheme, with no pattern, work?

Aierie (changes requested):
I think the logic for fetching customizations looks good, added some suggestions for improvements on the layout/styling/design side, and note about an error in the conditional display of "Non-transferable" vs "Transferable". 

Also, wonder if you've noticed this? I'm getting the [grid illusion](https://en.wikipedia.org/wiki/Grid_illusion) from the way we display prepaid cards, which might be distracting.

packages/web-client/app/css/card-pay/balances.css
burieberry: Another option: I like using the `auto-fill` type grid (example: https://boxel.stack.cards/cards-balances-cs-452/#/scenarios/media-registry/bunny_records), where it wraps if the screen gets smaller. Since the cards have specific widths, you do something like: `grid-template-columns: repeat(auto-fill, 22.75rem);` and then restrict the max-width of the container.

packages/web-client/app/components/card-pay/prepaid-card/index.hbs
```diff
@@ -44,8 +44,19 @@
...
+
+        {{#if (eq @reloadable true)}}
+          <div data-test-prepaid-card-reloadable>Reloadable</div>
+        {{else if (eq @reloadable false)}}
+          <div data-test-prepaid-card-non-reloadable>Non-reloadable</div>
+        {{/if}}
+
+        {{#if (eq @transferrable true)}}
+          <div data-test-prepaid-card-transferrable>Transferrable</div>
+        {{else if (eq @reloadable false)}}
```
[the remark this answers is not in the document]
burieberry: Eek, there was a ticket for this on linear at the top of the priority list, so I worked on that yesterday and it's merged.

burieberry (commented):
Looks good overall. If you don't want to deal with css, you can keep the current grid, and I can always go over it afterwards if needed. There's only the conflict regarding how we want ti handle the reloadable/transferrable attributes, based on the PR I merged. The card customization styles weren't reflected on the grid when I ran this, but I suppose it's WIP.

packages/web-client/app/components/card-pay/prepaid-card/index.hbs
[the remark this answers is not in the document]
backspace: TRUE, thank you, though this is now obsolete

Aierie:
Since the error is obsolete, I think this is fine to merge as is

packages/web-client/tests/acceptance/issue-prepaid-card-test.ts
backspace: ah that’s an autofix that I neglected to remove after moving the handling into `mirage/config.js`, thanks!

Aierie (approved):

packages/web-client/app/components/card-pay/prepaid-card/index.hbs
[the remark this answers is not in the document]
backspace: re the ticket, I changed its parent to this one, but I failed to assign it to myself… sorry for the duplicated work!

packages/web-client/app/css/card-pay/balances.css
backspace: thanks for the auto-fill suggestion. I have sadness about how it makes the right edge not align with the button, but I’m just going to leave it for now, if we get more guidance about how to handle responsiveness we can return to it.

autofill-screencast 2021-07-29 09-40-39

backspace:
> …  The card customization styles weren't reflected on the grid when I ran this, but I suppose it's WIP.

They should be rendering! Unless they’re older and don’t reflect the corrected JSON:API structure. We could pair to look closer at this if you like, @burieberry?

lukemelia:
> They should be rendering! Unless they’re older and don’t reflect the corrected JSON:API structure. We could pair to look closer at this if you like, @burieberry?

JSON files are written to S3 by our worker process, so if you are only running the hub server locally, and not the worker as well, it could result in this. Running the worker should process all the queued jobs when it starts up.

PR #1860, opened 2021-07-21
Add underscore-prefix exclusion on unused vars

backspace:
This saves us having to manually disable linting for
no-unused-vars when following the convention that variables
starting with an underscore are unused.


Aierie (approved):
This is convenient, thanks!

lukemelia (approved):
🙌

PR #1864, opened 2021-07-21
Update withdrawal source card

burieberry:
- Updated UI (cs-1271)
- Added footnote about gas fee for claiming tokens (cs-1235)

Script:
script-1
script-2

UI:
edit
view


lukemelia (approved):
Looks good. As a heads up, we will in the future need to list a user's merchant safe (if they have one) as a possible withdrawal source. We can deal with that when we get there, but it may have API implications on the components extracted here.

burieberry:
> Looks good. As a heads up, we will in the future need to list a user's merchant safe (if they have one) as a possible withdrawal source. We can deal with that when we get there, but it may have API implications on the components extracted here.

Ok, that explains why there's a selection under "merchant" in the script outline. Good to know, we can add that when we come to that 👍

PR #1865, opened 2021-07-22
Internal/package management tweaks

Aierie:
Remove Boxel's release script + update docs, and remove an obsolete (as of v3) lerna option.


lukemelia (approved):
👍

lukemelia:
One thing to separately consider is our changelog management approach for the monorepo.

PR #1867, opened 2021-07-22
Enhancement/chain change modal UI improvements

Aierie:
CS-1344

Make the currently very basic chain change modal look more like this:

image

And use a less demanding? tone for copywriting.

## Layer 1 pic
image

## Layer 2 pic
image

### DIFF
--- a/packages/web-client/app/components/card-pay/chain-change-modal/index.css
+++ b/packages/web-client/app/components/card-pay/chain-change-modal/index.css
@@ -40,7 +40,6 @@
   outline-style: solid;
 }
 
-
 .chain-change-modal__card-title {
   font: 700 1.125rem/calc(24 / 18) var(--boxel-font-family);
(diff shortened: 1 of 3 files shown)


backspace (approved):
I tried this out locally, looks good!

burieberry (approved):
@Aierie So you're doing the network messages separately (last 2 screenshots) instead of one demanding message (first screenshot) right? I'm good with that change. My only very minor recommendation for the button (non-blocking): `Disconnect and Reload` since we write the CTA text in title case

Aierie:
> @Aierie So you're doing the network messages separately (last 2 screenshots) instead of one demanding message (first screenshot) right? I'm good with that change. My only very minor recommendation for the button (non-blocking): `Disconnect and Reload` since we write the CTA text in title case

Ah, the demanding text was more on my initial placeholder copy. Will modify the casing.

PR #1869, opened 2021-07-22
add color style declaration to workflow-thread component

Aierie:
Adds the color style declaration to the workflow thread component so we don't have to specify it for each workflow we create. Note: The bug in CS-1327 did not happen in Chrome because Chrome's user agent stylesheet was overriding our text color via their custom property `-internal-light-dark`.

EDIT: moved the color declaration to the Boxel component so that the color declaration is where the background declaration is.


lukemelia (approved):
A nice straightforward change now that we have boxel in the monorepo!

PR #1870, opened 2021-07-22
Pass the customizationDid to SDK prepaid card creation

lukemelia:
I don't understand why Typescript wasn't complaining about this.


backspace (approved):
nice, I didn’t realise it would be this easy! 🎉

burieberry (approved):

PR #1871, opened 2021-07-22
make dashboard component more flexible

Aierie:
- handle case for dashboards with too little content (when height of content does not fill parent container, body named block can fill all available height by setting height:100%)
- also remove the dashboard body container overflow


Aierie:
Case we are aiming to avoid:

image

PR #1881, opened 2021-07-24
adding mainnet addresses

habdelra:
Hold off until we get CPXD addresses


habdelra:
we now have CPXD addresses 🎉

PR #1883, opened 2021-07-26
Make pattern URLs have a FQDN

lukemelia:


lukemelia:
I've updated these in the staging and production databases. Anyone who has already run seeds locally should update their local database @backspace @Aierie @burieberry

lukemelia:
@burieberry can you confirm that you saw this before I merge?

burieberry:
> @burieberry can you confirm that you saw this before I merge?

Thanks, looks good

PR #1884, opened 2021-07-26
Set up web-client with sentry for staging and production

lukemelia:
- [x] Add breadcrumbs for milestone completions

### DIFF
--- a/packages/web-client/app/routes/boom.ts
+++ b/packages/web-client/app/routes/boom.ts
@@ -1,0 +1,11 @@
+import Controller from '@ember/controller';
+import Route from '@ember/routing/route';
+
+export default class BoomRoute extends Route {
+  setupController(controller: Controller, model: any, transition: any) {
+    super.setupController(controller, model, transition);
+
+    //@ts-ignore intentional error
+    raiseAnUnhandledExceptionOnPurpose();
+  }
+}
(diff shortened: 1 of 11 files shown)


packages/web-client/app/routes/boom.ts
```diff
@@ -0,0 +1,11 @@
...
+import Route from '@ember/routing/route';
+
+export default class BoomRoute extends Route {
+  setupController(controller: Controller, model: any, transition: any) {
+    super.setupController(controller, model, transition);
+
+    //@ts-ignore intentional error
+    raiseAnUnhandledExceptionOnPurpose();
+  }
+}
```
backspace, line 11: Is this meant to be a permanent inclusion? I thought maybe it would be just for testing to ensure Sentry received something.

backspace (approved):
I’m not sure about the permanent inclusion of the exception-triggering route, but otherwise this looks like a good starting configuration 🎉

packages/web-client/app/routes/boom.ts
lukemelia: Its purpose is to verify Sentry functionality. I want to verify it on staging and prod as well. I'm inclined to leave it there after that to make verifying any sentry config changes easy but am open to other views.

PR #1885, opened 2021-07-26
web: Update prepaid issuance to use new SDK response

backspace:
This fixes the issue prepaid card workflow for the SDK’s
updated PrepaidCard.create interface, which as of #1878
returns a PrepaidCardSafe instead of just the address
of the prepaid card. This is something I requested for
#1855 and it’ll be used more thoroughly there, but this
is a minimal interim change to handle the new return type.

I’ve used this locally to issue a new prepaid card:

image


lukemelia (approved):
💯

PR #1887, opened 2021-07-26
Add page titles for each route

Aierie:
Page title defaults to Cardstack.

Card Pay page titles follow the tabs, omitting "Cardstack" from the title. These seem pretty long already!
- Card Balances · Card Pay
- Token Suppliers · Card Pay
- Merchant Services · Card Pay

### DIFF
--- a/packages/web-client/app/templates/card-pay/balances.hbs
+++ b/packages/web-client/app/templates/card-pay/balances.hbs
@@ -1,3 +1,5 @@
+{{page-title "Card Balances"}}
+
 <section class="card-pay-dashboard__section" role="tabpanel" aria-labelledby="card-pay.balances">
   <CardPay::DashboardPanel @panel={{@model.panel}}>
     {{#each @model.panel.sections as |section|}}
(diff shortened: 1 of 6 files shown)


packages/web-client/app/templates/card-pay/balances.hbs
```diff
@@ -1,3 +1,5 @@
+{{page-title "Card Balances" prepend=true}}
```
backspace, line 1: according to [the documentation](https://github.com/ember-cli/ember-page-title#page-title-helper), `prepend` has this as a default, so might as well remove it here and elsewhere?

backspace (changes requested):
This seems like a good start, though I wonder whether it might be possible to add more to the title when the modals are open… maybe a future improvement, though. Other than the redundant `prepend`, 👍🏻🚀

lukemelia (approved):
I don't mind the explicit prepend.

packages/web-client/app/templates/card-pay/balances.hbs
Aierie: Done!

PR #1888, opened 2021-07-26
web: Add balance refresh after prepaid issuance

backspace:
Local screenshots from before and after issuance with this change:

before-@cardstack:web-client 2021-07-26 17-00-39

after-@cardstack:web-client 2021-07-26 17-06-17

This brings the prepaid card issuance workflow in line with the others, though I wonder whether it would be preferable to “freeze” the balance shown in the opening card with the value before the workflow has completed, as it seems strange that it changes when the operation has been completed. But that’s a UX concern outside the scope of this fix, as certainly one would expect the balance shown after the confirmation to be the new lower number.


lukemelia (approved):
🚢

PR #1893, opened 2021-07-27
Update deposit source card and cover no-balance edge-case

burieberry:
- Disable radio option if balance is 0 or undefined
- If neither option is selectable, disable CTA and display validation message (The wording for the message might change)
- CSS for disabled radio options
- Preselect DAI option if user has DAI balance
- Preselect CARD option if user has no DAI but has CARD balance


lukemelia (approved):
:clap: for the integration test!

PR #1895, opened 2021-07-28
Fix/network aware messages

Aierie:
This PR fixes CS-1107. We delay rendering of workflows until both layer 1 and layer 2 have either completed reconnection, or identified that they don't need to reconnect, ensuring that the correct messages are displayed each time.

This does raise design questions, as discussed with @lukemelia yesterday - what should we show during while the networks are reconnecting? I've left this to address in a separate PR.

### DIFF
--- a/packages/web-client/app/utils/web3-strategies/layer2-chain.ts
+++ b/packages/web-client/app/utils/web3-strategies/layer2-chain.ts
@@ -117,7 +117,8 @@ export default abstract class Layer2ChainWeb3Strategy
         this.#exchangeRateApi = await getSDK('ExchangeRate', this.web3);
         this.#safesApi = await getSDK('Safes', this.web3);
         this.#hubAuthApi = await getSDK('HubAuth', this.web3, config.hubURL);
-        this.updateWalletInfo(accounts, this.chainId);
+        await this.updateWalletInfo(accounts, this.chainId);
+        this.isInitializing = false;
       } catch (e) {
         console.error(
           'Error initializing layer 2 wallet and services. Wallet may be connected to an unsupported chain'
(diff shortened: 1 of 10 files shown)


packages/web-client/app/utils/web3-strategies/layer2-chain.ts
```diff
@@ -102,16 +104,13 @@ export default abstract class Layer2ChainWeb3Strategy
         console.error('Error in display_uri callback', err);
         return;
       }
+      // if we get here when a user loads a page, then it means that the user did not have
+      // a connection from local storage. We can safely say they are initialized
+      this.isInitializing = false;
```
Aierie, line 109: Because `provider.enable` doesn't resolve until a user connects their wallet, `this.isInitializing = false` under `provider.enable` will never run if the user is not connected, and they will never see the workflow. This catches the case where a user does not need to reconnect.

lukemelia (approved):

packages/web-client/app/utils/web3-strategies/layer2-chain.ts
```diff
@@ -144,7 +143,16 @@ export default abstract class Layer2ChainWeb3Strategy
...
+
+    yield this.provider.enable();
+
+    // we only get here if the user has an account
+    // we wait for the logic in accountsChanged to run and fetch the user's depots
+    // before considering initialization done
+    // the connector is from WalletConnect and knows of accounts before the
+    // the accountsChanged event is triggered
+    yield this.waitForAccountDeferred.promise;
+    this.isInitializing = false;
```
Aierie, line 155: Just realized we can actually move this into the `accountsChanged` block since we have a different way of setting `isInitializing` for the non-reconnecting case. Will move this in, so the code is more straightforward.

PR #1896, opened 2021-07-28
increase txn timeout to be more suitable for mainnet realities

habdelra:
mainnet txns that run normal to fast in terms of gas price can take upwards of 10 minutes

### DIFF
--- a/packages/cardpay-sdk/sdk/utils/general-utils.ts
+++ b/packages/cardpay-sdk/sdk/utils/general-utils.ts
@@ -17,7 +17,7 @@ export async function networkName(web3: Web3): Promise<string> {
 export function waitUntilTransactionMined(
   web3: Web3,
   txnHash: string,
-  duration = 60 * 5 * 1000
+  duration = 60 * 10 * 1000
 ): Promise<TransactionReceipt> {
   let endTime = Number(new Date()) + duration;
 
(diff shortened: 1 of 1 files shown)


packages/cardpay-sdk/sdk/utils/general-utils.ts
```diff
@@ -31,7 +31,7 @@ export function waitUntilTransactionMined(
       if (receipt) {
         resolve(receipt);
       } else if (Number(new Date()) > endTime) {
-        throw new Error(`Transaction took too long to complete, waited ${duration / 1000} seconds`);
+        throw new Error(`Transaction ${txnHash} took too long to complete, waited ${duration / 1000} seconds`);
```
Aierie, line 34: Adding the transaction hash here will break our string matching of error messages in the prepaid card preview card:

https://github.com/cardstack/cardstack/blob/67e6bf31d2b6d3911192ac7ae559cc99ddba1b50/packages/web-client/app/components/card-pay/issue-prepaid-card-workflow/preview/index.ts#L75-L77

Maybe we can prepend this error with `TRANSACTION_TIMEOUT:` or something like that? If so, I can update the card to match that instead.

habdelra: I'll just add it to the end

habdelra: but yeah, i love your thinking. I think the error codes idea that you had is a better fit.

Aierie: Thanks! 😄

PR #1897, opened 2021-07-28
Separate authentication step in Prepaid Card Issuance flow

lukemelia:
https://www.loom.com/share/cc669b252b704c4f8474133dac12076e
Note: If you are already hub-authenticate the Auth mini-card and the message the precedes it are not included in the workflow.

### DIFF
--- a/packages/web-client/app/components/card-pay/issue-prepaid-card-workflow/authentication/index.hbs
+++ b/packages/web-client/app/components/card-pay/issue-prepaid-card-workflow/authentication/index.hbs
@@ -3,6 +3,7 @@
 >
   <Boxel::ActionChin
     @state={{this.authState}}
+    @disabled={{@frozen}}
   >
     <:default as |d|>
       {{#if (and this.error this.authenticationTask.isIdle)}}
--- a/packages/web-client/app/components/card-pay/issue-prepaid-card-workflow/authentication/index.hbs
+++ b/packages/web-client/app/components/card-pay/issue-prepaid-card-workflow/authentication/index.hbs
@@ -1,0 +1,53 @@
+<ActionCardContainer
+  @isComplete={{@isComplete}}
+>
+  <Boxel::ActionChin
+    @state={{this.authState}}
+  >
+    <:default as |d|>
+      {{#if (and this.error this.authenticationTask.isIdle)}}
+        <d.ActionStatusArea
+          @icon="card-wallet-app-icon"
+          style={{css-var status-icon-size="2.5rem"}}
+          data-test-failed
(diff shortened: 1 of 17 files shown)


packages/web-client/app/components/card-pay/issue-prepaid-card-workflow/authentication/index.hbs
```diff
@@ -0,0 +1,53 @@
+<ActionCardContainer
+  @isComplete={{@isComplete}}
+>
+  <Boxel::ActionChin
+    @state={{this.authState}}
+  >
+    <:default as |d|>
+      {{#if (and this.error this.authenticationTask.isIdle)}}
```
Aierie, line 8: I think we should handle `@frozen` layer 2 not being connected.

Aierie (changes requested):
Overall this looks good! The happy path works nicely for me, and I think the UI is nice. I received the prompt to auth the first time I ran this, and after that it didn't appear again. I was able to successfully create a prepaid card.

Summarizing what I understand for this PR, just in case I missed/misunderstood something:
- Responsibility for authenticating with the hub is moved earlier in the workflow, so users don't get two consecutive requests when they try to create a prepaid card. We no longer need to check that the user is authenticated when they click the button to create a prepaid card.
- We don't want to show the authentication message and card if the user is already authenticated. For this, we added a network-aware card, and added the ability for network-aware cards and messages to detect if the user is authenticated via the hub auth service.
- The is-network-initialized helper detects if the hub auth service is initializing in order to prevent the race condition we saw with layer 1 and layer 2.
- To ensure that the user doesn't go on with the workflow with an invalid auth token, we check upon initialization of the hub auth service that the user's auth token is valid. 

There are two cases (actually maybe just one, since disconnection also causes `@frozen` to be true) where the authentication card should be disabled, I've mentioned them in the comment below.

packages/web-client/app/components/card-pay/issue-prepaid-card-workflow/authentication/index.hbs
lukemelia: I actually removed @frozen because the memorialized state does not have any interactive elements. Do you think we still need it?

Aierie: Yep, we need it for the default state (user performs action that cancels workflow prior to authenticating, eg. disconnecting layer 2).

burieberry (approved):
Works great!

Aierie (approved):

PR #1900, opened 2021-07-28
Make reloadable and transferrable card display properties dynamic

burieberry:


lukemelia (approved):

Aierie:
@backspace , flagging that there is some overlap on the prepaid card UI front between this and #1855.

PR #1901, opened 2021-07-29
User rejection identification

Aierie:
CS-1319

- Adds a utility that checks if an error's message matches the user rejection error message.
- Uses the utility in the prepaid card preview card (to be used in the withdrawal flow as well)

This code is dependent on CardWallet returning the correct error message. As such, we will not be able to use this for other wallets connected via WalletConnect on Layer 1.


burieberry (approved):

backspace (approved):
I was able to exercise this successfully 🎉

image

PR #1904, opened 2021-07-29
set web client and boxel test ports to different values

Aierie:
This is so that running `yarn test` in the monorepo root will not fail because Boxel and Web Client both want to use port 7357. We could in addition (or as an alternative?) run tests in series, not parallel. I've found that running 2 ember test instances simultaneously can be very taxing on my laptop.

### DIFF
--- a/packages/boxel/.ember-cli.js
+++ b/packages/boxel/.ember-cli.js
@@ -1,7 +1,7 @@
 /* eslint-disable */
 'use strict';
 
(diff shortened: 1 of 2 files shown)


lukemelia (approved):
You're changes look fine. Not sure what's up with the tests.

Aierie:
> Your changes look fine. Not sure what's up with the tests.

It looks like accessibility tests are failing again, maybe ember release is using a different version. Will fix those tomorrow.

PR #1906, opened 2021-07-29
web: Add wallet navigation sidebar

backspace:
This is on hold because it’s almost all placeholders for now, but it shouldn’t be hard to bring back when it’s more useful:

image

Some styling updates are still needed etc


backspace:
I’m going to close this, it can be retrieved if it’s useful if/when this becomes a priority.

PR #1908, opened 2021-07-30
Enhancement/prepaid card preview UI

Aierie:
CS-1413

Updated prepaid card creation button text to read Create (previously Continue), and Retry after unsuccessful tries. Default error text is also updated.

Updated default state of the prepaid card creation card

Updated error state of the prepaid card creation card


lukemelia (approved):
:shipit:

PR #1910, opened 2021-07-30
web: Add prepaid listing refresh after creation

backspace:
To reproduce the SDK behaviour where `Safe.viewSafes` will return a new `PrepaidCard.create`,
this makes the test strategy store the newly-created `PrepaidCardSafe` in the same map that
`test__simulateAccountSafes` uses.

Here it is in action:

screencast 2021-07-30 10-26-45

### DIFF
--- a/packages/web-client/app/services/layer2-network.ts
+++ b/packages/web-client/app/services/layer2-network.ts
@@ -114,6 +114,9 @@ export default class Layer2Network
       customizationDid,
       options
     );
+
+    taskFor(this.viewSafes).perform(this.walletInfo.firstAddress!);
+
     return address;
   }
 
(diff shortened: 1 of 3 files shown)


lukemelia (approved):
✨

packages/web-client/app/services/layer2-network.ts
```diff
@@ -114,6 +114,9 @@ export default class Layer2Network
       customizationDid,
       options
     );
+
+    taskFor(this.viewSafes).perform(this.walletInfo.firstAddress!);
```
Aierie, line 118: Took me a while to understand that this was necessary because a consumer was reading the last successful value of this task, do you think a comment about this would be good?

lukemelia: Would it be clearer if we wrapped it in a "refreshSafes()` method?

lukemelia: Merging as is, but let's do either a comment or intention-revealing method in a separate PR, whichever you like best.

backspace: ya, I don’t love the way this works at the moment; the other place this is invoked is a `did-insert` in `balances.hbs`. Both of them pass the `firstAddress` like this. So `refreshSafes` would be a nice encapsulation, I’ll do that, thanks.

PR #1911, opened 2021-07-30
Update to token bridging to layer1

burieberry:
When withdrawing tokens, I was stuck in the roach motel. The funds were taken from my depot, but I didn't receive them in my Kovan wallet after the claiming step. In Etherscan, my Sokol wallet appeared as the `receiverAddress` instead of my Kovan wallet. When I made the changes in this PR, though, now everything works as expected. I changed the `receiverAddress` to be the Layer1 wallet address instead of Layer2. Does this look right? The existing withdrawal and claiming work for Luke, so ¯\_(ツ)_/¯

### DIFF
--- a/packages/web-client/app/components/card-pay/withdrawal-workflow/transaction-amount/index.ts
+++ b/packages/web-client/app/components/card-pay/withdrawal-workflow/transaction-amount/index.ts
@@ -1,6 +1,7 @@
 import { action } from '@ember/object';
 import Component from '@glimmer/component';
 import { tracked } from '@glimmer/tracking';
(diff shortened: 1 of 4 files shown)


habdelra (approved):
Yes that's right, the receiver address should be the layer 1 address that will receive the layer 1 tokens

habdelra:
it may be the case that luke was testing with the same wallet address in both layer 1 and layer 2

lukemelia:
>it may be the case that @lukemelia was testing with the same wallet address in both layer 1 and layer 2

Confirm!

lukemelia (approved):
⚡⚡⚡⚡⚡-> 🐛

PR #1912, opened 2021-07-31
web: Add deposit transaction-viewing URL when available

backspace:
Here it is with manual pauses, as it can otherwise be too fast to catch:

screencast 2021-08-03 13-06-18

Code to produce a pause for manual testing:

image


lukemelia:
Looks good. Can we do this for the unlock step as well? (Separate PR is fine)

Aierie:
Yep, this looks good to me. I think we won't need a design for 2 different block explorer urls, as the Etherscan link is for layer 1 and Blockscout is for layer 2 and we've only needed one at a time to provide a way to check on a transaction's status.

Aierie:
Also added CS-1461 which describes a similar task for the withdrawal flow.

PR #1915, opened 2021-08-03
add error message to token bridge/transaction status card

Aierie:
CS-896

This error message is to be shown when minting of tokens fails, would be good to have specific UI for a failed progress step, but we can probably leave that for later. If this is ok, will do the same for the withdrawal flow.

image

### DIFF
--- a/packages/web-client/app/components/card-pay/deposit-workflow/transaction-status/index.hbs
+++ b/packages/web-client/app/components/card-pay/deposit-workflow/transaction-status/index.hbs
@@ -36,7 +36,7 @@
             <Boxel::Button @as="anchor" @size="extra-small" @href={{this.blockscoutUrl}} target="_blank" rel="noopener" data-test-blockscout-button>
               View on Blockscout
             </Boxel::Button>
-          {{else if this.errored }}
+          {{else if this.errored}}
             Failed
           {{else}}
             Processing...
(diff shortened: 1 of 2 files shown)


packages/web-client/app/components/card-pay/deposit-workflow/transaction-status/index.hbs
```diff
@@ -36,11 +36,18 @@
             <Boxel::Button @as="anchor" @size="extra-small" @href={{this.blockscoutUrl}} target="_blank" rel="noopener" data-test-blockscout-button>
               View on Blockscout
             </Boxel::Button>
+          {{else if this.errored }}
```
backspace, line 39:
```suggestion
          {{else if this.errored}}
```

backspace (approved):

packages/web-client/app/components/card-pay/deposit-workflow/transaction-status/index.hbs
```diff
@@ -36,11 +36,18 @@
             <Boxel::Button @as="anchor" @size="extra-small" @href={{this.blockscoutUrl}} target="_blank" rel="noopener" data-test-blockscout-button>
               View on Blockscout
             </Boxel::Button>
+          {{else if this.errored}}
```
jurgenwerk, line 39: Just for consistency's sake I'd prefer if this was `{{else if this.error}}` (even if the props of the actual error object aren't being accessed, like [here for example](https://github.com/cardstack/cardstack/blob/main/packages/web-client/app/components/card-pay/issue-prepaid-card-workflow/authentication/index.hbs#L9))

jurgenwerk (approved):

packages/web-client/app/components/card-pay/deposit-workflow/transaction-status/index.hbs
jurgenwerk: Oops, didn't realize this is merged already 😅

Aierie: All good, can change this when I PR a similar change for the withdrawal workflow. Thanks for pointing this out.

PR #1917, opened 2021-08-03
Make prepaid card issuance workflow confirmation card retries fetching card customization data

lukemelia:
- there may be a delay in our background workers persisting this


backspace (approved):
There’s no design guidance but a possible future improvement would be to display a ⚠️ or the like if the fetch fails even after all the retries, with a tooltip/title displaying details of the error.

PR #1919, opened 2021-08-03
Add an "initializing" state to header's layer 1 and 2 connection buttons

lukemelia:
https://user-images.githubusercontent.com/353/128046974-362b5add-9490-4bbe-a434-177c52da76c9.mov


Aierie (approved):
Works nicely for me!

PR #1927, opened 2021-08-04
web: Add prepaid card issuer name truncation

backspace:
This also adds a `title` attribute to the element so the full name can still be seen:

screencast 2021-08-04 12-36-26


lukemelia (approved):
👍

Aierie (approved):
This works well. I think the title attribute might have some issues on mobile, so we may need to revisit that in the future.

PR #1928, opened 2021-08-04
[CS-1508] - Add firebase push function

danibonilha:
Adds the push firebase function to the repo so it can be tracked, this is the current deployed version. I tweaked as minimum as possible, just enough to make work in the new environment (Node 14) that being said, there are lots of improvements to make in this code ('ll comment out some of them), but just wanted to make sure we have a starting point.

Ps: Deploy action will be created in a different PR


packages/firebase-functions/push/functions/package.json
```diff
@@ -0,0 +1,42 @@
...
+    "@types/cors": "^2.8.6",
+    "@types/hapi__joi": "^16.0.4",
+    "@types/node": "^16.4.10",
+    "concurrently": "^5.0.0",
+    "eslint": "^5.6.0",
+    "eslint-config-standard": "^12.0.0",
+    "eslint-plugin-import": "^2.14.0",
+    "eslint-plugin-node": "^7.0.1",
+    "eslint-plugin-promise": "^4.0.1",
+    "eslint-plugin-standard": "^4.0.0",
```
danibonilha, line 27: Not sure if we need those deps

habdelra (approved):

PR #1931, opened 2021-08-04
web: Add merchant creation workflow skeleton

backspace:
This includes some components from prepaid card issuance, I’ve updated the
relevant issues to highlight that.

### DIFF
--- a/packages/web-client/app/components/card-pay/create-merchant-workflow/index.ts
+++ b/packages/web-client/app/components/card-pay/create-merchant-workflow/index.ts
@@ -1,0 +1,147 @@
+import Component from '@glimmer/component';
+import { getOwner } from '@ember/application';
+import { inject as service } from '@ember/service';
+import { WorkflowMessage } from '@cardstack/web-client/models/workflow/workflow-message';
+import { Workflow, cardbot } from '@cardstack/web-client/models/workflow';
+import { Milestone } from '@cardstack/web-client/models/workflow/milestone';
+import { WorkflowCard } from '@cardstack/web-client/models/workflow/workflow-card';
+import PostableCollection from '@cardstack/web-client/models/workflow/postable-collection';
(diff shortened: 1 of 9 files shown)


packages/web-client/app/components/card-pay/create-merchant-workflow/index.ts
```diff
@@ -0,0 +1,147 @@
...
+          message: `Hello, nice to see you!`,
+        }),
+        new WorkflowMessage({
+          author: cardbot,
+          message: `To receive payment through Card Pay, you need to create a merchant account.
+          All you need is to choose a name for your merchant account and sign a transaction to create your merchant on the ${c.layer2.fullName}.`,
+        }),
+        new NetworkAwareWorkflowMessage({
+          author: cardbot,
+          message: `Looks like you've already connected your ${c.layer2.fullName} wallet, which you can see below.
```
lukemelia, line 37:
```suggestion
          message: `Looks like you’ve already connected your ${c.layer2.fullName} wallet, which you can see below.
```

lukemelia (approved):
🙆

Aierie (approved):
Looks good to me! 

More of a design comment - it seems odd that the message and header introduces this step as "giving a merchant a name", when we giving it more (color, id, assigning its manager) + creating it.

image

packages/web-client/app/components/card-pay/create-merchant-workflow/index.ts
backspace: agreed, thanks. This came from a copypaste and I see it in a couple of other places, I’ll open another PR for that.

backspace:
> More of a design comment - it seems odd that the message and header introduces this step as "giving a merchant a name", when we giving it more (color, id, assigning its manager) + creating it.

I agree. I’ll make a subissue about this.

PR #1933, opened 2021-08-04
Display block confirm count feedback in L1 to L2 bridging

burieberry:
CS-1487

### DIFF
--- a/packages/web-client/app/components/card-pay/deposit-workflow/transaction-status/index.ts
+++ b/packages/web-client/app/components/card-pay/deposit-workflow/transaction-status/index.ts
@@ -9,14 +9,23 @@ import Layer2Network from '@cardstack/web-client/services/layer2-network';
 import { TokenSymbol } from '@cardstack/web-client/utils/token';
 import { WorkflowCardComponentArgs } from '@cardstack/web-client/models/workflow/workflow-card';
 import { currentNetworkDisplayInfo as c } from '@cardstack/web-client/utils/web3-strategies/network-display-info';
+import { TransactionReceipt } from 'web3-core';
+import { task } from 'ember-concurrency-decorators';
+import { taskFor } from 'ember-concurrency-ts';
+import { TaskGenerator } from 'ember-concurrency';
 
 class CardPayDepositWorkflowTransactionStatusComponent extends Component<WorkflowCardComponentArgs> {
+  totalBlockCount = 12;
   @service declare layer1Network: Layer1Network;
   @service declare layer2Network: Layer2Network;
   @reads('args.workflowSession.state.depositSourceToken')
   declare selectedTokenSymbol: TokenSymbol;
-  @tracked completedCount = 1;
-  @tracked error = false;
+  @reads('args.workflowSession.state.relayTokensTxnReceipt')
+  declare relayTxnReceipt: TransactionReceipt;
+  @tracked completedStepCount = 1;
+  @tracked bridgeError = false;
+  @tracked blockCountError = false;
+  @tracked blockCount = 0;
 
   get layer2BlockHeightBeforeBridging(): BN | undefined {
     return this.args.workflowSession.state.layer2BlockHeightBeforeBridging;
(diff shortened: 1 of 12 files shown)


packages/web-client/app/components/card-pay/deposit-workflow/transaction-status/index.ts
```diff
@@ -49,10 +58,25 @@ class CardPayDepositWorkflowTransactionStatusComponent extends Component<Workflo
...
-      this.completedCount = 3;
+      this.completedCount = 3; // steps 2 and 3 complete together
+      this.currentStep = this.completedCount;
+      this.waitForBlockConfirmations(transactionReceipt);
+    } catch (e) {
+      this.bridgeError = true;
+    }
+  }
+
+  async waitForBlockConfirmations(transactionReceipt: TransactionReceipt) {
```
lukemelia, line 69: I know this isn't ready for review yet, but I just wanted to encourage you to convert this method to an ember-concurrency task, in order to handle component teardown well.

burieberry:
There's still more work to do here in terms of tests, UI, and error-handling, but let me know if the block confirmations part is correct @habdelra. I updated my PR to follow your comment on linear (https://linear.app/cardstack/issue/CS-1487#comment-4cfb77c5). Not sure if the "waiting for transaction to be mined" status is necessary, since that seems to be resolved in the previous card.

habdelra:
yes this approach looks correct, 👍 . yeah I agree with the comment about the "waiting for transaction to be mined", i don't thin kit's super important to have that given the block countdown--it was just something to entertain the user while they wait.

packages/web-client/app/components/card-pay/deposit-workflow/transaction-status/index.ts
```diff
@@ -8,14 +8,23 @@ import { reads } from 'macro-decorators';
 import { TokenSymbol } from '@cardstack/web-client/utils/token';
 import { WorkflowCardComponentArgs } from '@cardstack/web-client/models/workflow/workflow-card';
 import { currentNetworkDisplayInfo as c } from '@cardstack/web-client/utils/web3-strategies/network-display-info';
+import { TransactionReceipt } from 'web3-core';
+import { task } from 'ember-concurrency-decorators';
+import { taskFor } from 'ember-concurrency-ts';
+import { TaskGenerator } from 'ember-concurrency';
 
 class CardPayDepositWorkflowTransactionStatusComponent extends Component<WorkflowCardComponentArgs> {
+  totalBlockCount = 12;
```
lukemelia, line 17: Perhaps this value should come from the SDK @habdelra @burieberry?

packages/web-client/app/components/card-pay/deposit-workflow/transaction-status/index.ts
```diff
@@ -36,6 +45,37 @@ class CardPayDepositWorkflowTransactionStatusComponent extends Component<Workflo
...
+        this.blockCountError = true;
+        console.error(e);
+      });
+  }
+
+  get displaySubstate() {
+    return taskFor(this.waitForBlockConfirmationsTask).isRunning;
+  }
+
+  get substate() {
```
lukemelia, line 60: `substate` is pretty generic. This might be clearer if it was named `progressSubstate` or `bridgingSubstate`. Feel free to ignore if you disagree.

lukemelia (approved):
Left a couple of suggestions.

packages/web-client/app/components/card-pay/deposit-workflow/transaction-status/index.ts
habdelra: good point. let's add that as a new constant in the SDK for kovan and mainnet

burieberry: > good point. let's add that as a new constant in the SDK for kovan and mainnet (it's a layer 1 artifact)

So I can add it to the `constants.ts` file in the SDK then. Did we pick this number 12 or is it a common thing?

Also, should I be adding the `waitUntilBlock` export to the sdk's `index.ts` with the other exports? Maybe something like `export * from './sdk/utils/general-utils';`? I noticed there's a `waitUntilTransactionMined` fn that we're not currently using in `web-client` but that we could be using.

habdelra: The number 12 came from here https://docs.tokenbridge.net/eth-xdai-amb-bridge/about-the-eth-xdai-amb.

And yes, you should add the `waitUntilBlock` to the exports in the main `index.ts` file. I think we should actually be explicit about which functions we want to export instead of using a `*`, since this is basically our API and we should consciously be deciding which things we want to expose from it. so if you'd like to use the `waitUntilTransactionMined` from the SDK, then let's add that as an explicit export too.

burieberry:
@lukemelia @habdelra Rebased and SDK changes are made. Please advise (regarding new SDK version release) or merge. How to release? Who needs to know? etc...

habdelra:
Looks good to me. I say merge it.

in terms of pushing to npm, i always run this command:
```
npx lerna publish --force-publish="*" --exact
```

Although, I have a bunch of stuff I'd like to get into the next SDK publish as well for the updated contract addresses. 
Generally in terms of notification I always send a message to Kiel in the wallet-internal channel that a new SDK version is available, because otherwise they don't have a way of knowing that they need to incorporate it into the react native app.

habdelra:
I merged the PR that I needed to get into the next SDK version. feel free to merge this and then push the SDK version to npm 👍

PR #1934, opened 2021-08-04
[CS-1511]: Add check-version function

danibonilha:
Adds checkVersion firebase function 

image

image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -90,3 +90,16 @@ packages/cli/node-tests/sample-cards/*/node_modules/**/*
 /packages/**/node_modules/@cardstack/**/*.js
 /packages/**/node_modules/@cardstack/**/*.js.map
 /packages/**/node_modules/@cardstack/**/*.d.ts
+
+# firebase functions 
+/packages/firebase-functions/**/node_modules
+/packages/firebase-functions/**/build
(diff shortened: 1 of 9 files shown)


lukemelia (commented):
Overall, looks good. I notice that the PR includes a yarn lockfile at `packages/firebase-functions/check-version/functions/yarn.lock`. None of the other packages in the repository have a lockfile -- I think this is because they use the one at the root as part of how yarn workspaces work. Is there a reason that this package would be different?

danibonilha:
> Overall, looks good. I notice that the PR includes a yarn lockfile at `packages/firebase-functions/check-version/functions/yarn.lock`. None of the other packages in the repository have a lockfile -- I think this is because they use the one at the root as part of how yarn workspaces work. Is there a reason that this package would be different?

Good catch, it seems lerna path is not configured correctly for nested paths, I'll add a fix = )

lukemelia:
@danibonilha Should the yarn.lock file be removed from the PR then?

danibonilha:
@lukemelia I've removed the nested yarn.lock, the one changed in this PR is the root one, adding the firebase dependencies.

habdelra (approved):
looks good to me 👍

PR #1936, opened 2021-08-05
Enhancement/withdrawal bridging error message

Aierie:
Addresses CS-1462.

- Adds error UI to the withdrawal status card 
- Adds component tests for the withdrawal status card and the deposit status card error states

image

### DIFF
--- a/packages/web-client/app/components/card-pay/withdrawal-workflow/transaction-status/index.ts
+++ b/packages/web-client/app/components/card-pay/withdrawal-workflow/transaction-status/index.ts
@@ -16,6 +16,7 @@ class CardPayWithdrawalWorkflowTransactionStatusComponent extends Component<Work
   @service declare layer2Network: Layer2Network;
 
   @tracked completedCount = 1;
+  @tracked error = false;
 
   constructor(owner: unknown, args: WorkflowCardComponentArgs) {
     super(owner, args);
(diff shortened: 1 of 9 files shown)


packages/web-client/app/components/card-pay/withdrawal-workflow/transaction-status/index.ts
```diff
@@ -16,21 +16,26 @@ class CardPayWithdrawalWorkflowTransactionStatusComponent extends Component<Work
...
+      let result = yield this.layer2Network.awaitBridgedToLayer1(
+        this.layer2BlockHeightBeforeBridging!,
+        this.args.workflowSession.state.relayTokensTxnHash
+      );
+      this.args.workflowSession.update('bridgeValidationResult', result);
+      this.layer2Network.refreshBalances();
+      this.completedCount = 2;
+      this.args.onComplete?.();
+    } catch (e) {
+      this.error = true;
```
lukemelia, line 37: It seems bad to blackhole the error object here. Perhaps we should log it to sentry?

lukemelia (changes requested):

packages/web-client/app/components/card-pay/withdrawal-workflow/transaction-status/index.ts
```diff
@@ -34,6 +34,9 @@ class CardPayWithdrawalWorkflowTransactionStatusComponent extends Component<Work
       this.completedCount = 2;
       this.args.onComplete?.();
     } catch (e) {
+      console.error('Failed to complete bridging to layer 1');
+      console.error(e);
+      Sentry.captureException(e);
```
Aierie, line 39: Does this look correct to you, @lukemelia? My understanding is Sentry doesn't automatically capture browser logs and requires manual capturing of caught exceptions.

Aierie: Based on: https://docs.sentry.io/platforms/javascript/guides/ember/usage/

lukemelia: I believe that console messages are captured as breadcrumbs but would not be treated as an error, so the approach you have here is good.

lukemelia (approved):

PR #1938, opened 2021-08-05
web: Extract hub authentication card component

backspace:
Review note: since this builds on the merchant creation skeleton in #1931, it’s set to merge into that for now, and includes a temporary override to make the GitHub Actions tests run; will remove that when #1931 is complete.


lukemelia (approved):
🤺

PR #1939, opened 2021-08-05
web: Add create merchant feature flag

backspace:
We didn’t discuss implementation details for this, but here’s a minimal approach.

This will conflict with #1931, but that’s easily fixed.


lukemelia (approved):
This is what I had in mind 👍

Aierie (approved):

PR #1940, opened 2021-08-06
Use websocket connections for layer 1 and layer 2

lukemelia:


packages/web-client/app/utils/wc-provider.ts
```diff
@@ -0,0 +1,526 @@
...
+
+import ProviderEngine from 'web3-provider-engine';
+import CacheSubprovider from 'web3-provider-engine/subproviders/cache';
+import FixtureSubprovider from 'web3-provider-engine/subproviders/fixture';
+import FilterSubprovider from 'web3-provider-engine/subproviders/filters';
+import HookedWalletSubprovider from 'web3-provider-engine/subproviders/hooked-wallet';
+import NonceSubprovider from 'web3-provider-engine/subproviders/nonce-tracker';
+import SubscriptionsSubprovider from 'web3-provider-engine/subproviders/subscriptions';
+import BlockTracker from './block-tracker';
+import { WebsocketProvider } from 'web3-core';
```
Aierie, line 27: `WebsocketProvider` here is only a type, there is no exported class.

packages/web-client/app/utils/wc-provider.ts
```diff
@@ -0,0 +1,526 @@
...
+    let socket = new WebSocket(this.rpcWss[this.chainId], []);
+    socket.addEventListener('close', function () {
+      console.log('close', ...arguments);
+    });
+    socket.addEventListener('message', function () {
+      console.log('message', ...arguments);
+    });
+    socket.addEventListener('open', function () {
+      console.log('open', ...arguments);
+    });
```
Aierie, line 119: We could probably remove this code, unless we want to keep it for some kind of logging?

Aierie (changes requested):
This works on https://github.com/cardstack/cardstack/pull/1940/commits/9f8b76c6e29c6b38da84780e2db83273479d6d9e, but the new import for `WebsocketProvider` in the latest commit (https://github.com/cardstack/cardstack/pull/1940/commits/4742860d038c9e7ea99f407623d89abc00320ff3) only references a type. I think web3 messed up their typing for their exports a little, because there is no named export for `WebsocketProvider` from `web3-core`.

packages/web-client/app/utils/wc-provider.ts
```diff
@@ -0,0 +1,526 @@
...
+  public websocketProvider!: WebsocketProvider;
+  public networkId!: number;
+  public infuraId?: string;
+
+  constructor(opts: ICardstackWalletConnectProviderOptions) {
+    super({
+      blockTracker: new BlockTracker({
+        provider: (new WebsocketProvider(
+          opts.rpcWss[opts.chainId!]
+        ) as unknown) as Provider,
```
Aierie, line 60: This WebsocketProvider doesn't have the reassigned sendAsync from the one below and will not work, even if the import is fixed.

Aierie (approved):
Nice! Tried this out for:
- connection
- reconnection
- disconnection
- connecting to the wrong network
- going through the whole Withdrawal flow

All works well with Rainbow Wallet for Layer 1.

PR #1943, opened 2021-08-06
web: Add source map upload for Sentry

backspace:
This updates deployment workflows to include steps to upload source maps to Sentry using the [Sentry Release](https://github.com/marketplace/actions/sentry-release) GitHub Action.

You can see an example release [here](https://sentry.io/organizations/cardstack/releases/web-client-d9cc44f912392094923cf39b55793eb047e21b6c%400.19.49/?project=5879708), with embedded artifacts that include the source maps.

image

[Here](https://sentry.io/organizations/cardstack/issues/2566752189/?environment=staging&project=5879708) is a manually-triggered issue from that release:

image

One unfortunate aspect is that the stacktrace for the application starts off as collapsed, as it is marked “non-app”; I have a thread going with Sentry support about how to address that. So to view the relevant code, you have to expand the first entry:

image

I made the arbitrary choice of this format for a Sentry release name, building on the existing pattern but adding the SHA:

```web-client-${GITHUB_SHA}@${LERNA_VERSION}```

A possible improvement would be to truncate the SHA at 7 characters, which would be less verbose, but the string-manipulation available in GitHub Actions references are minimal, so it might need to look something like [this](https://github.com/winnipegpolicecauseharm/wpch-ghost-theme/blob/3667bd021288956339fd3456d2a15b62c2b8ad61/.github/workflows/destroy-preview.yml#L17).

### DIFF
--- a/.github/workflows/manual-web-client.yml
+++ b/.github/workflows/manual-web-client.yml
@@ -17,6 +17,7 @@ jobs:
     steps:
       - uses: actions/checkout@v2
         with:
+          fetch-depth: 0
           ref: ${{ github.event.inputs.ref }}
 
       - uses: volta-cli/action@v1
@@ -53,3 +54,28 @@ jobs:
 
       - name: Deploy
         run: yarn deploy:web-client:${{ github.event.inputs.environment }}
+      - name: Get current lerna version
+        id: lerna-version
+        uses: saulonunesdev/lerna-get-version-action@v1.0.4
+      - name: Create decompressed copies of assets for Sentry
+        shell: bash
+        run: |
+          cd packages/web-client/tmp
+          mkdir -p deploy-dist-uncompressed/assets
+          cd deploy-dist/assets
+          echo *.js *.map | tr ' ' '\n' | xargs -n1 -I{} cp "{}" "../../deploy-dist-uncompressed/assets/{}.br"
+          cd ../../deploy-dist-uncompressed/assets
+          brotli --rm --decompress *.br
+      - name: Create Sentry release
+        uses: getsentry/action-release@v1
+        env:
+          SENTRY_AUTH_TOKEN: ${{ secrets.WEB_CLIENT_SENTRY_AUTH_TOKEN }}
+          SENTRY_ORG: cardstack
+          SENTRY_PROJECT: web-client
+        with:
+          environment: ${{ github.event.inputs.environment }}
+          ignore_empty: true
+          ignore_missing: true
+          sourcemaps: packages/web-client/tmp/deploy-dist-uncompressed
+          version: web-client-${{ github.sha }}@${{ steps.lerna-version.outputs.lerna-version }}
+          url_prefix: https://${{ fromJSON('["app-staging.stack.cards", "app.cardstack.com"]')[github.event.inputs.environment != 'staging'] }}
--- a/.github/workflows/push-main.yml
+++ b/.github/workflows/push-main.yml
@@ -155,7 +155,7 @@ jobs:
     steps:
       - uses: actions/checkout@v2
         with:
-          fetch-depth: 1
+          fetch-depth: 0
       - uses: volta-cli/action@v1
       - name: Set up yarn cache
         uses: actions/cache@v2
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

**Comment by lukemelia** [comment]
Could it be that we're uploading the source maps compressed and it expects them uncompressed?

**Comment by backspace** [comment]
> Could it be that we're uploading the source maps compressed and it expects them uncompressed?

oh wow, that does seem to be the case! I was viewing the source maps in the browser vs downloading 🙃 I’ll look at how to upload uncompressed copies, thanks.

image

**Comment by backspace** [comment]
I was able to expand a source map from a Sentry release as Brotli… it’s unfortunate that Sentry seems [unable to expand Brotli](https://github.com/getsentry/sentry-javascript/issues/2065#issuecomment-571979966).

Sentry’s source map validator [rejects](https://sourcemaps.io/report/1628610522023_https%3A%2F%2Fapp-staging-assets-cardstack.s3.amazonaws.com%2Fassets%2Fchunk.c9828302d1a9a295a42d.js) one of the Javascript assets:
Source Map Validator 2021-08-10 08-53-17

But the file _does_ include `sourceMappingURL`:

sourcemappingurl

Currently looking at options for uploading uncompressed files.

**Comment by backspace** [comment]
I finally have a real stacktrace showing in Sentry:

image

However, it uses the [`keep` option](https://github.com/DockYard/ember-cli-deploy-compress#keep) for `ember-cli-deploy-compress`, which writes the compressed files separately as `chunk-whatever.br`, meaning the assets won’t be served with compression. Instead of this, I’ll look at duplicating the `deploy-dist` directory, decompressing the Brotli files within, and uploading those to Sentry.

File: .github/workflows/manual-web-client.yml [hunk 1]
```diff
@@ -53,3 +54,39 @@ jobs:
...
+        id: assets_domain
+        run: |
+          if [$ENVIRONMENT == 'staging']
+          then
+              echo "::set-output name=domain::app-staging.stack.cards"
+          else
+              echo "::set-output name=domain::app.cardstack.com"
+          fi
+        env:
+          ENVIRONMENT: ${{ github.event.inputs.environment }}
```
backspace, line 79: This is a clunky compensation for the lack of a ternary operator, as taken from [here](https://github.community/t/do-expressions-support-ternary-operators-to-change-their-returned-value/18114/2). Is there anywhere the S3 assets domain could be derived from instead? I couldn’t find it defined anywhere.

.github/workflows/push-main.yml
```diff
@@ -172,6 +172,31 @@ jobs:
...
+          SENTRY_AUTH_TOKEN: ${{ secrets.WEB_CLIENT_SENTRY_AUTH_TOKEN }}
+          SENTRY_ORG: cardstack
+          SENTRY_PROJECT: web-client
+        with:
+          environment: staging
+          ignore_empty: true
+          ignore_missing: true
+          sourcemaps: packages/web-client/tmp/deploy-dist-uncompressed
+          version: web-client-${{ github.sha }}@${{ steps.lerna-version.outputs.lerna-version }}
+          url_prefix: https://app-staging.stack.cards
```
backspace, line 199: This is quite similar to the steps in `manual-web-client`, is it worth extracting into an in-repository action akin to `find-pr-number`?

.github/workflows/manual-web-client.yml
backspace: Another less-verbose-but-still-strange approach would be an adaptation of [this](https://github.community/t/do-expressions-support-ternary-operators-to-change-their-returned-value/18114/4):

```
${{ fromJSON('["app-staging.stack.cards", "app.cardstack.com"]')[github.events.input.environment == 'staging'] }}
```

.github/workflows/manual-web-client.yml
```diff
@@ -53,3 +54,28 @@ jobs:
...
+          SENTRY_AUTH_TOKEN: ${{ secrets.WEB_CLIENT_SENTRY_AUTH_TOKEN }}
+          SENTRY_ORG: cardstack
+          SENTRY_PROJECT: web-client
+        with:
+          environment: ${{ github.event.inputs.environment }}
+          ignore_empty: true
+          ignore_missing: true
+          sourcemaps: packages/web-client/tmp/deploy-dist-uncompressed
+          version: web-client-${{ github.sha }}@${{ steps.lerna-version.outputs.lerna-version }}
+          url_prefix: https://${{ fromJSON('["app-staging.stack.cards", "app.cardstack.com"]')[github.event.inputs.environment != 'staging'] }}
```
backspace, line 81: This strange formation is a [workaround](https://github.community/t/do-expressions-support-ternary-operators-to-change-their-returned-value/18114/4) for the lack of a ternary operator. But is there a better way to determine the domain for S3 assets based on the environment? I couldn’t find anywhere to extract these from. The `url_prefix` is required for Sentry to properly look up the artifacts uploaded as part of the release when resolving a stacktrace.

backspace:
Okay, this is actually working now, apart from the “Crashed in non-app” problem mentioned in the edited opening comment, so it’s ready for review, @lukemelia. Thanks to @Aierie for the pairing today that made me realise the `url_prefix` was required.

I have exercised the staging deployment workflow in now-removed commits that overrode the branches filter.

The `manual-web-client` workflow hasn’t been exercised for the production environment, so when that happens, there might be a need to be followup. [Manually deploying](https://github.com/cardstack/cardstack/runs/3305294942?check_suite_focus=true) on staging produced an [issue](https://sentry.io/organizations/cardstack/issues/2566752189/?environment=staging&project=5879708) with the expected stacktrace.

.github/workflows/push-main.yml
lukemelia: I think we can leave it for @pcjun97 to evaluate as he continues to refine our deployment process.

lukemelia (approved):

backspace:
I’m in communication with Sentry support about “Crashed in non-app” but I’ll merge this for now as it’s still helpful with that problem.

PR #1946, opened 2021-08-09
Handle layer one wallet workflow disconnect

jurgenwerk:
Ticket: [CS-1502](https://linear.app/cardstack/issue/CS-1502/when-workflow-epilogue-shows-includes-layer-1-wallet-card-as-a-summary)

This change affects the layer one connect card component when it's in the epilogue (i.e the deposit flow).

The problem is that when the wallet gets disconnected in a completed deposit flow, it shows the connect prompt, which doesn't make a lot of sense when the workflow is completed:

image

This PR aims to:
1. Show the wallet summary instead
2. Simplify the logic for conditional showing of the prompt/summary, and contents in the chin

The result is that we now see a wallet summary:

image

⁉️ QUESTION:

In my previous [PR](https://github.com/cardstack/cardstack/pull/1923) I addressed the same problem but for layer 2. In that case I chose to simply hide the action chin once the wallet disconnects. But in this case, the chin has some more button logic included, and the acceptance tests expect the chin to always be present, so I kept the chin for now. 

Do we want to:
1. Remove the chin when the wallet disconnects in this case (layer one card) the same way we do now with layer two card
2. Add the chin back in layer two card and keep the same logic as here
3. Not worry about this inconsistency given it's a bit of an edge case

### DIFF
--- a/packages/web-client/app/components/card-pay/layer-one-connect-card/index.hbs
+++ b/packages/web-client/app/components/card-pay/layer-one-connect-card/index.hbs
@@ -17,6 +17,7 @@
       @title={{this.connectedWalletProvider.name}}
       @imgUrl={{this.connectedWalletProvider.logo}}
       @dataTestId="layer-one-connect"
+      data-test-layer-1-wallet-summary
     >
       <CardPay::FieldStack>
         <CardPay::LabeledValue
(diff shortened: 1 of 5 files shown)


lukemelia:
I think we should go with "(1) Remove the chin when the wallet disconnects in this case (layer one card) the same way we do now with layer two card"

lukemelia (approved):
Looks good except the change for your open question. 👍

Aierie (approved):
Yep, looks good to me.

jurgenwerk:
@lukemelia @Aierie I added another commit where I hide the chin if the user disconnect the wallet. Can you please review it? Thanks!

lukemelia (approved):
:shipit:

PR #1951, opened 2021-08-10
Refactor how workflow checks for complete milestones...

lukemelia:
...to avoid being too eager in checking postable completeness.

This means that `includeIf` methods will be called closer to when the postable is due to be displayed.


Aierie (approved):
Nice!

PR #1952, opened 2021-08-10
Add Check Balance milestone to withdrawal workflow

lukemelia:


packages/web-client/app/components/card-pay/withdrawal-workflow/index.ts
```diff
@@ -2,21 +2,46 @@ import Component from '@glimmer/component';
...
 
 const FAILURE_REASONS = {
   DISCONNECTED: 'DISCONNECTED',
 } as const;
 
+class NotEnoughBalanceWorkflowMessage extends WorkflowPostable {
+  constructor() {
+    super({ name: 'cardbot' }, () => {
+      let balance = this.layer1Network.defaultTokenBalance;
+      return !!balance && balance <= new BN(0.01);
```
Aierie, line 30: `new BN(0.01)` will become `new BN('0')`. See:

https://github.com/indutny/bn.js#usage

Aierie: Also unsure whether using numeric operators instead of `bn.lte` is a good thing with bn.js, there is no clear mention in the documentation or issues about how this is implemented.

Aierie (commented):
Checked to see the context of use of #1951. #1951 solves the problem where we don't want to check on state before it is set in a previous milestone, but there is a possibility that layer 1 balances might not be fetched when we call `includeIf` (we are not waiting for balances to be fetched to consider layer 1 connected and hence the card + milestone completed). I think it would be good to change this to ensure consistency, but this does mean we will be slower in network initialization/first connection. 

Also left a comment about BN.js use

packages/web-client/app/components/card-pay/withdrawal-workflow/index.ts
lukemelia: Correct. Discovered both these issues in my development cycle.

lukemelia:
Closing in favor of #1958

PR #1953, opened 2021-08-10
Extract and reuse disconnection-cta component in merchant-creation flow

burieberry:
CS-1531

- Some refactor to reuse the disconnection-cta in workflows
- Add missing tests


jurgenwerk (approved):
LGTM! 👌

PR #1954, opened 2021-08-11
Enhancement/propagate connection events

Aierie:
This PR fixes CS-1490 by ensuring that connection events are propagated to other tabs. 

## Communicating connection events between tabs
- Using broadcast channels, as we were doing for disconnections, we send a message with a type of 'CONNECTED' to the layer-specific channel (one for layer 1 and another for layer 2).
- WalletConnect does not store its session in LocalStorage quickly enough, so we are unable to leverage that to establish connections. Instead we send the whole session over through the broadcast channel. MetaMask does not require this.

## Keeping cross-tab communication a responsibility of the ConnectionManager rather than leaking it into Layer1Chain
- Make ConnectionManager long-lived, and have it delegate to strategies for WalletConnect and MetaMask. Necessary to keep it long-lived as opposed to created upon connection + destroyed upon disconnection, so that it can listen for connection events.
- Have the now long-lived ConnectionManager post messages to, and listen for messages from the layer-specific broadcast channel. 

## Full flow for a layer 1 connection event being propagated to another tab and a session being established in that tab
- In tab A, the user initiates connection.
- After the user has confirmed that they want to connect on their wallet, a connection is established between wallet and dApp. The provider's `accountsChanged` event is fired with the newly connected account.
- The ConnectionManager posts a message to the broadcast channel. This message contains information about which provider (WalletConnect or MetaMask) was used and optionally (for WalletConnect), the session
- In tab B, the broadcast channel's message event listener picks up on the message. If the receiving ConnectionManager is not connected (there is no ConnectionStrategy), then it will emit a `cross-tab-connection` event to its consumer with the whole message.
- The consumer (layer 1 chain in the other tab) initiates connection using the information in the message (`providerId` and possibly `session`).

### DIFF
--- a/packages/web-client/app/utils/typed-channel.ts
+++ b/packages/web-client/app/utils/typed-channel.ts
@@ -1,0 +1,31 @@
+export class TypedChannel<MessageType = any> {
+  channelName: string;
+  private broadcastChannel: BroadcastChannel;
+
+  constructor(channelName: string) {
+    this.channelName = channelName;
+    this.broadcastChannel = new BroadcastChannel(channelName);
+  }
+
+  addEventListener(
+    event: keyof BroadcastChannelEventMap,
+    callback: (ev: MessageEvent<MessageType>) => any
+  ) {
+    this.broadcastChannel.addEventListener(event, callback);
+  }
+
+  removeEventListener(
+    event: keyof BroadcastChannelEventMap,
+    callback: (ev: MessageEvent<MessageType>) => any
+  ) {
+    this.broadcastChannel.removeEventListener(event, callback);
+  }
+
+  close() {
+    this.broadcastChannel.close();
+  }
+
+  postMessage(message: MessageType) {
+    this.broadcastChannel.postMessage(message);
+  }
+}
--- a/packages/web-client/app/utils/web3-strategies/layer-1-connection-manager.ts
+++ b/packages/web-client/app/utils/web3-strategies/layer-1-connection-manager.ts
@@ -16,6 +16,7 @@ import { TypedChannel } from '../typed-channel';
 const GET_PROVIDER_STORAGE_KEY = (chainId: number) =>
   `cardstack-chain-${chainId}-provider`;
 const WALLET_CONNECT_BRIDGE = 'https://safe-walletconnect.gnosis.io/';
+
 interface ConnectionManagerOptions {
   chainId: number;
   networkSymbol: NetworkSymbol;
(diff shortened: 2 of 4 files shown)


Aierie:
Nice to have: some way of Typechecking what goes into broadcast channel messages.

Aierie:
As we capture more things to Sentry, it might be useful to add tags so that we can explicitly mark an event as being part of a certain workflow/functionality of the app, though I'm not sure what the best way to do this is.

packages/web-client/app/utils/typed-channel.ts
```diff
@@ -0,0 +1,31 @@
...
+  }
+
+  close() {
+    this.broadcastChannel.close();
+  }
+
+  postMessage(message: MessageType) {
+    this.broadcastChannel.postMessage(message);
+  }
+}
```
backspace, line 31: nice, I hope this wrapper ends up being helpful! 🤩

packages/web-client/app/utils/web3-strategies/layer-1-connection-manager.ts
```diff
@@ -11,87 +9,148 @@ import { Emitter, SimpleEmitter } from '../events';
...
 import { action } from '@ember/object';
 import { getConstantByNetwork, networkIds } from '@cardstack/cardpay-sdk';
+import { NetworkSymbol } from './types';
+import Web3 from 'web3';
+import { TypedChannel } from '../typed-channel';
 
 const GET_PROVIDER_STORAGE_KEY = (chainId: number) =>
   `cardstack-chain-${chainId}-provider`;
 const WALLET_CONNECT_BRIDGE = 'https://safe-walletconnect.gnosis.io/';
-
```
backspace, line 18: spurious deletion? 😯

backspace (approved):
nice work, I exercised this locally (on Chrome and Safari 14.1.2) and it kept the tabs in sync as expected 🎉

safari-wallet-update-screencast 2021-08-12 07-53-44

packages/web-client/app/utils/typed-channel.ts
Aierie: For now it is only a subset of broadcast channel's functionality, we can add on as needed!

lukemelia (approved):
Looks great.... AND... I am concerned about the amount of code and interactions here that are not covered by tests. What would this look like if was designed to be testable and had thorough tests?

Aierie:
> Looks great.... AND... I am concerned about the amount of code and interactions here that are not covered by tests. What would this look like if was designed to be testable and had thorough tests?

I'm not very used to designing things for testability, but this is how I'm envisioning it. Without having realistic web3/network-related services available, meaningful tests for this could cover:
- Ensuring the right callbacks are called (or not called) when a broadcast channel message is picked up, depending on its payload/message
- Ensuring that successful connecting and reconnecting always sets web3, the strategy, and the providerId
- Ensuring that reset cleans up state properly (strategy, providerId)

PR #1955, opened 2021-08-11
Merchant info endpoint groundwork

jurgenwerk:
Ticket: [CS-1432](https://linear.app/cardstack/issue/CS-1432/hub-endpoint-to-create-merchantinfo-generates-and-returns-a-did)

This PR includes all the boilerplate code of adding a new route and a very basic endpoint functionality (insert to DB via POST). 

This is only a part of the bigger feature. I think splitting this feature up into smaller PRs is more convenient to review. 

Missing things (will be added later, probably in a separate PR):
1. Validations
2. DID creation 
3. More comprehensive tests
4. Refactor / DRY up (currently some code is duplicated - serialization, validation stuff)

### DIFF
--- a/packages/hub/config/structure.sql
+++ b/packages/hub/config/structure.sql
@@ -591,6 +591,23 @@ CREATE TABLE graphile_worker.migrations (
 ALTER TABLE graphile_worker.migrations OWNER TO postgres;
 
 --
+-- Name: merchant_infos; Type: TABLE; Schema: public; Owner: postgres
+--
+
+CREATE TABLE public.merchant_infos (
+    id uuid NOT NULL,
+    name text NOT NULL,
+    slug text NOT NULL,
+    color text NOT NULL,
+    text_color text NOT NULL,
+    owner_address text NOT NULL,
+    created_at timestamp without time zone DEFAULT CURRENT_TIMESTAMP NOT NULL
+);
+
+
+ALTER TABLE public.merchant_infos OWNER TO postgres;
+
+--
 -- Name: pgmigrations; Type: TABLE; Schema: public; Owner: postgres
 --
 
@@ -724,6 +741,14 @@ ALTER TABLE ONLY graphile_worker.migrations
 
 
 --
+-- Name: merchant_infos merchant_infos_pkey; Type: CONSTRAINT; Schema: public; Owner: postgres
+--
+
+ALTER TABLE ONLY public.merchant_infos
+    ADD CONSTRAINT merchant_infos_pkey PRIMARY KEY (id);
+
+
+--
 -- Name: pgmigrations pgmigrations_pkey; Type: CONSTRAINT; Schema: public; Owner: postgres
 --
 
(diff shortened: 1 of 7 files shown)


packages/hub/config/structure.sql
```diff
@@ -590,6 +590,22 @@ CREATE TABLE graphile_worker.migrations (
...
 ALTER TABLE graphile_worker.migrations OWNER TO postgres;
 
+--
+-- Name: merchant_infos; Type: TABLE; Schema: public; Owner: postgres
+--
+
+CREATE TABLE public.merchant_infos (
+    id uuid NOT NULL,
+    merchant_name text NOT NULL,
+    merchant_slug text NOT NULL,
```
jurgenwerk, line 600: In the specification, the name for this field is `unique id`. That sounds a bit confusing to me, because `id` is the actually the unique id. So I took the liberty to use the "slug" nomenclature. Let me know if that's ok!

packages/hub/db/migrations/20210809113449561_merchant-infos.js
```diff
@@ -0,0 +1,19 @@
...
+    id: { type: 'uuid', primaryKey: true },
+    merchant_name: { type: 'string', notNull: true },
+    merchant_slug: { type: 'string', notNull: true },
+    color: { type: 'string', notNull: true },
+    text_color: { type: 'string', notNull: true },
+    owner_address: { type: 'string', notNull: true },
+    created_at: { type: 'timestamp', notNull: true, default: pgm.func('current_timestamp') },
+  });
+
+  pgm.createIndex(TABLE, 'merchant_slug', { unique: true });
```
jurgenwerk, line 14: Unique index assures there will be no collisions here. 

However, I just noticed this will allow inserting same strings in different case (upper/lower) which will be a problem. I'm thinking about forcing lowercase before persisting. Thoughts?

jurgenwerk:
@lukemelia can you check and see if this PR looks ok - especially the fields and their properties in the migration? In the meanwhile I'll continue working on the DID part.

packages/hub/db/migrations/20210809113449561_merchant-infos.js
lukemelia: +1 to unique index. I suggest considering using the `citext` module. I believe it is supported on AWS Postgres RDS.

lukemelia (approved):
On the right track. I left a couple of notes about the database questions. Once addressed, good to merge. 👍

packages/hub/db/migrations/20210809113449561_merchant-infos.js
jurgenwerk: Thanks, didn't know this exists! 

I'm a bit reluctant to include new DB extensions unless absolutely necessary (there could be installation issues in various environments, and it's usually something extra to take care of during db provisioning). In this case it can be a simple `toLowercase` when inserting the payload and when filtering. Let me know if you think there's a better solution, but I'm pretty happy with the one I mentioned (will implement this a bit later).

jurgenwerk:
@lukemelia I changed the field names as per your suggestion.

I think this is good to merge now, will do it soon if you agree?

AFAIK the merge will automatically deploy to staging, is that right? Are there any other manual steps included, for example the migration? I guess this is also a good time for me to get familiar with [waypoint](https://linear.app/cardstack/issue/CS-1449/doc-staging-waypoint-setup-last-updated-august-2-2021) so I can learn to do hub deploys 🙃

PR #1956, opened 2021-08-11
Fix lint breaking

danibonilha:
`service-account.json` will be decrypted while deploying it, so this file will only exist during the deployment, that's why we need to use the node `require` so it won't complain that it doesn't exist, open to hear if there's a better solution for this, for the time being I'm disabling the eslint rule.

### DIFF
--- a/packages/firebase-functions/push/functions/src/index.ts
+++ b/packages/firebase-functions/push/functions/src/index.ts
@@ -5,6 +5,8 @@ import { register } from './routes/register';
 import cors from 'cors';
 import { IResponse } from './types';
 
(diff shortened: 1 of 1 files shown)


lukemelia:
Could/should we consider having a placeholder file there that is replaced in the deployment process?

danibonilha:
> Could/should we consider having a placeholder file there that is replaced in the deployment process?

@lukemelia  I thought about it, but then we would need to keep track on git, and to avoid committing the credentials I think it's safer to have it in the `.gitignore` and add it during deployment, lmk what you think

lukemelia (approved):

PR #1957, opened 2021-08-12
Enhancement/account changed event

Aierie:
Part of CS-1286.

- Removes chainId from WalletInfo as this is hardcoded in many places, not read, and potentially misleading.
- Adds an `account-changed` event to Layer 1 and 2, that is only emitted if a new account is connected while the dApp already has an existing account for that layer.

### DIFF
--- a/packages/web-client/app/utils/web3-strategies/layer1-chain.ts
+++ b/packages/web-client/app/utils/web3-strategies/layer1-chain.ts
@@ -207,7 +207,16 @@ export default abstract class Layer1ChainWeb3Strategy
   }
 
   private async updateWalletInfo(accounts: string[]) {
-    this.walletInfo = new WalletInfo(accounts);
+    let newWalletInfo = new WalletInfo(accounts);
+    if (this.walletInfo.isEqualTo(newWalletInfo)) {
+      return;
+    }
+
+    if (this.walletInfo.firstAddress && newWalletInfo.firstAddress) {
+      this.simpleEmitter.emit('account-changed');
+    }
+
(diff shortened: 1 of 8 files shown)


packages/web-client/app/utils/web3-strategies/layer1-chain.ts
```diff
@@ -202,7 +202,16 @@ export default abstract class Layer1ChainWeb3Strategy
   }
 
   private updateWalletInfo(accounts: string[]) {
-    this.walletInfo = new WalletInfo(accounts);
+    let newWalletInfo = new WalletInfo(accounts);
+    if (this.walletInfo.isEqualTo(newWalletInfo)) {
```
jurgenwerk, line 206: Before this condition, the `this.refreshBalances();` was always called (in case there are accounts). But now, it won't be called in case the `walletInfo` is equal due to the early return. I'm not sure but can this be problematic?

jurgenwerk (commented):
Just a minor question, otherwise LGTM!

packages/web-client/app/utils/web3-strategies/layer1-chain.ts
Aierie: You're right, this is a behaviour change. I think we aren't using `updateWalletInfo` to force balance refreshes at the moment, and we have a separate method to do it, so I would lean towards using an early return regardless.

jurgenwerk: 👍

lukemelia (approved):
I like this change. 👍

PR #1959, opened 2021-08-12
Enhancement/account change workflow cancelation

Aierie:
CS-1286

Cancels workflows upon the  `account-changed` event defined in #1957.


jurgenwerk (approved):
This looks ok to me. 🚀

PR #1962, opened 2021-08-12
[CS-1593] - Create separated linter workflow 

danibonilha:
Talking with @lukemelia we decided to split the linter action into it's own workflow, because it ensures that everything is working properly against the whole project and if a package doesn't have any tests we don't need to create a specific workflow for it, that's why I'm also removing some PR's check workflows, since those only had the linter action in it.


lukemelia (approved):
🚤  Quick turnaround!

PR #1968, opened 2021-08-12
web: Add registerMerchant call in create merchant workflow

backspace:
This is significantly copied from the issue prepaid card component that makes the call, will need to adapt/backfill component tests.


backspace:
closing in favour of #1969, I got the branch name wrong 😬

PR #1969, opened 2021-08-12
web: Add registerMerchant to merchant creation

backspace:
This includes placeholders for merchant customisation and the actual
choosing of the prepaid card.

### DIFF
--- a/packages/web-client/app/components/card-pay/create-merchant-workflow/prepaid-card-choice/index.ts
+++ b/packages/web-client/app/components/card-pay/create-merchant-workflow/prepaid-card-choice/index.ts
@@ -1,0 +1,116 @@
+import Component from '@glimmer/component';
+import { action } from '@ember/object';
+import { inject as service } from '@ember/service';
+import { tracked } from '@glimmer/tracking';
+
+import WorkflowSession from '@cardstack/web-client/models/workflow/workflow-session';
+import Layer2Network from '@cardstack/web-client/services/layer2-network';
+import MerchantCustomizationService, {
+  MerchantCustomization,
+} from '@cardstack/web-client/services/merchant-customization';
+import { isLayer2UserRejectionError } from '@cardstack/web-client/utils/is-user-rejection-error';
+import Resolved from '@cardstack/web-client/utils/resolved';
+import {
+  RegisterMerchantOptions,
+  TransactionHash,
+} from '@cardstack/web-client/utils/web3-strategies/types';
+
+import { task, TaskGenerator } from 'ember-concurrency';
+import { taskFor } from 'ember-concurrency-ts';
+import { reads } from 'macro-decorators';
+
+interface CardPayCreateMerchantWorkflowPrepaidCardChoiceComponentArgs {
+  workflowSession: WorkflowSession;
+  onComplete: () => void;
+  isComplete: boolean;
+}
+
+export default class CardPayCreateMerchantWorkflowPrepaidCardChoiceComponent extends Component<CardPayCreateMerchantWorkflowPrepaidCardChoiceComponentArgs> {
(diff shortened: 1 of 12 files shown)


packages/web-client/app/components/card-pay/create-merchant-workflow/prepaid-card-choice/index.ts
```diff
@@ -0,0 +1,119 @@
...
+  TransactionHash,
+} from '@cardstack/web-client/utils/web3-strategies/types';
+
+import { task, TaskGenerator } from 'ember-concurrency';
+import { taskFor } from 'ember-concurrency-ts';
+import { reads } from 'macro-decorators';
+
+// http://ember-concurrency.com/docs/typescript
+// infer whether we should treat the return of a yield statement as a promise
+type Resolved<T> = T extends PromiseLike<infer R> ? R : T;
```
lukemelia, line 23: This is also defined in app/components/card-pay/issue-prepaid-card-workflow/preview/index.ts. Let's extract it and define it in one place.

packages/web-client/app/components/card-pay/create-merchant-workflow/prepaid-card-choice/index.ts
```diff
@@ -0,0 +1,119 @@
...
+      .perform()
+      .catch((e) => console.error(e));
+  }
+
+  @task *createTask(): TaskGenerator<void> {
+    let { workflowSession } = this.args;
+
+    try {
+      this.chinInProgressMessage =
+        'Preparing to create your custom prepaid card…';
```
lukemelia, line 50: This message is not correct.

lukemelia (changes requested):

packages/web-client/app/components/card-pay/create-merchant-workflow/prepaid-card-choice/index.ts
```diff
@@ -0,0 +1,119 @@
...
+          this.txHash = txHash;
+          this.chinInProgressMessage =
+            'Waiting for the transaction to be finalized…';
+        },
+      };
+
+      let prepaidCards = (yield taskFor(
+        this.layer2Network.refreshSafes
+      ).perform()).filterBy('type', 'prepaid-card');
+      let placeholderPrepaidCard = prepaidCards[0]!;
```
Aierie, line 69: Given that this card should allow the user to select a prepaid card, would expect the fetching to occur earlier in the lifeycle of the card and have the fetched prepaid cards be stored as a property of this component. Probably can be done when there is UI follow-up!

Aierie (approved):
I think the refactors + copy change that Luke has suggested would be good! Otherwise, this seems like a good setup for the prepaid card selection UI that will follow.

packages/web-client/app/components/card-pay/create-merchant-workflow/prepaid-card-choice/index.ts
backspace: 👍🏻 I extracted it to `utils/resolved` 😀

packages/web-client/app/components/card-pay/create-merchant-workflow/prepaid-card-choice/index.ts
backspace: hah indeed, updated now to `Preparing to create merchant…`

packages/web-client/app/components/card-pay/create-merchant-workflow/prepaid-card-choice/index.ts
backspace: ya this is just choosing a placeholder to be passed in to the SDK call, fetching can indeed happen earlier to list the available cards

PR #1970, opened 2021-08-13
Internal/layer 1 connection manager tests

Aierie:


packages/web-client/tests/unit/layer-1-connection-manager-test.ts
```diff
@@ -0,0 +1,253 @@
...
+    assert.ok(
+      sokolConnectionManager,
+      'Sokol connection manager is instantiated'
+    );
+
+    assert.ok(
+      kovanConnectionManager,
+      'Kovan connection manager is instantiated'
+    );
+  });
```
backspace, line 23: With these kinds of assertions, I’d expect at least one more that exercises the opposite case, presumably that it can’t be instantiated for some other string.

packages/web-client/tests/unit/layer-1-connection-manager-test.ts
```diff
@@ -0,0 +1,253 @@
...
+    );
+    assert.ok(
+      !connectionManager.provider,
+      'There is no provider at first instantiation'
+    );
+    assert.ok(
+      !connectionManager.providerId,
+      'There is no providerId at first instantiation'
+    );
+  });
```
backspace, line 40: I wonder if this is a case of testing the implementation instead of the interface/behaviour? With unit tests there’s a danger of writing assertions that are too tightly coupled to implementation details that aren’t really that important. Often this is the case when assertions are being made on the private properties of the test under class.

packages/web-client/tests/unit/layer-1-connection-manager-test.ts
```diff
@@ -0,0 +1,253 @@
...
+      session: 'SESSION',
+    };
+
+    connectionManager.broadcastChannel.test__simulateMessageEvent(message);
+
+    assert.ok(
+      onCrossTabConnection.notCalled,
+      'The cross-tab-connection callback was not called'
+    );
+  });
```
backspace, line 165: These two tests are more the kind of thing I think it’s useful to test: that the public interface behaves as expected. 🎉

backspace (commented):
I made some high-level commentary about what kinds of tests/assertions are the most useful; there are some here that I think may be unnecessary/constricting, but I’ll it that up to you and/or other reviewers how important you think it is 🤓

packages/web-client/tests/unit/layer-1-connection-manager-test.ts
Aierie: Wait, I just realised I made a mistake and Sokol should not be allowed to be instantiated. Thanks for the reminder, I need to tighten the typing on this (and the layer 1 chain).

But otherwise, I would have been relying on Typescript to ensure that only certain strings (valid layer 1 networks) could be used to instantiate this.

Aierie: Is Typescript enough, or should I also add assertions within the constructor to prevent this?

packages/web-client/tests/unit/layer-1-connection-manager-test.ts
Aierie: I would like to do more of these tests, however without being able to mock connection behaviour + events in a realistic way for WalletConnect + MetaMask, I think it's hard to do meaningful tests for the other public interfaces (connect, reconnect, disconnect, reset).

packages/web-client/tests/unit/layer-1-connection-manager-test.ts
Aierie: **providerId**: Used in the UI to decide whether to show MetaMask or WalletConnect-specific images/copy so this is also part of the interface. 

**provider**: I think this is more towards implementation.

**Strategy**: Definitely implementation.

Aierie: I think keeping the test for web3.setProvider is good but we can omit the check for provider and strategy being defined.

packages/web-client/tests/unit/layer-1-connection-manager-test.ts
Aierie: I feel like Typescript is enough, and there probably is no necessity for a test for successful instantiation in that case, so this test was removed.

packages/web-client/tests/unit/layer-1-connection-manager-test.ts
Aierie: Dropped assertions for provider and strategy

packages/web-client/tests/unit/layer-1-connection-manager-test.ts
backspace: agreed! 🤓

packages/web-client/tests/unit/layer-1-connection-manager-test.ts
backspace: nice 🤩

packages/web-client/tests/unit/layer-1-connection-manager-test.ts
backspace: maybe there’s a way to encapsulate those for mocking, but it’s always hard to know how far/what direction to go etc 🧐

packages/web-client/app/utils/web3-strategies/layer-1-connection-manager.ts
```diff
@@ -109,7 +118,12 @@ export class ConnectionManager {
   }
 
   createStrategy(providerId: WalletProviderId) {
-    if (providerId === 'metamask') {
+    if (config.environment == 'test') {
```
lukemelia, line 121: When you see this pattern in a class under test: "if test env then instantiate this 'fake' collaborator, else instantiate the 'real' collaborator", it is a good sign that you should make this collaborator a constructor argument rather than have the class under test be responsible for instantiating it.

packages/web-client/tests/unit/layer-1-connection-manager-test.ts
```diff
@@ -0,0 +1,204 @@
...
+
+  test('The onConnect method of the layer 1 connection manager sets the providerId in storage', async function (assert) {
+    let connectionManager = new ConnectionManager('kovan');
+    let web3 = {
+      setProvider: () => {},
+    };
+
+    let onDisconnect = Sinon.spy();
+    connectionManager.on('disconnected', onDisconnect);
```
lukemelia, line 166: Why do you need to set up a disconnected handler in a test about onConnect?

Aierie: Sorry, I don't. Copy paste error.

packages/web-client/app/utils/web3-strategies/layer-1-connection-manager.ts
Aierie: I don't think we can make this collaborator (`ConnectionStrategy` subclasses) a constructor argument of `ConnectionManager` since `ConnectionManager` is long-lived and we can't tell which strategy will be used until the moment of connection/reconnection though.

lukemelia: You could inject a factory in that case.

PR #1972, opened 2021-08-13
Merchant account card UI

burieberry:
Styled template component only. 
Not used in the app yet, but looks like this:

merchant-card


packages/web-client/app/components/card-pay/merchant-card/index.hbs
```diff
@@ -0,0 +1,29 @@
...
+  data-test-merchant-card-logo={{@logo}}
+  data-test-merchant-card={{@id}}
+>
+  <header class="merchant-card__header">
+    <div class="merchant-card__header-network">on {{@network}}</div>
+    <div class="merchant-card__header-address">{{truncate-middle @address}}</div>
+  </header>
+  <section class="merchant-card__content">
+    <div class="merchant-card__merchant">
+      <div class="merchant-card__logo">{{truncate @name 1 false}}</div>
```
lukemelia, line 13: I know it is small, but we should keep an eye out to see if this should be its own component.

packages/web-client/app/components/card-pay/merchant-card/index.css
```diff
@@ -0,0 +1,98 @@
...
+  background: var(--merchant-logo);
+  width: 5rem;
+  height: 5rem;
+  margin: auto;
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  color: var(--boxel-light);
+  font-size: 3rem;
+  border-radius: 50%;
```
lukemelia, line 67: We need the ability to customize the text color here. Perhaps `var(--merchant-logo-text-color)`?

packages/web-client/app/components/card-pay/merchant-card/index.hbs
```diff
@@ -0,0 +1,29 @@
+<article
+  class="merchant-card"
+  style={{css-var merchant-logo=@logo}}
+  data-test-merchant-card-logo={{@logo}}
+  data-test-merchant-card={{@id}}
+>
```
lukemelia, line 6: Add splattributes? We're definitely going to need a way to add a click handler, for example.

lukemelia (changes requested):

packages/web-client/app/components/card-pay/merchant-card/index.hbs
burieberry: Added a note for this

lukemelia (approved):
:shipit:

PR #1973, opened 2021-08-16
Rename ExchangeRate SDK to LayerTwoOracle SDK

lukemelia:
Once this is released, cardwallet should update to use the new name.

The next version after that will drop the ExchangeRate SDK code.

The reason we are making this change is because we need to introduce some exchange rate functionality that derives from Layer 1. We'll call that SDK LayerOneOracle.


habdelra (approved):
looks good 👍

PR #1974, opened 2021-08-16
CS-1598 create custodial wyre wallet for customers and retrieve deposit address

habdelra:
This introduces a service to create a custodial wyre wallet for our customers if they don't already have one and to retrieve the associated custodial wallet's deposit address.

I opted to create a route with a singular inflection instead of a plural one, because the route `/api/custodial-wallet` needs to be accompanied by an authenticated session which ties the request to a specific wallet. There is never a case with this endpoint where you are retrieving more than one wallet. You are only ever retrieving the wallet for which your session is tied to.


habdelra:
It just occurred to me there is a case that I'm not handling here: where the custodial wallet actually already exists in wyre, but not in our database (perhaps because the hub DB was dropped and recreated). I'll add logic to lookup the custodial address in wyre first before creating it.

.vscode/settings.json
```diff
@@ -15,6 +15,12 @@
   "typescript.tsdk": "node_modules/typescript/lib",
   "cSpell.ignoreWords": [
     "ambabi",
-    "home"
+    "qrqzupcr"
```
lukemelia, line 18: Hmm, what's going on here?

packages/hub/routes/custodial-wallet.ts
```diff
@@ -0,0 +1,82 @@
...
+  }
+
+  async get(ctx: Koa.Context) {
+    if (!ensureLoggedIn(ctx)) {
+      return;
+    }
+
+    let userAddress = ctx.state.userAddress.toLowerCase();
+    let db = await this.databaseManager.getClient();
+    let result = await db.query('SELECT id, wyre_wallet_id, deposit_address FROM custodial_wallets WHERE id = $1', [
```
lukemelia, line 24: It is interesting that the `id` is the user's authenticated layer 2 address. My assumption before looking closely at this code is that a field named "id" would a primary key that is either a randomly generated UUID or an auto-incrementing integer. It might be clearer if this field was `user_address` or something like that and designated as the primary key.

lukemelia (approved):
I left one suggestion. Feel free to ignore if you disagree.

.vscode/settings.json
habdelra: it's just adding a word to the spellcheck ignore (it's from the ethereum address of my test). I like to have completely clean warnings, including spell check warnings), as I often have typos in my code so spell checking is really nice in vscode

packages/hub/routes/custodial-wallet.ts
habdelra: sure that's fine. an ethereum address is a really great natural primary key, when it makes sense to use it as a primary we we shouldn't shy away from it.

lukemelia (approved):

packages/hub/routes/custodial-wallet.ts
```diff
@@ -0,0 +1,82 @@
...
+      {
+        status: '401',
+        title: 'No valid auth token',
+      },
+    ],
+  };
+  ctx.status = 401;
+  ctx.type = 'application/vnd.api+json';
+  return false;
+}
```
Aierie, line 76: I think we have 3 identical definitions of ensureLoggedIn in 3 separate files now, might be good to refactor into some kind of utility. Not sure where this would live though.

packages/hub/node-tests/routes/custodial-wallets-test.ts
```diff
@@ -0,0 +1,179 @@
...
+      },
+    });
+    let dbManager = await container.lookup('database-manager');
+    db = await dbManager.getClient();
+    await db.query('DELETE FROM custodial_wallets');
+    // Add foil test data to assert we never get this result
+    await db.query(
+      'INSERT INTO custodial_wallets (user_address, wyre_wallet_id, deposit_address) VALUES ($1, $2, $3)',
+      [`foil-${stubUserAddress}`.toLowerCase(), `foil-${stubWyreWalletId}`, `foil-${stubDepositAddress}`]
+    );
```
Aierie, line 89: The concept of foil test data is new to me. Is this to prevent the case where `/api/custodial-wallet` is getting a random (the first?) wallet from the db, and not the one we request?

packages/hub/routes/custodial-wallet.ts
habdelra: sounds good. I'll make an auth utils for this

packages/hub/node-tests/routes/custodial-wallets-test.ts
habdelra: yup, that's right.

habdelra:
it occurs to me that we don't actually even need a DB for this feature. we could just call wyre directly to get the custodial wallet for a user address. I'm not sure why I thought I needed a DB

PR #1976, opened 2021-08-16
Extract Merchant component

burieberry:
As I was working on memorialized view, ended up making a `Merchant` component and decided to make it a separate PR.

**Usage:**
Basic usage:
```
<CardPay::Merchant @name={{@name}} />
```

```
<CardPay::Merchant
      @name={{@name}}
      @address={{@address}}
      @text="Merchant Account"
      @size="large"
      @vertical={{true}}
      @logoBackground="#ff5050"
      @logoTextColor="#fff"
/>
```
`@name`: merchant name

**Optional args:**
`@address`: merchant address
`@text`: short description
`@size`: `large`, `small` (if no arg given, it'll be default size)
`@vertical`: `true` (default is horizontal flex)

**Examples:**
With just `@name` arg (default):
merchant-default

With `@name`, `@text`, `@size="large"`, `@vertical={{true}}` args:
merchant-large-vertical

With `@name`, `@address`, `@size="small"` args:
merchant-small

### DIFF
--- a/packages/web-client/app/components/card-pay/merchant/index.hbs
+++ b/packages/web-client/app/components/card-pay/merchant/index.hbs
@@ -1,0 +1,30 @@
+<div
+  class={{cn "merchant" merchant--vertical=@vertical}}
+  style={{css-var
+    merchant-logo-background=(or @logoBackground "var(--boxel-red)")
+    merchant-logo-text-color=(or @logoTextColor "var(--boxel-light)")
+  }}
+  data-test-merchant={{@name}}
+  data-test-merchant-logo-background={{@logoBackground}}
+  data-test-merchant-logo-text-color={{@logoTextColor}}
+  ...attributes
+>
+  <div class={{cn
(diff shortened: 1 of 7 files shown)


packages/web-client/app/components/card-pay/merchant/index.hbs
```diff
@@ -0,0 +1,30 @@
...
+  data-test-merchant-logo-background={{@logoBackground}}
+  data-test-merchant-logo-text-color={{@logoTextColor}}
+  ...attributes
+>
+  <div class={{cn
+    "merchant__logo"
+    merchant__logo--sm=(eq @size "small")
+    merchant__logo--lg=(eq @size "large")
+  }}>
+    {{truncate @name 1 false}}
```
lukemelia, line 17: Not a big deal, but this would be more readable if we created a `first-char` helper.

lukemelia (approved):

PR #1977, opened 2021-08-17
[BREAKING] Remove ExchangeRate SDK (please move to LayerTwoOracle SDK)

lukemelia:


habdelra:
Did you push an SDK version with the new LayerTwoOracle and get confirmation that the mobile team is using that?

lukemelia:
>Did you push an SDK version to npm with the new LayerTwoOracle and get confirmation that the mobile team is using that?

Yes

PR #1979, opened 2021-08-17
[hub] Merchant info offchain upload

jurgenwerk:
Ticket: [CS-1432](https://linear.app/cardstack/issue/CS-1432/hub-endpoint-to-create-merchantinfo-generates-and-returns-a-did)

This PR implements uploading the merchant info JSON-API resource to S3 via a worker.

I tested it locally and confirmed the documents appear in the storage.cardstack.com bucket
inside the `merchant-info/` folder.

### DIFF
--- a/packages/hub/main.ts
+++ b/packages/hub/main.ts
@@ -23,6 +23,7 @@ import PrepaidCardPatternsRoute from './routes/prepaid-card-patterns';
 import PrepaidCardCustomizationSerializer from './services/serializers/prepaid-card-customization-serializer';
 import PrepaidCardCustomizationsRoute from './routes/prepaid-card-customizations';
 import PersistOffChainPrepaidCardCustomizationTask from './tasks/persist-off-chain-prepaid-card-customization';
+import PersistOffChainMerchantInfoTask from './tasks/persist-off-chain-merchant-info';
 import MerchantInfosRoute from './routes/merchant-infos';
 import MerchantInfoSerializer from './services/serializers/merchant-info-serializer';
 import { AuthenticationUtils } from './utils/authentication';
@@ -50,6 +51,7 @@ export function wireItUp(registryCallback?: RegistryCallback): Container {
   registry.register('boom-route', BoomRoute);
   registry.register('session-route', SessionRoute);
   registry.register('persist-off-chain-prepaid-card-customization', PersistOffChainPrepaidCardCustomizationTask);
+  registry.register('persist-off-chain-merchant-info', PersistOffChainMerchantInfoTask);
   registry.register('prepaid-card-customizations-route', PrepaidCardCustomizationsRoute);
   registry.register('prepaid-card-customization-serializer', PrepaidCardCustomizationSerializer);
   registry.register('prepaid-card-color-schemes-route', PrepaidCardColorSchemesRoute);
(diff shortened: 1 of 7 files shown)


packages/hub/utils/aws-config.ts
```diff
@@ -16,7 +16,10 @@ export default async function awsConfig({ roleChain = [] }) {
   let accessKeyId = config.get('aws.config.credentials.AccessKeyId');
   let secretAccessKey = config.get('aws.config.credentials.SecretAccessKey');
   if (accessKeyId || secretAccessKey) {
-    result.credentials = config.get('aws.config.credentials');
+    result.credentials = {
```
jurgenwerk, line 19: This change fixes 2 bugs:

1. When testing locally, AWS failed to upload, returning a cryptic error `aws TypeError: Cannot read property 'byteLength' of undefined`

I discovered this is happening because the properties from the config, `AccessKeyId` and `SecretAccessKey`, should be given as `accessKeyId` and `secretAccessKey` while providing credentials to the call ([credentials should not be capitalized](https://github.com/aws/aws-sdk-js-v3/issues/2411#issuecomment-864127937)).

After changing to camelCase, the upload call started working.

2. We were attempting to write to a frozen object

Since `config.get('aws.config.credentials')` returns a frozen object, there was an exception thrown when the function [attempted to write](https://github.com/cardstack/cardstack/blob/9356b70c52e63d267b6a661b71d346c59c9c9a03/packages/hub/utils/aws-config.ts#L34)  to `result.credentials` from the `assumedRole`. The fix is to use a mutable object.

packages/hub/services/serializers/merchant-info-serializer.ts
```diff
@@ -19,27 +19,56 @@ interface JSONAPIDocument {
...
-        color: content.color,
-        textColor: content.textColor,
-        ownerAddress: content.ownerAddress,
+  async serialize(content: MerchantInfo | string): Promise<JSONAPIDocument> {
+    let data: MerchantInfo;
+    let did: string;
+
+    if (typeof content === 'string') {
+      did = encodeDID({ type: 'MerchantInfo', uniqueId: content });
+      data = await this.loadMerchantInfo(content);
```
lukemelia, line 28: I would prefer for the serializer to have a narrower responsibility and more constrained constructor parameters by excluding data loading from it's responsibility.

packages/hub/main.ts
```diff
@@ -218,6 +220,10 @@ export async function bootWorker() {
         let task = await container.instantiate(PersistOffChainPrepaidCardCustomizationTask);
         return task.perform(payload, helpers);
       },
+      'persist-off-chain-merchant-info': async (payload: any, helpers: Helpers) => {
+        let task = await container.instantiate(PersistOffChainMerchantInfoTask);
+        return task.perform(payload, helpers);
+      },
```
lukemelia, line 226: In the category of boilerplate cleanup, we could likely add a macro-like function to generate methods of this type.

lukemelia (changes requested):
Overall looks great. I'd like to see the serializer change made before merging. Other comment is optional.

packages/hub/main.ts
jurgenwerk: Good suggestion. I'll add a new ticket and address this later so that we can merge this and unblock several other tickets that depend on this PR.

packages/hub/services/serializers/merchant-info-serializer.ts
jurgenwerk: Makes sense. I took some inspiration from how the `PrepaidCardCustomizationSerializer` loads the data but I agree it would be good to let the serializer do only the JSON API serialization work. I'm thinking about moving the loading part in the worker task

jurgenwerk:
@lukemelia apologies for a bunch of new code changes, but can you please review my last 3 commits? In those I tried to address your feedback and also do some refactoring. So, these changes aim to:

1. Introduce a separate service responsible for querying the DB. 

I think it's a bit nicer compared to using arbitrary db calls all around the codebase, and more convenient for testing in the console. We can do the same for card customizations

2. Relieve the serializer of fetching data

After thinking about your suggestion, I think the responsibility for fetching the data belongs to the offchain persist worker. So I moved the fetching part there while letting the serializer only do the JSON API serializing

lukemelia (approved):
These changes are the cat's meow! 🐈  Nice work.

PR #1980, opened 2021-08-17
Merchant creation form edit state

Aierie:
This PR adds UI for the edit state of the merchant customization card, with a function mocking the check for uniqueness by checking if the id ends with y.

https://user-images.githubusercontent.com/39579264/130231271-84f90aa0-2b1c-4edd-bfbd-a30c48fead76.mov

Will check with design re: copy for validation messages and helper text.

### DIFF
--- a/packages/web-client/app/components/card-pay/create-merchant-workflow/merchant-customization/index.hbs
+++ b/packages/web-client/app/components/card-pay/create-merchant-workflow/merchant-customization/index.hbs
@@ -36,68 +36,68 @@
         />
       </CardPay::LabeledValue>
     {{else}}
-    <CardPay::FieldStack>
-      <CardPay::LabeledValue
-        @fieldMode="edit"
-        @label="Merchant name"
-        data-test-merchant-customization-merchant-name-field
-      >
-        <Boxel::Input
-          @value={{this.merchantName}}
-          @onInput={{this.onMerchantNameInput}}
-          @invalid={{this.merchantNameValidationMessage}}
-          @errorMessage={{this.merchantNameValidationMessage}}
-          @required={{true}}
-          autocomplete="off"
-          autocorrect="off"
-          autocapitalize="off"
-          spellcheck="false"
-          {{on "blur" this.validateMerchantName}}
-        />
-      </CardPay::LabeledValue>
-      <CardPay::LabeledValue
-        @fieldMode="edit"
-        @label="Merchant ID"
-        data-test-merchant-customization-merchant-id-field
-      >
-        <CardPay::CreateMerchantWorkflow::MerchantCustomization::ValidationStateInput
-          @state={{this.merchantIdInputState}}
-          @value={{this.merchantId}}
-          @onInput={{this.onMerchantIdInput}}
-          @errorMessage={{this.merchantIdValidationMessage}}
-          @helperText="This will be used to uniquely identify your merchant in Cardstack products and cannot be changed after your merchant is created."
-          autocomplete="off"
-          autocorrect="off"
-          autocapitalize="off"
-          spellcheck="false"
-          {{on "blur" this.validateMerchantId}}
-        />
-      </CardPay::LabeledValue>
-      <CardPay::LabeledValue
-        @fieldMode="edit"
-        @label="Custom color"
-        data-test-merchant-customization-color-field
-      >
-        {{!-- template-lint-disable require-input-label --}}
-        <div class="merchant-customization__color">
-          <input
-            type="color"
-            value={{this.merchantBgColor}}
-            class="merchant-customization__color-input"
-            {{on "input" this.onMerchantBgColorInput}}
-          >
-          {{this.merchantBgColor}}
-        </div>
-      </CardPay::LabeledValue>
-      <CardPay::LabeledValue @label="Manager">
-        <CardPay::AccountDisplay
-          @wrapped={{true}}
-          @address={{this.layer2Network.walletInfo.firstAddress}}
-          data-test-merchant-customization-manager-address
-        />
-      </CardPay::LabeledValue>
-    </CardPay::FieldStack>
-  {{/if}}
+      <CardPay::FieldStack>
+        <CardPay::LabeledValue
+          @fieldMode="edit"
+          @label="Merchant name"
+          data-test-merchant-customization-merchant-name-field
+        >
+          <Boxel::Input
+            @value={{this.merchantName}}
+            @onInput={{this.onMerchantNameInput}}
+            @invalid={{this.merchantNameValidationMessage}}
+            @errorMessage={{this.merchantNameValidationMessage}}
+            @required={{true}}
+            autocomplete="off"
+            autocorrect="off"
+            autocapitalize="off"
+            spellcheck="false"
+            {{on "blur" this.validateMerchantName}}
+          />
+        </CardPay::LabeledValue>
+        <CardPay::LabeledValue
+          @fieldMode="edit"
+          @label="Merchant ID"
+          data-test-merchant-customization-merchant-id-field
+        >
+          <CardPay::CreateMerchantWorkflow::MerchantCustomization::ValidationStateInput
+            @state={{this.merchantIdInputState}}
+            @value={{this.merchantId}}
+            @onInput={{this.onMerchantIdInput}}
+            @errorMessage={{this.merchantIdValidationMessage}}
+            @helperText="This will be used to uniquely identify your merchant in Cardstack products and cannot be changed after your merchant is created."
+            autocomplete="off"
+            autocorrect="off"
+            autocapitalize="off"
+            spellcheck="false"
+            {{on "blur" this.validateMerchantId}}
+          />
+        </CardPay::LabeledValue>
+        <CardPay::LabeledValue
+          @fieldMode="edit"
+          @label="Custom color"
+          data-test-merchant-customization-color-field
+        >
+          {{!-- template-lint-disable require-input-label --}}
+          <div class="merchant-customization__color">
+            <input
+              type="color"
+              value={{this.merchantBgColor}}
+              class="merchant-customization__color-input"
+              {{on "input" this.onMerchantBgColorInput}}
+            >
+            {{this.merchantBgColor}}
+          </div>
+        </CardPay::LabeledValue>
+        <CardPay::LabeledValue @label="Manager">
+          <CardPay::AccountDisplay
+            @wrapped={{true}}
+            @address={{this.layer2Network.walletInfo.firstAddress}}
+            data-test-merchant-customization-manager-address
+          />
+        </CardPay::LabeledValue>
+      </CardPay::FieldStack>
+    {{/if}}
   </ActionCardContainer::Section>
   <Boxel::ActionChin
     @state={{if @isComplete "memorialized" "default"}}
--- a/packages/web-client/app/components/card-pay/create-merchant-workflow/merchant-customization/index.hbs
+++ b/packages/web-client/app/components/card-pay/create-merchant-workflow/merchant-customization/index.hbs
@@ -9,16 +9,14 @@
       @label={{if @isComplete "Merchant Name" "Preview"}}
       data-test-merchant-customization-merchant-preview
     >
-      {{#let (trim this.merchantName) as |trimmedName|}}
-        <CardPay::Merchant
-          class={{unless trimmedName "merchant-customization__enter-name"}}
-          @name={{or trimmedName "Enter merchant name"}}
-          @logoBackground={{this.merchantBgColor}}
-          @logoTextColor={{this.merchantTextColor}}
-          @size="large"
-          @vertical={{true}}
-        />
-      {{/let}}
+      <CardPay::Merchant
+        class={{unless this.trimmedMerchantName "merchant-customization__enter-name"}}
+        @name={{or this.trimmedMerchantName "Enter merchant name"}}
+        @logoBackground={{this.merchantBgColor}}
+        @logoTextColor={{this.merchantTextColor}}
+        @size="large"
+        @vertical={{true}}
+      />
     </CardPay::LabeledValue>
     {{#if @isComplete}}
       <CardPay::LabeledValue @label="Merchant ID" data-test-merchant-customization-merchant-id-field>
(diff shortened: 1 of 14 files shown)


packages/web-client/app/components/card-pay/create-merchant-workflow/merchant-customization/index.ts
```diff
@@ -0,0 +1,151 @@
...
+
+function findDomainNameErrors(value: string) {
+  // international domain names start with xn--
+  if (value.startsWith('xn--')) return 'Merchant ID cannot start with xn--';
+  else if (/[^a-z0-9-]/.test(value))
+    return 'Merchant ID can only contain lowercase alphabets, numbers, and hyphens';
+  else if (/^[^a-zA-Z0-9]/.test(value) || /[^a-zA-Z0-9]$/.test(value))
+    return 'Merchant ID must start and end with lowercase alphabet or number';
+  else if (value.length > 63)
+    return `Merchant ID must be at most 63 characters, currently ${value.length}`;
```
Aierie, line 141: @backspace noted yesterday that it might not be possible to allocate the full 63 characters to an ID if it was meant to be used as a subdomain, we would need to allocate `63 - domainLength` instead. 

@lukemelia do you know what the domain would be, to set this?

Aierie: Will set this to a nice round number smaller than 63 for now (50).

packages/web-client/app/components/card-pay/create-merchant-workflow/merchant-customization/index.ts
```diff
@@ -0,0 +1,151 @@
+import { action } from '@ember/object';
+import Component from '@glimmer/component';
+import { tracked } from '@glimmer/tracking';
+import Layer2Network from '@cardstack/web-client/services/layer2-network';
+import { inject as service } from '@ember/service';
+import { WorkflowCardComponentArgs } from '@cardstack/web-client/models/workflow/workflow-card';
+import { didCancel, restartableTask, timeout } from 'ember-concurrency';
+import { taskFor } from 'ember-concurrency-ts';
+import { mostReadable, random as randomColor } from '@ctrl/tinycolor';
```
Aierie, line 9: Selected this package because it has first-class Typescript support, and provides a random color implementation that picks "nice" colors, and also has a nice method for the most readable color.

packages/web-client/tests/integration/components/card-pay/create-merchant-workflow/merchant-customization-test.ts
```diff
@@ -0,0 +1,273 @@
...
+      assert.dom(SAVE_DETAILS_BUTTON).isEnabled();
+
+      for (let invalidEntry of MERCHANT_ID_INVALID_INPUTS) {
+        await fillIn(merchantIdInput, invalidEntry.value);
+        await waitUntil(
+          () =>
+            (document.querySelector(
+              '[data-test-validation-state-input]'
+            ) as HTMLElement).dataset.testValidationStateInput === 'invalid'
+        );
```
backspace, line 201: I was able to remove this `waitUntil` and have the tests still pass because `fillIn` includes a `settled` call? That sometimes doesn’t work if asynchronous processes are happening outside Ember’s test waiters, but did you find that to be the case? (If it is indeed superfluous, that’s likely the case for other instances.)

packages/web-client/app/components/card-pay/create-merchant-workflow/merchant-customization/index.ts
```diff
@@ -0,0 +1,151 @@
...
+    this.merchantIdValidationMessage = '';
+    return true;
+  }
+}
+
+function findDomainNameErrors(value: string) {
+  // international domain names start with xn--
+  if (value.startsWith('xn--')) return 'Merchant ID cannot start with xn--';
+  else if (/[^a-z0-9-]/.test(value))
+    return 'Merchant ID can only contain lowercase alphabets, numbers, and hyphens';
```
backspace, line 137: Is this string something defined elsewhere that you’re just incorporating here? I’m not used to seeing “alphabets” like this, though it’s concise at least… maybe we can ask for design/script input on the exact wording for this

packages/web-client/app/components/card-pay/create-merchant-workflow/merchant-customization/index.ts
```diff
@@ -0,0 +1,151 @@
...
+      return false;
+    }
+
+    // replace w call to api
+    let merchantIdExists: boolean = yield checkIfMerchantIdExists(value);
+
+    this.lastCheckedMerchantId = value;
+    if (merchantIdExists) {
+      this.merchantIdValidationMessage =
+        'This merchant ID is already taken, please choose another ID.';
```
backspace, line 124: We probably need to sweep the application for this (and/or maybe have an i18n-esque strings file to make it easier to scan) but the other validation messages don’t end with `.` so maybe this shouldn’t?

packages/web-client/app/components/card-pay/create-merchant-workflow/merchant-customization/index.ts
```diff
@@ -0,0 +1,151 @@
...
+      if (didCancel(e)) {
+        return;
+      }
+      console.error(e);
+    }
+  }
+
+  @restartableTask *validateMerchantIdTask() {
+    // debounce
+    yield timeout(500);
```
backspace, line 105: I think this should be shorter in tests, ideally set in `config/environment` as with `threadAnimationInterval`, because it’s slowing the test suite unnecessarily:

slow-test-validation-screencast 2021-08-20 10-11-43

Here it is with 10 instead:

fast-test-validation-screencast 2021-08-20 10-15-33

Perhaps the delays still seen in this faster one can also be eliminated, though it’s likely the 750 and therefore soon to be removed 🤔

packages/web-client/app/components/card-pay/create-merchant-workflow/merchant-customization/index.ts
```diff
@@ -0,0 +1,151 @@
...
+      // all fields populated
+      this.merchantBgColor &&
+      this.merchantName &&
+      this.merchantId &&
+      // unique merchant name + valid id characters
+      this.merchantIdInputState === 'valid' &&
+      // no other validation errors
+      !this.merchantNameValidationMessage &&
+      !this.merchantBgColorValidationMessage
+    );
```
backspace, line 32: a space of much opinion, but I lean toward avoiding comments in favour of extraction, so it could look something like this:

```
return
  this.allFieldsPopulated &&
  this.merchantIdValid &&
  this.noValidationMessages
```

packages/web-client/app/components/card-pay/create-merchant-workflow/merchant-customization/index.hbs
```diff
@@ -1,27 +1,28 @@
...
   @header="Merchant Account"
   @isComplete={{@isComplete}}
-  data-test-layout-customization-is-complete={{@isComplete}}
+  data-test-merchant-customization-is-complete={{@isComplete}}
 >
   <ActionCardContainer::Section @title="Choose a name and ID for the merchant">
     <CardPay::LabeledValue
       class="merchant-customization__name"
       @label={{if @isComplete "Merchant Name" "Preview"}}
+      data-test-merchant-customization-merchant-preview
```
backspace, line 10: I think this can be addressed separately but we should figure out what will and won’t work here:

image

image

image
[DIĜIR](https://en.wikipedia.org/wiki/Dingir)

image

image
[BELL](http://www.unicode-symbol.com/u/0007.html) is considered valid but appears blank

packages/web-client/app/components/card-pay/create-merchant-workflow/merchant-customization/index.hbs
```diff
@@ -31,20 +32,92 @@
...
+      data-test-merchant-customization-edit
+        {{on "click" @onIncomplete}}
+      >
+        Edit
+      </m.ActionButton>
+    </:memorialized>
   </Boxel::ActionChin>
-</ActionCardContainer>
\ No newline at end of file
+</ActionCardContainer>
```
backspace, line 123: [EditorConfig for VS Code](https://marketplace.visualstudio.com/items?itemName=EditorConfig.EditorConfig) should eliminate this kind of spurious diff 🤓

backspace (changes requested):
I left some suggestions and IMO-blocking comments, but this is looking good regardless 🤩

burieberry (approved):
Love this! Tested on all 3 main browsers, all good. Also that `tinycolor` package seems great.

packages/web-client/app/components/card-pay/create-merchant-workflow/merchant-customization/index.hbs
burieberry: hmm we should update the `first-char` helper to account for these. Currently it only accepts string type.

burieberry:
> Love this! Tested on all 3 main browsers, all good. Also that `tinycolor` package seems great.

Ah, adding a `y` still gives id validation error, though.

Aierie:
> > Love this! Tested on all 3 main browsers, all good. Also that `tinycolor` package seems great.
> 
> Ah, adding a `y` still gives id validation error, though.

Thanks! Yep, that was intentional, to give a feel for the validation w delay. I can replace the mock uniqueness validation with a no-op for now.

packages/web-client/tests/acceptance/create-merchant-test.ts
```diff
@@ -135,10 +136,21 @@ module('Acceptance | create merchant', function (hooks) {
...
+    );
+    await fillIn(
+      `[data-test-merchant-customization-merchant-id-field] input`,
+      'a-valid-id'
+    );
+    await waitUntil(
+      () =>
+        (document.querySelector(
+          '[data-test-validation-state-input]'
+        ) as HTMLElement).dataset.testValidationStateInput === 'valid'
```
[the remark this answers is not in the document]
Aierie: Ah, I'll change to the conventionally-used helper function then. Thanks!

packages/web-client/tests/integration/components/card-pay/create-merchant-workflow/merchant-customization-test.ts
```diff
@@ -0,0 +1,273 @@
...
+let PREVIEW = '[data-test-merchant-customization-merchant-preview]';
+let MERCHANT = '[data-test-merchant]';
+let MERCHANT_NAME_FIELD =
+  '[data-test-merchant-customization-merchant-name-field]';
+let MERCHANT_ID_FIELD = '[data-test-merchant-customization-merchant-id-field]';
+let COLOR_FIELD = '[data-test-merchant-customization-color-field]';
+let MANAGER = '[data-test-merchant-customization-manager-address]';
+let SAVE_DETAILS_BUTTON = '[data-test-merchant-customization-save-details]';
+let EDIT_BUTTON = '[data-test-merchant-customization-edit]';
+let COMPLETED_SELECTOR = '[data-test-merchant-customization-is-complete]';
```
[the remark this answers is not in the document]
Aierie: We have discussed page objects before. I'm quite interested in seeing how the code would look in a prototype!

packages/web-client/tests/integration/components/card-pay/create-merchant-workflow/merchant-customization-test.ts
Aierie: I think I had to add the `waitUntil` because some assertions were failing. I think this might have to do with what you mentioned earlier, resetting to a valid state prior to checking that an invalid one is detected and correctly handled.

packages/web-client/app/components/card-pay/create-merchant-workflow/merchant-customization/index.ts
Aierie: You're right, will remove the full stop.

packages/web-client/app/components/card-pay/create-merchant-workflow/merchant-customization/index.ts
Aierie: Yep, will be asking design for updated copy in a bit, including the validation/error messages.

packages/web-client/app/components/card-pay/create-merchant-workflow/merchant-customization/index.hbs
burieberry: @backspace What do you recommend for checking for emojis etc inside a string? Regex?

packages/web-client/tests/integration/components/card-pay/create-merchant-workflow/merchant-customization-test.ts
Aierie: This one is unnecessary! The rest are necessary as the save details button is disabled while the validation is running.

packages/web-client/app/components/card-pay/create-merchant-workflow/merchant-customization/index.hbs
Aierie: I found this:

https://medium.com/@giltayar/iterating-over-emoji-characters-the-es6-way-f06e4589516

which suggests using for-of loops or `Array.from(str)`, these implementations account for unicode. I can make this change @burieberry

backspace: ah I didn’t realise this was about `first-char`! I thought it was just the font. Nice find re for/of 🎉

burieberry: > I found this:
> 
> https://medium.com/@giltayar/iterating-over-emoji-characters-the-es6-way-f06e4589516
> 
> which suggests using for-of loops or `Array.from(str)`, these implementations account for unicode. I can make this change @burieberry

Sure, sounds good!

Aierie: https://github.com/cardstack/cardstack/pull/1980/commits/389374446a2b2f60449e8c886cccf2daf83666a0 solves the emoji/multi-code-unit issue, but characters like BELL that don't have a visual representation are a bit trickier. Will create a follow up PR for some input-filtering and maybe whitespace-collapsing.

Aierie:
Follow up issues:
CS-1674, CS-1675

backspace (approved):
thanks for all the changes 🎉

PR #1982, opened 2021-08-17
Fix withdrawal workflow bugs

Aierie:
Fixes bugs in withdrawal workflow as described in CS-1624:
- Check balance step never resolving when workflow is opened without an existing layer 1 connection
- App crashing when disconnecting from layer 1 in withdrawal workflow *when account does not have enough balance to claim tokens


Aierie:
image

With this fix, the app no longer crashes on disconnection, but check balance card looks odd - similar to other existing cards that refer to wallet + depot details.

Aierie:
Slight tangent, but another example of a wallet-related card in an odd state after disconnection:

image

backspace (approved):
I approve with the caveat that I wasn’t able to reproduce the targeted bugs locally 😯

Aierie:
> I approve with the caveat that I wasn’t able to reproduce the targeted bugs locally 😯

Not sure why the first wouldn't be replicable, but for the second I forgot to add a caveat, you need to be using an account with close to 0 ETH on layer 1, which will make the check balance card poll for you having enough ETH.

PR #1983, opened 2021-08-17
web: Change workflows to use info banner icon

backspace:
The withdrawal workflow was referencing a non-existent icon
but they are all meant to use the same one. I have removed
the icon formerly called “deposit” as it’s no longer used
and is shadowed by “payment” in Boxel.

Thanks to @burieberry for clarifying that the icon was
out of date in the first draft of this!


burieberry (approved):

lukemelia (approved):
👌

PR #1984, opened 2021-08-17
Merchant account and prepaid-card select cards memorialized UI

burieberry:
Focused on the memorialized fields with existing data. Also displaying the merchant creation fee and merchant card address in the Prepaid Card Choice card. Will handle the card selector field separately.

merchant-account
prepaid-card-choice

With more data:
more-data


Aierie (approved):
This works nicely! In the future we could find a good way to apply the field typography/style without using `CardPay::LabeledValue`, but overriding its layout seems okay for now.

PR #1985, opened 2021-08-18
Console command + readme improvements

jurgenwerk:
Added the `yarn console` for convenience and a couple of examples to help with app debugging and testing.

### DIFF
--- a/packages/hub/package.json
+++ b/packages/hub/package.json
@@ -71,6 +71,7 @@
   },
   "scripts": {
     "start": "node ./bin/server.ts",
+    "console": "node --experimental-repl-await ./bin/console.ts",
     "db:setup:local": "createdb hub_development && node-pg-migrate up && node ./bin/seed-db.ts && node ./bin/dump-db-structure.js && node ./bin/setup-test-db.js",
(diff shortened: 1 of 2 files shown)


packages/hub/package.json
```diff
@@ -71,6 +71,7 @@
   },
   "scripts": {
     "start": "node ./bin/server.ts",
+    "console": "node --experimental-repl-await ./bin/console.ts",
```
jurgenwerk, line 74: `--experimental-repl-await` flag allows await to be used "freely", without the async function

lukemelia (approved):
Nice!

Note: A yarn script for this also exists in the monorepo root-level package.json (https://github.com/cardstack/cardstack/blob/main/package.json#L61). I'm not sure if there is an elegant way to manage yarn scripts at each level.

jurgenwerk:
Oh, I see, didn't check there. I suppose it doesn't hurt to have this on the hub level as well.

Aierie (approved):
👍

burieberry (approved):

backspace (approved):

PR #1990, opened 2021-08-19
Adding Wyre webhook to move customer purchased DAI to admin account for prepaid card provisioning

habdelra:
Note that wyre will invoke our callback URL with a content-type of `application/json`, which until this PR, our server rejected. I had to add a new KOA middleware to process JSON (as opposed to JSONAPI). For lack of a better name I'm calling this route "/callbacks"--since if we have any other webhooks we need to process, very likely they will be sent to us as `application/json`.

Also the reservation and prepaid card provisioning systems are not ready yet, so i just added placeholders for how I anticipate that will plug in to this feature.


packages/hub/routes/wyre-callback.ts
```diff
@@ -0,0 +1,349 @@
...
+  prepaidCardInventory = inject('prepaid-card-inventory', { as: 'prepaidCardInventory' });
+  databaseManager: DatabaseManager = inject('database-manager', { as: 'databaseManager' });
+
+  constructor() {
+    autoBind(this);
+  }
+
+  async post(ctx: Koa.Context) {
+    let request = ctx.request.body;
+    log.info(`Received wyre callback: ${JSON.stringify(request, null, 2)}`);
```
lukemelia, line 48: Are there any security issues to consider in logging this request body?

packages/hub/services/json-middleware.ts
```diff
@@ -0,0 +1,73 @@
...
+// @ts-ignore
+import mimeMatch from 'mime-match';
+import KoaBody from 'koa-body';
+import { Memoize } from 'typescript-memoize';
+import { CardstackError } from '../utils/error';
+import { SessionContext } from './authentication-middleware';
+import { inject } from '../di/dependency-injection';
+import WyreCallbackRoute from '../routes/wyre-callback';
+
+const ROUTE_PREFIX = '/callbacks';
```
lukemelia, line 13: `callbacks-middleware` or `webhooks-middleware` would be a more descriptive name for this service.

lukemelia (approved):
Looks good. Left a few suggestions.

packages/hub/routes/wyre-callback.ts
habdelra: No, there is no personally identifiable info, no names, no credit card numbers, no addresses. Just wyre generated identifiers, and transaction amounts.

packages/hub/services/json-middleware.ts
habdelra: Sounds good.

PR #1994, opened 2021-08-20
hub: Introduce .env file for hub environment variables

jurgenwerk:
This PR is a suggestion to solve a developer experience type of problem I've been dealing with recently. 

To run the hub locally, at least one custom environment variable is needed: `SERVER_SECRET`. Currently, developers can provide this inline, for example `SERVER_SECRET=abc yarn start:worker`, or save the env variable export to their shell config, or modify development.js file, for example.  

Solutions like this are too brittle or cumbersome IMO. I think a better experience is to be able to save the custom environment variables to a git untracked file and run yarn commands without needing to provide these variables. 

I wanna get your opinion on introducing the `dotenv` library which reads from the `.env` file and populates the environment variables. An example of an .env file:

```
 LOG_LEVELS='*=debug'
 SERVER_SECRET=abc123
 AWS_PROFILE=cardstack_prod
```

Alternatively, we can perhaps use `custom-environment-variables.json` for this, and add it to `.gitignore`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,7 +9,7 @@
 /packages/web-client/deploy/dist
 /packages/cardpay-subgraph/build
 /packages/cardpay-subgraph/generated
-/packages/hub/.env
+/packages/*/.env
 .eslintcache
 .vagrant
 .nvmrc
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,7 @@
 /packages/web-client/deploy/dist
 /packages/cardpay-subgraph/build
 /packages/cardpay-subgraph/generated
+/packages/hub/.env
 .eslintcache
 .vagrant
 .nvmrc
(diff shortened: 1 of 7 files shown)


lukemelia:
I'm open to this. I think it would be a nice improvement.

Aierie:
To prevent `.env` from being cleaned up when we run `yarn clean`, we should also add it to the list of files to exclude:

https://github.com/cardstack/cardstack/blob/ae4d1167aaf445dcc9132aca678b7bd80a2e523f/package.json#L46

jurgenwerk:
@Aierie Good catch, thanks! 👍

It seems that we're mostly on board with this idea.

I'll do some more testing and appease the linter, then switch from draft to a real PR.

lukemelia:
Do we need a README update for this?

jurgenwerk:
@lukemelia good catch, will add a line in the readme.

PR #1995, opened 2021-08-20
Demonstrate a more idiomatic use of resources

ef4:
Here's an example of a more idiomatic way to take advantage of resources. The key idea is that a resource encapsulates what you want, and all the state for getting it, without the consumer needing to worry about any of that.

It would also be possible to split the `fetchJSON` method into a separate `Resource` in its own right, so it can be shared by other resources. It wasn't being reused at the moment, so I didn't bother. 

This implementation is restart and teardown safe, even though it doesn't use ember-concurrency-based generators. The idea is that these async functions do always run to completion, but if the app or test suite has moved on, nobody is observing our resource instance anymore anyway. Each time a Resource's arguments change, it gets replaced by a new instance.

I left a comment about a type error that was revealed by this refactor. `alsoKnownAs` is a `string[]`, not a `string`, so I just took the first one. I don't know what the real desired semantics are.

This also points toward how to answer @backspace's other question about intentionally triggering an update. When you're implementing your own `Resource` class you can just put a method on it, and call it from the component.

### DIFF
--- a/packages/web-client/app/components/card-pay/prepaid-card-safe/index.ts
+++ b/packages/web-client/app/components/card-pay/prepaid-card-safe/index.ts
@@ -1,10 +1,7 @@
 import Component from '@glimmer/component';
-import { inject as service } from '@ember/service';
-
-import CardCustomization from '@cardstack/web-client/services/card-customization';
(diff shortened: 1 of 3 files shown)


backspace:
very helpful, thank you 🎉

PR #2000, opened 2021-08-23
Remove spendBalance field from PrepaidCard

Aierie:
This field is no longer used since recent subgraph change that removed it from the schema.


lukemelia (approved):

lukemelia:
FYI @habdelra this subgraph change broke the web-client and mobile app.

habdelra:
My bad! I missed that

PR #2003, opened 2021-08-23
Fix/misc UI

Aierie:
Fixes some UI bugs that I missed when introducing modifications to the layer two connect card and action chin:

## Layer Two Connect Card @ 360px width
Before
image

After
image

## Disabled memorialized action chin with steps
Before
image

image

After
image

image

### DIFF
--- a/packages/boxel/addon/components/boxel/action-chin/index.css
+++ b/packages/boxel/addon/components/boxel/action-chin/index.css
@@ -51,7 +51,7 @@
   --boxel-action-chin-padding-vertical: var(--boxel-sp-xs);
 }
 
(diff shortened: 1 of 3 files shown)


Aierie:
> Thanks, that memorialized disabled buttons fix has been on my radar for a while 👍
> 
> Should we try to stick to boxel sizing variables (`--boxel-sp-sm` or `--boxel-sp`) instead of the `padding: 1em;`? Not a big deal though

Yep, 1em was what the app store placeholders used so I copied it over, but you're right, we should try to use the sizing variables where possible. i can make that change.

Aierie:
Needs https://github.com/cardstack/cardstack/pull/2005 for CI to pass.

PR #2005, opened 2021-08-23
allow multiple versions of test waiters via dependency-lint config

Aierie:


backspace (approved):
I look forward to using this in #1987, thanks.

backspace:
I’m going to merge so I can do that now 🤓

packages/boxel/tests/dummy/config/dependency-lint.js
```diff
@@ -5,5 +5,6 @@ module.exports = {
     '@embroider/util': '0.39.1 || 0.40.x || 0.41.x',
     'ember-fetch': '6.x || 8.x',
     'ember-test-selectors': '2.1.0 || 5.x',
+    '@ember/test-waiters': '*',
```
lukemelia, line 8: It's usually better to constrain this to the specific versions we have in play here so that we re-evaluate this if a new version is added to the mix.

backspace: ah understood… this was lax in anticipation of [this PR](https://github.com/salsify/ember-cli-dependency-lint/pull/22) being merged which will obviate any local need 😯

PR #2006, opened 2021-08-24
hub: Update merchant attributes to be dasherised

backspace:
While I don’t think this is required by the specification, it’s now established
in other serialisers.


jurgenwerk (approved):
Good catch. 👍

Aierie (approved):

PR #2007, opened 2021-08-24
update copy for merchant ID field as per design feedback

Aierie:


burieberry (approved):
Much better wording!

Aierie:
> Much better wording!

Thanks, will merge after fixing tests!

PR #2009, opened 2021-08-24
web: Fix component name

backspace:
I looked for a linting rule that would help with this but
didn’t find one. It must be happening to others though!


Aierie (approved):
Thanks for fixing this (sorry I missed it)!

backspace:
I’ve done the same thing, and it was missed in review, hence my thirst for linting or the like 😐

PR #2010, opened 2021-08-24
Fix $NaN in prepaid card creation

Aierie:
Fixes commas in the string causing problems with the SDK's currency utilities and producing an NaN. This change still directly uses the output of the SDK's USD formatting, which doesn't use commas to group/separate groups of numbers.

Before
image

After
image

@burieberry, is it important that we have a consistent 3-number separation using commas (eg. 1,234.56 and 123,456)? If so, this may be something we can add to the SDK currency utils (with some coordination with the React Native team, who also use these currency utils). If you think this is important, I'll create a Linear task for this.

### DIFF
--- a/packages/web-client/app/components/card-pay/issue-prepaid-card-workflow/layout-customization/display-view/index.hbs
+++ b/packages/web-client/app/components/card-pay/issue-prepaid-card-workflow/layout-customization/display-view/index.hbs
@@ -14,7 +14,7 @@
         @mockBalance={{true}}
         @mockOptions={{true}}
         @balance="123,456"
-        @usdBalance="1,234.56"
+        @usdBalance="1234.56"
         @address={{placeholder-address}}
         @network={{network-display-info "layer2" "fullName"}}
       />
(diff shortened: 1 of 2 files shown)


burieberry:
> Fixes commas in the string causing problems with the SDK's currency utilities and producing an NaN. This change still directly uses the output of the SDK's USD formatting, which doesn't use commas to group/separate groups of numbers.
> 
> Before
> image
> 
> After
> image
> 
> @burieberry, is it important that we have a consistent 3-number separation using commas (eg. 1,234.56 and 123,456)? If so, this may be something we can add to the SDK currency utils (with some coordination with the React Native team, who also use these currency utils). If you think this is important, I'll create a Linear task for this.

Is this only for USD or for SPEND as well? It seems like the larger amounts would be hard to read without the commas.

burieberry (approved):

Aierie:
> > Fixes commas in the string causing problems with the SDK's currency utilities and producing an NaN. This change still directly uses the output of the SDK's USD formatting, which doesn't use commas to group/separate groups of numbers.
> > Before
> > image
> > After
> > image
> > @burieberry, is it important that we have a consistent 3-number separation using commas (eg. 1,234.56 and 123,456)? If so, this may be something we can add to the SDK currency utils (with some coordination with the React Native team, who also use these currency utils). If you think this is important, I'll create a Linear task for this.
> 
> Is this only for USD or for SPEND as well? It seems like the larger amounts would be hard to read without the commas.

The comma issue only happens with USD at the moment, because we format SPEND ourselves.

PR #2011, opened 2021-08-24
handle not enough balance for gas in RevenuePool.claim

Aierie:
Is this check is necessary for `RevenuePool.claim`? Can/does the funding for the gas come from somewhere else, or from another token? 

This was the only case I found where we move tokens around and don't check if the amount + gas is enough.

### DIFF
--- a/packages/cardpay-sdk/sdk/revenue-pool/base.ts
+++ b/packages/cardpay-sdk/sdk/revenue-pool/base.ts
@@ -127,6 +127,14 @@ export default class RevenuePool {
       0,
       tokenAddress
     );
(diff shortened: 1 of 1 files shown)


habdelra:
So the idea is that the cost of the gas will be deducted straight from the merchant's safe. so in normal use that is probably always going to work. The only way in which this could fail (and your check is necessary) is exactly the logic that you wrote where the merchant 's safe balance after the claim is less than the cost of gas. for xdai, that means about $0.001 USD. However, if this protocol was ever deployed into an environment were gas was expensive (like mainnet), then this could be a real concern. I think its fine to leave this check in there. it does cover a real edge case.

habdelra (approved):
Looks good

PR #2012, opened 2021-08-24
web: Convert card customisation-fetching to useResource

backspace:
I removed #1995 from #1987 so it would be focused, this brings it back.

### DIFF
--- a/packages/web-client/app/resources/card-customization.ts
+++ b/packages/web-client/app/resources/card-customization.ts
@@ -1,0 +1,99 @@
+import { timeout } from 'ember-concurrency';
+import { tracked } from '@glimmer/tracking';
+import { getResolver } from '@cardstack/did-resolver';
+import { Resolver } from 'did-resolver';
+import { Resource } from 'ember-resources';
+import config from '../config/environment';
+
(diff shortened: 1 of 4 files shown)


backspace:
looking into test failures 🧐

Aierie (approved):
Looks good!

packages/web-client/app/resources/card-customization.ts
```diff
@@ -0,0 +1,102 @@
...
+
+export class CardCustomization extends Resource<Args> {
+  // our output values
+  @tracked issuerName: string | undefined;
+  @tracked background: string | undefined;
+  @tracked patternColor: string | undefined;
+  @tracked textColor: string | undefined;
+  @tracked patternUrl: string | undefined;
+
+  // our derived state so you can check when the output values are ready, if you
```
Aierie, line 25: Unsure about use of the word, is this derived state? It isn't calculated based on other state in here.

backspace: I think it’s a reference to [similar properties](http://ember-concurrency.com/docs/derived-state) on Ember Concurrency tasks. But now that I think about it I think these comments were more to educate me than to be committed, so I’ll remove them 😆

PR #1890, opened 2021-07-27
Add Try Again button to Issue Prepaid Card action-chin, and add more granular progress messages

lukemelia:
https://www.loom.com/share/42b192708fd24f78b833d1325f5cbc80

PR #2016, opened 2021-08-26
web: Add nonce-preservation for merchant creation

backspace:
This almost entirely copied from the implementation in
#1890.

screencast 2021-08-27 13-47-24

### DIFF
--- a/packages/web-client/app/components/card-pay/create-merchant-workflow/prepaid-card-choice/index.css
+++ b/packages/web-client/app/components/card-pay/create-merchant-workflow/prepaid-card-choice/index.css
@@ -15,3 +15,8 @@
   flex-grow: 0;
   flex-shrink: 0;
 }
+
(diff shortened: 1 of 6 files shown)


Aierie (approved):
The UI functions as expected!

However in testing this, I realised there are probably some details to work out for canceling transactions in general - canceling the transaction doesn't prevent the user from pressing "confirm" on Card Wallet and creating a merchant.

Aierie:
https://linear.app/cardstack/issue/CS-1714/user-canceling-transaction-from-dapp-still-allows-confirming-of

lukemelia (approved):
It would be interesting to see if we can think of a way to extract this pattern for easier re-use. There's no obvious approach to me, though.

backspace:
ya I’d love for this to be generalisable but it wasn’t obvious to me either, if it comes up a third time I’ll see about extraction possibilities

PR #2017, opened 2021-08-26
Add prepaid-card selection to create merchant workflow

burieberry:
CS-1596

- Update milestones in merchant creation workflow to include prepaid card selection
- Do not pre-select prepaid card
- Disable CTA before card is selected
- Display prepaid card options in dropdown
- Default, in-progress, and memorialized states UI
- Styling and tests
- CardPicker component

Note: The background image on the prepaid cards flicker while scrolling in the dropdown menu but I'm not sure how that can be fixed. Also sometimes when I click on the dropdown trigger, the first card gets selected immediately.

select-card

selected-card

change-card


packages/web-client/app/components/card-pay/create-merchant-workflow/prepaid-card-choice/index.hbs
```diff
@@ -27,7 +28,29 @@
...
+          <CardPay::CardPicker::CardOption
+            @card={{this.selectedPrepaidCard}}
+            data-test-prepaid-card-choice-selected-card
+          />
+          {{#if (and (not @isComplete) (not-eq this.creationState "in-progress"))}}
+            <CardPay::CardPicker
+              @cards={{this.prepaidCards}}
+              @chooseCard={{this.choosePrepaidCard}}
+              @selectedCard={{this.selectedPrepaidCard}}
+              @selectedCardAddress={{this.selectedPrepaidCard.address}}
```
backspace, line 42: Just looking at this invocation and how the component is defined, it looks like `selectedCardAddress` could just be removed and accessed in the component as `@selectedCard.address`? 🤔

packages/web-client/app/components/card-pay/create-merchant-workflow/prepaid-card-choice/index.hbs
```diff
@@ -27,7 +28,29 @@
...
     </CardPay::LabeledValue>
     <CardPay::LabeledValue @label="Pay Using Prepaid Card" class="prepaid-card-choice__field-label">
-      {{!-- TODO: card picker --}}
+      {{#if this.selectedPrepaidCard}}
+        <div class="prepaid-card-choice__change-card">
+          <CardPay::CardPicker::CardOption
+            @card={{this.selectedPrepaidCard}}
+            data-test-prepaid-card-choice-selected-card
+          />
+          {{#if (and (not @isComplete) (not-eq this.creationState "in-progress"))}}
```
backspace, line 37: There’s no real standard on this but to me this is close to the point where it makes sense to extract as a getter in the component, something like `{{#if this.showSelectedPrepaidCard}}`

packages/web-client/app/components/card-pay/create-merchant-workflow/prepaid-card-choice/index.ts
```diff
@@ -58,7 +60,21 @@ export default class CardPayCreateMerchantWorkflowPrepaidCardChoiceComponent ext
...
+  }
+
+  @action choosePrepaidCard(card: PrepaidCardSafe) {
+    this.selectedPrepaidCard = card;
+  }
+
   @action createMerchant() {
+    if (this.isCtaDisabled) {
+      return;
+    }
```
backspace, line 77: They shouldn’t be able to trigger this action if the button is disabled, is there another reason to have this guard? 😯

packages/web-client/app/components/card-pay/create-merchant-workflow/prepaid-card-choice/index.ts
```diff
@@ -58,7 +60,21 @@ export default class CardPayCreateMerchantWorkflowPrepaidCardChoiceComponent ext
     this.merchantRegistrationFee = yield this.layer2Network.strategy.merchantRegistrationFee();
   }
 
+  get prepaidCards() {
+    return this.layer2Network.safes.value.filterBy(
+      'type',
+      'prepaid-card'
+    ) as PrepaidCardSafe[];
+  }
```
backspace, line 68: Should this exclude prepaid cards whose balance is less than the required amount of SPEND? ([This PR](https://github.com/cardstack/cardstack/pull/2027) should help with that)

packages/web-client/tests/integration/components/card-pay/create-merchant-workflow/prepaid-card-choice-test.ts
```diff
@@ -84,26 +103,115 @@ module(
...
+          `[data-test-prepaid-card-choice-selected-card] [data-test-prepaid-card="${prepaidCardAddress2}"]`
+        )
+        .exists();
+      assert
+        .dom(`[data-test-prepaid-card-choice-selected-card]`)
+        .containsText(prepaidCardAddress2);
+      assert
+        .dom(`[data-test-prepaid-card-choice-selected-card]`)
+        .containsText('500 SPEND');
+    });
```
backspace, line 188: Nice tests! One more I’d like to see is something that asserts that `registerMerchant` is called with the address of the chosen card; you can see an example of that kind of assertion [here](https://github.com/cardstack/cardstack/blob/4b99e277936acd394149cb9dc52580f6d3dd987a/packages/web-client/tests/integration/components/card-pay/deposit-workflow/transaction-amount-test.ts#L77-L99), elsewhere in this file there are tests that stub the service call.

It’s true that the acceptance test verifies this implicitly but I think it would be nice to see an explicit verification 🤓

backspace (changes requested):
I find this fun to use, which is always a good sign. I did encounter the first-card-is-automatically-selected problem but I think that could be documented and worked on separately. I made some suggestions but the only thing I consider a requirement before merging is asserting on the correct address being passed to `registerMerchant`. Nice work 🤩

backspace:
#2029 includes some new assertions that reference card positions that will shift as part of the new milestone in this PR, so I expect some breakage 😯

packages/web-client/app/components/card-pay/card-picker/index.css
```diff
@@ -0,0 +1,91 @@
...
+  max-height: var(--options-menu-height);
+  background-color: var(--boxel-light);
+  border: 1px solid rgba(0, 0, 0, 0.15);
+  border-radius: var(--boxel-border-radius);
+  box-shadow: 0 15px 30px rgba(0, 0, 0, 0.25);
+  overflow-y: auto;
+}
+
+.card-picker-dropdown .ember-basic-dropdown-content.ember-basic-dropdown--transitioning-in {
+  animation: drop-fade-below var(--boxel-transition);
```
Aierie, line 57: I think the animations are why the first card is autoselected. The autoselection happens only when you click at the bottom of the dropdown (~20px). If you change `drop-fade-below` to start at `-100px` instead, you will have the autoselect throughout the whole dropdown trigger.

The quick fix for this (and I think it's ok while animating), is to do `pointer-events: none` while the dropdown is opening.

Aierie: If this isn't great, there might be some PRs in ember-power-select that solve this, because others have faced similar issues.

Aierie (approved):
I don't have much to add to what @backspace has gone through, this works great!

Aierie:
Also had a look at the flicker issue, and it:

1. Happens on Chrome but not Safari or Firefox for me
2. Looks like it could be a very deep rabbit hole. Many solutions where people don't understand why or how a fix works, many fixes that may or may not work depending on the case.

packages/web-client/app/components/card-pay/create-merchant-workflow/prepaid-card-choice/index.ts
burieberry: > They shouldn’t be able to trigger this action if the button is disabled, is there another reason to have this guard? 😯

Typescript gives an error

packages/web-client/app/components/card-pay/card-picker/index.css
burieberry: I tried the above suggestions, but still seeing the autoselection happen sometimes.

packages/web-client/app/components/card-pay/create-merchant-workflow/prepaid-card-choice/index.ts
burieberry: We will need to fade them out / make them not selectable

packages/web-client/tests/integration/components/card-pay/create-merchant-workflow/prepaid-card-choice-test.ts
burieberry: Added this (starts at line 246 below)

packages/web-client/app/components/card-pay/card-picker/index.hbs
```diff
@@ -0,0 +1,20 @@
...
+as |card|>
+  <CardPay::CardPicker::CardOption
+    class={{cn
+      "card-picker-dropdown__option"
+      card-picker-dropdown__option--selected=(eq card.address @selectedCardAddress)
+    }}
+    @card={{card}}
+    data-test-card-picker-dropdown-option={{card.address}}
+  />
+</PowerSelect>
```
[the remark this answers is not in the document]
burieberry: Not sure perfectly understanding this, but feel free add a ticket about it 👍

packages/web-client/app/components/card-pay/card-picker/index.css
Aierie: I'm finding it hard to reproduce with the current change. Maybe you can create an issue for this?

packages/web-client/tests/integration/components/card-pay/create-merchant-workflow/prepaid-card-choice-test.ts
backspace: excellent, thanks!

PR #2018, opened 2021-08-27
Merchant payment landing page payment request card

Aierie:
This PR adds UI that a user sees when a merchant requests payment from them via Card Pay.

There is a fair bit of CSS here that is taken from various components within the repo, including:

- `CardPay::Merchant`
- `CardPay::BalanceDisplay`
- `ActionCardContainer`

Did not modify these components and opted to pluck out the styles because of subtle differences + the need for responsive design that I'm not sure we want to apply the same way in our workflows.

## Full width (720px), non-deep-linkable view
image

## 360px (breakpoint for media query for more compact view), deep-linkable view
image

image


lukemelia:
This is more of a design question, but is there any point to having the QR code behind a toggle rather than showing both via scroll?

lukemelia:
What is the plan regarding adding routing for this and adding tests?

Aierie:
Discussed w @lukemelia re: CS-1652, will focus on answering design questions for this ticket as there are some things tbd re: hosting + routing. There are no plans for this branch to be merged into main, but leaving it up as a way to discuss design + implementation issues.

burieberry:
> This is more of a design question, but is there any point to having the QR code behind a toggle rather than showing both via scroll?

This is the page that opens in the browser after the user scans the merchant payment code with their phones or maybe they opened that link some other way. The main view of this page is the responsive mobile one with the button, which will lead the user to the Card Wallet app to complete the payment.

It's also possible that the user opened the payment link on their desktop, in that case they'll need to scan the QR code to open it on their phones to proceed with the payment.

Aierie:
Since we've decided to have this page on the DApp, this work should be merged into the web client's codebase. This PR will only cover the card, the route + template will follow in a separate one.

packages/web-client/tests/integration/components/card-pay/merchant-payment-request-card-test.ts
```diff
@@ -0,0 +1,119 @@
...
+    });
+
+    test('It renders the non-deep-link view correctly', async function (assert) {
+      await render(hbs`
+        <CardPay::MerchantPaymentRequestCard
+          @amount={{this.amount}}
+          @usdAmount={{this.usdAmount}}
+          @merchant={{this.merchant}}
+          @paymentURL={{this.paymentURL}}
+          @canDeepLink={{false}}
```
Aierie, line 49: Using an argument rather than a media query to determine whether to show the "Pay Merchant" link, because this depends more on platform than on things that are easily detected via media query.

packages/web-client/app/components/card-pay/merchant-payment-request-card/payment-request/index.hbs
```diff
@@ -0,0 +1,95 @@
...
+      </span>
+
+      <button
+        class="payment-request__link-view-toggle"
+        type="button"
+        {{on "click" (set this "showAsQR" (not this.showAsQR))}}
+        data-test-payment-request-link-view-toggle
+      >
+        {{#if this.showAsQR}}
+          Show Payment Link
```
Aierie, line 70: Does anyone have any thoughts on what the copy for this should be? "Show Payment Link" is potentially confusing, I think, since the payment link is already shown (`.payment-request__url`).

packages/web-client/app/components/card-pay/merchant-payment-request-card/payment-request/index.hbs
```diff
@@ -0,0 +1,95 @@
...
+        <Boxel::Button
+          class="payment-request__deep-link"
+          @as="anchor"
+          @kind="primary"
+          @size="touch"
+          href={{@paymentURL}}
+          data-test-payment-request-deep-link
+        >
+          Pay Merchant
+        </Boxel::Button>
```
Aierie, line 57: Both of these use the `@paymentURL` right now, am unsure if the clicked deep link is different from the QR code.

packages/web-client/app/components/card-pay/merchant-payment-request-card/index.hbs
```diff
@@ -0,0 +1,32 @@
...
+    <a
+      class="merchant-payment-request-card__testflight-link"
+      href={{this.testFlightLink}}
+      target="_blank"
+      rel="noopener noreferrer"
+      data-test-payment-request-beta-access-link
+    >
+        Download TestFlight
+        {{svg-jar "caret-right-large" class="merchant-payment-request-card__testflight-link-icon"}}
+    </a>
```
backspace, line 21: Could be done separately but what do you think of extracting this to a component that renders differently based on the environment? As I expect this will be replaced with an App Store button when out of beta, but the beta would still need to link to TestFlight

packages/web-client/app/components/card-pay/merchant-payment-request-card/payment-request/index.css
```diff
@@ -0,0 +1,89 @@
...
+}
+
+.payment-request__link-view-toggle {
+  appearance: none;
+  -webkit-appearance: none;
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  padding: var(--boxel-sp-xs);
+  border-radius: 100px; /* same as Boxel button border radius */
```
backspace, line 79: Could this be extracted into a variable for consumption here and elsewhere?

packages/web-client/app/components/card-pay/merchant-payment-request-card/payment-request/index.hbs
backspace: “Show as Payment Button”? maybe meh

backspace (approved):

packages/web-client/public/images/icons/qr-code.svg
```diff
@@ -0,0 +1,31 @@
...
+      <path d="M66.667,73.334v2.627h1.051V74.385h1.576v1.576h1.051V73.334Z" transform="translate(-57.734 -63.35)" />
+      <path d="M40,10v1.051h2.1v2.627h1.051V10Z" transform="translate(-35.271 -10)" />
+      <path d="M42.1,40v2.1H40v3.153h2.1v2.627h1.051V44.2h-2.1V43.153h2.1v-2.1H44.2V42.1h1.051V40Z"
+        transform="translate(-35.271 -35.271)" />
+      <rect width="1.051" height="1.576" transform="translate(4.729 11.034)" />
+      <rect width="1.576" height="1.051" transform="translate(2.102 6.831)" />
+      <path d="M14.729,23.333V25.96H10v3.153h1.051v-2.1H15.78V23.333Z" transform="translate(-10 -21.231)" />
+    </g>
+  </g>
+</svg>
```
jurgenwerk, line 31: I think it would be good to feed this into a svg minimizer.

jurgenwerk (approved):

packages/web-client/app/components/card-pay/merchant-payment-request-card/index.hbs
Aierie: I think extracting this early does help with limiting the surface area we need to change when we need a different rendering, if we are disciplined in how we use it and specify that it is for that purpose only.

packages/web-client/app/components/card-pay/merchant-payment-request-card/payment-request/index.hbs
Aierie: Could use that! Will ask about this when I post a video for design review.

PR #2020, opened 2021-08-27
web: Add slug uniqueness check for create merchant

backspace:
This adds a Mirage handler for the uniqueness check endpoint that
checks the Mirage schema for an existing `MerchantInfo`. It also
includes a message for when the validation call fails, with a
placeholder string that’s easily replaced:
> There was an error validating merchant ID uniqueness

### DIFF
--- a/packages/web-client/app/services/merchant-info.ts
+++ b/packages/web-client/app/services/merchant-info.ts
@@ -15,6 +15,10 @@ interface PersistMerchantInfoTaskParams {
   textColor: string;
 }
 
+interface CheckMerchantSlugUniquenessTaskParams {
+  slug: string;
+}
+
 export default class MerchantInfoService extends Service {
   @service declare hubAuthentication: HubAuthentication;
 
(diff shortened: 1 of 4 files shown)


packages/web-client/app/services/merchant-info.ts
```diff
@@ -67,6 +71,30 @@ export default class MerchantInfoService extends Service {
...
+      `${config.hubURL}/api/merchant-infos/validate-slug/${params.slug}`,
+      {
+        method: 'GET',
+        headers: {
+          Authorization: 'Bearer: ' + this.hubAuthentication.authToken,
+          Accept: 'application/vnd.api+json',
+          'Content-Type': 'application/vnd.api+json',
+        },
+      }
+    );
```
Aierie, line 88: We might need to do a try catch + clear the hub auth token if stuff fails here too. This pattern is now pretty common across the DApp, wonder if we can a good way to abstract it.

Aierie (approved):
This works nicely!

packages/web-client/app/services/merchant-info.ts
backspace: ya could be something like a `fetch` wrapper in the hub service, akin to [this](https://github.com/hashicorp/nomad/blob/86fca8f513e6b2a0d9a6adaf2c473465516ec4f6/ui/app/services/token.js#L95-L101)

PR #2023, opened 2021-08-30
make merchant creation fee in spend an sdk constant

Aierie:
This PR moves the constant for merchant creation fee into the SDK.


backspace:
How does this relate to [`merchantRegistrationFee`](https://github.com/cardstack/cardstack/blob/683595bbd6294be3adb42478f3ab6590bcfe6fc5/packages/web-client/app/services/layer2-network.ts#L138)? 😯

Aierie:
That seems to make more sense, will update 

> How does this relate to [`merchantRegistrationFee`](https://github.com/cardstack/cardstack/blob/683595bbd6294be3adb42478f3ab6590bcfe6fc5/packages/web-client/app/services/layer2-network.ts#L138)? 😯

That seems to make more sense, will close this PR to rework the cancelation to use that

habdelra:
This value is actually coming from on chain--please do not make a 2nd copy of this in the SDK, We should query the protocol to get this value

habdelra:
Specifically is is available on chain here: https://github.com/cardstack/card-protocol-xdai/blob/main/contracts/RevenuePool.sol#L27

PR #2024, opened 2021-08-30
web: Add prepaid refetch after merchant creation

backspace:


lukemelia (approved):
It would be nice to have this be reactive in some way, but this works in the absence of that 👍

backspace:
indeed, this is the now second place where a manual fetch happens, I wonder whether we could have wallet events or something in the SDK that we can subscribe to, as it seems like the SDK would know when on-chain operations will cause safes to change. There’s an issue for that [here](https://linear.app/cardstack/issue/CS-1670/feature-request-can-we-subscribe-to-card-wallet-account-events-via-the).

PR #2028, opened 2021-08-31
Prepaid card workflow persistence

jurgenwerk:
This PR adds a persistence feature to prepaid card workflow:

1. It restores the state when a user reloads the window in the middle of the workflow (completed cards)
2. It restores the state of a finished workflow (for inspection and to keep track of completed workflows)
3. It restores cancelled workflows
4. It prevents restoration in case the user is disconnected from the wallet or is unauthenticated or has changed the wallet address

The way it works is that when a user clicks on the "Issue prepaid card", the url becomes `some-domain.com/card-pay/balances?flow=issue-prepaid-card&flow-id=idDtMNYqfeAcj6kpoiBtkZ`. Notice the `flow-id` param - it maps to `workflowPersistenceId` which is used to restore and persist the workflow session state, which is persisted in `localStorage`:

image

Then, when a workflow is initialized, and we can find its persisted state, we check which cards have been completed and then mark them as such; the completed cards will be immediately rendered using the restored session state and next cards will be offered. 

To support this, a couple of components have been adjusted to read from the workflow session state rather than rely on their local state to get their data. 

In case you review PRs commit by commit - there's quite a few of commits here and some of them partly overwrite the previous ones, which makes the review a bit more difficult, but luckily not too much. I wanted to rebase/squash more aggressively but other authors have contributed code in the middle so I decided to keep that history.

### DIFF
--- a/packages/cardpay-sdk/sdk/prepaid-card/base.ts
+++ b/packages/cardpay-sdk/sdk/prepaid-card/base.ts
@@ -592,12 +592,6 @@ export default class PrepaidCard {
     };
   }
 
-  async getPrepaidCardSafesFromTxHash(txHash: string): Promise<PrepaidCardSafe[]> {
-    let prepaidCardAddresses = await this.getPrepaidCardsFromTxn(txHash);
-    await waitUntilTransactionMined(this.layer2Web3, txHash);
-    return await this.resolvePrepaidCards(prepaidCardAddresses);
-  }
-
   async convertFromSpendForPrepaidCard(
     prepaidCardAddress: string,
     minimumSpendBalance: number,
--- a/packages/cardpay-sdk/sdk/prepaid-card/base.ts
+++ b/packages/cardpay-sdk/sdk/prepaid-card/base.ts
@@ -592,6 +592,12 @@ export default class PrepaidCard {
     };
   }
 
+  async getPrepaidCardSafesFromTxHash(txHash: string): Promise<PrepaidCardSafe[]> {
+    let prepaidCardAddresses = await this.getPrepaidCardsFromTxn(txHash);
+    await waitUntilTransactionMined(this.layer2Web3, txHash);
+    return await this.resolvePrepaidCards(prepaidCardAddresses);
+  }
+
   async convertFromSpendForPrepaidCard(
     prepaidCardAddress: string,
     minimumSpendBalance: number,
--- a/packages/web-client/app/components/card-pay/issue-prepaid-card-workflow/index.ts
+++ b/packages/web-client/app/components/card-pay/issue-prepaid-card-workflow/index.ts
@@ -20,6 +20,7 @@ import RouterService from '@ember/routing/router-service';
 import { faceValueOptions } from './workflow-config';
 import { currentNetworkDisplayInfo as c } from '@cardstack/web-client/utils/web3-strategies/network-display-info';
 import HubAuthentication from '@cardstack/web-client/services/hub-authentication';
+import { next } from '@ember/runloop';
 
 const FAILURE_REASONS = {
   DISCONNECTED: 'DISCONNECTED',
@@ -377,7 +378,9 @@ class IssuePrepaidCardWorkflowComponent extends Component {
       const errors = workflow.restorationErrors(persistedState);
 
       if (errors.length > 0) {
-        workflow.cancel(errors[0]);
+        next(this, () => {
+          workflow.cancel(errors[0]);
+        });
       } else {
         workflow.restoreFromPersistedWorkflow();
       }
--- a/packages/web-client/app/components/card-pay/issue-prepaid-card-workflow/index.ts
+++ b/packages/web-client/app/components/card-pay/issue-prepaid-card-workflow/index.ts
@@ -272,7 +272,9 @@ class IssuePrepaidCardWorkflow extends Workflow {
   restoreFromPersistedWorkflow() {
     this.session.restoreFromStorage();
 
-    const lastCompletedCardName = this.session.state.lastCompletedCardName;
+    const [lastCompletedCardName] = (
+      this.session.state.completedCardNames || []
+    ).slice(-1);
 
     // TODO: Only attempt to restore when both layer 1 and 2 are connected
     if (lastCompletedCardName) {
--- a/packages/web-client/app/components/card-pay/issue-prepaid-card-workflow/index.ts
+++ b/packages/web-client/app/components/card-pay/issue-prepaid-card-workflow/index.ts
@@ -62,6 +62,7 @@ class IssuePrepaidCardWorkflow extends Workflow {
         }),
         new WorkflowCard({
           author: cardbot,
+          cardName: 'LAYER2_CONNECT',
           componentName: 'card-pay/layer-two-connect-card',
           async check() {
             let layer2Network = this.workflow?.owner.lookup(
--- a/packages/web-client/app/components/card-pay/issue-prepaid-card-workflow/index.ts
+++ b/packages/web-client/app/components/card-pay/issue-prepaid-card-workflow/index.ts
@@ -2,7 +2,11 @@ import Component from '@glimmer/component';
 import { getOwner } from '@ember/application';
 import { inject as service } from '@ember/service';
 import { WorkflowMessage } from '@cardstack/web-client/models/workflow/workflow-message';
-import { Workflow, cardbot } from '@cardstack/web-client/models/workflow';
+import {
+  Workflow,
+  cardbot,
+  WorkflowName,
+} from '@cardstack/web-client/models/workflow';
 import { Milestone } from '@cardstack/web-client/models/workflow/milestone';
 import { WorkflowCard } from '@cardstack/web-client/models/workflow/workflow-card';
 import PostableCollection from '@cardstack/web-client/models/workflow/postable-collection';
@@ -25,7 +29,8 @@ const FAILURE_REASONS = {
 
 class IssuePrepaidCardWorkflow extends Workflow {
   workflowPersistenceId?: string;
-  name = 'Prepaid Card Issuance';
+  name = 'PREPAID_CARD_ISSUANCE' as WorkflowName;
+
   milestones = [
     new Milestone({
       title: `Connect ${c.layer2.fullName} wallet`,
--- a/packages/web-client/app/components/card-pay/issue-prepaid-card-workflow/index.ts
+++ b/packages/web-client/app/components/card-pay/issue-prepaid-card-workflow/index.ts
@@ -276,6 +276,7 @@ class IssuePrepaidCardWorkflow extends Workflow {
 
     // TODO: Only attempt to restore when both layer 1 and 2 are connected
     if (lastCompletedCardName) {
+      this.isRestored = true;
       const postables = this.milestones
         .flatMap((m: any) => m.postableCollection.postables)
         .concat(this.epilogue.postables);
--- a/packages/web-client/app/components/card-pay/issue-prepaid-card-workflow/index.ts
+++ b/packages/web-client/app/components/card-pay/issue-prepaid-card-workflow/index.ts
@@ -9,8 +9,10 @@ import PostableCollection from '@cardstack/web-client/models/workflow/postable-c
 import NetworkAwareWorkflowCard from '@cardstack/web-client/components/workflow-thread/network-aware-card';
 import NetworkAwareWorkflowMessage from '@cardstack/web-client/components/workflow-thread/network-aware-message';
 import Layer2Network from '@cardstack/web-client/services/layer2-network';
+import WorkflowPersistence from '@cardstack/web-client/services/workflow-persistence';
 import { action } from '@ember/object';
 import BN from 'bn.js';
+import RouterService from '@ember/routing/router-service';
 
 import { faceValueOptions } from './workflow-config';
 import { currentNetworkDisplayInfo as c } from '@cardstack/web-client/utils/web3-strategies/network-display-info';
@@ -22,6 +24,7 @@ const FAILURE_REASONS = {
 } as const;
 
 class IssuePrepaidCardWorkflow extends Workflow {
+  workflowPersistenceId?: string;
   name = 'Prepaid Card Issuance';
   milestones = [
     new Milestone({
@@ -258,19 +261,51 @@ class IssuePrepaidCardWorkflow extends Workflow {
     }),
   ]);
 
-  constructor(owner: unknown) {
+  constructor(owner: unknown, workflowPersistenceId: string) {
     super(owner);
+    this.workflowPersistenceId = workflowPersistenceId;
+
     this.attachWorkflow();
+  }
+
+  restoreFromPersistedWorkflow() {
+    this.session.restoreFromStorage();
+
+    const lastCompletedCardName = this.session.state.lastCompletedCardName;
+
+    // TODO: Only attempt to restore when both layer 1 and 2 are connected
+    if (lastCompletedCardName) {
+      const postables = this.milestones
+        .flatMap((m: any) => m.postableCollection.postables)
+        .concat(this.epilogue.postables);
+
+      const index = postables.mapBy('cardName').indexOf(lastCompletedCardName);
+
+      postables.slice(0, index + 1).forEach((p: any) => {
+        p.isComplete = true;
+      });
+    }
   }
 }
 
 class IssuePrepaidCardWorkflowComponent extends Component {
   @service declare layer2Network: Layer2Network;
+  @service declare workflowPersistence: WorkflowPersistence;
+  @service declare router: RouterService;
 
   workflow!: IssuePrepaidCardWorkflow;
+
   constructor(owner: unknown, args: {}) {
     super(owner, args);
-    this.workflow = new IssuePrepaidCardWorkflow(getOwner(this));
+
+    const workflow = new IssuePrepaidCardWorkflow(
+      getOwner(this),
+      this.router.currentRoute.queryParams.workflowPersistenceId!
+    );
+
+    workflow.restoreFromPersistedWorkflow();
+
+    this.workflow = workflow;
   }
 
   @action onDisconnect() {
--- a/packages/web-client/app/components/card-pay/issue-prepaid-card-workflow/index.ts
+++ b/packages/web-client/app/components/card-pay/issue-prepaid-card-workflow/index.ts
@@ -305,11 +305,13 @@ class IssuePrepaidCardWorkflow extends Workflow {
       author: cardbot,
       componentName: 'workflow-thread/default-cancelation-cta',
       includeIf() {
-        return ([
-          FAILURE_REASONS.RESTORATION_UNAUTHENTICATED,
-          FAILURE_REASONS.RESTORATION_L2_ADDRESS_CHANGED,
-          FAILURE_REASONS.RESTORATION_L2_DISCONNECTED,
-        ] as String[]).includes(String(this.workflow?.cancelationReason));
+        return (
+          [
+            FAILURE_REASONS.RESTORATION_UNAUTHENTICATED,
+            FAILURE_REASONS.RESTORATION_L2_ADDRESS_CHANGED,
+            FAILURE_REASONS.RESTORATION_L2_DISCONNECTED,
+          ] as String[]
+        ).includes(String(this.workflow?.cancelationReason));
       },
     }),
   ]);
--- a/packages/web-client/app/components/card-pay/issue-prepaid-card-workflow/index.ts
+++ b/packages/web-client/app/components/card-pay/issue-prepaid-card-workflow/index.ts
@@ -17,18 +17,21 @@ import WorkflowPersistence from '@cardstack/web-client/services/workflow-persist
 import { action } from '@ember/object';
 import BN from 'bn.js';
 import RouterService from '@ember/routing/router-service';
-
 import { faceValueOptions } from './workflow-config';
 import { currentNetworkDisplayInfo as c } from '@cardstack/web-client/utils/web3-strategies/network-display-info';
+import HubAuthentication from '@cardstack/web-client/services/hub-authentication';
 
 const FAILURE_REASONS = {
   DISCONNECTED: 'DISCONNECTED',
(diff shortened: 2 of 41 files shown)


packages/web-client/app/components/card-pay/issue-prepaid-card-workflow/index.ts
```diff
@@ -101,6 +101,7 @@ class IssuePrepaidCardWorkflow extends Workflow {
           },
         }),
         new NetworkAwareWorkflowCard({
+          cardName: 'HUB_AUTH',
```
jurgenwerk, line 107: I added these names to all cards so that we are able to reference the last completed card. I was thinking about persisting the `componentName` but I figured there will be problems in case someone changes the component name.

packages/web-client/app/routes/card-pay/balances.ts
```diff
@@ -43,6 +44,26 @@ const BALANCES_PANEL = {
 };
 
 export default class CardPayBalancesRoute extends Route {
+  queryParams = {
+    flow: {
+      refreshModel: true,
+    },
+  };
+
+  beforeModel(transition: any) {
```
jurgenwerk, line 53: Adds a fresh `workflowPersistenceId` in case one isn't present yet.

packages/web-client/app/templates/card-pay/balances.hbs
```diff
@@ -8,10 +8,12 @@
           <CardPay::DashboardPanel::Section @section={{section}}>
             <:cta>
               <Boxel::Button
-                {{on "click" (set this "flow" section.workflow)}}
                 @kind="primary"
                 @size="touch"
                 @disabled={{section.isCtaDisabled}}
+                @as="link-to"
```
jurgenwerk, line 14: I figured it would be cleaner to use `LinkTo`s (now when there's more query params) and deal with query params in route hooks.

packages/web-client/app/routes/card-pay/balances.ts
```diff
@@ -43,6 +44,26 @@ const BALANCES_PANEL = {
...
+    const { flow, workflowPersistenceId } = transition.to.queryParams;
+
+    if (flow && !workflowPersistenceId) {
+      transition.abort(); // https://github.com/emberjs/ember.js/issues/17118
+      this.transitionTo(transition.to.name, {
+        queryParams: {
+          flow: flow,
+          workflowPersistenceId: uuidv4(), // TODO: Something shorter for a prettier URL?
```
jurgenwerk, line 61: `http://localhost:4200/card-pay/balances?flow=issue-prepaid-card&workflowPersistenceId=ea310aea-c724-4c14-adf3-b9dd51329304` is a pretty long URL. Should we use a shorter param name and shorter hash?

jurgenwerk:
@lukemelia here's my work in progress - I'll likely still write some more comments but if you have time you can start looking at this. Thanks!

packages/web-client/app/templates/card-pay/balances.hbs
lukemelia: I think moving back to actions approach is likely better.

packages/web-client/app/routes/card-pay/balances.ts
lukemelia: How about `flow-id` for the query param for the and using [short-uuid](https://www.npmjs.com/package/short-uuid) for the ID?

lukemelia:
I have pushed changes to implement the requirement to render completed postables immediately when restoring rather than having them revealed with a delay.

packages/web-client/app/models/workflow/workflow-card.ts
```diff
@@ -58,6 +62,10 @@ export class WorkflowCard extends WorkflowPostable {
       this.workflow?.emit('visible-postables-will-change');
       this.isComplete = true;
     }
+
+    if (this.isComplete && this.cardName) {
+      this.workflow?.session.update('lastCompletedCardName', this.cardName);
+    }
```
lukemelia, line 66: This wasn't behaving properly when the `check` call above it succeeded asynchronously. I have pushed a commit for this PR to address this.

packages/web-client/app/services/workflow-persistence.ts
```diff
@@ -0,0 +1,23 @@
+import { default as Service } from '@ember/service';
+
+export default class WorkflowPersistence extends Service {
```
lukemelia, line 3: Perhaps we should consider test behavior for this that uses a mock localstorage?

packages/web-client/app/components/card-pay/issue-prepaid-card-workflow/index.ts
```diff
@@ -258,19 +261,51 @@ class IssuePrepaidCardWorkflow extends Workflow {
...
 
-  constructor(owner: unknown) {
+  constructor(owner: unknown, workflowPersistenceId: string) {
     super(owner);
+    this.workflowPersistenceId = workflowPersistenceId;
+
     this.attachWorkflow();
   }
+
+  restoreFromPersistedWorkflow() {
```
lukemelia, line 272: TODO: promote this to the Workflow model

packages/web-client/app/components/card-pay/issue-prepaid-card-workflow/index.ts
```diff
@@ -258,19 +261,51 @@ class IssuePrepaidCardWorkflow extends Workflow {
...
+        .flatMap((m: any) => m.postableCollection.postables)
+        .concat(this.epilogue.postables);
+
+      const index = postables.mapBy('cardName').indexOf(lastCompletedCardName);
+
+      postables.slice(0, index + 1).forEach((p: any) => {
+        p.isComplete = true;
+      });
+    }
+  }
```
lukemelia, line 290: This doesn't cover the case of canceled workflows yet.

packages/web-client/app/components/card-pay/issue-prepaid-card-workflow/index.ts
```diff
@@ -258,19 +261,51 @@ class IssuePrepaidCardWorkflow extends Workflow {
...
   constructor(owner: unknown, args: {}) {
     super(owner, args);
-    this.workflow = new IssuePrepaidCardWorkflow(getOwner(this));
+
+    const workflow = new IssuePrepaidCardWorkflow(
+      getOwner(this),
+      this.router.currentRoute.queryParams.workflowPersistenceId!
+    );
+
+    workflow.restoreFromPersistedWorkflow();
```
lukemelia, line 308: This should probably become something that is done as part of the constuctor.

packages/web-client/app/templates/card-pay/balances.hbs
jurgenwerk: Sure, will change back to use that.

packages/web-client/app/routes/card-pay/balances.ts
jurgenwerk: Good suggestion, will try that.

packages/web-client/app/models/workflow/workflow-card.ts
jurgenwerk: Great, thanks!!

packages/web-client/app/services/workflow-persistence.ts
jurgenwerk: Yeah, sounds good. Will consider this when I start writing tests.

packages/web-client/app/components/card-pay/issue-prepaid-card-workflow/index.ts
jurgenwerk: Forgot about these. Will add this

Aierie:
Thanks for working on this, I think this feature would be very nice to have! 

With this current PR, restoring a workflow may lead to unexpected states because previous assumptions may no longer be true. Eg. if the user is not connected or doesn't have enough balance to create a prepaid card, we should not allow them past the layer 2 connection card, which isn't addressed here. Another case is if the balance of the user is now lower than the face value selected, we should stop the restoration there.

Have had some small discussions w @backspace on how workflow cards make assumptions about state that was previously set and conditions that have to be met. I think prior to this being added, it would be good to add a standardized way for cards to declare their assumptions about `workflowSession`, `layer1Network`, and `layer2Network`, and check that they are true, so that we can check them when restoring a workflow.

jurgenwerk:
@Aierie thanks! That's a good point. We'll definitely need to implement some means of checking whether the workflow can be restored – I added this to the TODO.

packages/cardpay-sdk/sdk/prepaid-card/base.ts
```diff
@@ -530,6 +530,12 @@ export default class PrepaidCard {
     };
   }
 
+  async getPrepaidCardSafesFromTxHash(txHash: string): Promise<PrepaidCardSafe[]> {
```
lukemelia, line 533: As of cardpay-sdk 0.23.0, the PrepaidCard sdk's `create` method has an overlaps that can accept a single txnHash argument that does what this method does. When you have a chance, please rebase and update your code to use that approach.

jurgenwerk: Nice! Will do.

jurgenwerk:
@lukemelia @Aierie thanks for all your help here! ⭐️

I changed the PR description and promoted this from draft to a real PR. Can you please review it again and take it for a spin to see if you can find any bugs? Also @backspace @burieberry 🙏

packages/web-client/app/components/card-pay/issue-prepaid-card-workflow/preview/index.ts
```diff
@@ -61,53 +62,99 @@ export default class CardPayPrepaidCardWorkflowPreviewComponent extends Componen
...
+    super(owner, args);
+  }
+
+  @action checkForPendingTransaction() {
+    if (this.args.workflowSession.state.txnHash) {
+      taskFor(this.issueTask).perform();
+    }
+  }
+
+  // TODO: Refactor long method
```
lukemelia, line 78: This TODO seems like a good idea 😄

packages/web-client/app/controllers/card-pay/balances.ts
```diff
@@ -2,16 +2,31 @@ import Controller from '@ember/controller';
...
-  @tracked flow: 'deposit' | null = null;
+  queryParams = ['flow', { workflowPersistenceId: 'flow-id' }];
+  @tracked flow: string | null = null;
+  @tracked workflowPersistenceId: string | null = null;
 
   get prepaidCards() {
     return this.layer2Network.safes.value?.filterBy('type', 'prepaid-card');
   }
+
+  @action setFlow(flow: string) {
```
lukemelia, line 20: Maybe we should rename this method since I have changed the implementation?

packages/web-client/app/models/workflow/workflow-session.ts
```diff
@@ -1,14 +1,37 @@
...
+
 export default class WorkflowSession {
+  workflow: any;
+  workflowPersistenceStorage: WorkflowPersistence | undefined;
+
+  constructor(workflow?: any) {
+    this.workflow = workflow;
+
+    // Allow WorkflowSession to be instantiated as a POJO (for unit tests)
+    if (this.workflow && typeof this.workflow.owner.lookup === 'function') {
```
lukemelia, line 16: Is there a reason for this guard? Wouldn't the owner always have a lookup method?

packages/web-client/app/models/workflow/workflow-session.ts
```diff
@@ -1,14 +1,37 @@
...
+      this.workflowPersistenceStorage = this.workflow.owner.lookup(
+        'service:workflow-persistence'
+      );
```
lukemelia, line 19: Also, as I think I commented elsewhere, it would be better for the Workflow class to be responsible for looking up/injecting the service, and then have this class just read it off the workflow instance.

packages/web-client/tests/acceptance/issue-persisted-prepaid-card-test.ts
```diff
@@ -0,0 +1,391 @@
...
+import { BN } from 'bn.js';
+import { faceValueOptions } from '@cardstack/web-client/components/card-pay/issue-prepaid-card-workflow/workflow-config';
+import Layer2TestWeb3Strategy from '@cardstack/web-client/utils/web3-strategies/test-layer2';
+import { toWei } from 'web3-utils';
+
+import WorkflowPersistence from '@cardstack/web-client/services/workflow-persistence';
+
+interface Context extends MirageTestContext {}
+
+module('Acceptance | issue prepaid card', function (hooks) {
```
lukemelia, line 20:
```suggestion
module('Acceptance | issue persisted prepaid card', function (hooks) {
```

lukemelia (changes requested):
This came out very nicely!

backspace:
I don’t think I’ll be able to get a full review done today unfortunately; I tried it out locally and it’s pretty exciting! I encountered some things though:
* I closed the workflow, changed my wallet, and revisited the link with the persistence key, and it didn’t cancel the workflow as I believe is expected
* when I set `localStorage.authToken = ''` and revisited a workflow I got this error in the console: You attempted to update `state` on `WorkflowSession`, but it had already been used previously in the same computation.  Attempting to update a value after using it in a computation can cause logical errors, infinite revalidation bugs, and performance issues, and is not supported.`
* it did successfully cancel the workflow, but the Start Again button just looped indefinitely: auth-loop-screencast 2021-09-10 08-34-14

Nice work though 🤩

jurgenwerk:
@backspace thanks for doing such a detailed QA! Will take a look and fix these issues.

packages/web-client/app/models/workflow/workflow-session.ts
```diff
@@ -1,14 +1,28 @@
...
+
 export default class WorkflowSession {
+  workflow: any;
+  workflowPersistenceStorage: WorkflowPersistence;
+
+  constructor(workflow?: any) {
+    this.workflow = workflow;
+    this.workflowPersistenceStorage = this.workflow.owner.lookup(
+      'service:workflow-persistence'
+    );
```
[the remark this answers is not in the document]
jurgenwerk: Yeah I think it would make sense - but not sure we can inject the service using the decorator, will probably have to lookup manually on the owner.

packages/web-client/app/components/card-pay/issue-prepaid-card-workflow/preview/index.ts
jurgenwerk: Yep - will get to it 🙂

packages/web-client/app/controllers/card-pay/balances.ts
jurgenwerk: Good catch. I think `transitionToBalances` would be more suitable.

packages/web-client/app/models/workflow/workflow-session.ts
jurgenwerk: Yeah that makes total sense. Will adjust it.

packages/web-client/app/models/workflow/workflow-card.ts
```diff
@@ -19,44 +19,60 @@ type FailureCheckResult = {
...
+
+    if (this.isComplete && this.cardName) {
+      const existingCompletedCardNames =
+        this.session?.state.completedCardNames || [];
+
+      if (!existingCompletedCardNames.includes(this.cardName)) {
+        this.session?.updateMany({
+          completedCardNames: [...existingCompletedCardNames, this.cardName],
+          completedMilestonesCount: this.workflow?.completedMilestoneCount,
+          milestonesCount: this.workflow?.milestones.length,
```
Aierie, line 73: I can't find where we're using `completedMilestonesCount` and `milestonesCount`, what are we using them for?

Aierie (changes requested):
Overall this looks good! 

One issue that I think is blocking:

image

Reloading a workflow with a changed balance that is now less than the face value should invalidate the selected face value, but it currently allows the user to continue. Also, a balance that is less than the required amount to continue with the workflow will not trigger a cancelation, because we bypass the `check` method of the Layer 2 connect card and set `isComplete` to true directly.

Another issue (non-blocking I think) to consider is that the `completedCardNames` array is not updated in `onIncomplete`, so users who have rolled back part of their workflow choices and load a workflow from persistence/reload the page might be a bit surprised.

packages/web-client/app/utils/web3-strategies/test-layer2.ts
```diff
@@ -33,16 +29,16 @@ import { taskFor } from 'ember-concurrency-ts';
...
 
 interface RegisterMerchantRequest {
   deferred: RSVP.Deferred<MerchantSafe>;
   onTxnHash?: (txnHash: TransactionHash) => void;
-  onNonce?: (nonce: string) => void;
-  nonce?: string;
+  onNonce?: (nonce: BN) => void;
+  nonce?: BN;
   infoDID: string;
 }
```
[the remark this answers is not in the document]
Aierie: Unsure though, given that this is a test class after all 😅

Aierie:
> Also, a balance that is less than the required amount to continue with the workflow will not trigger a cancelation, because we bypass the check method of the Layer 2 connect card and set isComplete to true directly.

I think asynchronous checks and their relationship to `isComplete` is a problem we should find a solid, reusable solution for, since workflows like the withdrawal one also use them.

packages/web-client/app/models/workflow/workflow-card.ts
jurgenwerk: I added them for convenience - to support the progress display on unfinished workflows for when we add this list (check the design below). Do you think it would be more suitable to calculate progress by looking at the finished cards instead?

image

jurgenwerk:
@Aierie Thanks for finding these bugs! Regarding the minimum balance value required, I think it might be worth considering doing the `check` before setting the `isComplete` on restore. Will look into this!

packages/web-client/app/models/workflow/workflow-card.ts
Aierie: Ah I see, thanks! I'm not sure what the best way to do it is, in an ideal world I would like to have the milestones/completion percentage be derived state

Aierie:
> @Aierie I was thinking about adding the balance check before restoring and possible simplifications. I suppose the most proper way of doing this would be:
> 
> Before restoring:
> 
> 1. Check if the current balance is large enough
> 2. If not, cancel the workflow (or roll back to the face value selection and show the error message), else restore fully
> 
> If we go with this solution, we will need to add some means of differentiation between restoring finished and unfinished workflows as a whole.
> 
> For finished workflows, we should just show the state as it was at that time and not perform any balance validations, and do the balance check for unfinished ones. This means we need to add extra logic to see whether the workflow has been completed or not, and make the balance validation dependent on that.
> 
> At this point I'm wondering how bad it would be if we just let the user continue past the face value selection step while restoring an incomplete workflow even though their balance is too low for prepaid card issuance? In this case, when they would try to create the card at the next step, they would get the low balance error and a chance to restart the flow there. Since this last step directly follows the face value selection step, I think not much would be lost between these steps. Granted, the whole workflow would be cancelled at this point, but I'm wondering if it makes sense to spend development time to support this validation as mentioned above. What are your thoughts?

I think this isn't critical to ensuring the safety of users' assets for this workflow. The withdrawal workflow however needs need async validation of having enough balance (ETH balance of the layer 1 account), and consequences of bypassing the balance check might be more troublesome there. We could defer this till we get to the withdrawal workflow, and think about how to solve both of these issues together then.

jurgenwerk:
Agree, other workflows will need a stricter check between steps for restoration. I'm in favour of deferring this for now.

packages/web-client/app/models/workflow/workflow-card.ts
jurgenwerk: Sounds good. We'll evaluate then - I'll leave this in for now.

packages/web-client/app/models/workflow/workflow-session.ts
jurgenwerk: Added this condition because of `new ConcreteWorkflow({});` in tests (where lookup isn't a function on the `{}`). I think it's better to remove this guard and change all `new ConcreteWorkflow({})` to `new ConcreteWorkflow(this.owner)`

lukemelia: OK, this seems fine. There is a way we could refactor to make our container construct our workflow instances and allow injections to work. I can pair with you on it for a future PR.

packages/web-client/app/components/card-pay/issue-prepaid-card-workflow/preview/index.ts
```diff
@@ -75,7 +75,6 @@ export default class CardPayPrepaidCardWorkflowPreviewComponent extends Componen
     }
   }
 
-  // TODO: Refactor long method
```
jurgenwerk: On second thought - the method is mostly just a sequence of tasks so even if we split it up into smaller methods it will still be just that 😀. So I'm fine with this method even when it's a big longer than usual. I did simplify it a bit though.

jurgenwerk:
New update:

1. Fixed bugs reported by @backspace 
2. Smaller refactor of the workflow <-> persistence relationship
3. [Got rid of the](https://github.com/cardstack/cardstack/pull/2028/commits/f9dc70d87da849807a70cef42b41bef11d845596) "same computation" warnings. I think using `next` to address these kinds of issues can be more of a bandaid fix but currently I'm unsure how to do this more properly. I can spend more time on this but would also appreciate any help here 😌
4. Updated the persisted `completedCardNames` when user goes back to edit a previously completed card. ⁉️ A question that just occurred to me: when rolling back  `completedCardNames`, should we also nullify the state set by the completed cards that we remove in this case? 

The PR (there are  7 new commits) is good to review again.

lukemelia (approved):

Aierie (approved):
Nice work ! I would like to create a child issue in https://linear.app/cardstack/issue/CS-1590/simple-workflow-persistence-for-issue-prepaid-card-flow to capture the issue around allowing users to go through the workflow with insufficient balance, and link it to the withdrawal workflow. Do you think that makes sense @jurgenwerk ?

jurgenwerk:
@Aierie yes – will add that!

PR #2029, opened 2021-08-31
web: Fix merchant info repersistance after change

backspace:
This addresses a bug where if a user clicked the button to create a merchant
but then rejected the request in Card Wallet, if they then changed their
merchant details and tried again, the old details would be attached to the
merchant safe. This updates the merchant customisation card to delete
a persisted merchant DID from the session if the details have changed
so a new one will be persisted in the following card.

Dealing with the [orphaned DID](https://linear.app/cardstack/issue/CS-1696/card-customisation-and-merchant-info-dids-can-be-orphaned) is out of scope for this PR.

This might produce confusion if the user goes back to edit their details but
leaves the slug the same, as it will then be invalid because the old details
have already been persisted.

The steady increase in implicit state shared between workflow cards makes
me wonder whether something like XState would be helpful here.


lukemelia (approved):
🙌

Aierie (approved):

PR #2031, opened 2021-09-01
web: Convert depot to be derived from safes

backspace:
This is a work in progress, probably the balances computed
properties can be converted as well.

This seemed to me a useful precursor to allowing a merchant safe to fund a withdrawal, as then all eligible types of safe can be chosen from in the workflow.


backspace:
Closing in favour of #2032

PR #2032, opened 2021-09-01
Single source of truth for depot and other safes

lukemelia:
Likely an alternative to #2031

### DIFF
--- a/packages/web-client/app/resources/depot.ts
+++ b/packages/web-client/app/resources/depot.ts
@@ -1,0 +1,53 @@
+import { Resource } from 'ember-resources';
+import { DepotSafe, Safe } from '@cardstack/cardpay-sdk/sdk/safes';
+import { Layer2Web3Strategy } from '@cardstack/web-client/utils/web3-strategies/types';
+import BN from 'bn.js';
+import { taskFor, TaskFunction } from 'ember-concurrency-ts';
+import { reads } from 'macro-decorators';
+
+interface Args {
+  named: {
+    strategy: Layer2Web3Strategy;
+    walletAddress: string;
+  };
+}
+
+export class Depot extends Resource<Args> {
+  @reads('args.named.strategy.viewSafesTask')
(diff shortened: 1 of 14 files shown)


packages/web-client/app/resources/depot.ts
```diff
@@ -0,0 +1,53 @@
...
+    await taskFor(this.viewSafesTask).perform(this.args.named.walletAddress);
+  }
+
+  get value() {
+    let safes = taskFor(this.viewSafesTask).lastSuccessful?.value;
+    let depotSafes = (safes || []).filter(
+      (safe: Safe) => safe.type === 'depot'
+    ) as DepotSafe[];
+
+    let depot = depotSafes[depotSafes.length - 1] ?? null;
```
backspace, line 35: Is it concerning at all if there’s more than one depot?

packages/web-client/app/utils/web3-strategies/layer2-chain.ts
```diff
@@ -230,15 +226,17 @@ export default abstract class Layer2ChainWeb3Strategy
   }
 
   async refreshBalances() {
-    return taskFor(this.fetchDepotTask).perform();
+    return taskFor(this.viewSafesTask).perform();
```
backspace, line 229: Maybe current instances of `safes.fetch` can be replaced with a call to `refreshBalances`? To make that the standard interface.

backspace (approved):
I look forward to using this 🎉

packages/web-client/app/resources/depot.ts
lukemelia: Yes, but I am going to trust the rest of the system to ensure that does not happen.

packages/web-client/app/utils/web3-strategies/layer2-chain.ts
lukemelia: Good idea.

PR #2034, opened 2021-09-02
rename existing e-concurency tasks to use the task suffix

Aierie:
We're using the `-Task` suffix in most parts of the DApp to easily distinguish tasks from generic async functions. This PR renames remaining tasks that do not follow this convention.

Manually tried:
- Merchant creation happy path
- Deposit happy path
- Withdrawal happy path
- Prepaid card happy path
- Withdrawal balance check failure


jurgenwerk (approved):
Looks good

PR #2035, opened 2021-09-02
stricter checks for auth token + validity

Aierie:
This PR:
- makes the hub auth service (and by extension the hub auth card) check whether an existing auth token is valid instead of simply returning if there is one, when `HubAuthenticationService.ensureAuthenticated` is called
- makes the return type of `HubAuthenticationService.ensureAuthenticated` void because we don't use the returned value anywhere
- makes `HubAuthenticationService.isAuthenticated` false whenever the `authToken` setter receives a falsey value

### DIFF
--- a/packages/web-client/app/services/hub-authentication.ts
+++ b/packages/web-client/app/services/hub-authentication.ts
@@ -37,20 +37,19 @@ export default class HubAuthentication extends Service {
 
   @task *initializeTask(): TaskGenerator<void> {
     yield waitForProperty(this.layer2Network, 'isInitializing', false);
-    if (!this.authToken) return;
-    if (!this.layer2Network.isConnected) {
-      this.storage.removeItem('authToken');
-      return;
-    }
-    if (!this.authToken) return;
-    let isAuthValid = yield this.layer2Network.checkHubAuthenticationValid(
-      this.authToken
-    );
-    if (isAuthValid) {
+    if (yield this.hasValidAuthentication()) {
       this.isAuthenticated = true;
     } else {
-      this.storage.removeItem('authToken');
+      this.authToken = null;
     }
+  }
+
+  async hasValidAuthentication() {
+    return Boolean(
+      this.layer2Network.isConnected &&
+        this.authToken &&
+        (await this.layer2Network.checkHubAuthenticationValid(this.authToken))
+    );
   }
 
   get authToken(): string | null {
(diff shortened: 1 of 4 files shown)


Aierie:
Will add tests for the hub auth service before un-drafting

lukemelia (commented):
Looks good so far.

packages/web-client/app/services/hub-authentication.ts
```diff
@@ -37,41 +37,66 @@ export default class HubAuthentication extends Service {
...
+
+    try {
+      let newAuthToken = await this.layer2Network.authenticate();
+
+      if (newAuthToken) {
+        this.authToken = newAuthToken;
+        this.isAuthenticated = true;
+      } else {
+        this.authToken = null;
+        throw new Error('Failed to fetch auth token');
```
backspace, line 88: image

This is a welcome change, as now when Card Wallet and web client are out of sync it doesn’t show authentication success while actually storing the token as `undefined` 🎉

packages/web-client/app/services/hub-authentication.ts
```diff
@@ -37,41 +37,66 @@ export default class HubAuthentication extends Service {
...
+
   get authToken(): string | null {
     return this.storage.getItem('authToken');
   }
 
   set authToken(val: string | null) {
     if (val) {
+      // we're not setting isAuthenticated to true here because
+      // it is possible to mistakenly set authToken to an arbitrary truthy value
+      // in an external consumer
```
backspace, line 63: I wonder if we could use the [`storage` event](https://developer.mozilla.org/en-US/docs/Web/API/Window/storage_event) to react to direct manipulation of the token in `localStorage` 🤔

packages/web-client/tests/unit/services/hub-authentication-test.ts
```diff
@@ -0,0 +1,234 @@
...
+    // this is the condition for initializing with an authenticated state
+    // assumption made that layer2Service.checkHubAuthenticationValid returns Promise<true>
+    window.TEST__AUTH_TOKEN = HUB_AUTH_TOKEN;
+    layer2Service = this.owner.lookup('service:layer2-network').strategy;
+    layer2Service.test__simulateAccountsChanged([layer2AccountAddress]);
+  });
+
+  // Test initialization
+  test('it can initialize with an authenticated state', async function (assert) {
+    hubAuthentication = this.owner.lookup('service:hub-authentication');
```
backspace, line 26: Does it work to move this into the `beforeEach`? I see it in the tests, but maybe there’s some subtle interaction that prevents that

backspace (approved):
🎉

packages/web-client/app/services/hub-authentication.ts
Aierie: I was thinking about this too. I think it probably is good to listen for clearing/removal of the token from local storage.

packages/web-client/tests/unit/services/hub-authentication-test.ts
Aierie: There are some cases where we want to set `window.TEST__AUTH_TOKEN` before we initialize a hub authentication service instance, so it is a bit awkward to move it into a `beforeEach`. I could organize stuff into modules (with hub auth instantiated vs without), though I am not sure if that's overkill

packages/web-client/tests/unit/services/hub-authentication-test.ts
```diff
@@ -0,0 +1,234 @@
...
+    sinon.stub(layer2Service, 'authenticate').throws(new Error('A test error'));
+    sinon
+      .stub(layer2Service, 'checkHubAuthenticationValid')
+      .returns(Promise.resolve(false));
+    let error = null;
+    try {
+      await hubAuthentication.ensureAuthenticated();
+    } catch (e) {
+      error = e;
+    }
```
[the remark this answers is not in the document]
Aierie: I found this, but it didn't support async stuff. Reviewing the docs + github comments because of your comment made me find this though:

https://api.qunitjs.com/assert/rejects/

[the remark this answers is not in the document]
Aierie: Will update to use rejects

packages/web-client/tests/unit/services/hub-authentication-test.ts
backspace: ya I don’t love having to organise tests around setup requirements vs conceptual differences, and it doesn’t seem worth it to me for a one-line savings

PR #2037, opened 2021-09-02
Add apple-app-site-association file

lukemelia:
This is part of making universal links work.

### DIFF
--- a/packages/web-client/public/.well-known/apple-app-site-association
+++ b/packages/web-client/public/.well-known/apple-app-site-association
@@ -1,0 +1,10 @@
+{
+  "applinks": {
+    "details": {
+      "appIDs: [
+        "QS5AFH4668.com.cardstack.cardpay"
+      ],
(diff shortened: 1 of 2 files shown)


packages/web-client/public/.well-known/apple-app-site-association
```diff
@@ -0,0 +1,10 @@
+{
+  "applinks": {
+    "details": {
+      "appIDs: [
+        "QS5AFH4668.com.cardstack.cardpay"
+      ],
+      "paths" => [ "*" ]
```
danibonilha, line 7: is the syntax with arrow or colon ? Does it matter ?

packages/web-client/public/.well-known/apple-app-site-association
```diff
@@ -0,0 +1,10 @@
...
+    }
+  }
+}
```
danibonilha, line 10: EOF

PR #2038, opened 2021-09-02
web: Add merchant safes as withdrawal source options

backspace:
This updates the withdrawal workflow to support merchant safes as sources. While reward safes don’t yet exist in the API, this is designed with that future addition in mind.

The `choose-balance` card lists the depot and merchant safes as possible sources:

image

The `transaction-amount` card now shows the chosen safe and balance:

image

In contrast to my previous acceptance test approach that involves a somewhat [murky interface](https://github.com/cardstack/cardstack/blob/a0d9e4c8b030a333cab375ea6f5439b3e91550b3/packages/web-client/app/utils/web3-strategies/test-layer2.ts#L409), I added optional parameters for `test__simulateBridgedToLayer1` that cause it throw an error when no matching call to `bridgeToLayer1` was made. I’m open to suggestions on a better way to accomplish this, I just want some assurance that the correct values are getting passed along through the workflow and make it into the SDK call at the end.

### DIFF
--- a/packages/web-client/app/components/card-pay/safe-chooser-dropdown/safe-option/index.hbs
+++ b/packages/web-client/app/components/card-pay/safe-chooser-dropdown/safe-option/index.hbs
@@ -1,0 +1,20 @@
+<div class='safe-option' ...attributes>
+  {{#if this.data.icon}}
+    {{svg-jar this.data.icon width="20" height="20" class='safe-option__logo'}}
+  {{else}}
+    <CardPay::MerchantLogo
+      class='safe-option__logo'
+      style={{css-var merchant-margin-right=0}}
+      @name={{first-char this.data.info.name}}
+      @logoBackground={{this.data.info.backgroundColor}}
+      @logoTextColor={{this.data.info.textColor}}
+    />
+  {{/if}}
+  <div class='safe-option__name-and-type'>
+    <span class='safe-option__name'>{{this.data.info.name}}</span>
+    {{#if this.data.type}}
+      <span class='safe-option__type'>{{this.data.type}}</span>
+    {{/if}}
+  </div>
+  <div class='safe-option__address'>{{@safe.address}}</div>
+</div>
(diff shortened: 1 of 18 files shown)


Aierie (changes requested):
## UI
Current CSS truncates the dropdown contents that hang outside the card, leading to safes appearing missing:

image

When I set overflow: visible on the card:

image


## Merchant balances not shown
This merchant safe should have 200 SPEND backed by ~1.9 DAI, but it's not shown. I think merchant safe balances need to be fetched this way (seems to work on the CLI): https://github.com/cardstack/cardstack/blob/a0d9e4c8b030a333cab375ea6f5439b3e91550b3/packages/cardpay-cli/revenue-pool.ts#L25-L33, but would be good to get @habdelra 's confirmation

image

Aierie:
We may need to use `RevenuePool.claim` to withdraw from a merchant safe as well, I think

https://github.com/cardstack/cardstack/blob/a0d9e4c8b030a333cab375ea6f5439b3e91550b3/packages/cardpay-cli/revenue-pool.ts#L35-L54

packages/web-client/app/components/card-pay/safe-chooser-dropdown/safe-option/index.hbs
```diff
@@ -0,0 +1,20 @@
+<div class='safe-option' ...attributes>
+  {{#if this.data.icon}}
+    {{svg-jar this.data.icon width="20" height="20" class='safe-option__logo'}}
+  {{else}}
+    <CardPay::MerchantLogo
+      class='safe-option__logo'
+      @name={{first-char this.data.info.name}}
+      @marginRight={{"0"}}
+      @logoBackground={{this.data.info.backgroundColor}}
+      @logoTextColor={{this.data.info.textColor}}
```
lukemelia, line 10: I think these would probably be cleaner as just css vars rather than a mix of component props and css vars

lukemelia (approved):
👍  One comment, but not a blocker

packages/web-client/app/components/card-pay/safe-chooser-dropdown/safe-option/index.hbs
backspace: `@logoBackground` and `@logoTextColor` were already there but I’ve changed to using a CSS variable for the right margin, thanks.

backspace:
Thanks for the investigation, @Aierie. I’ll address the truncation problem in a followup PR.

PR #2040, opened 2021-09-03
web: Fix Power Select auto-choosing bug

backspace:
When I run this locally, it fixes the problem where the prepaid
card chooser in the merchant creation workflow looks like
it’s not opening but is really selecting the first card and
closing.

Thanks to @johanrd for documenting and @Aierie for finding:
https://github.com/cibernox/ember-power-select/issues/1408

### DIFF
--- a/packages/web-client/app/styles/app.css
+++ b/packages/web-client/app/styles/app.css
@@ -31,3 +31,11 @@ ol {
   margin: 0;
   padding: 0;
 }
+
+/*
+  Fixes a problem where Power Select would auto-choose the first item
+  https://github.com/cibernox/ember-power-select/issues/1408
+*/
+.ember-power-select-dropdown {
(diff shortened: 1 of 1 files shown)


backspace:
This does move the dropdown down but I think it’s worth it!

1px-screencast 2021-09-03 10-04-49

Aierie:
This fixes one part of the issue, I think another part is the dropdown taking up space and pushing itself under the mouse as it scrolls. Wondering if we should not allow the dropdown to take up space in that way.

Aierie:
single-click-select

This selection is done in a single click as the dropdown pushes itself up

Aierie (approved):

burieberry (approved):
It's still a bit funky for me but let's try it

backspace:
ya further revisions are still possible of course!

PR #2041, opened 2021-09-03
drop unused global lerna in ci

ef4:


ef4:
@lukemelia is there already a solution for testing this kind of change? When workflows are edited, they don't actually run because the underlying projects didn't change.

mattmcmanus (approved):
Nevermind Luke. We figured out an approach.

PR #2042, opened 2021-09-03
[CS-1495] - Update apple-app-site-association

danibonilha:
Apple changed the structure of `apple-app-site-association`, unfortunately it does not have any docs regarding this besides this [WWDC video](https://developer.apple.com/videos/play/wwdc2019/717/), the new format only supports iOS 13+, so in other to support previous versions we need to have both patterns.

`"apps": [],`, `"paths":`  and `"appID"` are for previous versions.
`"appIDs"` and `components` are for 13+


lukemelia (approved):
Let's give it a try!

PR #2044, opened 2021-09-03
move mocha out of top level

ef4:
In prep for merging the ef4-refactor branch we are cleaning up the top level of monorepo so that package-specific concerns live as much as possible in those packages.


lukemelia (approved):
Good summer cleaning!

PR #2046, opened 2021-09-03
checks must run when yarn.lock changes

ef4:
Changing the top-level yarn.lock can break any package. So it's necessary for each of these workflows to run when it changes.

This will make CI slower when changing dependencies, but there's really no finer-grained way to tell which packages need to be tested when yarn.lock changes.

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,7 +11,6 @@
   "dependencies": {
     "@actions/core": "^1.2.6",
     "@actions/github": "^4.0.0",
-    "@types/console-ui": "^2.2.3",
     "@types/filenamify-url": "^1.0.1",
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -11,7 +11,6 @@
   "dependencies": {
     "@actions/core": "^1.2.6",
     "@actions/github": "^4.0.0",
-    "@types/console-ui": "^2.2.3",
```
ef4, line 14: This was going to be a separate PR but snuck in here. As long as the tests passed, it's fine. This looks completely unused.

lukemelia (approved):
👍 Good catch!

PR #2049, opened 2021-09-03
Display merchants on merchant dashboard

burieberry:


packages/web-client/app/resources/merchant-info.ts
```diff
@@ -0,0 +1,100 @@
...
+  private async fetchMerchantInfo(
+    infoDID: string,
+    waitForInfo = false
+  ): Promise<void> {
+    let did = await this.#didResolver.resolve(infoDID);
+    let alsoKnownAs = did?.didDocument?.alsoKnownAs;
+
+    if (alsoKnownAs) {
+      let jsonApiDocument = await this.fetchJson(alsoKnownAs[0], waitForInfo);
+      console.log('json', jsonApiDocument);
```
backspace, line 63: I’m guessing this is from iteration in development and could be removed?

packages/web-client/app/templates/card-pay/balances.hbs
```diff
@@ -1,81 +1,74 @@
...
+              <CardPay::PrepaidCardSafe @safe={{prepaidCard}} />
+            {{/each}}
           </section>
-        {{/if}}
-      </aside>
+        </section>
+      </section>
     {{/if}}
   </section>
-</section>
```
backspace, line 71: I’m not sure what these changes are about, maybe just rearrangements due to CSS?

backspace (approved):
There’s a likely-not-meant-to-be-included `console.log` but this looks good to me. I think if we have another repetition of the DID-resolving/`fetchJson` pattern we could extract some common functions for resources, but for now the duplication seems fine.

packages/web-client/app/resources/merchant-info.ts
lukemelia: @backspace good catch. Please push a commit to remove this and then merge this PR. :shipit:

PR #2050, opened 2021-09-06
Prototype/fix accessing visible postables too early

Aierie:
The impact of the problem addressed in https://github.com/cardstack/cardstack/pull/2050/commits/c460425f42c90c38c6c5b4b17d374beb5bb2cfa8 is janky animations, and evaluating whether a postable needs to be shown before some important state is retrieved. As postables' `includeIf`s are only evaluated once, this may lead to divergence from the intended path for a workflow

### DIFF
--- a/packages/web-client/app/models/workflow/workflow-card.ts
+++ b/packages/web-client/app/models/workflow/workflow-card.ts
@@ -52,6 +52,7 @@ export class WorkflowCard extends WorkflowPostable {
   }
 
   @action async onComplete() {
+    if (this.isComplete) return;
     let checkResult = await this.check();
     if (checkResult.success) {
       // visible-postables-will-change starts test waiters in animated-workflow.ts
--- a/packages/web-client/app/components/card-pay/issue-prepaid-card-workflow/face-value/index.hbs
+++ b/packages/web-client/app/components/card-pay/issue-prepaid-card-workflow/face-value/index.hbs
@@ -3,6 +3,7 @@
   @header="Prepaid card funding"
   @isComplete={{@isComplete}}
   {{did-insert this.prepareFaceValueOptions}}
+  data-test-face-value-card
 >
   <ActionCardContainer::Section @title="Choose the face value of your prepaid card" class="face-value-card__section">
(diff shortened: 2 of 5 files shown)


packages/web-client/app/components/card-pay/layer-two-connect-card/index.ts
```diff
@@ -36,6 +36,8 @@ class CardPayLayerTwoConnectCardComponent extends Component<CardPayLayerTwoConne
   constructor(owner: unknown, args: CardPayLayerTwoConnectCardComponentArgs) {
     super(owner, args);
     if (this.isConnected) {
+      // this ends up running after rendering the whole workflow up to first incomplete postable
+      // which means test waiters start after they have been stopped
       next(this, () => {
         this.args.onComplete?.();
       });
```
Aierie, line 43: We use this pattern in other places too for cards that are completed from the start (eg. view only cards), would be good to review the use of `next` calling `onComplete` in the constructor - could lead to odd consequences.

jurgenwerk:
@Aierie thanks for these fixes! At a first glance it looks ok. I have some changes to push into the target branch but then I think these branches will get diverged. Let me clean up my branch first and then I'll test this one on top of it!

packages/web-client/app/models/workflow/workflow-card.ts
```diff
@@ -52,6 +52,8 @@ export class WorkflowCard extends WorkflowPostable {
   }
 
   @action async onComplete() {
+    // not too sure if this is the right fix, but it seems to make sense to not be able to complete a completed card
+    if (this.isComplete) return;
```
jurgenwerk, line 56: Do you know of a case where this happens?

Aierie: layer two connect card is the place, hence the comment in the same commit

Aierie: Comment removed as it would be confusing out of context - this referenced the layer two connect card calling `onComplete` in `next`, leading to it starting test waiters after a restored workflow had finished rendering.

PR #2051, opened 2021-09-06
prevent autoselection of first prepaid card

Aierie:
This circumvents our previous problems by changing the trigger of the dropdown to a click instead - problems were caused by the mousedown event opening the dropdown, allowing elements that slid under the cursor to become the target of a click or mouseup event, and triggering click/mouseup event listeners before the user was ready to respond. For autoselection, this was a dropdown option; for difficulty opening when clicking the bottom border, this was the dropdown option container.

https://github.com/cibernox/ember-power-select/issues/1282

### DIFF
--- a/packages/web-client/app/components/card-pay/card-picker/index.css
+++ b/packages/web-client/app/components/card-pay/card-picker/index.css
@@ -55,7 +55,6 @@
 
 .card-picker-dropdown .ember-basic-dropdown-content.ember-basic-dropdown--transitioning-in {
   animation: drop-fade-below var(--boxel-transition);
-  pointer-events: none;
 }
 
 .card-picker-dropdown .ember-basic-dropdown-content.ember-basic-dropdown--transitioning-out {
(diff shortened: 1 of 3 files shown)


Aierie:
If we choose to use this fix, I think it might be worth figuring out how to apply it across all dropdowns, since opening a dropdown on click vs mousedown feels like it should be consistent across an app.

Aierie:
@backspace do let me know if this works for the dropdown you're working on?

Aierie:
Usage of the argument in source:

https://github.com/cibernox/ember-power-select/blob/3a73a70293c721e6368a8ce293e2652869930342/addon/components/power-select.hbs#L29

Documentation on Ember Basic Dropdown trigger events:

https://ember-basic-dropdown.com/docs/trigger-events

backspace (approved):
This works for me locally. I’ll try it in the `safe-chooser-dropdown` in the PR I’m working on but I haven’t been seeing the auto-select problem nearly as much with it 🤔

PR #2052, opened 2021-09-06
Refactor SDK args (onTxnHash, nonce, onNonce) to a single TransactionOptions arg

lukemelia:
- this is a BREAKING change to the SDK's API
- this prepares for the future addition of a txnHash option

Once merged cardstack/cardwallet will need an update here:
 https://github.com/cardstack/cardwallet/blob/develop/cardstack/src/screens/SendSheetDepot.tsx#L253-L255

### DIFF
--- a/packages/cardpay-cli/bridge.ts
+++ b/packages/cardpay-cli/bridge.ts
@@ -58,9 +58,9 @@ export async function claimLayer1BridgedTokens(
   let blockExplorer = await getConstant('blockExplorer', web3);
 
   console.log(`Claiming layer 1 bridge tokens for message ID ${messageId}...`);
-  await tokenBridge.claimBridgedTokens(messageId, encodedData, signatures, (txnHash) =>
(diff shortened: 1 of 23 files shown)


habdelra:
I'm ok with collapsing the various params into a single pojo, but I'm not sure that we actually want to have the txnHash in here too--as just providing a txnHash means that you don't actually need to provide any other param. otherwise what does it mean if you provide a txnHash, a nonce, and a safeAddress? it's a little more difficult divining the caller's intent when you can pass params that don't necessarily go together. let's use TS function overloading instead so that we can provide clear groupings of params that work together and we can more clearly determine the caller's intent.

lukemelia:
Great point! I'll refactor accordingly.

lukemelia:
@habdelra Looks like all the overload changes will be in #2053. OK to merge this one now?

habdelra (approved):
👍  looks good to merge

PR #2054, opened 2021-09-07
Extract boxel org title

Aierie:
Extracts a Boxel::OrgTitle component from Boxel::OrgHeader as this is also used in a hamburger menu (see top left of each image).

image

image

Also in the merchant payment request landing page:

image

### DIFF
--- a/packages/boxel/addon/components/boxel/org-header/usage.hbs
+++ b/packages/boxel/addon/components/boxel/org-header/usage.hbs
@@ -20,6 +20,7 @@
     <Args.String
       @name="title"
       @value={{this.title}}
+      @required={{true}}
       @onInput={{fn (mut this.title)}}
     />
     <Args.String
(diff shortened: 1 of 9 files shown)


Aierie:
Where the org header is used in web client, we are using the CSS variable API as follows. This seems to be the only consumer in web-client at the moment. Given that, we could also limit the CSS API and only allow preset sizes (there are 2 sizes we need to support, 2.5rem and the smaller 2.125rem for mobile).

https://github.com/cardstack/cardstack/blob/a3dbcbfc28517783735cdb3dff50f958a816cfd5/packages/web-client/app/components/card-pay/header/index.css#L1-L8

packages/boxel/addon/components/boxel/org-header/usage.hbs
```diff
@@ -1,11 +1,11 @@
 <Freestyle::Usage @name="Boxel:OrgHeader">
   <:description>
-    Usually shown at the top of a dashboard
+    Usually shown at the top of a dashboard. See also Boxel::OrgTitle
```
backspace, line 3: Would it make sense to have a similar “See also” on `org-title`? Also, can this be a link?

backspace (approved):
I noted a possible improvement but it’s non-blocking IMO 🤓

packages/boxel/addon/components/boxel/org-header/usage.hbs
Aierie: Yes, will do both. Thanks!

jurgenwerk (approved):
LGTM!

PR #2057, opened 2021-09-07
Card Compiler

mattmcmanus:


mattmcmanus:
@lukemelia @habdelra I think we're ready to get this merged. It's a monster! If you think it would be helpful to do a walk through, I'm happy to do it.

lukemelia:
@mattmcmanus I am definitely interested in a walkthrough, but it doesn't need to be a blocker to merging this PR. If you and Ed both feel confident, go for it, and then let's schedule time for a walkthrough.

habdelra:
ditto w/ what luke said

PR #2058, opened 2021-09-07
web: Extract CardPay::MerchantLogo component

backspace:
I need to be able to use this separately in #2038.

Left is with this change, right is `main`:

image


Aierie (approved):
This looks good. Separating a CSS circle as its own component does mean that it is subject to use in containers that it doesn't fit in, and can end up being squished in:

image

Not a big deal for now (this was artificially contorted), but something to be aware of I think.

backspace:
hmm 😳  do you think there’s a way to mitigate this?

Aierie:
> hmm 😳 do you think there’s a way to mitigate this?

I have only ever briefly thought about it, so not very sure about all the situations where this could happen. I think the most common cause would be `flex-shrink`.

backspace:
okay, well hopefully there’s a way to deal if it comes up hehe thx

PR #2059, opened 2021-09-07
Feature/hamburger menu

Aierie:
This adds a basic hamburger menu meant to be used for the merchant payment request page. At present, this hamburger menu is not animated. See https://github.com/cardstack/cardstack/pull/2060#discussion_r703737782 for context of use

# The hamburger menu @ 480px
480px hamburger


backspace (approved):
I tried this in the context of #2060 with an extra commit to add a template that renders it, looks good. I wonder if:
* it should be in Boxel
* there’ll be occasion for a non-link `MenuItem`, but that can be addressed in the future if it comes up

Aierie:
> I tried this in the context of #2060 with an extra commit to add a template that renders it, looks good. I wonder if:
> 
> * it should be in Boxel
> * there’ll be occasion for a non-link `MenuItem`, but that can be addressed in the future if it comes up

Unsure if it should be in Boxel. I think we can extract it in the future, if necessary. re: non-link `MenuItem`s, am thinking it could follow `Boxel::Button`'s `@as` API, but will also leave this till we need to use it and requirements are clearer.

PR #2060, opened 2021-09-07
Feature/merchant payment request page

Aierie:
Documenting states, will add to Linear issue and link here when done.

---

Previous list of TODOs:

- [x] replace hardcoded merchant info with properties of the component class
- [x] replace "#" with real links
- [x] Replace mock data with data from query params
- [x] Determine route and create it, move this work into said route
- [x] PR to extract `fetchJson` used in resources into a utility function
- [x] Improve validation + error handling when fetching the merchant safe
- [x] PR to expose SDK viewSafe method that does not require a web3 instance
- [x] Improve validation + error handling when getting the amount + currency
- [x] PR to add constants for currency codes to the SDK, or use a shared library/other resource to ensure consistency between mobile and web? (SDK already has them, native-currencies.ts)
- [x] PR to add merchant payment url generation to SDK (check with mobile team about https://github.com/cardstack/cardwallet/blob/b336a7463f64b4c46b80a7ef82868e2f49993fa1/cardstack/src/utils/merchant-utils.ts#L81-L93)
- [x] Validation of payment urls in the SDK
- [x] Show the clickable link only when on an iOS browser
- [x] Tests
- [x] Replace validation and url generation here with utilities from SDK (requires https://github.com/cardstack/cardstack/pull/2075)
- [x] Error state/404/other when merchant fetching fails (remember to edit tests)

Requires some discussion:
1. What situations can a universal link fail in, and what is a good user experience when it fails? Deferring this, we should probably have a 404/standard error page
2. ~~When should we display the payment link vs QR code? As per discussion w @topwebtek7, only show it on iOS browser for now. Will use ember-useragent or ua-parser-js.~~
3. How should we handle currencies other than SPEND and USD in this UI? At the moment, don't show them, and send the amount + currency directly to the wallet which will know better how to handle them. Might improve this in a separate PR.


packages/web-client/app/components/card-pay/merchant-payment-request-page/index.hbs
```diff
@@ -0,0 +1,70 @@
...
+        </:header>
+        <:body as |MenuItem|>
+          <MenuItem @url="#">
+            About
+          </MenuItem>
+          <MenuItem @url="#">
+            Help
+          </MenuItem>
+        </:body>
+      </CardPay::HamburgerMenu>
```
Aierie, line 56: This is where and how the hamburger menu is used

Aierie: We are controlling the visibility of the hamburger menu via media query at the moment, but toggling whether it's open or not via Javascript. It's possible to set `this.hamburgerMenuOpen` to `false` via a JS media query(`window.matchMedia`), but unsure if necessary.

packages/cardpay-sdk/sdk/safes/base.ts
```diff
@@ -128,6 +128,14 @@ const safesQuery = `
   }
 `;
 
+// TODO: replace this with a formal implementation
+export async function viewSafe(network: 'xdai' | 'sokol', safeAddress: string): Promise<Safe | undefined> {
+  let {
+    data: { safe },
+  } = await query(network, safeQuery, { id: safeAddress });
+  return processSafeResult(safe as GraphQLSafeResult);
+}
```
Aierie, line 137: @habdelra on this page, we need to display merchant customization details to a user that may not be connected to layer 2, using a merchant safe address and network extracted from the URL.

I'd like to open a PR to expose a way to view a safe without needing a web3 instance, would this be okay? Wondering if there is a better way, because this seems to break the SDK's conventions.

habdelra: I'm totally open to that

Aierie: Cool, will do that tomorrow then. Thanks!

lukemelia (approved):
My name is Luke Melia and I endorse this PR.

PR #2064, opened 2021-09-08
extract utility function to fetch resources from off-chain storage

Aierie:
This is to be used in https://github.com/cardstack/cardstack/pull/2060.

Will replace the code below:

https://github.com/cardstack/cardstack/blob/ab7416041ef3d1ab481a475a0f39ef7a956853b8/packages/web-client/app/routes/pay.ts#L32

with

`let jsonApiDocument = await fetchJson(alsoKnownAs[0], true)`

### DIFF
--- a/packages/web-client/app/resources/card-customization.ts
+++ b/packages/web-client/app/resources/card-customization.ts
@@ -1,14 +1,12 @@
-import { timeout } from 'ember-concurrency';
 import { tracked } from '@glimmer/tracking';
 import { getResolver } from '@cardstack/did-resolver';
 import { Resolver } from 'did-resolver';
 import { Resource } from 'ember-resources';
-import config from '../config/environment';
 import * as Sentry from '@sentry/browser';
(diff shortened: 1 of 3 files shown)


packages/web-client/app/utils/fetch-json.ts
```diff
@@ -0,0 +1,45 @@
...
+}
+
+/**
+ * Fetches resources stored in off-chain (S3) storage.
+ *
+ * @param resource The URL for the desired resource
+ * @param shouldRetry Whether to retry if fetching fails
+ * @returns The resource stored in off-chain storage
+ */
+export async function fetchJson(
```
backspace, line 18: Since this is specific to off-chain storage, maybe it could be named `fetchOffChainJson`?

backspace (approved):
Nice to see this extraction, thanks. I wouldn’t insist on the renaming but I think it might be a bit more clear.

PR #2065, opened 2021-09-08
sdk: Update to v0.8.0 of protocol

habdelra:
- includes support for new gas policy: "raw" gas payments via prepaid cards
- updated to accept a parameter for the prepaid card market in which to list newly created cards within


tintinthong:
Just wanna understand this. Is an sdk upgrade dependent on any smart contract state like `cardpayVersion()`

habdelra:
Yup, it’s totally dependent on that. The SDK is future compatible so that we can have smooth App Store approvals. The idea is that we build the SDK for the version of the protocol that will be deployed (using the cardpayVersion() number). The new functionality, though, is dormant until the protocol is actually deployed. So the idea is that we incorporate this SDK update in the mobile app before it is sent for approval. After approval we then deploy the smart contracts, the cardpay version update on chain will trigger the new version of the SDK to be used. So the idea is the future functionality “unlocks” itself as part of the protocol update.

semver is also important. For patch versions of the cardpay protocol the SDK will revert to the closest patch version functionality that is available—so we aren’t forced to rev all the SDK functionality for every patch. But for minor versions of the protocol, which can be breaking, the SDK will throw an exception if there is no available code for the minor version. (Same for major versions). 

All of the contract ABI’s are versioned (there is a separate version folder for the current version and one for the future version). About half of the actual SDK functions are capabilities. Some SDK capabilities are not directly dependent on our on chain contracts, those API’s are not versioned. The rest of the SDK capabilities are versioned, like the prepaid card API’s. For those we have a folder which each version as a separate module, where there is a base.ts module for the shared functionality. Some version modules might just be shims in the case where the contract has no appreciable changes between the versions, and others might be wholly different (it just depends on the type of change). 

We’ll keep around the old version until the change has made its way to xdai (however since we have no real xdai clients, I’m just cleaning out the old stuff immediately). Testing has proven to be quite difficult with this approach. I think that when we merge the protocol repo into the mono repo that will be much easier to deal with. For now it’s just very painful and we have had a few hours of downtime because of how painful it is to test against future versions of the protocol.

PR #2066, opened 2021-09-08
Avoid "NaN" display when rendering USD value, and speed up conversion

lukemelia:


packages/web-client/app/services/token-to-usd.ts
```diff
@@ -39,8 +39,10 @@ export default class TokenToUsd extends Service {
   usdConverters = new UsdConverters();
   #registeredHelpers: Set<TokenToUsdHelper> = new Set();
 
-  @task({ maxConcurrency: 1, drop: true }) *pollTask(): any {
+  @task({ maxConcurrency: 1, restartable: true }) *pollTask(): any {
+    console.log('pollTask start');
     while (this.shouldPoll) {
+      console.log('pollTask loop');
```
backspace, line 45: forgotten? 😯

backspace (approved):
👍🏻 apart from the likely-forgotten logging statements

Aierie (approved):
Thanks for fixing this! In case I'm missing something - my understanding of this fix is that it allowing the task to be restarted allows new instances of the helper that reference tokens not fetched in a previous polling loop to trigger a fetch of the necessary converters.

lukemelia:
> Thanks for fixing this! In case I'm missing something - my understanding of this fix is that it allowing the task to be restarted allows new instances of the helper that reference tokens not fetched in a previous polling loop to trigger a fetch of the necessary converters.

You've hit the nail on the head. 👍  There is also an optimization that avoids restarting the task if the list of the symbolsToUpdate has not changed.

PR #2068, opened 2021-09-09
Add registration functions for reward program

tintinthong:
Added two functions: `RegisterRewardProgram` and `RegisterRewardee`

- The `RegisterRewardProgram` API is used to register a reward program using a prepaid card. The call can specify an EOA admin account -- it defaults to the owner of the prepaid card itself. The reward program admin will then be able to manage the reward program using other api functions like`lockRewardProgram`, `addRewardRule`, etc. A fee of 500 spend is charged when registering a reward program. Currently, tally only gives rewards to a single reward program (sokol: "0x4767D0D74356433d54880Fcd7f083751d64388aF"). 

- The `RegistereRewardee` API is used to register a rewardee for a reward program using a prepaid card. The purpose of registering is not to "be considered to receive rewards" rather to "be able to claim rewards that have been given". By registering, the owner of the prepaid card is given ownership of a reward safe that will be used to retrieve rewards from the reward pool. A rewardee/eoa is eligible to only have one reward safe for each reward program; any attempts to re-register will result in a revert error. A fee of 500 spend is charged when registering a rewardee.

Demo: https://www.loom.com/share/1540ea3042814576868e5a46842c248c

### DIFF
--- a/packages/cardpay-sdk/sdk/prepaid-card/base.ts
+++ b/packages/cardpay-sdk/sdk/prepaid-card/base.ts
@@ -479,7 +479,7 @@ export default class PrepaidCard {
   async registerRewardProgram(txnHash: string): Promise<TransactionReceipt>;
   async registerRewardProgram(
     prepaidCardAddress: string,
-    rewardProgramID: string,
+    rewardProgramId: string,
     txnOptions?: TransactionOptions,
     contractOptions?: ContractOptions
   ): Promise<TransactionReceipt>;
@@ -511,7 +511,7 @@ export default class PrepaidCard {
         )
     );
 
-    let rewardProgramID = randomHex(20);
+    let rewardProgramId = randomHex(20);
     let prepaidCardMgr = await this.getPrepaidCardMgr();
     let owner = await prepaidCardMgr.methods.getPrepaidCardOwner(prepaidCard).call();
 
@@ -524,7 +524,7 @@ export default class PrepaidCard {
         let payload = await this.getRegisterRewardProgramPayload(
           prepaidCardAddressOrTxnHash,
           admin ?? owner,
-          rewardProgramID,
+          rewardProgramId,
           REWARD_PROGRAM_REGISTRATION_FEE_IN_SPEND,
           rateLock
         );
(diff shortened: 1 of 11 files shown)


packages/cardpay-sdk/sdk/prepaid-card/base.ts
```diff
@@ -33,6 +33,8 @@ const TIMEOUT = 1000 * 60 * 5;
 // stay below the AML limit of $10,000 USD.
 export const MAXIMUM_PAYMENT_AMOUNT = 10000 * 100;
 export const MAX_PREPAID_CARD_AMOUNT = 10;
+const REWARD_PROGRAM_REGISTRATION_FEE_IN_SPEND = 500;
+const REWARDEE_REGISTRATION_FEE_IN_SPEND = 500;
```
habdelra, line 37: are these constants available on chain? if so we should refer to those in case chris decides to change the fees

packages/cardpay-sdk/sdk/prepaid-card/base.ts
```diff
@@ -603,6 +605,238 @@ export default class PrepaidCard {
...
+      REWARD_PROGRAM_REGISTRATION_FEE_IN_SPEND,
+      (issuingToken, balanceAmount, requiredTokenAmount, symbol) =>
+        new Error(
+          `Prepaid card does not have enough balance to register reward program. The issuing token ${issuingToken} balance of prepaid card ${prepaidCardAddressOrTxnHash} is ${fromWei(
+            balanceAmount
+          )} ${symbol}, payment amount in issuing token is ${fromWei(requiredTokenAmount)} ${symbol}`
+        )
+    );
+
+    let rewardProgramId = randomHex(20);
```
habdelra, line 643: do we need to check for a program ID collision on chain?

habdelra (approved):
I had a couple minor nit picks, but otherwise this looks good 👍

habdelra:
@tintinthong i did some refactoring on how safe txn are signed and executed in this PR https://github.com/cardstack/cardstack/pull/2099. I don't think this branch will TS compile if its merged as is. You should rebase with main and cleanup the compile errors.

tintinthong:
- added reward manager
- queried onchain fees
- generate reward program id with collision check
- rebase on refactoring for safe tx

PR #2072, opened 2021-09-09
Add in-memory caching for off-chain JSON fetched via DID

lukemelia:


backspace (approved):
Some formatting lint to address but looks good. I wondered whether it would make a difference as it’s still asynchronous but it definitely does. Before:

many-s3-screencast 2021-09-09 15-47-07

After:

json-cache-screencast 2021-09-09 22-32-37

I suspect the selected-item-constantly-rerendering problem is still happening but it’s now imperceptible!

PR #2074, opened 2021-09-10
derive depot from safes instead of making it separate resource

Aierie:
We were triggering the `viewSafes` task twice because it was referenced in two different resources being used in the same page, which caused an error when changing accounts in the withdrawal flow - see https://github.com/cardstack/cardstack/runs/3560882551?check_suite_focus=true#step:6:285.

I tried nesting resources (using `useResource(this, Safes, () => ({...}))`) in the `Depot` resource, but this still triggers two fetches. Making the depot derived state of the `Safes` resource solves this issue so we won't need https://github.com/cardstack/cardstack/pull/2038/commits/dcbc50ecaa9d4ef47b26aeabd0bfd6c966607468#.


Aierie:
Might be good to make a `Depot` class that encapsulates this logic, but wasn't sure so I implemented the bare bones first.

backspace:
wow, nice, thanks… did you see that dcbc50e didn’t even fix it, it must have been that with some other of the tiny changes we made, but _which_ 😭

Aierie:
> wow, nice, thanks… did you see that [dcbc50e](https://github.com/cardstack/cardstack/commit/dcbc50ecaa9d4ef47b26aeabd0bfd6c966607468) didn’t even fix it, it must have been that with some other of the tiny changes we made, but _which_ 😭

Ah I missed that ! I'm glad we don't need to pick through them now though 😛

PR #2075, opened 2021-09-10
Enhancement/payment request resources

Aierie:
Paired w @topwebtek7 

This PR adds utilities to facilitate merchant payment in both the DApp and the wallet:
- Generating merchant payment URL
- Checking if a merchant payment URL is valid
- Checking if a currency is supported
- Checking if a given amount is valid for a currency

### DIFF
--- a/packages/cardpay-sdk/sdk/currency-utils.ts
+++ b/packages/cardpay-sdk/sdk/currency-utils.ts
@@ -12,6 +12,32 @@ export const isSupportedCurrency = (nativeCurrency: string) => {
   return has(supportedNativeCurrencies, `${nativeCurrency}`);
 };
 
+export const isValidAmountForCurrency = (nativeCurrency = 'USD', amount: BigNumberish) => {
+  let currency = get(supportedNativeCurrencies, nativeCurrency);
+  if (!currency) {
+    return false;
+  }
+
+  if (typeof (amount as unknown) === 'string') {
+    amount = Number(amount);
+  } else if (amount instanceof BigNumber) {
+    amount = amount.toNumber();
+  }
+
+  if (typeof amount !== 'number' || isNaN(amount)) {
+    return false;
+  }
+
+  let maxDecimals = currency.decimals;
+  let numberOfDecimals = `${amount}`.split('.')[1]?.length;
+
+  if (numberOfDecimals > maxDecimals) {
+    return false;
+  }
+
+  return true;
+};
+
 type BigNumberish = number | string | BigNumber;
 
 interface Asset {
--- a/packages/cardpay-sdk/sdk/currency-utils.ts
+++ b/packages/cardpay-sdk/sdk/currency-utils.ts
@@ -5,8 +5,12 @@
  */
 
 import BigNumber from 'bignumber.js';
-import { get, isNil } from 'lodash';
+import { get, has, isNil } from 'lodash';
 import supportedNativeCurrencies from './native-currencies';
+
+export const isSupportedCurrency = (nativeCurrency: string) => {
+  return has(supportedNativeCurrencies, `${nativeCurrency}`);
+};
 
 type BigNumberish = number | string | BigNumber;
 
--- a/packages/cardpay-sdk/sdk/currency-utils.ts
+++ b/packages/cardpay-sdk/sdk/currency-utils.ts
@@ -12,32 +12,6 @@ export const isSupportedCurrency = (nativeCurrency: string) => {
   return has(supportedNativeCurrencies, `${nativeCurrency}`);
 };
 
-export const isValidAmountForCurrency = (nativeCurrency = 'USD', amount: BigNumberish) => {
-  let currency = get(supportedNativeCurrencies, nativeCurrency);
-  if (!currency) {
-    return false;
-  }
-
-  if (typeof (amount as unknown) === 'string') {
-    amount = Number(amount);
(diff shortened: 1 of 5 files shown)


packages/cardpay-sdk/sdk/currency-utils.ts
```diff
@@ -5,9 +5,39 @@
...
+  if (typeof amount !== 'number' || isNaN(amount)) {
+    return false;
+  }
+
+  let maxDecimals = currency.decimals;
+  let numberOfDecimals = `${amount}`.split('.')[1]?.length;
+
+  if (numberOfDecimals > maxDecimals) {
+    return false;
+  }
```
Aierie, line 36: The immediate context of use for this is a merchant requesting payment from someone. If the amount has too many decimal places for the given currency, we could:
- Not display the amount/not handle the amount, OR
- Round/truncate the amount

@lukemelia do you have any thoughts about what we should do?

packages/cardpay-sdk/sdk/utils/merchant.ts
```diff
@@ -0,0 +1,45 @@
...
+  currency = 'SPD',
+}: MerchantPaymentURLParams) => {
+  const handleAmount = amount ? `amount=${amount}&` : '';
+
+  return `${domain}/pay/${network}/${merchantSafeID}?${handleAmount}currency=${currency}`;
+};
+
+export const isValidMerchantPaymentUrl = (merchantPaymentUrl: string) => {
+  let url = new URL(merchantPaymentUrl);
+  if (url.protocol !== 'cardwallet:' || url.hostname) {
```
topwebtek7, line 34: @Aierie isn't it should be like `if (url.protocol !== 'cardwallet:' || !url.hostname) {`?
just wonder you missded `!` before `url.hostname`

Aierie: I think we're not expecting a hostname. However, I realised upon trying this in both the browser and node environments that node and browser implementations of `URL` are different.

For:

```
new URL('cardwallet://pay/sokol/0xE73604fC1724a50CEcBC1096d4229b81aF117c94').hostname
```

Browser returns "", node returns "pay". I'll remove this check

packages/cardpay-sdk/sdk/currency-utils.ts
lukemelia: I'd say whichever is easier from an engineering POV, with a preference for the latter. There is an old adage in software: "Be strict with your output and liberal with your input."

packages/cardpay-sdk/sdk/utils/merchant.ts
```diff
@@ -0,0 +1,45 @@
+import { isAddress } from 'web3-utils';
+
+export function validateMerchantId(value: string) {
+  if (!value.trim().length) return 'This field is required';
+  if (/[^a-z0-9]/.test(value))
+    return 'The Merchant ID can only contain lowercase letters or numbers, no special characters';
+  else if (value.length > 50)
```
lukemelia, line 7: Let's at least name this magic number. Is anything else using this "50" number?

packages/cardpay-sdk/sdk/utils/merchant.ts
```diff
@@ -0,0 +1,45 @@
...
+};
+
+export const isValidMerchantPaymentUrl = (merchantPaymentUrl: string) => {
+  let url = new URL(merchantPaymentUrl);
+  if (url.protocol !== 'cardwallet:') {
+    return false;
+  }
+
+  let parts = url.pathname.replace(/^\/*/, '').split('/');
+  if (parts.length !== 3) {
```
lukemelia, line 39: Is this necessary if we have the checks below on line 44?

lukemelia (changes requested):

lukemelia:
OK to merge once you've processed my comments.

packages/cardpay-sdk/sdk/utils/merchant.ts
Aierie: Nope. This is basically a round number < 63, which is the domain name limit.

packages/cardpay-sdk/sdk/utils/merchant.ts
Aierie: I'm not sure if the path might be longer than 3 parts, eg. `cardwallet://pay/sokol/address/somethingelse`, this is to prevent that case.

packages/cardpay-sdk/sdk/currency-utils.ts
Aierie: Knowing that CS-1736 exists, I think it is better UX to be able to edit a rounded amount!

Aierie: Will remove this method, since the main value of this method was 31-36

backspace:
Maybe [`URLPattern`](https://web.dev/urlpattern/) could be of use here if route recognition continues to be a task 😯

Aierie:
> Maybe [`URLPattern`](https://web.dev/urlpattern/) could be of use here if route recognition continues to be a task 😯

Thanks! I had a look at this, it would be great but has limited browser support and no Node support, and the polyfill seems to need some module/export rework at the moment 😢

backspace:
Harsh! Well it’s p new, hopefully in the future it’ll work better, I’m glad it’s on its way at least.

PR #2078, opened 2021-09-11
Pin waypoint cli version

pcjun97:
Pin waypoint to 0.4.0 (same version as the running server) to prevent breaking changes.


lukemelia:
Good find @pcjun97 !

PR #2082, opened 2021-09-13
lock web3 version

habdelra:
The drifting web3 version is causing type-checking headaches with consumers of the SDK


danibonilha (approved):
LGTM

PR #2083, opened 2021-09-13
web: Add generics for token utility classes

backspace:
I [need](https://github.com/cardstack/cardstack/blob/web/merchant-withdrawal-cs-1658/packages/web-client/app/components/card-pay/withdrawal-workflow/choose-balance/index.ts#L27) a way to say that `token.symbol` where `token: TokenBalance` would be a particular subtype, so I added generics for the relevant token utility classes to be able to specify what type of token was expected, if any. Hopefully this makes sense! I could probably use `as` but it seemed helpful to me to make things explicit.


backspace:
I see this has a CI failure; I think it could still be reviewed in a general way to see if it even makes sense, I’ll address the test failure tomorrow.

Aierie (approved):

PR #2084, opened 2021-09-13
web: Change depot to be derived from safes

backspace:
This is extracted from #2038 (via #2074) to facilitate a smaller review.


lukemelia (approved):
Looks OK. Should the Depot resource be deleted?

Aierie (approved):

backspace:
Good point, @lukemelia, thanks, removed now!

PR #2086, opened 2021-09-14
Suggestion/type refactor

Aierie:


backspace:
I did wonder at some of the duplication in the file… this doesn’t seem hard to read to me really though 🤔

PR #2087, opened 2021-09-14
Enhancement/improve merchant payment page

Aierie:
This PR addresses:
- [CS-1830]: makes the url generated on the merchant payment request page start with `https://wallet.cardstack.com` in production, and `https://wallet-staging.stack.cards` in staging and dev. Did not add these as constants to the SDK for now due to difficulties getting imports to work.
- [CS-1837]: makes incomplete routes (eg. `https://wallet.cardstack.com/pay/sok`) match the `pay-error` page.

### DIFF
--- a/packages/web-client/app/router.ts
+++ b/packages/web-client/app/router.ts
@@ -16,7 +16,7 @@ Router.map(function () {
   this.route('pay', {
     path: '/pay/:network/:merchant_safe_id',
   });
-  this.route('pay-error', {
+  this.route('pay-missing-route', {
     path: '/pay/:*',
   });
   this.route('boom');
--- a/packages/web-client/app/router.ts
+++ b/packages/web-client/app/router.ts
@@ -16,5 +16,8 @@ Router.map(function () {
   this.route('pay', {
     path: '/pay/:network/:merchant_safe_id',
   });
+  this.route('pay-error', {
+    path: '/pay/:*',
+  });
   this.route('boom');
 });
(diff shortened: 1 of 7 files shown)


packages/web-client/app/router.ts
```diff
@@ -16,5 +16,8 @@ Router.map(function () {
   this.route('pay', {
     path: '/pay/:network/:merchant_safe_id',
   });
+  this.route('pay-error', {
```
jurgenwerk, line 19: Would it be better to name this `pay-missing-route` or something like that? Just to make it a bit clearer.

jurgenwerk (approved):
LGTM! Just a very minor comment.

packages/web-client/app/router.ts
Aierie: At the moment this is reusing Ember's error substate route (that is rendered when we fail to fetch a merchant safe) https://guides.emberjs.com/release/routing/loading-and-error-substates/#toc_error-substates, is there a good way to specify that without creating a new route? Should we create a new route for this case?

lukemelia: Having a declared route named the same as implicit routes like this should be avoided, IMO

lukemelia: You can use the `templateName` property of a route to reuse the template.

packages/web-client/app/routes/pay-missing-route.ts
```diff
@@ -0,0 +1,6 @@
+import Route from '@ember/routing/route';
+import '../css/pay.css';
+
+export default class PayErrorRoute extends Route {
```
lukemelia, line 4: Rename this class name for clarity.

Aierie: Oops, thanks!

PR #2090, opened 2021-09-14
Use setOwner to cleanup workflow's service usage

lukemelia:


backspace (approved):
I appreciate this better integration with Ember conventions 🎉

PR #2091, opened 2021-09-14
web: Add validation enforcement of withdrawal limits

backspace:
This adds checks that the withdrawal amount is within the limits returned by `TokenBridgeHomeSide.getWithdrawalLimits` and includes the relevant limit in the validation string if not.

validation-screencast 2021-09-16 14-37-30

This necessitated some adjustments to the validation function, which had a `min` option that was being used in the deposit transaction amount card to enforce that the value had to be higher than 0, which isn’t a true minimum. `min` is now optional and `> 0` is enforced in its absence. This didn’t have test representation so I added another test for that card and refactored out the duplicated setup.

The validation function now also accepts a `balance` to distinguish between the maximum from the withdrawal limit vs the maximum from the balance.

We [decided](https://github.com/cardstack/cardstack/pull/2100/files/6d06633303d6c6cdbc31651efd85f14624b59e65#r710214248) to not add the bridged tokens to the SDK’s `native-currencies` list, so there’s some string manipulation with the result of the call to `convertRawAmountToNativeDisplay` to display a bridged token in the validation message if that’s the context. It’s a bit unsightly but it works and seems okay to me, but other approaches are possible.

### Test failure

I see there’s a test failure for the acceptance test that asserts that a prepaid card workflow can’t be restored if the authentication token has been cleared. I don’t understand how that could be related, but I’ll look more into it tomorrow. I think this can be reviewed regardless, not to merge until it’s addressed.

### DIFF
--- a/packages/web-client/app/components/card-pay/deposit-workflow/transaction-amount/index.ts
+++ b/packages/web-client/app/components/card-pay/deposit-workflow/transaction-amount/index.ts
@@ -124,7 +124,7 @@ class CardPayDepositWorkflowTransactionAmountComponent extends Component<Workflo
 
   validate() {
     this.validationMessage = validateTokenInput(this.amount, {
-      min: new BN(0),
+      tokenSymbol: this.currentTokenSymbol,
       max: this.currentTokenBalance,
     });
   }
--- a/packages/web-client/app/components/card-pay/deposit-workflow/transaction-amount/index.ts
+++ b/packages/web-client/app/components/card-pay/deposit-workflow/transaction-amount/index.ts
@@ -125,7 +125,7 @@ class CardPayDepositWorkflowTransactionAmountComponent extends Component<Workflo
   validate() {
     this.validationMessage = validateTokenInput(this.amount, {
       tokenSymbol: this.currentTokenSymbol,
-      max: this.currentTokenBalance,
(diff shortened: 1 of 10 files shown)


packages/web-client/app/components/card-pay/deposit-workflow/transaction-amount/index.ts
```diff
@@ -124,7 +124,7 @@ class CardPayDepositWorkflowTransactionAmountComponent extends Component<Workflo
 
   validate() {
     this.validationMessage = validateTokenInput(this.amount, {
-      min: new BN(0),
+      tokenSymbol: this.currentTokenSymbol,
       max: this.currentTokenBalance,
     });
```
Aierie, line 129: Maybe this should also use the `balance` validation option? For a more consistent validation message.

Aierie (approved):
Looks good! I wonder if we should include the withdrawal limits task within `initializeTask`, but am not sure if it's necessary.

packages/web-client/app/components/card-pay/deposit-workflow/transaction-amount/index.ts
backspace: Agreed, done in bf2428e, thanks

PR #2094, opened 2021-09-15
Fix/misc payment page issues

Aierie:
Fix some issues I noticed when trying out the payment page in staging:
- Clicking on the deep link in an iOS browser simply refreshes the page, it doesn't attempt to open the app. Apple's documentation says that clicking on a link with the same domain as the current window will not open deep links in apps. As such, the clickable link was changed to a `cardwallet:` custom protocol link.
- Hamburger menu icon was squished to invisibility in Safari's rendering.


jurgenwerk (approved):
LGTM! 🚀

PR #2096, opened 2021-09-15
Fix dropdown oversized content

jurgenwerk:
Ticket: [CS-1844](https://linear.app/cardstack/issue/CS-1844/safe-chooser-dropdown-truncates)

Problem:

image

After the fix (added the scrollbar):

image

### DIFF
--- a/packages/boxel/addon/components/boxel/card-container/index.css
+++ b/packages/boxel/addon/components/boxel/card-container/index.css
@@ -2,11 +2,20 @@
   position: relative;
   background-color: var(--boxel-light);
   border-radius: var(--boxel-border-radius);
-  overflow: hidden;
   transition:
     max-width var(--boxel-transition),
     border-radius var(--boxel-transition),
     box-shadow var(--boxel-transition);
+}
+
+.boxel-card-container > *:first-child {
+  border-top-left-radius: var(--boxel-border-radius);
+  border-top-right-radius: var(--boxel-border-radius);
+}
+
+.boxel-card-container > *:last-child {
+  border-bottom-left-radius: var(--boxel-border-radius);
+  border-bottom-right-radius: var(--boxel-border-radius);
 }
 
 .boxel-card-container--boundaries {
--- a/packages/boxel/addon/components/boxel/card-container/index.css
+++ b/packages/boxel/addon/components/boxel/card-container/index.css
@@ -7,15 +7,6 @@
     border-radius var(--boxel-transition),
     box-shadow var(--boxel-transition);
 }
-
-.boxel-card-container > *:first-child {
-  border-top-left-radius: var(--boxel-border-radius);
-  border-top-right-radius: var(--boxel-border-radius);
-}
-
-.boxel-card-container > *:last-child {
-  border-bottom-left-radius: var(--boxel-border-radius);
-  border-bottom-right-radius: var(--boxel-border-radius);
 }
 
 .boxel-card-container--boundaries {
--- a/packages/web-client/app/components/card-pay/safe-chooser-dropdown/index.css
+++ b/packages/web-client/app/components/card-pay/safe-chooser-dropdown/index.css
@@ -19,13 +19,14 @@
 }
 
 .safe-chooser-dropdown .ember-basic-dropdown-content {
+  max-height: 300px;
+  overflow: auto;
   position: absolute;
   z-index: 1;
   width: 100%;
   background-color: var(--boxel-light);
   border: 1px solid var(--boxel-purple-300);
   border-radius: var(--boxel-border-radius-sm);
-  overflow: hidden;
 }
 
 .safe-chooser-dropdown .ember-basic-dropdown-content.ember-basic-dropdown--transitioning-in {
--- a/packages/web-client/app/components/card-pay/safe-chooser-dropdown/index.css
+++ b/packages/web-client/app/components/card-pay/safe-chooser-dropdown/index.css
@@ -19,7 +19,7 @@
 }
 
 .safe-chooser-dropdown .ember-basic-dropdown-content {
-  max-height: 300px;
+  max-height: calc(var(--field-height) * 4.5);
   overflow: auto;
   position: absolute;
   z-index: 1;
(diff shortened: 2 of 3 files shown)


packages/web-client/app/components/card-pay/safe-chooser-dropdown/index.css
```diff
@@ -19,13 +19,14 @@
 }
 
 .safe-chooser-dropdown .ember-basic-dropdown-content {
+  max-height: 300px;
```
jurgenwerk, line 22: Just an arbitrary value that looked good. Questions:
1. Should this be a variable?
2. Should we add this to all dropdowns we have? Otherwise we'll probably have the same issues elsewhere

packages/boxel/addon/components/boxel/card-container/index.css
```diff
@@ -2,13 +2,22 @@
   position: relative;
   background-color: var(--boxel-light);
   border-radius: var(--boxel-border-radius);
-  overflow: hidden;
```
jurgenwerk, line 5: In case the dropdown content is too big to be fitted inside the container edge, `overflow: hidden;` will cut it off:

image

Removing `overflow: hidden;` fixes this issue):

image

However, removing that produces a new bug - the rounded corners of the card container stop being round 🤯:

image

I think `overflow: hidden;` was here to make the border radius work when there are children elements present, [as discussed here](https://stackoverflow.com/questions/10995294/border-radius-not-working).

So to support both elements that need to span out of the container (dropdowns etc), and keeping the border radius, I added the `.boxel-card-container > *:first-child` and `.boxel-card-container > *:last-child` rules for border radius.

packages/web-client/app/components/card-pay/safe-chooser-dropdown/index.css
backspace: I tend to favour extraction of magic numbers, but there are a lot of them in the CSS already so it wouldn’t be a departure to keep this. I think dropdowns in particular might be ready for some generalised styles but I’m unsure how to accomplish that with the way the styles are per-component at the moment 🤔

backspace (approved):
Thanks for looking into this!

packages/web-client/app/components/card-pay/safe-chooser-dropdown/index.css
Aierie: I think it would be nice to have a way to share generalised styles too! Would be nice to have something that can carry more intent/package things more nicely than variable prefixes.

packages/boxel/addon/components/boxel/card-container/index.css
Aierie: We use `CardContainer` very liberally at the moment - not just as a workflow card but as anything that needs the border radius + background + shadow, `first-child` might not always refer to the header, `last-child` might not always refer to the chin; likewise they may not be attached to the top/bottom. What do you think about making these changes in the `CardPay::ActionCardContainer` component instead?

packages/web-client/app/components/card-pay/safe-chooser-dropdown/index.css
Aierie: (but yep also don't mind the magic number as is)

lukemelia: One other thought on this is that ideal dropdown max height seems dependent on screen size. Perhaps it should be in vh, perhaps with a min-max?

packages/boxel/addon/components/boxel/card-container/index.css
lukemelia: Shouldn't dropdowns be doing something to remove themselves from DOM hierarchy to avoid this?

jurgenwerk: @Aierie thanks, I think you're right - better to put this in `.action-card-container` then!

@lukemelia Yeah I think that's the default behaviour for power select, however we use `@renderInPlace={{true}}`, which renders it right there in the component. If I remove this, I get an odd error:

image

It looks like we're using `@renderInPlace={{true}}` in quite a few other places as well, but I'm not sure why 🤔

packages/web-client/app/components/card-pay/safe-chooser-dropdown/index.css
jurgenwerk: What about expressing it like this:

`max-height: calc(var(--field-height)*4.5)`.

This way we just rely on the variable which is expressed in rem which reacts to the screen size.

I think displaying 4 items would be enough. I added 0.5 so that it's clearer there is more to see in the dropdown (and that one should scroll down to reveal it)

image

jurgenwerk:
@backspace @Aierie @lukemelia added two commits that address your suggestions - let me know if you have any further comments - if not I'll squash and merge

Aierie:
> @backspace @Aierie @lukemelia added two commits that address your suggestions - let me know if you have any further comments - if not I'll squash and merge

Looks good to me! Thanks

Aierie (approved):

lukemelia (approved):
👌

packages/boxel/addon/components/boxel/card-container/index.css
burieberry: > @Aierie thanks, I think you're right - better to put this in `.action-card-container` then!
> 
> @lukemelia Yeah I think that's the default behaviour for power select, however we use `@renderInPlace={{true}}`, which renders it right there in the component. If I remove this, I get an odd error:
> 
> image
> 
> It looks like we're using `@renderInPlace={{true}}` in quite a few other places as well, but I'm not sure why 🤔

I've dealt with this error as well and `@renderInPlace={{true}}` was the workaround. I tried the suggestions in this thread, which stopped the error message but then the dropdown did not appear anywhere on the UI: https://discuss.emberjs.com/t/console-error-when-making-addon-component/18360/16 This one recommends going back to an older version: https://github.com/cibernox/ember-basic-dropdown/issues/614. 

There's a card in Merchant Creation workflow that also had this dropdown issue (where the user selects a prepaid card from the dropdown). I had removed overflow for that specific card. We should only be using the CardContainer for isolated and embedded cards, but the header and footer are sometimes optional or hidden in various card modes so let's keep an eye on overflow and card borders.

jurgenwerk: Got it. Thanks for chiming in.

PR #2098, opened 2021-09-16
ci: Update workflow with composite actions and add discord notification

pcjun97:
With [this change](https://github.com/actions/runner/issues/646#issuecomment-901336347) to GitHub Action, we can now use nested composite actions. I've updated the deploy workflow for `web-client`, `hub`, and `cardie` to utilize this.

Another use case we can now support easily is a simple action to push notification to discord with the help of cardie. This feature is included in this PR as well.

Closes:
[CS-1329](https://linear.app/cardstack/issue/CS-1329/notify-hub-and-web-client-deployments-in-discord)
[CS-1786](https://linear.app/cardstack/issue/CS-1786/move-common-workflow-steps-to-composite-actions)


lukemelia (approved):
🙌 Great to see Cardie B evolving!

PR #2100, opened 2021-09-16
sdk: Add definitions for Card Pay currencies

backspace:
This will allow the use of Card Pay currencies with the formatting
utilities. I also included a better error message when attempting to
format an unknown currency, which previously resulted in:

`Cannot destructure property 'decimals' of 'nativeSelected'`

I have use for this in #2091 to format currency values in validation messages. The specifics of each configuration are copied from Ethereum with the names and labels changed, which is a guess, I’m open to suggestions.


packages/cardpay-sdk/sdk/native-currencies.ts
```diff
@@ -11,6 +11,50 @@ export default {
...
+    alignment: 'left',
+    assetLimit: 0.001,
+    currency: 'CARD.CPXD',
+    decimals: 18,
+    label: 'CARD.CPXD',
+    mask: '[09999999999]{.}[999999999999999999]',
+    placeholder: '0',
+    smallThreshold: 0.003,
+    symbol: '',
+  },
```
habdelra, line 57: Let's remove the CPXD tokens. The CPXD tokens are really just analogs for their mainnet counterparts. There will only ever be one way to format a CARD or DAI token regardless of whether it is the layer1 version or the layer 2, CPXD version.

packages/cardpay-sdk/sdk/native-currencies.ts
```diff
@@ -11,6 +11,50 @@ export default {
...
+  CARD: {
+    alignment: 'left',
+    assetLimit: 0.001,
+    currency: 'CARD',
+    decimals: 18,
+    label: 'CARD',
+    mask: '[09999999999]{.}[999999999999999999]',
+    placeholder: '0',
+    smallThreshold: 0.003,
```
habdelra, line 33: A CARD token is very much smaller than an ETH token in value. It's much more similar to USD in terms of value. I think a small threshold for this can be 1.

packages/cardpay-sdk/sdk/native-currencies.ts
```diff
@@ -11,6 +11,50 @@ export default {
...
+    decimals: 18,
+    label: 'Dai',
+    mask: '[09999999999]{.}[999999999999999999]',
+    placeholder: '0',
+    smallThreshold: 0.003,
+    symbol: '',
+  },
+  CARD: {
+    alignment: 'left',
+    assetLimit: 0.001,
```
habdelra, line 27: A CARD token is very much smaller than an ETH token in value. Let's try with an asset limit of 0.1 because its so small

packages/cardpay-sdk/sdk/native-currencies.ts
backspace: ya I originally didn’t have them but I wanted the validation string to match the symbol displayed in the field and I hesitated to use hackish-seeming string replacement:

image
 
I could add an optional `currencyDisplayOverride` parameter, or do you think I should just change the string after I get it back?

habdelra: Hmmm, I guess I don’t have a real strong feeling either way. And I assume the UI designs are actually requesting that we show the .CPXD suffix in the text fields?

habdelra (approved):

packages/cardpay-sdk/sdk/native-currencies.ts
backspace: ya, exactly. I think I’ll just use string replacement for now and see what people think 😬 thanks

PR #2102, opened 2021-09-16
Load depot balances via Assets API

lukemelia:
Previously, depot balances came from the Safes API, which uses The Graph, an index that has a delay. This can result in workflows reporting incorrect depot balances immediately following a deposit or withdrawal, which is problematic. 

I'm not entirely happy with how this came out, so feel free to make suggestions for improved architecture if you've got them.

### DIFF
--- a/packages/web-client/app/components/card-pay/issue-prepaid-card-workflow/index.ts
+++ b/packages/web-client/app/components/card-pay/issue-prepaid-card-workflow/index.ts
@@ -22,6 +22,9 @@ import {
   WorkflowMessage,
   WorkflowName,
 } from '@cardstack/web-client/models/workflow';
+import { task } from 'ember-concurrency-decorators';
+import { taskFor } from 'ember-concurrency-ts';
+import { tracked } from '@glimmer/tracking';
 
 export const faceValueOptions = [500, 1000, 2500, 5000, 10000, 50000];
 export const spendToUsdRate = 0.01;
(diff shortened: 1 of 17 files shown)


packages/web-client/app/components/card-pay/issue-prepaid-card-workflow/index.ts
```diff
@@ -362,36 +365,41 @@ class IssuePrepaidCardWorkflowComponent extends Component {
...
-  workflow!: IssuePrepaidCardWorkflow;
+  @tracked workflow: IssuePrepaidCardWorkflow | null = null;
 
   constructor(owner: unknown, args: {}) {
     super(owner, args);
 
-    const workflow = new IssuePrepaidCardWorkflow(getOwner(this));
-    const persistedState = workflow.session.getPersistedData()?.state ?? {};
-    const willRestore = Object.keys(persistedState).length > 0;
+    let workflow = new IssuePrepaidCardWorkflow(getOwner(this));
```
jurgenwerk, line 373: I tend to use `const` as a default for everything but it looks like the convention here is to use `let`. I'm fine with either tbh, will remember to use `let` from now on.

jurgenwerk (approved):
I have limited knowledge about the APIs here but it looks reasonable to me. I also tested this locally - twice. I tried to issue a prepaid card when my balance was too low, then added funds and went back to the prepaid card issuance immediately, and the new balance was picked up correctly (where previously it told me my balance was too low because it wasn't refreshed).

backspace (approved):
I don’t love it either but I think it’s a fine solution until we can perhaps improve things in the SDK to consistently manage the disparate sources for us.

packages/web-client/app/components/card-pay/issue-prepaid-card-workflow/index.ts
lukemelia: Confirmed on let over const for locals. This preference in much of Ember community originated here: https://madhatted.com/2016/1/25/let-it-be.

PR #2107, opened 2021-09-16
Persist merchant creation workflow data

burieberry:
CS-1760

- [x] workflow card names
- [x] restoration errors
- [x] persistence id and transitioning changes
- [x] merchant customization card
- [x] next steps card
- [ ] prepaid card choice card changes
- [ ] tests


packages/web-client/app/components/card-pay/create-merchant-workflow/next-steps/index.ts
```diff
@@ -8,7 +8,9 @@ class CardPayCreateMerchantWorkflowNextStepsComponent extends Component<Workflow
   @service declare router: RouterService;
 
   @action returnToDashboard() {
-    this.router.transitionTo({ queryParams: { flow: null } });
+    this.router.transitionTo({
+      queryParams: { flow: null, worklowPersistenceId: null },
```
lukemelia, line 12:
```suggestion
      queryParams: { flow: null, workFlowPersistenceId: null },
```

burieberry: Had to change this to `'flow-id': null`  instead of `workflowPersistenceId: null ` for both workflows. Otherwise clicking on that button gave an error and told me to use 'flow-id'.

packages/web-client/app/components/card-pay/create-merchant-workflow/merchant-customization/index.ts
```diff
@@ -24,12 +24,29 @@ export default class CardPayCreateMerchantWorkflowMerchantCustomizationComponent
   @tracked merchantIdValidationMessage = '';
   @tracked merchantBgColorValidationMessage = '';
 
+  constructor(owner: unknown, args: WorkflowCardComponentArgs) {
+    super(owner, args);
+
+    let { merchantName, merchantId, merchantBgColor } =
+      this.args.workflowSession.state;
+
+    if (merchantName && merchantId && merchantBgColor) {
```
jurgenwerk, line 33: Maybe safer to use `isPresent` for all these args? Just to make sure they aren't empty strings

jurgenwerk (approved):
I tested this and it works great! 👍

I think it's fine to merge. There's a TODO comment left though?

burieberry:
> I tested this and it works great! 👍
> 
> I think it's fine to merge. There's a TODO comment left though?

The TODO comment is something that relates to all workflows about the inconsistency I mentioned about the confirmation request sent on the Card Wallet app. If user clicks on "Create" and then refreshes the workflow after a confirmation req is sent to their phone, the request sent remains active, which might be confusing to the user. When the page is refreshed we re-enable the "Create" button if there was no txnHash created yet. It may not be an actual problem though, since this doesn't seem to effect the result of the workflow.

PR #2108, opened 2021-09-17
Deposit flow workflow persistence

lukemelia:


jurgenwerk:
I tested this and I'm noticing a couple of problems:

1. There's a compute warning. In previous PRs we addressed this using runloop `next()`s

image

2. When I trigger the process in the Token bridge card, and reload the window during "Bridge tokens from Kovan testnet to Sokol testnet" or "Mint tokens on Sokol: DAI.CPXD", I get an error soon thereafter:

image

image

3. In case of failure in the last step like the one describe above, I think we should cancel the workflow

packages/web-client/app/components/card-pay/deposit-workflow/transaction-amount/index.ts
```diff
@@ -129,64 +154,69 @@ class CardPayDepositWorkflowTransactionAmountComponent extends Component<Workflo
...
+          this.currentTokenSymbol,
+          layer2Address,
+          this.amountAsBigNumber,
+          (txnHash) => {
+            session.update('relayTokensTxnHash', txnHash);
+          }
+        );
+      }
+      session.update('relayTokensTxnReceipt', transactionReceipt);
+      session.update('depositedAmount', this.amountAsBigNumber.toString());
```
jurgenwerk, line 215: Could use `updateMany` to update both at the same time. Actually, I think it's confusing we have both `update` and `updateMany`. Could just be one update method that accepts an object with keys and values (to remove `update` and rename `updateMany` to `update`)

jurgenwerk (changes requested):
I think the code looks ok, just a minor comment. 👍  I found a critical error though - noted it in the PR comments.

lukemelia:
Good catches @jurgenwerk. I'll get those issues resolved.

packages/web-client/app/components/card-pay/deposit-workflow/transaction-amount/index.ts
lukemelia: Left this along for now. I have an idea I'd like us to consider about this API.

lukemelia:
Fixed. The issue was that we were writing a `BN` instance to the session, which is OK when it was in memory, not OK when it needs to round-trip to localstorage.

packages/web-client/app/utils/web3-strategies/layer1-chain.ts
```diff
@@ -303,6 +309,14 @@ export default abstract class Layer1ChainWeb3Strategy
     );
   }
 
+  async resumeRelayTokens(
+    txnHash: TransactionHash
+  ): Promise<TransactionReceipt> {
+    if (!this.web3) throw new Error('Cannot unlock tokens without web3');
```
jurgenwerk, line 315: In which case can this happen - the absence of `this.web3`?

jurgenwerk (approved):
LGTM! 🚀

packages/web-client/app/utils/web3-strategies/layer1-chain.ts
lukemelia: Good question. I think it might be there to avoid a typescript "!"

PR #2109, opened 2021-09-17
Web/improve merchant payment currency handling

Aierie:
This allows the payment page to display currencies supported by Card Wallet (assuming that `packages/cardpay-sdk/sdk/native-currencies.ts` properly represents native currencies supported in Card Wallet). It also rounds supported native currencies + SPEND to an appropriate amount of decimals, rounding up each time.

### DIFF
--- a/packages/web-client/tests/acceptance/pay-test.ts
+++ b/packages/web-client/tests/acceptance/pay-test.ts
@@ -6,9 +6,11 @@ import sinon from 'sinon';
 
 import { MirageTestContext } from 'ember-cli-mirage/test-support';
 import {
+  convertAmountToNativeDisplay,
   formatUsd,
   generateMerchantPaymentUrl,
   MerchantSafe,
+  roundAmountToNativeCurrencyDecimals,
   spendToUsd,
 } from '@cardstack/cardpay-sdk';
 import { getResolver } from '@cardstack/did-resolver';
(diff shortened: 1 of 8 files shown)


packages/web-client/tests/acceptance/pay-test.ts
```diff
@@ -192,6 +229,49 @@ module('Acceptance | pay', function (hooks) {
...
+    let expectedUrl = generateMerchantPaymentUrl({
+      domain: universalLinkDomain,
+      network,
+      merchantSafeID: merchantSafe.address,
+      currency: jpySymbol,
+      amount: Number(roundedJpyAmount),
+    });
+    assert.dom(QR_CODE).hasAttribute('data-test-styled-qr-code', expectedUrl);
+    assert.dom(PAYMENT_URL).containsText(expectedUrl);
+  });
```
backspace, line 273: I don’t know where this belongs exactly (probably not this PR!) but I’ve never seen this representation of the local currency:

image

It would usually be `C$100` or `$100 CAD` but not both (nor is `CA` really a thing when it comes to currency, that I’ve seen)

backspace (approved):
🎉

packages/web-client/tests/acceptance/pay-test.ts
Aierie: This is leftover from rainbow. Let's change it if we can settle on a better way to represent it (and if it doesn't break any assumptions of mobile design), issue in CS-1897.

PR #2111, opened 2021-09-17
remove async module callbacks

Aierie:
Qunit complains about these


backspace (approved):
thanks for finding these, I’ve been wondering about this message for a few days now

PR #2116, opened 2021-09-20
Comma grouping of numbers

Aierie:
This PR adds separator commas to various currency amounts in the DApp. This was implemented by:
- Adding a new `formatCurrencyAmount` function to the SDK's currency utils that formats a `BigNumberish` to a string with separators + the specified amount of decimals
- Changing `format-token-amount` to `format-wei-amount` to reflect that it is supposed to act on wei amounts
- Making `format-wei-amount` use the new `formatCurrencyAmount` utility
- Making `format-usd` use the older `convertAmountToNativeDisplay` utility and removes the `formatUsd` utility, hence providing it with separator commas. This changes its behaviour when dealing with amounts < 1 (see tests), and removes the ability to specify prefix/suffixes.
- Adding a `format-amount` helper/utility to format SPEND

There are a few issues with this PR:
- Which components should be formatting amounts, and which components should rely on their parents/context formatting amounts? 
- Should we be using something similar to `convertAmountToNativeDisplay` for tokens (which has responsibility for adding the correct suffixes and prefixes) and bypassing the necessity for a `format-wei-amount` helper? If we do so what is a good escape hatch for customization?

### DIFF
--- a/packages/cardpay-sdk/sdk/currency-utils.ts
+++ b/packages/cardpay-sdk/sdk/currency-utils.ts
@@ -343,3 +343,7 @@ export const formatUsd = (value: BigNumberish, options: FormatUsdOptions = { sym
   }
   return result;
 };
+
+export const formatCurrencyAmount = (amount: BigNumberish, decimalPlaces = 2) => {
+  return new BigNumber(amount).toFormat(decimalPlaces);
+};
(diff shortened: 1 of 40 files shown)


backspace (approved):
image

Nice!

re your question of which components should be formatting amounts, I tend to favour pushing as much of that into the leaf nodes as possible, I don’t love having to remember to pass in pre-formatted things, it makes it easier to produce inconsistent rendering

lukemelia (approved):

jurgenwerk (approved):
LGTM!

PR #2117, opened 2021-09-20
drop dependency on `esm`

ef4:
The `esm` package is a polyfill for ES modules in node. It is unmaintained now that node has native support. Node's native support however is stricter and hard to adopt. Our best immediate path forward is to allow typescript to emit commonjs instead of ES modules.

This was all motivated by the need to support Node 14+ features that allow customizing the `exports` key in package.json. `esm` does not implement this which breaks the card compiler.


ef4:
Requested review from @pcjun97 because I don't know enough about the test coverage in cardie. It seems like it was probably not really relying on `esm` at all, because it has ts-node instead.

ef4:
Requested review from @habdelra because I don't know for sure if there are other CLI commands that aren't covered by CI that I could be breaking here.

habdelra:
@ef4 no prob. I'll check out this branch and smoke test the CLI

tsconfig.json
```diff
@@ -1,8 +1,7 @@
 {
   "compilerOptions": {
     "target": "es2017",
-    "module": "es6",
-    "moduleResolution": "Node",
+    "module": "commonjs",
```
ef4, line 4: This is the big cross-cutting change here. The shared typescript compile will now emit commonjs instead of esm.

habdelra (approved):
I did a smoke test of the CLI in this branch and everything works just fine 👍

PR #2118, opened 2021-09-20
web: Change Sentry to remove host for source maps

backspace:
As documented here, this will cause Sentry to strip the
domain when uploading source maps:
https://docs.sentry.io/platforms/javascript/guides/gatsby/sourcemaps/uploading/multiple-origins/

As a bonus, this simplifies the web client deployment action because we no longer need to care about deployment domain.

An [example](https://sentry.io/organizations/cardstack/issues/2661042522/?project=5879708&query=is%3Aunresolved&statsPeriod=14d) stacktrace on `wallet-staging.stack.cards` before this change:

image

When I temporarily overrode the `push-main` workflow to run on this branch and used `~` as the Sentry URL prefix, the stacktrace [showed](https://sentry.io/organizations/cardstack/issues/2558539178/?project=5879708&query=is%3Aignored&statsPeriod=14d) properly:

image

### DIFF
--- a/.github/actions/deploy-web-client/action.yml
+++ b/.github/actions/deploy-web-client/action.yml
@@ -9,9 +9,6 @@ inputs:
   environment:
     description: Deployment environment
     required: true
-  sentry_url_prefix:
-    description: Adds a prefix to source map urls after stripping them
-    required: true
 
 runs:
   using: composite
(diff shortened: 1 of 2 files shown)


lukemelia (approved):
👍👍👍

PR #2121, opened 2021-09-21
Deploys should run when deploy code has changed

ef4:
Without this change, you can easily introduce breakage into push-main.yml or waypoint.hcl and not notice until much later when one of the projects is updated.

This change ensures that all the project deploys will run if you touch their deploy code.


lukemelia:
@pcjun97 please take a look

PR #2122, opened 2021-09-21
ci: Fix waypoint secrets in push-main

pcjun97:


lukemelia (approved):
Whoops, that would do it!

PR #2133, opened 2021-09-22
Cancel workflow if hub auth fails during merchant and prepaid card creation

burieberry:
Initial implementation - Added a Restart Workflow button and functionality to the card for any failure other than user rejection

Changes - Made the workflow cancellation specific to the hub auth failures mentioned in the ticket (CS-1707). Also, instead of just displaying a restart button, called for the cancellation of the workflow. This means the workflow will be frozen and the DefaultCancelationCta component will appear.

I'm not sure if this type of cancellation from within a card is a good idea, though. Another option would be to follow the event listener pattern to listen for hub auth token nullification. 

I also need to find a good place to add the tests


lukemelia (approved):
Pending a green build, looks good

backspace (approved):
🎉

PR #2135, opened 2021-09-23
log websocket events to Sentry

Aierie:
Currently the bug in CS-1800 does not affect a large amount of users and there is little useful information in Sentry on the root cause of this issue. This PR adds some logging about websocket status changes, in hopes that we get more information about what we're trying to fix.


lukemelia (approved):
👍

PR #2137, opened 2021-09-23
web: Add current card name to workflow persistence

backspace:
The design for #2127 workflow restoration includes display names for the persisted workflow and its current card, as well as a progress indicator for how many milestones have been completed. The current persistence scheme includes a `name` and `state.{completedMilestonesCount,milestonesCount}` that are sufficient to render everything but the current card name. At first I tried reconstructing the workflow from the persisted state to calculate the current card but it was brittle and wasteful, so this has `WorkflowCard`s store their names in the session upon `reveal` instead.

I’m unsure whether `onRevealed` makes sense and whether assuming an `AnimatedWorkflow` is okay.

This only stores the internal `name` so far, I’ll be extracting display names from templates… 😯

### DIFF
--- a/packages/web-client/app/components/card-pay/issue-prepaid-card-workflow/confirmation/index.hbs
+++ b/packages/web-client/app/components/card-pay/issue-prepaid-card-workflow/confirmation/index.hbs
@@ -1,5 +1,5 @@
 <ActionCardContainer
-  @header="Prepaid card"
+  @header={{@displayName}}
   @isComplete={{@isComplete}}
   class="issue-prepaid-card-workflow-confirmation"
 >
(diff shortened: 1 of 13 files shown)


lukemelia:
A few thoughts:

`onRevealed` seems OK, and assuming an AnimatedWorkflow is not great but is OK. Maybe we can refactor onRevealed to fire in both cases?

Seems like we should refactor cards to derive display names from the .ts file so that it is easier to capture in a consistent way.

backspace:
I couldn’t find a place to call `onRevealed` in a non-animated workflow but I’m probably missing something.

re moving the display names to `.ts`, that’s what I did in cfd7919 for just prepaid card issuance, thoughts? I don’t love that `cardDisplayName` will need to be repeated for every (for instance) `layer-two-connect-card` instantiation, but that’s already the case for `cardName`, so…

backspace:
I’m closing this as it turns out the current milestone is to be displayed, not the current card name! I’ll open another with a similar approach.

PR #2140, opened 2021-09-23
CS-1377 Create prepaid card reservation service

habdelra:
This introduces 3 new server routes to support prepaid card reservations:
- GET `/api/inventory` This returns the prepaid card inventory taking into account prepaid cards that have been reserved
- POST/GET `/api/reservations` This allows the client to create a reservation for a prepaid card SKU, as well as to view the reservation (and reservation status)
- POST/GET `/api/order` This allow the client to create a wyre order/reservation correlation, as well as to view the order (and associated reservation)

`GET /api/reservations/:reservation_id` will allow the client to see the provisioning status. The client will need to poll this after they have submitted the wyre order. The status will update as the various wyre callbacks are invoked, and then finally the prepaid card will be assigned and reflected in this response. 

Note that this requires `PROVISIONER_SECRET` env var to be set in order to work properly. This value can be found in the AWS secrets manager for the pertinent environment.

- [ ] Also, this will require a subgraph deployment after being merged


packages/hub/routes/utils/orders.ts
```diff
@@ -0,0 +1,181 @@
...
+  reservation   \          |
+                 V         V
+                {PROVISIONING}
+                     /    \
+  provisioned-mined /      \ provision-error
+                   /        \
+                  V          V
+             {COMPLETE}    {ERROR-PROVISIONING}
+
+*/
```
habdelra, line 56: This state machine is important for understanding how prepaid cards are provisioned

habdelra: (amazingly github co-pilot did most of the drawing here)

mattmcmanus:
@habdelra I have  PR up for some changes to how we create the hub server and utilize things in tests. I think it's simplifies things a bunch but the order of merges matter here. If this is merged first, I can add the changes to mine. https://github.com/cardstack/cardstack/pull/2142

habdelra:
thanks for the heads up!

PR #2142, opened 2021-09-24
Refactor hub server creation

mattmcmanus:
These are some changes that came out of my compiler-server merge that are self contained. This will make it easier to reason about how the server is created. I also change how the server is handled in tests. Rather than booting and listening on a port, it passed a callback to superagent. This allows us to remove a lot of the complicated even listening and other shenanigans.


habdelra (approved):
Nice, looks like a very pleasant refactor

packages/hub/main.ts
```diff
@@ -79,58 +85,74 @@ export function wireItUp(registryCallback?: RegistryCallback): Container {
...
+      .use(environment)
+      .use(cors({ origin: '*', allowHeaders: 'Authorization, Content-Type, If-Match, X-Requested-With' }))
+      .use(httpLogging);
+
+    app.use(((await container.lookup('authentication-middleware')) as AuthenticationMiddleware).middleware());
+    app.use(((await container.lookup('development-proxy-middleware')) as DevelopmentProxyMiddleware).middleware());
+    app.use(((await container.lookup('api-router')) as ApiRouter).routes());
+    app.use(((await container.lookup('callbacks-router')) as CallbacksRouter).routes());
+
+    function onError(err: Error, ctx: Koa.Context) {
```
habdelra, line 113: @mattmcmanus it's a little tiny thing, but since you are already in there, could you add this? It's a huge quality of life improvement when unexpected errors occur:
```js 
logger.error(`Unhandled error:`, err);
```

mattmcmanus: Added

PR #2144, opened 2021-09-24
add meta (created and updated dates) to persisted data for a workflow

Aierie:
Will update this after #2134 is merged.

### DIFF
--- a/packages/web-client/tests/unit/models/workflow/workflow-session-test.ts
+++ b/packages/web-client/tests/unit/models/workflow/workflow-session-test.ts
@@ -1,6 +1,8 @@
 import { module, test } from 'qunit';
 import { setupTest } from 'ember-qunit';
-import WorkflowSession from '@cardstack/web-client/models/workflow/workflow-session';
+import WorkflowSession, {
+  buildState,
+} from '@cardstack/web-client/models/workflow/workflow-session';
 import WorkflowPersistence from '@cardstack/web-client/services/workflow-persistence';
 import Ember from 'ember';
 import BN from 'bn.js';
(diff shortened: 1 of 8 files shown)


backspace (approved):
This will definitely improve #2127. I’m happy to see the addition of `meta` as `state`’s chaos has been gnawing at me, I wonder if `completedCardNames`, `completedMilestonesCount`, and `milestonesCount` belong there too? 🤔

Aierie:
> This will definitely improve #2127. I’m happy to see the addition of meta as state’s chaos has been gnawing at me, I wonder if completedCardNames, completedMilestonesCount, and milestonesCount belong there too? 🤔

I think they might belong in `meta`, `isCanceled` and `cancelationReason` probably do too?

backspace:
YES 🎉

`localStorage` is going to be a jumble of evolving schema 😳

backspace:
hmm but writing to `meta` from outside the persistence service would be sad times 🤔

lukemelia:
Yes, and Meta should a typescript type that is added to SupportedTypes for the WorkflowSession

Aierie:
Moved `meta` to `state` since we want to add other properties to meta.

packages/web-client/tests/unit/models/workflow/workflow-session-test.ts
```diff
@@ -4,6 +4,8 @@ import WorkflowSession from '@cardstack/web-client/models/workflow/workflow-sess
 import WorkflowPersistence from '@cardstack/web-client/services/workflow-persistence';
 import Ember from 'ember';
 import BN from 'bn.js';
+import sinon from 'sinon';
+import Sinon from 'sinon';
```
backspace, line 8:
```suggestion
import { SinonFakeTimers } from 'sinon';
```

Maybe this would be a bit less strange-looking 🤔

packages/web-client/tests/unit/models/workflow/workflow-session-test.ts
```diff
@@ -574,4 +605,76 @@ module('Unit | WorkflowSession model', function (hooks) {
...
+      meta.createdAt,
+      startDateString,
+      `Created date has the start ISO date string: ${startDateString}`
+    );
+    assert.equal(
+      meta.updatedAt,
+      updatedDateString,
+      `Updated date has an updated ISO date string: ${updatedDateString}`
+    );
+  });
```
backspace, line 679: Should there be a test that writes arbitrary data to `meta`?

backspace (approved):
If we’re agreeing that those other properties should be part of `meta`, what do you think of including that in this PR? I could try it myself too if it’s too much 🤓

packages/web-client/tests/unit/models/workflow/workflow-session-test.ts
Aierie: There probably should be, will add, thanks

Aierie:
> If we’re agreeing that those other properties should be part of `meta`, what do you think of including that in this PR? I could try it myself too if it’s too much 🤓

I think it makes sense to add those to this PR, but wasn't sure if it was too much. Will add them

packages/web-client/tests/unit/models/workflow/workflow-session-test.ts
```diff
@@ -606,8 +607,91 @@ module('Unit | WorkflowSession model', function (hooks) {
...
+      workflowPersistenceId: ID,
+    });
+    assert.throws(() => {
+      subject.setValue('meta', 'something');
+    }, 'Please use setMeta to set meta values');
+
+    assert.throws(() => {
+      subject.setValue({ meta: 'something-else' });
+    }, 'Please use setMeta to set meta values');
+  });
```
backspace, line 627: love it, thanks

backspace (approved):
this is a great step in improving the interface for persisted workflows, thanks

PR #2161, opened 2021-09-27
Preselect prepaid card if there is only 1; disable low-balance cards

burieberry:
CS-1896

Prepaid card selection card-picker:
- If the user only has 1 card (with sufficient balance), preselect that card
- Disable and fade out low balance cards


Aierie (approved):
This works well!

Possible improvement is sorting the prepaid cards so that disabled ones are at the bottom.

lukemelia (approved):
👍

PR #2164, opened 2021-09-28
More robust workflow state comparison

Aierie:
- `WorkflowSession` recognizes `Workflow` in Typescript
- Deserializes actual state to `Record<string, SupportedType>` rather than serializing the expected state so that `deepEqual` will not be affected by the order of properties specified, or string formatting.


Aierie:
Didn't apply these changes in `workflow-session-test.ts` because it feels like there is value in asserting against strings in the lower level implementation, while here it seems like we just want values to match.

jurgenwerk (approved):
LGTM!

lukemelia:
I like the idea and most of the implementation. There are a few things I'd like to review together with you.

PR #2165, opened 2021-09-28
Workflow persistence boilerplate refactor

jurgenwerk:
This PR:

1. Extracts the restoration logic out of the components and moves it to the `Workflow` model 
2. Moves the responsibility of getting the URL's `flow-id` query param from the `Workflow` model to the component 
3. Delays assigning the workflow to workflow components to the point where the async restoration validation checks and the restoration itself have been finished. This is to avoid unnecessary recomputation on restore, and allows us to remove some of the `next`s that we use  in the restoration flow to mitigate the "recomputation warning"
4. Replaces `MockWorkflow`, `ConcreteWorkflow`, `StubWorkflow` with a `WorkflowStub` so that we don't use so many different names for the same thing

### DIFF
--- a/packages/web-client/app/models/workflow.ts
+++ b/packages/web-client/app/models/workflow.ts
@@ -232,9 +232,7 @@ export abstract class Workflow {
       this.session.getMeta().isCancelled &&
       this.session.getMeta().cancelationReason
     ) {
-      next(this, () => {
-        this.cancel(this.session.getMeta().cancelationReason);
-      });
+      this.cancel(this.session.getMeta().cancelationReason);
     }
   }
 }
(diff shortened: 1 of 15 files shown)


packages/web-client/app/models/workflow.ts
```diff
@@ -51,10 +50,29 @@ export abstract class Workflow {
   workflowPersistence: WorkflowPersistence;
   workflowPersistenceId?: string;
 
-  constructor(owner?: any) {
+  abstract restorationErrors(state: any): string[];
```
jurgenwerk, line 53: My reasoning was that since the concrete workflows need to implement validation and restoration checks, `abstract` makes sense here. However, this comes with extra burden of needing to implement these methods on every instantiation, which is reflected in the later commit where I introduce `WorkflowStub` that is used in unit and integration tests.

lukemelia (approved):

burieberry (approved):

PR #2180, opened 2021-09-29
add workflow name to workflow session test

Aierie:
The name persisted when `WorkflowSession.persistToStorage` is called is taken from a workflow session's workflow.

Alternatively, we could cast the expected persisted state to `WorkflowPersistencePersistedData`.

### DIFF
--- a/packages/web-client/tests/unit/models/workflow/workflow-session-test.ts
+++ b/packages/web-client/tests/unit/models/workflow/workflow-session-test.ts
@@ -42,12 +42,13 @@ module('Unit | WorkflowSession model', function (hooks) {
   test('get string value', function (assert) {
     let workflowPersistence = new WorkflowPersistence();
     workflowPersistence.persistData(ID, {
-      name: 'EXAMPLE',
+      name: 'PREPAID_CARD_ISSUANCE',
       state: {
         myKey: '{ "value": "myValue" }',
       },
     });
     let subject = new WorkflowSession({
+      name: 'PREPAID_CARD_ISSUANCE',
       workflowPersistence,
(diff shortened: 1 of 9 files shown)


Aierie:
Unsure if https://github.com/cardstack/cardstack/pull/2180/commits/7cc5be3cfd66a008f8a19b45d7f001247d340200 is needed

packages/web-client/tests/unit/models/workflow/workflow-session-test.ts
```diff
@@ -42,12 +42,13 @@ module('Unit | WorkflowSession model', function (hooks) {
   test('get string value', function (assert) {
     let workflowPersistence = new WorkflowPersistence();
     workflowPersistence.persistData(ID, {
-      name: 'EXAMPLE',
+      name: 'PREPAID_CARD_ISSUANCE',
```
backspace, line 45: weird, this is one of the first things I tried, thinking that it was because it wasn’t a known `name`. I guess I did something else wrong!

backspace (approved):
I don’t know about the `abstract`ing either, but thanks for figuring this out

backspace:
In the interest of a smaller PR, I just pulled the one commit into #2127. But if the `abstract` work is sensible it could be a separate PR. Thank you!!!! 🎉

packages/web-client/tests/unit/models/workflow/workflow-session-test.ts
Aierie: The change that matters is setting the `name` property on the object that we pass to the `WorkflowSession` constructor, this is just changed to be consistent/less confusing.

PR #2184, opened 2021-09-29
Add updating of workflow count on storage event

backspace:
I’m opening this is a separate PR to get feedback on whether this awkwardness is worth it for getting testing around having a `storage` event trigger a recalculation of the count of restorable workflows.

The test is as close as I could get to the true situation of another tab updating `localStorage` with a newly-persisted or -updated workflow. It appends a blank `iframe` as a child of the test page and updates its `localStorage`, which will trigger the `storage` event handler in the workflow persistence service. Noteworthy is that in tests, `localStorage` isn’t even being used for workflow persistence, but this technique works to be able to react to the `storage` event, which seems hard to produce artificially.

Here it is in action:

storage-event-screencast 2021-09-29 15-29-33

### DIFF
--- a/packages/web-client/app/services/workflow-persistence.ts
+++ b/packages/web-client/app/services/workflow-persistence.ts
@@ -24,22 +24,40 @@ export default class WorkflowPersistence extends Service {
   @service declare router: RouterService;
 
   #storage!: Storage | MockLocalStorage;
+  __storage: MockLocalStorage | undefined;
+
+  #handleStorageEvent: () => void;
 
   @tracked persistedDataIds: string[] = [];
 
   constructor() {
     super(...arguments);
 
-    let entries;
+    let entries: Record<string, string>;
 
     if (config.environment === 'test') {
       this.#storage = new MockLocalStorage();
+      this.__storage = this.#storage;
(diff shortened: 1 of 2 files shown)


lukemelia:
Not worth it IMO for the reasons we discussed in Sprint planning: solves for tabs with localstorage-specific complexity and doesn't address cross-decice syncing.

backspace:
so do you think it’s not worth having a listener for `storage` at all? Or just not have it tested?

lukemelia:
I was thinking not to have it at all. Open to other perspectives though.

Aierie:
I think not having it makes the workflow queue pretty confusing for someone who has multiple tabs open

Aierie:
Thinking specifically that this is confusing:


https://user-images.githubusercontent.com/39579264/135502143-257bd9bd-d71b-476f-8094-b27235f678d3.mov

burieberry (approved):
Since this is already done and is working for the same browser, we might as well add it.

lukemelia:
>Since this is already done and is working for the same browser, we might as well add it.

OK, I'm convinced.

PR #2192, opened 2021-09-30
Fix conflicting SVG part IDs

lukemelia:
ember-svg-jar was minifying the id attributes within our SVGs to the point that they conflict with the IDs in other SVGs when they are both inlined into the page at the same time. This PR to disables minifying those IDs and gives unique IDs to the SVGs that have this issue (likely from having the same damage done to them by ImageOptim).


Aierie:
The config specified in this issue (https://github.com/ivanvotti/ember-svg-jar/issues/170) can also provide unique ids, though we might run into a similar problem as they did - ids being unique per file, but not per instance of inline svg, leading to `url(#id)` resolving to nothing when the originally resolved element was removed from the DOM.

burieberry (approved):
yeah I've seen this happen. thanks!

PR #2193, opened 2021-09-30
Workflow tracker list UI improvements

burieberry:
CS-2044

### DIFF
--- a/packages/web-client/app/components/card-pay/header/workflow-tracker/dropdown-list/index.ts
+++ b/packages/web-client/app/components/card-pay/header/workflow-tracker/dropdown-list/index.ts
@@ -3,15 +3,43 @@ import { inject as service } from '@ember/service';
 import { action } from '@ember/object';
 
 import WorkflowPersistence from '@cardstack/web-client/services/workflow-persistence';
+import { tracked } from '@glimmer/tracking';
 
 interface CardPayHeaderWorkflowTrackerDropdownListArgs {}
 
 export default class CardPayHeaderWorkflowTrackerDropdownList extends Component<CardPayHeaderWorkflowTrackerDropdownListArgs> {
   @service declare workflowPersistence: WorkflowPersistence;
+  element!: HTMLElement;
+  initialPosition?: number;
+  @tracked currentPosition?: number;
+
+  @action getScrollPosition(element: HTMLElement) {
+    this.element = element;
+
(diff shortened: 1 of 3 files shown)


packages/web-client/app/components/card-pay/header/workflow-tracker/dropdown-list/index.ts
```diff
@@ -3,15 +3,62 @@ import { inject as service } from '@ember/service';
...
+  }
+
+  get isScrolled() {
+    return Boolean(
+      this.currentPosition && this.currentPosition !== this.initialPosition
+    );
+  }
+
+  lockBodyScroll() {
+    let body = document.getElementsByTagName('body')[0];
```
Aierie, line 43: this is the same as `document.body`

packages/web-client/app/components/card-pay/header/workflow-tracker/dropdown-list/index.ts
```diff
@@ -3,15 +3,62 @@ import { inject as service } from '@ember/service';
...
+    element.onblur = this.unlockBodyScroll;
+    element.onmouseover = this.lockBodyScroll;
+    element.onmouseout = this.unlockBodyScroll;
+  }
+
+  get isScrolled() {
+    return Boolean(
+      this.currentPosition && this.currentPosition !== this.initialPosition
+    );
+  }
```
Aierie, line 40: isScrolled can be determined via `this.element.scrollTop > 0` and set in the event listener, so we don't need to use the children's positions to determine whether the contents of an element are scrolled or not:

https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollTop

Aierie (approved):
This works nicely!

packages/web-client/app/components/card-pay/header/workflow-tracker/dropdown-list/index.ts
burieberry: thanks!

burieberry: > `isScrolled` can be determined via `this.element.scrollTop > 0` and set in the event listener, so we don't need to use the children's positions to determine whether the contents of an element are scrolled or not:
> 
> https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollTop

However, this doesn't tell me which list is scrolling.

PR #2197, opened 2021-10-04
Workflow cancellation refactor

jurgenwerk:
This PR:

1. Adds restoration cancellation messages (more context [here in discord](https://discord.com/channels/584043165066199050/855105564643819530/894502990361296916))
2. Renames occurrences of "cancelled" to "canceled"
3. Filters out canceled workflows from the active workflow selector

QUESTION ⁉️ 🚨: There is a potential bug introduced in (2) but we can decide to ignore it. The problem is that the currently persisted workflows have the property written as `isCancelled` while now we use `isCanceled`. This means the currently saved workflows that are canceled will still be shown in the workflow selector UI. Can someone give us an idea to what extent this feature has been exercised on staging and production? If we want to 100% support this filter for historically persisted canceled workflows then we can add another check (to support both `isCancelled` and `isCanceled`).

### DIFF
--- a/packages/web-client/app/components/card-pay/create-merchant-workflow/index.ts
+++ b/packages/web-client/app/components/card-pay/create-merchant-workflow/index.ts
@@ -27,6 +27,7 @@ import RouterService from '@ember/routing/router-service';
 import WorkflowPersistence from '@cardstack/web-client/services/workflow-persistence';
 import { formatAmount } from '@cardstack/web-client/helpers/format-amount';
 import { tracked } from '@glimmer/tracking';
+import { isPresent } from '@ember/utils';
 
 const FAILURE_REASONS = {
   UNAUTHENTICATED: 'UNAUTHENTICATED',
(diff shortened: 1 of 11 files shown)


packages/web-client/app/components/card-pay/create-merchant-workflow/index.ts
```diff
@@ -300,22 +301,12 @@ class CreateMerchantWorkflow extends Workflow {
...
-            FAILURE_REASONS.DISCONNECTED,
-            FAILURE_REASONS.ACCOUNT_CHANGED,
-            FAILURE_REASONS.NO_PREPAID_CARD,
-            FAILURE_REASONS.INSUFFICIENT_PREPAID_CARD_BALANCE,
-            FAILURE_REASONS.RESTORATION_UNAUTHENTICATED,
-            FAILURE_REASONS.RESTORATION_L2_ACCOUNT_CHANGED,
-            FAILURE_REASONS.RESTORATION_L2_DISCONNECTED,
-          ] as String[]
-        ).includes(String(this.workflow?.cancelationReason));
+        return isPresent(this.workflow?.cancelationReason);
```
jurgenwerk, line 309: We can show the cta for all known failures

packages/web-client/app/components/card-pay/withdrawal-workflow/index.ts
```diff
@@ -359,13 +359,33 @@ with Card Pay.`,
...
+      },
+    }),
     new WorkflowCard({
       author: cardbot,
       componentName: 'workflow-thread/default-cancelation-cta',
       includeIf() {
-        return (Object.values(FAILURE_REASONS) as String[]).includes(
-          String(this.workflow?.cancelationReason)
-        );
+        return isPresent(this.workflow?.cancelationReason);
```
jurgenwerk, line 388: We can show the cta for all known failures

packages/web-client/tests/acceptance/persistence-view-restore-test.ts
```diff
@@ -134,9 +134,18 @@ module('Acceptance | persistence view and restore', function () {
       });
 
       workflowPersistenceService.persistData('canceled', {
-        name: 'PREPAID_CARD_ISSUANCE',
+        name: `PREPAID_CARD_ISSUANCE`,
         state: buildState({
           meta: {
+            completedCardNames: [
```
jurgenwerk, line 140: Workflows can get canceled in the middle of the progress and their whole state will get persisted.

backspace (approved):
Looks good, thanks. I’ve been concerned about the constantly-shifting schema as we iterate on this, I wonder if once we release it to beta testers we can ask them for a one-time `localStorage.clear()` 🤔

lukemelia:
We haven't released to production yet, so I think we are OK.

lukemelia (approved):

backspace:
Some persistence has been released at least! 😯

image

lukemelia:
> Some persistence has been released at least! 😯

Ah, so it has! I guess the clearing localstorage bit is the way to go for this release than.

PR #2198, opened 2021-10-04
[Refactor]: Split currencies and add types

danibonilha:
This PR adds types to the currencies, also it splits them into "contexts", I was wondering about splitting  into Fiat/Crypto instead of Native/Crypto, but since SPD is not fiat I ended up using Native/Crypto. This adds the ability to use one of those or all together, also it's type safe and we get rid of a couple of `anys` in the mobile side. This is a suggestion though, feel free to change whatever is needed to match the expected behavior.


Aierie (commented):
I think it's great to have these types! Not sure how they will evolve in the future but seems good enough for now. I'm wondering: 

1. What does it mean for a currency to be native? Crypto is clear, but native is quite vague and seems like it was tied to Rainbow's business logic. Also, ETH is included as a native currency in Card Wallet: https://github.com/cardstack/cardwallet/blob/550604029566351b7a4bfff781864705d2ad7b20/src/references/native-currencies.json#L2-L12, might cause issues with types
2. When we change something in `currencies.ts` (I renamed `native-currencies` to reflect its role more clearly), do we need folks from both mobile and web team to review? If so, could we leave a note at the top of the file about this?

topwebtek7 (approved):
Looks good to me but needs web dapp team reviews, I think

danibonilha:
> I think it's great to have these types! Not sure how they will evolve in the future but seems good enough for now. I'm wondering:
> 
> 1. What does it mean for a currency to be native? Crypto is clear, but native is quite vague and seems like it was tied to Rainbow's business logic. Also, ETH is included as a native currency in Card Wallet: https://github.com/cardstack/cardwallet/blob/550604029566351b7a4bfff781864705d2ad7b20/src/references/native-currencies.json#L2-L12, might cause issues with types
> 2. When we change something in `currencies.ts` (I renamed `native-currencies` to reflect its role more clearly), do we need folks from both mobile and web team to review? If so, could we leave a note at the top of the file about this?

1. I believe Rainbow doesn't split anything, and having it as Native, means it's the currency the user can select across the whole app, and I think on mainnet they only support ETH from crypto currencies besides fiat, but @topwebtek7 probably has a better understanding about this than me. Good catch that rainbow uses ETH alongside with "fiat", we can't forget that in the app.
2. Thanks for renaming it. I think it would be good to have shared reviews, at least while we don't have integration tests, not sure though where to add the comments since there are a couple of files dealing with currencies, we could probably just shared in the channel, whenever anyone messes with currencies, let's double check on mobile side. 

If it looks good, feel free to merge it.

PR #2201, opened 2021-10-04
web: Add Card Space profile card

backspace:
This adds a card that represents a Card Space with placeholder defaults and the option to mark it “Profile Preview”.

An example with values passed in including images already present in the repository:

image

I didn’t want to commit any external references but here it is with more-representative images:

buck-Cardstack 2021-10-05 11-48-35

The placeholders:

image

This also adds a `/card-space/profile-card-temp` for ease of review until the profile card shows elsewhere in the application, but this is easily removed if desired.


packages/web-client/app/components/card-space/profile-card/index.css
```diff
@@ -0,0 +1,108 @@
...
+  font: 600 var(--boxel-font-sm);
+  color: var(--boxel-purple-500);
+}
+
+.card-space-profile-card__category {
+  display: inline-block;
+  margin-top: var(--boxel-sp-xxxs);
+  padding: 2px var(--boxel-sp-xxxs) 2px var(--boxel-sp-xxxs);
+  color: var(--boxel-purple-300);
+  font: 600 0.55rem var(--boxel-font-family);
```
backspace, line 97: Is there a possibility of having `--boxel-font-xxs`? I looked at adding it but I couldn’t understand the ratios between sizes (and line heights).

lukemelia: I don't see why not, though I have doubts about whether we should use fonts that small.

lukemelia (approved):
Nice CSS work!

Aierie (changes requested):
Not entirely sure about this, but I think the profile preview card is different from the Card Space user page profile card, the text "Profile Preview" seems to refer to it being the profile preview others get of your space, not to a preview of your profile card. I think the wording is a bit confusing - there are two similarly named cards, and you don't know the purpose of the profile preview as opposed to the profile card. Would be good to confirm with design.

Usage of profile preview card below:

https://app.zeplin.io/project/6155e2f8f897cbaf80bf7e38/screen/6155e32df8effdb23b00ae9b


Profile preview card side by side with the (Card Space user page) profile card below:

image

It would be good to fix this card's size, given the intention to use it in a grid and have all information available at a glance. Have asked design about this, and they've agreed that we should add a content restriction at the description field for the profile + profile preview (~2 lines at most). Sorry that I didn't update the linear issue with this context! (https://app.zeplin.io/project/6155e8fcfc362bb18ff50b67/screen/6155e9f0fe8362b2c5001cc8)

jurgenwerk (approved):

packages/web-client/app/components/card-space/profile-card/index.css
backspace: ya I would generally avoid it but it’s in the design… are you the person who set up the font variables and ratios, @burieberry? What should I do 😯

PR #2203, opened 2021-10-05
Remove markdown code blocks

Aierie:
CS-2063

- Changed to a markdown renderer that allows disabling of the code blocks feature in its API (commonmark required walking and replacing in the AST)
- Replace a bunch of single quotes with apostrophes ~~since this new renderer does not replace them automatically.~~

EDIT: turns out it does, its default is not commonmark but commonmark + some extra stuff and using the commonmark config removes the quote replacement functionality. Thanks @jurgenwerk for highlighting the importance of the quotes. I probably made a mistake the first time round trying this in the node REPL.


jurgenwerk (commented):
Wouldn't it be better if the apostrophe processing/replacing happened in the helper?

Aierie:
> Wouldn't it be better if the apostrophe processing/replacing happened in the helper?

It would be a better developer experience. I think replacing single quotes in text/applying spell checking is a separate problem though, and it feels like it's not a very easy one - though I haven't explored enough to say for sure.

jurgenwerk:
It would be an easy mistake to use a single quote instead of the apostrophes and then the formatting will not work as intended. That's why I think it's worth exploring adding this processing step to the helper instead

jurgenwerk (approved):
LGTM!

PR #2204, opened 2021-10-05
Allow completed flows to be restored via tracker UI

Aierie:
Tried restoring all 4 completed workflows, I think the deposit workflow's transaction status card was the only one which needed fixing.

### DIFF
--- a/packages/web-client/app/components/card-pay/deposit-workflow/transaction-status/index.ts
+++ b/packages/web-client/app/components/card-pay/deposit-workflow/transaction-status/index.ts
@@ -56,7 +56,11 @@ class CardPayDepositWorkflowTransactionStatusComponent extends Component<Workflo
 
   constructor(owner: unknown, args: WorkflowCardComponentArgs) {
     super(owner, args);
-    if (!this.completedLayer2TxnReceipt) {
+    if (this.completedLayer2TxnReceipt) {
+      this.completedStepCount = 3;
+      this.blockCount = this.totalBlockCount;
+      this.args.onComplete?.();
+    } else {
       taskFor(this.waitForBlockConfirmationsTask)
         .perform()
         .catch((e) => {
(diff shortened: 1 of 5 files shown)


packages/web-client/app/components/card-pay/deposit-workflow/transaction-status/index.ts
```diff
@@ -56,7 +56,11 @@ class CardPayDepositWorkflowTransactionStatusComponent extends Component<Workflo
 
   constructor(owner: unknown, args: WorkflowCardComponentArgs) {
     super(owner, args);
-    if (!this.completedLayer2TxnReceipt) {
+    if (this.completedLayer2TxnReceipt) {
+      this.completedStepCount = 3;
```
backspace, line 60: This is outside the scope of this PR but I hope we can find a way to automate this kind of thing, it’s always worrisome to me to have to manually intervene like this and only find out because of manual testing 😳

backspace (approved):
Since we’re looking at the restoration UI, what do you think about the items being links instead of buttons? I usually hate when things are buttons because it takes away my ability to open in a new tab 🤓

Aierie:
> Since we’re looking at the restoration UI, what do you think about the items being links instead of buttons? I usually hate when things are buttons because it takes away my ability to open in a new tab 🤓

That makes sense! Though it does make it easier for the user to create multiple tabs with the same workflow open... especially since there is no way to differentiate workflows that are on the same milestone from each other.

lukemelia (approved):

packages/web-client/app/components/card-pay/deposit-workflow/transaction-status/index.ts
lukemelia: The general approach is that restorable card state should all be derived from workflowSession state. If it is, there should be no need for this kind of procedural code.

PR #2205, opened 2021-10-05
Update issue prepaid card flow cancelation message for insufficient funds

Aierie:
image

Not sure about https://github.com/cardstack/cardstack/pull/2205/commits/9b930cb9aa9be6363c059bc2530895a82d1c2ba2 because of the possibility of evolving schemas / workflow session property names, and also because I'm unsure if being true to the moment of cancelation is better than reflecting the latest value.


backspace (approved):
That minimum amount is quite lengthy, maybe in the future we’ll have standards around rounding/how many decimals to show 😯

lukemelia:
Why is the minimum amount so precise?

Aierie:
It is a little too precise to be useful since exchange rates will vary over time. I've updated the message to show a rounded amount along with its relationship to USD, does this seem better?

image

PR #2208, opened 2021-10-05
web: Add merchant safes as issuance source options

backspace:
This follows a pattern similar to #2038. Cards in the issuance workflow had hard-coded assumptions that the depot safe would be the source for funding the prepaid card, so this adds the `SafeChooser` and persists `prepaidFundingSafeAddress` to the workflow.

image

The amended `test__simulateIssuePrepaidCardForAmountFromSource` name reflects that it’s ensuring both the face value and source have made it through the workflow. The name is unwieldy but it’s only used once so I left it.

### DIFF
--- a/packages/web-client/tests/acceptance/issue-prepaid-card-test.ts
+++ b/packages/web-client/tests/acceptance/issue-prepaid-card-test.ts
@@ -50,9 +50,10 @@ const MIN_SPEND_AMOUNT = Math.min(...faceValueOptions);
 const MIN_AMOUNT_TO_PASS = new BN(
   toWei(`${Math.ceil(MIN_SPEND_AMOUNT / 100)}`)
 );
-const FAILING_AMOUNT = new BN(
-  toWei(`${Math.floor(MIN_SPEND_AMOUNT / 100) - 1}`)
+const FAILING_AMOUNT_IN_ETHER = new BN(
+  `${Math.floor(MIN_SPEND_AMOUNT / 100) - 1}`
 );
+const FAILING_AMOUNT = new BN(toWei(`${FAILING_AMOUNT_IN_ETHER}`));
 const SLIGHTLY_LESS_THAN_MAX_VALUE_IN_ETHER =
   Math.floor(Math.max(...faceValueOptions) / 100) - 1;
 const SLIGHTLY_LESS_THAN_MAX_VALUE = new BN(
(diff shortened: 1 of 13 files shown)


packages/web-client/app/utils/token.ts
```diff
@@ -153,8 +153,20 @@ export class TokenDisplayInfo<T extends TokenSymbol> implements DisplayInfo {
...
 
-  static iconFor(symbol: TokenSymbol) {
-    return _tokenDisplayInfoMap[symbol].icon;
+  static iconFor(symbol: string) {
+    switch (symbol) {
+      case 'ETH':
+        return _tokenDisplayInfoMap['ETH'].icon;
+      case 'DAI':
+      case 'DAI.CPXD':
+      case 'XDAI':
```
Aierie, line 162: What token corresponds to this case? Do we accept XDAI?

packages/web-client/app/components/card-pay/balance-view-banner/index.ts
```diff
@@ -0,0 +1,66 @@
...
+  }
+
+  get tokenIcon() {
+    return TokenDisplayInfo.iconFor(this.args.token);
+  }
+
+  get summaryBalanceLabel() {
+    if (this.args.safe.type === 'merchant') {
+      if (this.merchantInfo.loading) {
+        return `…`;
```
Aierie, line 50: Would it be good to show "Merchant" while the merchant info is loading? Also why is there a difference in capitalization between merchant and depot 🤔

packages/web-client/app/components/card-pay/balance-view-banner/index.ts
```diff
@@ -0,0 +1,66 @@
...
+      } else {
+        return `Merchant ${this.merchantInfo.name}`;
+      }
+    } else if (this.args.safe.type === 'depot') {
+      return 'DEPOT';
+    } else {
+      return 'Unknown';
```
Aierie, line 57: If we know that this list is always displaying safes, is "Safe" a good fallback?

packages/web-client/tests/acceptance/issue-prepaid-card-test.ts
```diff
@@ -119,6 +123,25 @@ module('Acceptance | issue prepaid card', function (hooks) {
...
+      accumulatedSpendValue: 100,
+      infoDID: MERCHANT_DID,
+    });
+
+    this.server.create('merchant-info', {
+      id: await getFilenameFromDid(MERCHANT_DID),
+      name: 'Mandello',
+      slug: 'mandello1',
+      did: MERCHANT_DID,
+      'owner-address': '0x182619c6Ea074C053eF3f1e1eF81Ec8De6Eb6E44',
```
Aierie, line 142: This could use the `layer2AccountAddress` variable

packages/web-client/app/components/card-pay/balance-view-banner/index.ts
backspace: maybe, probably? My true desire was to have the component throw an error if it was passed a type of safe it didn’t know how to render but I couldn’t get the tests to pass even with try/catch! But since this is only about a string representation (and icon) it’s probably fine to just have a generic label.

packages/web-client/app/utils/token.ts
Aierie: I guess we can clean this up when we start moving towards supporting more tokens? Since we should move towards a single source of truth for token display stuff by then.

PR #2209, opened 2021-10-06
Workflow selector fix cs 2042

jurgenwerk:
Ticket: [CS-2042](https://linear.app/cardstack/issue/CS-2042/polish-workflow-tracker-scrolling-interaction)

This PR changes the following:

1. Makes the dropdown stick to the opener button

- I did this change because I feel that dropdowns should stick to their opener button

Previously (the dropdown follows you while you scroll the page):

image

After the change (the dropdown will stay anchored to the top while scrolling down):

image

2. Fixes the scrollbar jumping behaviour

- Video of the bug is here: https://www.loom.com/share/d10e9e7e04e04e808f1298d942ba3e97
- The reason for the body scrollbar appearing/disappearing is the scroll locking code which is adding and removing `has-modal` to the body, which is removing the body scrollbar
- In my opinion, the scroll locking logic isn't needed anymore. Not sure if this is a remnant of a previous design implementation, but now it seems the scrolling works perfectly fine. Here's the video of this PR: https://www.loom.com/share/314ba6290fd34c05b4d01a5c92f5494a. The bonus of removing this logic is that it also fixes the scrollbar bug described above.

### DIFF
--- a/packages/web-client/app/components/card-pay/header/workflow-tracker/dropdown-list/index.ts
+++ b/packages/web-client/app/components/card-pay/header/workflow-tracker/dropdown-list/index.ts
@@ -3,51 +3,15 @@ import { inject as service } from '@ember/service';
 import { action } from '@ember/object';
 
 import WorkflowPersistence from '@cardstack/web-client/services/workflow-persistence';
-import { tracked } from '@glimmer/tracking';
 
 interface CardPayHeaderWorkflowTrackerDropdownListArgs {}
 
 export default class CardPayHeaderWorkflowTrackerDropdownList extends Component<CardPayHeaderWorkflowTrackerDropdownListArgs> {
   @service declare workflowPersistence: WorkflowPersistence;
 
-  initialPosition?: number;
-  @tracked currentPosition?: number;
-
   @action clearCompletedWorkflows() {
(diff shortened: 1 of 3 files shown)


Aierie (approved):
This UI feels better to me, though I haven't put much thought into this problem before this and am probably biased as a dev. Feels like input from @burieberry about the motivations for choosing the previous approach would help!

packages/web-client/app/components/card-pay/header/workflow-tracker/dropdown-list/index.ts
```diff
@@ -1,53 +1,23 @@
...
-
-    this.initialPosition = item?.getBoundingClientRect().top;
-
-    element.onscroll = () => {
-      this.currentPosition = item?.getBoundingClientRect().top;
-    };
-    element.onfocus = this._lockBodyScroll;
-    element.onblur = this._unlockBodyScroll;
-    element.onmouseover = this._lockBodyScroll;
-    element.onmouseout = this._unlockBodyScroll;
```
burieberry, line 37: We'd like to lock the body scroll while the user is interacting with the modal. Removing these removes that feature.

burieberry (approved):

PR #2210, opened 2021-10-06
adds scroll shadow back

Aierie:
Scroll shadow is at the top:

scroll shadow

### DIFF
--- a/packages/web-client/app/components/card-pay/header/workflow-tracker/dropdown-list/index.ts
+++ b/packages/web-client/app/components/card-pay/header/workflow-tracker/dropdown-list/index.ts
@@ -1,6 +1,7 @@
 import Component from '@glimmer/component';
 import { inject as service } from '@ember/service';
 import { action } from '@ember/object';
+import { tracked } from '@glimmer/tracking';
 
 import WorkflowPersistence from '@cardstack/web-client/services/workflow-persistence';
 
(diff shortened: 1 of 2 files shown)


packages/web-client/app/components/card-pay/header/workflow-tracker/dropdown-list/index.ts
```diff
@@ -1,13 +1,19 @@
...
 import WorkflowPersistence from '@cardstack/web-client/services/workflow-persistence';
 
 interface CardPayHeaderWorkflowTrackerDropdownListArgs {}
 
 export default class CardPayHeaderWorkflowTrackerDropdownList extends Component<CardPayHeaderWorkflowTrackerDropdownListArgs> {
   @service declare workflowPersistence: WorkflowPersistence;
+  @tracked isScrolled = false;
+
+  @action checkIfScrolled(e: Event) {
+    this.isScrolled = (e.target as HTMLElement).scrollTop > 0;
```
Aierie, line 15: @burieberry I had a look at the issue with `scrollTop` not recognizing which list was scrolling, it doesn't make a difference for this implementation since the shadow in the previous implementation was always from the first (active workflow) header.

jurgenwerk (approved):
Thanks!

packages/web-client/app/components/card-pay/header/workflow-tracker/dropdown-list/index.ts
burieberry: Yeah, I think initially I was adding the shadows separately but then changed it

PR #2220, opened 2021-10-07
Hide workflow tracker button when wallets are not connected

burieberry:
CS-2085

### DIFF
--- a/packages/web-client/app/components/card-pay/header/index.hbs
+++ b/packages/web-client/app/components/card-pay/header/index.hbs
@@ -25,8 +25,10 @@
         @onConnect={{@layer2Connect}}
       />
     </li>
-    <li class="card-pay-header__item">
-      <CardPay::Header::WorkflowTracker />
-    </li>
+    {{#if (and @isLayer1Connected @isLayer2Connected)}}
+      <li class="card-pay-header__item">
+        <CardPay::Header::WorkflowTracker />
+      </li>
+    {{/if}}
   </ul>
 </Boxel::OrgHeader>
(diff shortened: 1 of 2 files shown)


packages/web-client/app/models/workflow.ts
```diff
@@ -86,6 +92,22 @@ export abstract class Workflow {
     this.cancelationMessages.setWorkflow(this);
   }
 
+  async _runBeforeRestorationChecks(promises: Promise<void>[]) {
```
jurgenwerk, line 95: Would it be better to use the hash notation to indicate this is private?

jurgenwerk (changes requested):
This PR solves the issue where the restoration process hangs when a wallet isn't connected, but I think it leaves a bit more to be desired.

If I understand correctly, in this PR we leave the user waiting for 10s in a state where the screen is overlayed and no modal is shown, and no indication of progress is show either. That probably doesn' feel too good to see.

Secondly, I think there's another issue. Let's imagine the user has their wallets disconnected, then tries to restore a couple of workflows, but (after the timeout hits) they all get cancelled, which wouldn't happen if the user had their wallets connected. It sounds to me that we have a deeper conceptual issue here that should be addressed differently.

Maybe this task slipped through the cracks but I think we discussed we probably don't even want to list the workflows that don't match the current context (L1 and L2 wallet addresses). I think that would probably help to get rid of the problems listed above.

Another possible fix is to urge the user to connect their wallets they need to restore a specific workflow before listing the workflows in the dropdown, or before trying to restore a workflow.

PR #2222, opened 2021-10-08
Introduce a version property to workflow session's meta state

lukemelia:
When a workflow is restored, we check the version persisted to the session and if it is older than the current workflow version, the flow is canceled and we show a message.


lukemelia:
image

PR #2224, opened 2021-10-10
Reduce workflow boilerplate

lukemelia:
For your consideration. I felt the workflow definition was getting a bit verbose and explored a couple of ways to reduce the boilerplate. This PR does this in two ways:

- wraps the conditional cancelation message pattern in a function
- assumes postable author is cardbot by default

### DIFF
--- a/packages/web-client/app/components/card-pay/issue-prepaid-card-workflow/index.ts
+++ b/packages/web-client/app/components/card-pay/issue-prepaid-card-workflow/index.ts
@@ -11,7 +11,6 @@ import BN from 'bn.js';
 import { currentNetworkDisplayInfo as c } from '@cardstack/web-client/utils/web3-strategies/network-display-info';
 import HubAuthentication from '@cardstack/web-client/services/hub-authentication';
 import {
-  cardbot,
   IWorkflowSession,
   Milestone,
   NetworkAwareWorkflowCard,
@@ -23,6 +22,8 @@ import {
   WorkflowMessage,
   WorkflowName,
   UNSUPPORTED_WORKFLOW_STATE_VERSION,
+  conditionalCancelationMessage,
+  defaultCancelationCard,
 } from '@cardstack/web-client/models/workflow';
 
 import { tracked } from '@glimmer/tracking';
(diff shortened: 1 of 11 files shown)


packages/web-client/app/components/card-pay/issue-prepaid-card-workflow/index.ts
```diff
@@ -294,85 +266,36 @@ class IssuePrepaidCardWorkflow extends Workflow {
...
-        return (
-          [
-            FAILURE_REASONS.DISCONNECTED,
-            FAILURE_REASONS.ACCOUNT_CHANGED,
-            FAILURE_REASONS.UNAUTHENTICATED,
-            FAILURE_REASONS.RESTORATION_UNAUTHENTICATED,
-            FAILURE_REASONS.RESTORATION_L2_ACCOUNT_CHANGED,
-            FAILURE_REASONS.RESTORATION_L2_DISCONNECTED,
-          ] as String[]
-        ).includes(String(this.workflow?.cancelationReason));
```
lukemelia, line 373: I am not sure why the prepaid card workflow uses this pattern while other workflows don't. Changing this one to use the standard check to show the default cancelation CTA did not break any tests. If anyone knows of a reason we need to preserve the logic removed here, let me know.

Aierie: Sorry I was unclear - all ctas, whether cancelation or epilogue, are not completed cards. They started that way for styling purposes, but this is something we hope to change in the issue I mentioned. 

In this case, the first cancelation CTA not being complete is going to prevent the second from being shown.

burieberry: So NextSteps boxes are not cards, they're just components. We've been treating them like action-cards which is actually not right but we never refactored it. In the case of user having insufficient funds, we only want to display the `insufficient-funds-cta` component and not the `default-cancellation-cta`.  Both of those components are using the `NextStepsBox` component underneath. It seems like because of the completion state on the next steps box (which actually should not have a completion state), everything seems to still be working.

burieberry:
> Is the takeaway here that we should merge this but we need a new Linear task for making the cancelation postables work on purpose and not by accident?

Haha yes, I think so

PR #2228, opened 2021-10-12
Add content on token-suppliers dashboard

burieberry:
token-balances-depot


lukemelia:
@Aierie can you rebase and update this PR to fit the new bridged token symbol approach?

lukemelia (commented):
Looks good, just needs to be rebased and updated to use bridge token symbols from SDK

Aierie:
> @Aierie can you rebase and update this PR to fit the new bridged token symbol approach?

Sure

packages/web-client/app/services/token-to-usd.ts
```diff
@@ -88,7 +88,18 @@ export default class TokenToUsd extends Service {
...
+    amount: BN
+  ): number | undefined {
+    if (symbol === 'DAI') {
+      symbol = 'DAI.CPXD';
+    }
+
+    if (symbol === 'CARD') {
+      symbol = 'CARD.CPXD';
+    }
+
```
Aierie, line 102: The way we're doing the usd conversion is a bit convoluted right now. We accept `.CPXD` here, but in `layer2-chain` we remove it. I'll clean it in a followup PR.

lukemelia (approved):

packages/web-client/app/services/token-to-usd.ts
burieberry: Thanks! merging this

PR #2235, opened 2021-10-15
Start using boxel components in cardhost and adding cards

burieberry:


packages/base-cards/string/edit.js
```diff
@@ -3,20 +3,21 @@ import { precompileTemplate } from '@ember/template-compilation';
...
-  return function (event) {
-    action(event.target.value);
+  return function (value) {
+    action(value);
   };
 });
 
 export default setComponentTemplate(
-  precompileTemplate('<input type="text" value="{{@model}}" {{on "input" (getTargetValue @set)}} ...attributes />', {
+  precompileTemplate('<BoxelInput type="text" @value="{{@model}}" @onInput={{getTargetValue @set}} ...attributes />', {
```
ef4, line 18: It looks like we don't need a helper here at all anymore and `@onInput={{getTargetValue @set}}` can become `@onInput={{@set}}`

packages/cardhost/app/routes/application.ts
```diff
@@ -0,0 +1,5 @@
+import Route from '@ember/routing/route';
+
+import '@cardstack/boxel/styles/boxel';
```
ef4, line 3: I would like to investigate how to keep these concerns at the component level instead.

ef4: That goes for all the font files too.

PR #2237, opened 2021-10-15
web: Fix display of restored withdrawal source

backspace:
The default value for the selected safe was being used
even when was already persisted in the workflow:

screencast 2021-10-15 10-07-16

What other subtleties like this might exist 😳


Aierie (approved):
🧹🪲!

PR #2280, opened 2021-10-27
Boxel style imports refactor

burieberry:
Changes:
- Each component imports its own style file, along with `global.css`
- Use templateOnly components
- Fix asset references, make it less brittle
- Finally able to update `embroider` and `ember-cli-babel` versions without the build breaking on boxel
- Use file-loader in cardhost for assets


packages/boxel/addon/components/boxel/apply-changes-toggle/index.css
```diff
@@ -12,8 +12,8 @@
 
 .boxel-apply-changes-toggle--active {
   background-image:
-    url('../@cardstack/boxel/images/icons/arrows.svg'),
-    url('../@cardstack/boxel/images/icons/arrows.svg');
+    url('/@cardstack/boxel/images/icons/arrows.svg'),
+    url('/@cardstack/boxel/images/icons/arrows.svg');
```
mattmcmanus, line 16: Is the slash correct here?

mattmcmanus (approved):
Nice changes. I only have the one question around the paths starting with `/`. Is that some sort of webpack trick?

packages/boxel/addon/components/boxel/apply-changes-toggle/index.css
burieberry: > Nice changes. I only have the one question around the paths starting with `/`. Is that some sort of webpack trick?

I don't know. It crashes if I remove the slash. It works with it somehow, without doing any relative paths.

PR #2295, opened 2021-10-29
web: Add timeout message for deposit delay

backspace:
This adds a message to the deposit transaction status card that shows after two minutes. It contains a link that updates as the card progresses, so it always links to the same place as the most-recently-displayed transaction status button.

image

### DIFF
--- a/packages/web-client/app/components/card-pay/deposit-workflow/transaction-status/index.css
+++ b/packages/web-client/app/components/card-pay/deposit-workflow/transaction-status/index.css
@@ -36,3 +36,11 @@
   align-items: center;
   justify-content: center;
 }
+
+.transaction-status__delay {
+  color: #707070;
+}
+
+.transaction-status__delay a {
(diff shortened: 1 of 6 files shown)


packages/web-client/app/components/card-pay/deposit-workflow/transaction-status/index.css
```diff
@@ -36,3 +36,11 @@
   align-items: center;
   justify-content: center;
 }
+
+.transaction-status__delay {
+  color: #707070;
```
backspace, line 41: The issue calls for grey text but I couldn’t find an existing Boxel colour that was readable 🤔

Aierie: It's a bit confusing but our most commonly used gray is a purple variant... `var(--boxel-purple-500);`, or `var(--boxel-purple-400);` might be good?

PR #2297, opened 2021-10-29
depositing can be canceled if user has not confirmed the transaction

Aierie:
Double depositing should not happen because of needing to approve the amount beforehand. However, users who leave the page after canceling in the DApp, and later approve on their wallet will not be able to continue the workflow properly on the DApp later.

### DIFF
--- a/packages/web-client/app/components/card-pay/deposit-workflow/transaction-amount/index.hbs
+++ b/packages/web-client/app/components/card-pay/deposit-workflow/transaction-amount/index.hbs
@@ -108,6 +108,11 @@
       <i.ActionButton data-test-deposit-button>
         Depositing
       </i.ActionButton>
+      {{#if this.cancelable}}
+        <i.CancelButton {{on "click" this.cancelDeposit}} data-test-deposit-cancel-button>
+          Cancel
+        </i.CancelButton>
+      {{/if}}
       {{#if this.depositTxnViewerUrl}}
         <i.InfoArea>
           <Boxel::Button @as="anchor" @kind="secondary-dark" @size="extra-small" @href={{this.depositTxnViewerUrl}} target="_blank" rel="noopener" data-test-deposit-etherscan-button>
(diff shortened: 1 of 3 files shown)


lukemelia (approved):
Code looks fine. I tested this with chrome-extension Metamask, and it's a little weird in as much as pressing the Cancel button doesn't reject the metamask signing request, and if you then initiate another request, metamask conceptually "stacks it" on top of the previous one, so the previous ("canceled") request appears after you confirm the new request. If you then reject the original request, the UI shows an error message. Something seems a bit funky.

Aierie:
Did more investigation since @lukemelia pointed out some potential issues. The scenarios below start with canceling the first deposit transaction in DApp UI, then clicking on deposit again without rejecting the existing transaction in MetaMask. This is how you get 2 stacked transactions that depend on the same "allowance". Scenarios:

1. Confirm both - then one transaction will fail silently because it exceeds the allowance. The UI will show you a success state. This is expected.
1. **Confirm the first one in the stack and do nothing with the second** -  Depositing never completes. It occurs because we have canceled the actions to be performed after confirming the first one in the stack. Reloading will advance the workflow.
1.  **Confirm the first one in the stack and reject the second** - UI will show an error state. As above, we have canceled the actions to be performed after confirming the first one in the stack, and are still listening for errors that come back from the second transaction.
1. Reject the first one in the stack and confirm the second - UI will show you a success state. 
1. Reject both - UI will show an error state, this is expected - but the one that matters/provides feedback to the UI is the second one shown to you.

Aierie:
I think properly canceling transactions/signing requests will involve dealing with nonces. Some wallets will cause problems because they (one of them being MetaMask) [ignore the nonce field when provided by an application (as opposed to a user manually entering it)](https://docs.metamask.io/guide/sending-transactions.html#transaction-parameters). 

It feels like we should show users a "retry" UI, not "cancel", since we don't have a way to actually cancel transactions reliably. Canceling a deposit transaction will have to be done via the wallet. If folks are agreeable to this, i'll create an issue for, or replace the work in this PR with it.

lukemelia:
@Aierie I am in agreement with your analysis.

Aierie:
Closing until this work is updated.

PR #2298, opened 2021-10-29
clean and make the workflow card config stricter

Aierie:
I tried, with cards that had registered their config, and some that had not:

- Adding random config
- Adding config belonging to another card
- Omitting required config
- Adding extra properties (in the main object, not the config) to a card

This solution detects all of these situations nicely.


backspace (approved):
I’m definitely not totally understanding but this seems like a more elegant solution. I wonder if we can figure out how to get automated testing around this eventually, but for now I’ll merge, thanks for the ongoing improvements 🎉

PR #2301, opened 2021-10-31
updating subgraph to handle v0.8.4 of protocol

habdelra:
i ran this subgraph and it index to the latest without error on my hosted instance


habdelra:
this has now been deployed to sokol staging subgraph

PR #2302, opened 2021-11-01
Reserved words service

Aierie:
Currently the reserved words service only detects words that exactly match the reserved words. This PR adds a reserved words service (since reserved words need to be used in multiple places), makes the profanity-matching a bit stricter (checks whether the string includes the reserved word, rather than matching it exactly), and removes a few words that will probably interfere with registration of usernames containing common non-profane english words, eg. `anal` in `analysis`.


backspace:
The [Scunthorpe problem](https://en.wikipedia.org/wiki/Scunthorpe_problem) 😬

Aierie:
We have the scunthorpe problem, in this PR's implementation..

jurgenwerk (approved):
Thanks for putting this in a service! It will come handy in other places too.

Regarding the existing profanity words list, I think it's quite narrow anyway and some words in there aren't even that bad. Some really terrible words are missing. Nasty stuff will always find a way to slip in 😐 I guess more important is to have the reserved words filter, such as "www", "ftp", "mail", and branded keywords.

PR #2306, opened 2021-11-01
web: Add Percy for visual diffs

backspace:
This adds full-width-only Percy snapshots at the end of acceptance tests for completed workflows, one for each type. It’s currently [configured](https://percy.io/Cardstack-1/-cardstack-web-client/settings) to only run on Chrome, as I think our interest is more in unknown regressions in our styling than browser-to-browser comparison.

There are some CSS overrides to hide frequently-changing strings and break workflows out of their containers so the snapshot can capture the entire thing. There are also some changes to prevent some strings from changing: using static example safe addresses instead of random ones and using a randomisation seed for merchant creation background colour in the test environment.


package.json
```diff
@@ -55,7 +55,8 @@
     "test:hub": "NODE_ENV=test cd packages/hub && yarn test",
     "test:did-resolver": "cd packages/did-resolver && yarn test",
     "test:discord-bot": "cd packages/discord-bot && yarn test",
-    "test:web-client": "cd packages/web-client && ember test",
+    "test:web-client": "cd packages/web-client && percy exec -- ember test",
+    "test:web-client:percy": "cd packages/web-client && ember test",
```
burieberry, line 59: Did you mean this the other way around? It doesn't look like the percy build ran because of the mixup

backspace: lol oops, fixed now, thank you

burieberry (approved):
Looks good. We can test how well this works when it's applied to #2280, which should be breaking a bunch of stuff.

PR #2310, opened 2021-11-02
Suggestion/further fixes

Aierie:
I think we technically also need a version bump for prepaid card issuance and withdrawal workflows, since these persist a "source token". Done in https://github.com/cardstack/cardstack/pull/2310/commits/dda01b3427eeb90ef2e9180aac65afae7c5a5d72

### DIFF
--- a/packages/web-client/app/components/card-pay/deposit-workflow/confirmation/index.ts
+++ b/packages/web-client/app/components/card-pay/deposit-workflow/confirmation/index.ts
@@ -35,9 +35,9 @@ class CardPayDepositWorkflowConfirmationComponent extends Component<WorkflowCard
 
   get receivedToken() {
     if (this.selectedTokenSymbol === 'CARD') {
-      return new TokenDisplayInfo('CARD.CPXD');
+      return new TokenDisplayInfo(this.layer2Network.bridgedCardTokenSymbol);
     } else {
-      return new TokenDisplayInfo('DAI.CPXD');
+      return new TokenDisplayInfo(this.layer2Network.bridgedDaiTokenSymbol);
     }
   }
 
--- a/packages/web-client/app/components/card-pay/layer-two-connect-card/index.ts
+++ b/packages/web-client/app/components/card-pay/layer-two-connect-card/index.ts
@@ -49,11 +49,11 @@ class CardPayLayerTwoConnectCardComponent extends Component<CardPayLayerTwoConne
   get balancesToShow() {
     return [
       {
-        symbol: 'DAI.CPXD',
+        symbol: this.layer2Network.bridgedDaiTokenSymbol,
         amount: this.layer2Network.defaultTokenBalance,
       },
       {
-        symbol: 'CARD.CPXD',
+        symbol: this.layer2Network.bridgedCardTokenSymbol,
         amount: this.layer2Network.cardBalance,
       },
     ].filter((o) => o.amount && !o.amount?.isZero());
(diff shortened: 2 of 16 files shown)


Aierie:
Didn't work on hardcoded mentions of `DAI.CPXD` in the UI... there are quite a few. We could change these, but it feels like they're not essential for this change? Downside is they might confuse folks who see them in development.

lukemelia:
IMO, the UI should continue to show .CPXD for bridged tokens  even in Sokol

packages/web-client/app/components/card-pay/layer-two-connect-card/index.ts
```diff
@@ -49,11 +49,11 @@ class CardPayLayerTwoConnectCardComponent extends Component<CardPayLayerTwoConne
   get balancesToShow() {
     return [
       {
-        symbol: 'DAI.CPXD',
+        symbol: this.layer2Network.bridgedDaiTokenSymbol,
```
lukemelia, line 52: This is for display purposes, which is why I left it. I guess I am OK either way and curious about other people's thoughts.

packages/web-client/app/components/card-pay/deposit-workflow/confirmation/index.ts
```diff
@@ -35,9 +35,9 @@ class CardPayDepositWorkflowConfirmationComponent extends Component<WorkflowCard
 
   get receivedToken() {
     if (this.selectedTokenSymbol === 'CARD') {
-      return new TokenDisplayInfo('CARD.CPXD');
+      return new TokenDisplayInfo(this.layer2Network.bridgedCardTokenSymbol);
```
lukemelia, line 38: This is also going to be awkwardly wrong on Sokol for now.

packages/web-client/app/utils/web3-strategies/layer2-chain.ts
```diff
@@ -123,7 +123,10 @@ export default abstract class Layer2ChainWeb3Strategy
   async fetchIssuePrepaidCardMinValues() {
     this.issuePrepaidCardSpendMinValue = Math.min(...faceValueOptions);
     this.issuePrepaidCardDaiMinValue = new BN(
-      await this.convertFromSpend('DAI', this.issuePrepaidCardSpendMinValue)
+      await this.convertFromSpend(
+        this.bridgedDaiTokenSymbol,
```
lukemelia, line 127: Are you sure that the Oracle from the SDK method accepts 'DAI.CPXD' and 
'CARD.CPXD'?

packages/web-client/app/components/card-pay/deposit-workflow/confirmation/index.ts
Aierie: It is 😓

packages/web-client/app/utils/web3-strategies/layer2-chain.ts
Aierie: No, it doesn't. This is for the mapping in `TokenContractInfo` which resolves to `daiCpxd` and `cardCpxd`.

packages/web-client/app/components/card-pay/layer-two-connect-card/index.ts
Aierie: The safe balance card shows `DAI/CARD` on Sokol, not `DAI.CPXD/CARD.CPXD`. I don't mind either, but we should probably be consistent about this.

Aierie:
Probably not a big deal, but I'm thinking that leaving the hardcoded `.CPXD`s in is going to make it hard to figure out when the `CPXD` is a bug/unexpected and when it's not, when we're developing on Sokol.

packages/web-client/app/utils/web3-strategies/layer2-chain.ts
Aierie: Leaving it at `DAI` would work for Sokol but break in xDai

lukemelia: OK, let's go with it.

PR #2311, opened 2021-11-02
Timeout after 1 minute when trying to authenticate with the hub

jurgenwerk:
Ticket: [cs-2295](https://linear.app/cardstack/issue/CS-2295/bug-card-issuance-hangs-at-hub-authen)

I haven't been able to reproduce this bug, but Kiel's idea about introducing a timeout makes a lot of sense. In case something we can't control goes wrong or hangs, the user gets notified and can retry after 1 minute.

### DIFF
--- a/packages/web-client/app/components/card-pay/hub-authentication/index.hbs
+++ b/packages/web-client/app/components/card-pay/hub-authentication/index.hbs
@@ -12,7 +12,13 @@
           style={{css-var status-icon-size="2.5rem"}}
           data-test-failed
         >
-          Authentication failed or was canceled.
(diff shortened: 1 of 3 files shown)


packages/web-client/app/components/card-pay/hub-authentication/index.hbs
```diff
@@ -12,7 +12,13 @@
           style={{css-var status-icon-size="2.5rem"}}
           data-test-failed
         >
-          Authentication failed or was canceled.
+          {{#if (eq this.error.message "AUTH_TIMEOUT")}}
+            <span>
```
jurgenwerk, line 16: Needed to add `span` otherwise the layout gets broken with cardstack support link:
image

PR #2313, opened 2021-11-02
Add percy to boxel

burieberry:
Added a snapshot on the docs page only, but might need to add more component-level snapshots for the various states of the components. Not sure if the `pr-boxel.yml` file setup is right, did it similar to #2306 . Ideally, percy should only run for PRs.


backspace (approved):
I tried adding an empty commit just to see and got some tiny differences in Edge, like this:

image

image

One thing I didn’t approach in my PR yet is choosing which browsers to run on. It’s set up to have diffs from Safari, Firefox, Chrome, and Edge. I think we might exceed our limits pretty quickly with all of those, it depends what we’re trying to address. My original fear that led me to pick up the Percy work was unintentional regressions that broke CSS, so we might be able to just use one browser to catch that. But maybe we want more coverage across browsers?

Apart from that tiny diff noise this looks good. An easy fix would just be to remove Edge from Percy diffing 😆

I wonder if we might want runs on `main` too, just in case other projects manage to break things somehow, though maybe that’s more a web client thing.

burieberry:
> I tried adding an empty commit just to see and got some tiny differences in Edge, like this:
> 
> image
> 
> image
> 
> One thing I didn’t approach in my PR yet is choosing which browsers to run on. It’s set up to have diffs from Safari, Firefox, Chrome, and Edge. I think we might exceed our limits pretty quickly with all of those, it depends what we’re trying to address. My original fear that led me to pick up the Percy work was unintentional regressions that broke CSS, so we might be able to just use one browser to catch that. But maybe we want more coverage across browsers?
> 
> Apart from that tiny diff noise this looks good. An easy fix would just be to remove Edge from Percy diffing 😆
> 
> I wonder if we might want runs on `main` too, just in case other projects manage to break things somehow, though maybe that’s more a web client thing.

Maybe for boxel, I'll just leave Safari on, since that's the one with the most css strangenesses that we don't tend to catch, and I'll hide everything but the components on that docs page from percy.

For web-client, percy coverage is more important, since so far it's the only place we use boxel in is web-client. We're aiming to make changes in the boxel components so that the card-pay styles will be the default and won't need overriding.

PR #2314, opened 2021-11-03
improve returned validation messages

Aierie:
This assumes use with the card space setup workflow, and is better for that case, but may be confusing outside of it.

https://user-images.githubusercontent.com/39579264/140021785-f6d1400c-b2e4-43e8-94b5-096907073169.mov


jurgenwerk:
Thanks @Aierie! Really appreciate you taking the time to put this in a more proper form.

PR #2315, opened 2021-11-03
Host token logos for Honeyswap

jurgenwerk:
Ticket: [CS-2334](https://linear.app/cardstack/issue/CS-2334/host-logo-for-cardcpxd-and-daicpxd-so-we-can-have-a-logouri)

Kiel needs Card.CPXD and DAI.CPXD token image links to upload to a DeFi. For convenience I added png versions which will probably be easier to consume for external services.

After deploy to production, these should then be available using the following links:

https://app.cardstack.com/images/logos/tokens/dai.cpxd.png
https://app.cardstack.com/images/logos/tokens/card.cpxd.png


Aierie (approved):

jurgenwerk:
Waiting for an answer in the ticket whether we even need to host svgs, or the pngs are enough.

PR #2316, opened 2021-11-03
depositing can be canceled if user has not confirmed the transaction

Aierie:
Updated #2297 to allow cancelation for now (with cancel button shown after a delay). Follow-up in https://linear.app/cardstack/issue/CS-2395/add-retry-ui-that-properly-handles-doubled-requests


jurgenwerk (approved):
Looks good!

PR #2317, opened 2021-11-03
Test/externally used assets exists

Aierie:


Aierie:
@backspace is this good to add to Percy?

Aierie:
> Could it work as a “component” test? This implementation looks like it wouldn’t actually render anything for Percy to see 🤔 but in a component test there could be a template that renders the images

We could visit each asset's page I think

backspace:
I think you can only `visit` routes recognised by the router:

image

Also, scarcity thinking, but it might save on our limited usage to have them all in one snapshot vs one for each 😆

Aierie:
> I think you can only `visit` routes recognised by the router:
> 
> image
> 
> Also, scarcity thinking, but it might save on our limited usage to have them all in one snapshot vs one for each 😆

Sounds good, will work on a PR to add Percy to this

PR #2322, opened 2021-11-04
Avoid issue where Infura is 1 block behind the chain head

jurgenwerk:
Ticket: [CS-2391](https://linear.app/cardstack/issue/CS-2391)

Check the discussion in the ticket. I did some testing and I believe this fixes the issue.

### DIFF
--- a/packages/cardpay-sdk/sdk/assets.ts
+++ b/packages/cardpay-sdk/sdk/assets.ts
@@ -1,6 +1,7 @@
 import ERC20ABI from '../contracts/abi/erc-20';
 import { AbiItem } from 'web3-utils';
 import Web3 from 'web3';
+import { safeContractCall } from './utils/general-utils';
 
 export interface IAssets {
   getNativeTokenBalance(userAddress?: string): Promise<string>;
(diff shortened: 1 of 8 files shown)


jurgenwerk:
@habdelra here's my solution. Do you think it would be better to solve this in the sdk?

habdelra:
We do have an API in the SDK that performs this same logic. (in the Assets.getBalanceForToken). I think this was originally written before that API existed. I just updated that API using your strategy

habdelra:
ok @jurgenwerk, my changes are in now as well. how do they look to you?

habdelra:
oh i spotted one more thing I want to do: to be consistent in our testing env, we should also wait one extra block in kovan (even though we dont' see the errors there)

habdelra (approved):
👍

habdelra:
@jurgenwerk I approved it, but i think you should approve my changes as well

habdelra:
@jurgenwerk I approved it, but i think you should approve my changes as well

jurgenwerk:
@habdelra changes look good! One other thing I wanna do - I'm gonna try to use the assets api in layer1 chain in `refreshBalances` so that we don't repeat the `getBlockNumber` logic. Will finish that tomorrow morning but this can be merged now if really needed.

habdelra:
i'm totally fine waiting until tomorrow for you to complete the work

lukemelia:
@habdelra with @jurgenwerk's blessing, I finished up the work on this. Can you give it quick look-see and then we can get it merged?

habdelra:
@lukemelia looks good. it would be nice, though, to have @jurgenwerk re-run his script to try to reproduce the bug again with these changes. but i don't think that has to stop us from merging this.

jurgenwerk:
I've been running my stress test on prod and I can't reproduce the bug anymore. Thanks @habdelra and @lukemelia! 🎉

PR #2329, opened 2021-11-04
CS-2402 change the cardbot command for air-drops to !card-drop

habdelra:


packages/hub/node-tests/bot-commands/card-me-test.ts
```diff
@@ -1,7 +1,7 @@
 import { Client as DBClient } from 'pg';
```
lukemelia, line 1: Rename this file to card-drop-test.ts?

lukemelia (approved):
Looks good except the test file name I commented on

packages/hub/node-tests/bot-commands/card-me-test.ts
habdelra: whoops!

PR #2330, opened 2021-11-04
web: Add guard to ignore task cancelation

backspace:
Task cancelation results in promise rejection, but we don’t
want to treat cancelation (from the component being closed
by clicking out of the workflow, for instance) as a failure.

Since this was a bug report about a console warning, it’s not
covered by testing, but I hacked the component test to
allow me to arbitrarily destroy the component:

image

Before this change, destroying the component produced the warning:

screencast 2021-11-04 15-48-06

Afterward, no warning:

screencast 2021-11-04 15-49-32


habdelra (approved):
👍

PR #2336, opened 2021-11-05
web: Update payments page marketing message

lukemelia:
This renames the page from “Merchant Services” to “Payments” and changes the route name as well.

image


backspace (approved):
These look like the first WebP images in the repository, do we have any standards about image formats? It looks [well-supported](https://caniuse.com/?search=webp) by now at least.

lukemelia:
Thanks to Percy, I see that I need to deal with the collapsed state too

PR #2341, opened 2021-11-05
web: Add timeout message for withdrawal tracking

backspace:
This is the withdrawal equivalent of #2295, though it’s simpler
because there are only two steps and the first is always complete
when the component renders.

This also changes the text colour [as suggested by @Aierie](https://github.com/cardstack/cardstack/pull/2295/files#r741135185) since
these components use the same CSS 🙃


jurgenwerk (approved):
LGTM!

PR #2343, opened 2021-11-05
web: Rename merchants to business accounts

backspace:
This is a user-facing change only, since the SDK will continue
to call it a merchant safe, and it seemed too dramatic to
change Hub routes and database names. It replaces mentions of
“merchant” with “business” and “business account” depending
on what makes the most sense. It also renames the query
parameter from create-merchant to create-business and changes
relevant user-facing strings in Hub and SDK.


backspace:
I see merge conflicts are present, will fix ASAP

lukemelia (approved):
LGTM

PR #2349, opened 2021-11-05
web: Add available balances listing

backspace:
This updates the Deposit & Withdrawal page to show depot and merchant safe balances and adds a withdrawal heading button alongside the deposit one:

image


packages/web-client/app/components/card-pay/safe-balances/index.hbs
```diff
@@ -0,0 +1,74 @@
...
+    <header aria-label='title' class='card-pay-safe-balances__title-container' data-test-safe-balances-title>
+      {{#if this.data.icon}}
+        {{svg-jar this.data.icon width='20' height='20' class='card-pay-safe-balances__logo' data-test-safe-balances-logo=this.data.icon}}
+      {{else}}
+        <CardPay::MerchantLogo
+          class='card-pay-safe-balances__logo'
+          @name={{first-char this.data.info.name}}
+          @logoBackground={{this.data.info.backgroundColor}}
+          @logoTextColor={{this.data.info.textColor}}
+        />
```
Aierie, line 31: Merchant logo sometimes get squished here if the merchant's name is long. We should add some defensive CSS to the merchant logo:

```
  flex-shrink: 0;
  flex-grow: 0;
```

packages/web-client/app/controllers/card-pay/deposit-withdrawal.ts
```diff
@@ -27,6 +27,12 @@ export default class CardPayDepositAndWithdrawalController extends Controller {
     this.workflowPersistenceId = null;
   }
 
+  get safes() {
+    return this.layer2Network.safes.value
+      ?.filter((s) => s.type === 'merchant' || s.type === 'depot')
+      .sortBy('type');
```
Aierie, line 33: Maybe a good time to introduce convenience getters for safes by types in the safe resource? I could follow up with this.

packages/web-client/app/components/card-pay/safe-balances/index.hbs
```diff
@@ -0,0 +1,74 @@
...
+      <div class='card-pay-safe-balances__balances-num' data-test-safe-balances-usd-total>
+        {{format-usd this.usdBalanceTotal}}
+      </div>
+    </h4>
+    {{#each this.tokenBalances as |b|}}
+      <div class='card-pay-safe-balances__token' data-test-safe-balances-token={{b.symbol}}>
+        <CardPay::BalanceDisplay
+          class='card-pay-safe-balances__token-balance-display'
+          @icon={{b.icon}}
+          @balance={{format-wei-amount b.balance 2 2}}
```
Aierie, line 66: Can we omit this change + the change in `format-wei-amount.ts`? https://github.com/cardstack/cardstack/pull/2367 should cover the decimal limit change.

Aierie (approved):
Overall i think this works well and is good to merge! There are some UX issues around finding merchants if we have many, but I don't think that we know enough about how this is going to be used to really say what the right way to optimize is - eg. sort by balance, sort by type, whether to display safes that have no balances.

packages/web-client/app/components/card-pay/safe-balances/index.hbs
backspace: hmm yes, that fixes the squishing, but then there’s overflow, so I should add truncation too, thanks

image

backspace: Actually I’m going to just open an issue for that, I don’t know how to handle it right now.

packages/web-client/app/controllers/card-pay/deposit-withdrawal.ts
backspace: yes, it could follow a pattern like `get issuePrepaidCardSourceSafes()` 🤔

packages/web-client/app/components/card-pay/safe-balances/index.hbs
backspace: 👍🏻 I’ve merged that in via `main` and updated the invocation here, thanks.

PR #2351, opened 2021-11-07
Remove register rewardee temporarily

tintinthong:
- `registerRewardee` had a an event changed in this [pr](https://github.com/cardstack/card-pay-protocol/pull/124).
- for simplcity, let's just remove the affected event from indexing and deploy a new protocol


tintinthong:
I keep on forgetting that the deploy to subgraph is literally the last thing we do in the step. Ignore this pr.

PR #2353, opened 2021-11-08
Customize info shown when connecting via WalletConnect

Aierie:


backspace (approved):
Thanks for fixing this!

For reference, before:
image


After:

image

PR #2354, opened 2021-11-08
3 - Add reward program lock and update reward program admin in SDK 

tintinthong:
- add `lockRewardProgram` and `isRewardProgramLocked`
- add `updateRewardProgramAdmin` and `rewardProgramAdmin`

`updateRewardProgramAdmin` will work as normal. `lockRewardProgram` will be affected by next card pay deploy (but its ok to merge first bcos its just a state variable) .


habdelra:
I'd like to propose that we start splitting out the PrepaidCardManager's base.ts. It feels like it's losing cohesion in that is is doing _a lot_ of different things, and the file size is getting pretty large. let's create a new SDK for the reward related prepaid card actions, as then split out any common functionality between the PrepaidCardManager and the new SDK so that both these modules can share.

habdelra (commented):
Let's create a new SDK for rewards related prepaid card actions

tintinthong:
Ok I'll add a refactor for that

tintinthong:
Also this is not a blocking pr for simple rewards.

tintinthong:
> I'd like to propose that we start splitting out the PrepaidCardManager's base.ts. It feels like it's losing cohesion in that is is doing a lot of different things, and the file size is getting pretty large. let's create a new SDK for the reward related prepaid card actions, as then split out any common functionality between the PrepaidCardManager and the new SDK so that both these modules can share.

Actually, wouldn't it be better to just move the prepaid card actions into the reward manager base since it is not that big at all. And the prepaid card actions for rewards do correspond to reward manager logic. Its more straightforward change.

tintinthong:
A deeper refactor perhaps is more appropriate in a different pr.

tintinthong:
this pr can go in before protocol deploy.

habdelra:
> Actually, wouldn't it be better to just move the prepaid card actions for rewards into the reward manager base (similar to prepaid card market is organised) since it is not that big at all. And the prepaid card actions for rewards do correspond to reward manager logic. Its more straightforward change.

yeah that sounds just fine.

habdelra (approved):
This looks very good. you'll just need to merge upstream changes into this PR and resolve the conflicts

PR #2355, opened 2021-11-08
2 - upgrade sdk 0.8.5 cardpay

tintinthong:


tintinthong:
@habdelra I think there are no abis on the sdk to update from this change right https://github.com/cardstack/card-pay-protocol/pull/134/files

habdelra:
@tintinthong we can remove v0.8.0 artifacts since the upgrade to v0.8.4 was pushed all the way to xdai

habdelra (commented):
Overall this looks good, but let's also cleanup the 0.8.0 artifacts since xdai's current version is now 0.8.4 and we have explicit support for that in the SDK. Also I merged upstream changes into this PR (to hopefully fix the failing tests), so you'll want to pull the latest in this branch

PR #2356, opened 2021-11-08
Add sdk for withdraw from reward safe

tintinthong:
- sdk to withdraw funds to safe owner
   - handles tricky thing of signing with a non-incremented nonce purely for gas estimation
   - relies on [change in relayer](https://github.com/cardstack/card-protocol-relay-service/pull/43)
   -  **Context of main problem:** `CannotEstimateGas` error occured due to failing gas esimate calls on the relayer -- the calls which use `requiredTxGas`. The `requiredTxGas` failed bcos it ran a simulation call which is different from the outer execution of the gnosis safe. The incremented nonce signed for the inner execution was not valid because there was NO outer execution of the gnosis safe, ie there was no outer execution to simulate the increase of the nonce.

### DIFF
--- a/packages/cardpay-sdk/sdk/constants.ts
+++ b/packages/cardpay-sdk/sdk/constants.ts
@@ -32,7 +32,7 @@ const SOKOL = {
   rpcNode: 'https://sokol.poa.network',
   rpcArchiveNode: 'https://sokol-archive.blockscout.com',
   rpcWssNode: 'wss://sokol.poa.network/wss',
-  relayServiceURL: 'https://relay-staging.stack.cards/api',
+  relayServiceURL: 'http://localhost:8000/api',
   subgraphURL: 'https://graph-staging.stack.cards/subgraphs/name/habdelra/cardpay-sokol',
   tallyServiceURL: 'https://tally-service-staging.stack.cards/api/v1',
   merchantUniLinkDomain: MERCHANT_PAYMENT_UNIVERSAL_LINK_STAGING_HOSTNAME,
--- a/packages/cardpay-sdk/sdk/constants.ts
+++ b/packages/cardpay-sdk/sdk/constants.ts
@@ -32,7 +32,7 @@ const SOKOL = {
   rpcNode: 'https://sokol.poa.network',
   rpcArchiveNode: 'https://sokol-archive.blockscout.com',
   rpcWssNode: 'wss://sokol.poa.network/wss',
-  relayServiceURL: 'http://localhost:8000/api',
+  relayServiceURL: 'https://relay-staging.stack.cards/api',
   subgraphURL: 'https://graph-staging.stack.cards/subgraphs/name/habdelra/cardpay-sokol',
   tallyServiceURL: 'https://tally-service-staging.stack.cards/api/v1',
   merchantUniLinkDomain: MERCHANT_PAYMENT_UNIVERSAL_LINK_STAGING_HOSTNAME,
--- a/packages/cardpay-sdk/sdk/reward-manager/base.ts
+++ b/packages/cardpay-sdk/sdk/reward-manager/base.ts
@@ -531,7 +531,7 @@ export default class RewardManager {
     }
 
     let rewardManager = await getSDK('RewardManager', this.layer2Web3);
-    let rewardManagerAddress = await this.address()
+    let rewardManagerAddress = await this.address();
 
     let from = contractOptions?.from ?? (await this.layer2Web3.eth.getAccounts())[0];
     let rewardSafeOwner = await rewardManager.getRewardSafeOwner(safeAddress);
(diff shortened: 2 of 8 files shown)


tintinthong:
This one has to really wait on protocol deploy for testing.

packages/cardpay-sdk/sdk/reward-manager/base.ts
```diff
@@ -35,6 +44,159 @@ export default class RewardManager {
     return await (await this.getRewardManager()).methods.getRewardSafeOwner(rewardSafeAddress).call();
   }
 
+  async withdraw(txnHash: string): Promise<TransactionReceipt>;
+  async withdraw(
+    safeAddress: string,
+    to: string,
```
alex-cardstack, line 50: On first glance, this looks good, the one comment I have is that the "to" is only allowed to be the owner of the reward safe at the moment. So we could calculate it, but maybe it still makes sense to pass it in in case we allow other addresses in future

tintinthong: So @alex-cardstack this is actually oversight by me when I reviewed the pr on card protocol. Reason why I put `to` is because the `withdraw` should go to ANY address not just the owner's address. We can set a ticket to fix this in the future.

tintinthong:
So I am stuck on this pr. Mainly because of gasEstimate issue. Funny thing there are two gnosis exec transactions here. And the innerExec has a gasPrice estimate of 250 gwei!! I am not sure if this is right. 

When experimenting with other things like the prepaid card (like when I specify facevalue of 50spend), I remember `CannotEstimateGas` error might either be generic or literally a gas estimate error. So im still debugging this but probably will need more time.

tintinthong:
@habdelra so I there is definitely a revert occuring on this on the relayer giving me GasEstimate error but I am having trouble finding the logic that is wrong. Can you help me eyeball this and check if I'm doing something weirdly wrong?

packages/cardpay-sdk/sdk/reward-manager/base.ts
```diff
@@ -496,6 +500,174 @@ export default class RewardManager {
...
+      )
+      .encodeABI();
+
+    if (nonce == null) {
+      nonce = getNextNonceFromEstimate(innerEstimate);
+      if (typeof onNonce === 'function') {
+        onNonce(nonce);
+      }
+    }
+    let estimate = await gasEstimate(this.layer2Web3, safeAddress, rewardManagerAddress, '0', payload, 0, tokenAddress);
```
tintinthong, line 623: `CannotEstimate` error  here. When I just hardcode the gas, I get safe transaction reverted.

packages/cardpay-sdk/sdk/reward-manager/base.ts
```diff
@@ -496,6 +500,174 @@ export default class RewardManager {
...
+
+    let gasCost = new BN(innerEstimate.dataGas).add(new BN(innerEstimate.baseGas)).mul(new BN(innerEstimate.gasPrice));
+    if (weiAmount.lt(gasCost)) {
+      throw new Error(
+        `Reward safe does not have enough to pay for gas when claiming rewards. The reward safe ${safeAddress} balance for token ${tokenAddress} is ${fromWei(
+          safeBalance
+        )}, amount being claimed is ${amount}, the gas cost is ${fromWei(gasCost)}`
+      );
+    }
+    let payload = (await this.getRewardManager()).methods
```
tintinthong, line 605: There probably is something wrong with this payload or the params of this payload.

tintinthong:
@alex-cardstack @habdelra this is good to review. I have added the changes that make it work.

packages/cardpay-sdk/sdk/constants.ts
```diff
@@ -32,7 +32,7 @@ const SOKOL = {
   rpcNode: 'https://sokol.poa.network',
   rpcArchiveNode: 'https://sokol-archive.blockscout.com',
   rpcWssNode: 'wss://sokol.poa.network/wss',
-  relayServiceURL: 'https://relay-staging.stack.cards/api',
+  relayServiceURL: 'http://localhost:8000/api',
```
alex-cardstack, line 35:
```suggestion
  relayServiceURL: 'https://relay-staging.stack.cards/api',
```

Should this be reverted back?

alex-cardstack (approved):
Looking good except maybe that url, good solution for the estimation issue though.

packages/cardpay-sdk/sdk/utils/signing-utils.ts
```diff
@@ -135,6 +135,55 @@ export async function signSafeTxAsBytes(
   return [await signTypedData(web3, owner, typedData)];
 }
 
+export async function fullSignatureTxAsBytes(
```
habdelra, line 138: could we rename this function. it's not clear to me as a caller of this util when I'd want to use `fullSignatureTxAsBytes` vs `signSafeTxAsBytes`. let's change the name of this function to indicate why it's different and when I'd use one vs the other

habdelra (approved):
👍

PR #2359, opened 2021-11-08
Fix/websocket stuff

Aierie:
Try to reconnect to websockets if they disconnect. If that fails, we disconnect and display an error message.

image

Coincidentally found out that this change should also fix the occasional issue with the DApp not responding when Card Wallet scans the Layer 2 QR code to connect - it happens when the websocket is disconnected, which happens after ~2 mins of inactivity.

https://github.com/cardstack/cardstack/pull/2359/commits/c0679fdc1731b902a0c972494f9a89dd08d3a03f is for removal, it provides a way to access one WalletConnect websocket via the console as `window.wsref`, in order to trigger closing of a websocket via `window.wsref.close()`. The implementation makes it hard to try this with Layer 1 and 2 connected via WalletConnect.

TODO:
- [x] tests - will try to fire the events to trigger websocket disconnection. 
- [ ] think about appropriate reconnection interval

### DIFF
--- a/packages/web-client/app/controllers/card-pay.ts
+++ b/packages/web-client/app/controllers/card-pay.ts
@@ -7,7 +7,7 @@ import { tracked } from '@glimmer/tracking';
 import { currentNetworkDisplayInfo as c } from '../utils/web3-strategies/network-display-info';
 import config from '../config/environment';
 
-interface ChainChangeModalOptions {
+interface NetworkProblemModalOptions {
   title: string;
   body: string;
   actionText: string;
@@ -23,7 +23,7 @@ export default class CardPayController extends Controller {
   @tracked isShowingLayer1ConnectModal = false;
   @tracked isShowingLayer2ConnectModal = false;
   @tracked needsReload = false;
-  @tracked chainChangeModalOptions: ChainChangeModalOptions | null = null;
+  @tracked networkProblemModalOptions: NetworkProblemModalOptions | null = null;
 
   constructor() {
     super(...arguments);
@@ -46,11 +46,11 @@ export default class CardPayController extends Controller {
   }
 
   @action showNetworkUnstableModal(network: keyof typeof c) {
-    if (this.chainChangeModalOptions) return;
-    this.showChainChangeModal({
+    if (this.networkProblemModalOptions) return;
+    this.showNetworkProblemModal({
       title: `Disconnected from ${c[network].fullName}`,
       body: `Sorry! Card Pay is disconnected from ${c[network].fullName}. You can restore the connection by refreshing the page.`,
-      onClose: this.hideChainChangeModal,
+      onClose: this.hideNetworkProblemModal,
       action: () => window.open(config.urls.discordSupportChannelUrl, '_blank'),
       actionText: 'Contact Cardstack Support',
       dismissable: true,
@@ -60,7 +60,7 @@ export default class CardPayController extends Controller {
   @action
   onLayer1Incorrect() {
     this.needsReload = true;
-    this.showChainChangeModal({
+    this.showNetworkProblemModal({
       title: `Please connect to ${c.layer1.fullName}`,
       body: `Card Pay uses ${c.layer1.fullName} as its Layer 1 network. To ensure the safety of your assets and transactions, this page will reload as soon as you change your wallet’s network to ${c.layer1.fullName} or disconnect your wallet.`,
       onClose: () => {},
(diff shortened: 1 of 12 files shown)


Aierie:
Will follow up with a PR for adding the network problem handling to Card Space.

packages/web-client/app/controllers/card-pay.ts
```diff
@@ -15,45 +16,53 @@ interface ChainChangeModalOptions {
...
+  }
+
   @action
   onLayer1Incorrect() {
     this.needsReload = true;
-    this.showChainChangeModal({
-      ...networkCorrectionMessages.layer1,
+    this.showNetworkProblemModal({
+      title: `Please connect to ${c.layer1.fullName}`,
+      body: `Card Pay uses ${c.layer1.fullName} as its Layer 1 network. To ensure the safety of your assets and transactions, this page will reload as soon as you change your wallet’s network to ${c.layer1.fullName} or disconnect your wallet.`,
```
jurgenwerk, line 65: I have similar thoughts about this copy as well. I think it's better we lay out the problem more precisely, and be more direct with what we need the user to do. My suggestion:

"We detected your wallet is connected to X network, while Y network is needed to manage your Card Pay assets and transactions. Please change your wallet's network to Y, and then reload the page."

packages/web-client/app/utils/wc-provider.ts
```diff
@@ -517,8 +519,35 @@ class WalletConnectProvider extends ExtendedProviderEngine {
     };
   }
 
+  bindSocketListeners() {
+    this.websocketProvider.on('close', this.onWebsocketClose.bind(this));
+    this.websocketProvider.on('connect', this.onWebsocketConnect.bind(this));
+  }
+
+  async maybeReconnect() {
+    setTimeout(() => {
```
jurgenwerk, line 528: How come `setTimeout` is needed?

Aierie: `setTimeout` is needed to delay execution of this code until the websocket provider has finished the callback for its close event, because part of that callback includes clearing all event listeners attached to it. We need to reattach the event listeners after that completes.

packages/web-client/app/controllers/card-pay.ts
Aierie: I agree that friendlier/less terse messaging makes sense here!

packages/web-client/app/utils/wc-provider.ts
lukemelia: This deserves to be a code comment.

packages/web-client/app/utils/wc-provider.ts
```diff
@@ -517,8 +519,39 @@ class WalletConnectProvider extends ExtendedProviderEngine {
...
+          return;
+        }
+        this.lastReconnection = Date.now();
+        this.websocketProvider.reset();
+        this.bindSocketListeners();
+        this.websocketProvider.connect();
+      } catch (e) {
+        this.emit('websocket-disconnected');
+      }
+    }, 0);
```
backspace, line 546: Maybe this is passé but it used to be that you’d see warnings when calling Ember-y things in a `setTimeout` callback, they’d need to be wrapped in `Ember.run` to be part of the runloop. `Ember.next` is a convenience function for this exact case: run with no delay but within a runloop.

packages/web-client/tests/acceptance/network-problems-test.ts
```diff
@@ -0,0 +1,101 @@
...
+
+    await visit('/card-pay/balances');
+  });
+
+  test('Layer 2 incorrect chain event triggers the showing of a modal', async function (assert) {
+    layer2Service.simpleEmitter.emit('incorrect-chain');
+
+    await settled();
+
+    assert.dom(MODAL_TITLE).containsText('Please connect to L2 test chain');
```
backspace, line 47: I think most tests use the `currentNetworkDisplayInfo` abstraction to generate the strings for comparison… I actually kind of like having the bare string, just wanted to highlight the possible discrepancy though 🤓

backspace (approved):
It all seems sad to me but if we have these flaky sockets, better to have explicit handling around that at least, thanks for the work on this.

packages/web-client/app/controllers/card-pay.ts
Aierie: Added https://linear.app/cardstack/issue/CS-2468/improve-error-messages-shown-when-network-problems-occur

PR #2361, opened 2021-11-08
[CS-1805] Further progress on indexing

ef4:


ef4:
Summarizing where we left off: we should consider implementing an `neq` filter operator analogous to `eq`, rather than relying on wrapping a whole expression in a `not` operator (which is often too blunt, because it affects the types and not just the values of the fields).

ef4:
Discussion points as we figure out next steps:

1. We need to implement the `Builder` interface such that it's backed by the search index. 
   - this gives us caching again
   - this lets us eliminate redundancy pre-indexing card cache capabilities
2. Challenge: we probably need inter-card invalidations right away, because indexing can't know one perfect card dependency order.
3. Challenge: the module defining features of the current card-cache are still needed (writing files that are visible to webpack and requirable by node).
4. The invalidation system needs to be able to target from one card to other cards *and* specific fields in other cards.
5. The reindexing / card compiling process needs to accept a list of invalidations (down to the field level) as an input so that it knows which parts can be left alone and which should be touched.
    - this may be something we can control at the Builder level rather than the Compiler level because it decides what to give back for `getRawCard` and `getCompiledCard`. But that might require making those APIs more field-level aware.
    - this is one place where we can choose to control versioning / snapshotting. This is under-designed but that's OK for now.
6. The index needs to represent error states for cards, and we need to figure out ways to shows those usefully to developers (as well as automatically resolve them when dependencies change and cause a card to unbreak). 

Next actions:
 - add a dependency column to the cards table that's capable of representing "I depend on card X, all fields" and "I depend card X, specific field".
 - add a test that:
   -  indexes a realm with a card with a broken dependency
   - sees the broken card giving a helpful error message (cardService.load throwing a good exception, which means API endpoint would also give helpful error response)
   - uses cardService.create to provide the missing dependency
   - sees that the original card is no longer broken (and we didn't need to do anything manual to get it to reindex)
 - at this point we can make the builder rely entirely on the search index

ef4:
More todos as we discover them:

 - [ ] move realm config into container, eliminating need for imperative `createRealm()`
 - [ ] implement `beginReplaceAll` and `endReplaceAll` and use them in the fs realm indexer, and then get rid of double `si.reset()` in test setup.
 - [ ] eliminate eager creation of card service in setupHub, in favor of layered separate setup functions, once of which gives you a synchronous card service.
 - [ ] update `lookupFieldsForCard` to throw context-sensitive errors in the same way that we did for `getParentCard`

PR #2364, opened 2021-11-09
Add reward rule

tintinthong:
- sdk for add reward rule 
- sdk for remove reward program (uses single owner safe to execute. its meant to be acting like a dao). And yes it also relies on gas tokens being available in that safe.


tintinthong:
Forked from pr 4

tintinthong:
@habdelra @alex-cardstack this is good to review

packages/cardpay-sdk/sdk/reward-manager/base.ts
```diff
@@ -308,6 +315,143 @@ export default class RewardManager {
...
+
+    if (typeof onTxnHash === 'function') {
+      await onTxnHash(txnHash);
+    }
+    return await waitForSubgraphIndexWithTxnReceipt(this.layer2Web3, txnHash);
+  }
+
+  async removeRewardProgram(txnHash: string): Promise<TransactionReceipt>;
+  async removeRewardProgram(
+    safeAddressIdOrTxnHash: string,
```
habdelra, line 388: when providing overloaded TS methods, the name of the params should reflect the specific overload--only the last implementation signature needs the "xxxORyyy" parameter. So this should be renamed to `safeAddress`

packages/cardpay-sdk/sdk/reward-manager/base.ts
```diff
@@ -308,6 +315,143 @@ export default class RewardManager {
...
+      throw new Error('rewardProgramId is required');
+    }
+    let { nonce, onNonce, onTxnHash } = txnOptions ?? {};
+    let safeAddress = safeAddressIdOrTxnHash;
+    let from = contractOptions?.from ?? (await this.layer2Web3.eth.getAccounts())[0];
+    let rewardManagerAddress = await getAddress('rewardManager', this.layer2Web3);
+    let rewardManager = await this.getRewardManager();
+
+    let payload = rewardManager.methods.removeRewardProgram(rewardProgramId).encodeABI();
+    let gasTokenAddress = await this.defaultGasToken();
```
habdelra, line 416: what? there is no "default gas token". I thought his was just going to be an EOA txn, which could optionally be a DAO in the future.

habdelra: it might even be the case where this is executed via the gnosis hosted relay server and not our relay server in that future scenario.

alex-cardstack (approved):
Looks good to me except for @habdelra's comments

packages/cardpay-sdk/sdk/reward-manager/base.ts
habdelra: crap--yeah that unsupported chain ID i think runs deep. I've seen it awhile back looking at an unrelated thing. but there is like a dep of a dep of a dep that has the network ID's hardcoded into it. because we don't do EOA txn's in layer 2 we normally woudn't see this.

So for EOA txn's it's totally ok to assume that the address has access to native coin--since that is the way ethereum actually works. This particular function is actually a function that I think that we as cardstack would only ever run (for the time being). Moreover, we would probably use blockscout to run this and not the SDK, and we defnite do have access to native coin. I think the answer here is to just remove this from the SDK. This is not a function that I would anticipate any of our current protocol clients (dap, cli, wallet app) would actually call.

tintinthong: >  I think the answer here is to just remove this from the SDK. This is not a function that I would anticipate any of our current protocol clients (dap, cli, wallet app) would actually call.

Makes lots of sense. I will remove it from the sdk.

tintinthong: @habdelra I removed the `removeRewardProgram`. Good to merge?

habdelra (approved):
👍

PR #2368, opened 2021-11-09
Update wallet tab content

jurgenwerk:
Ticket: [CS-2407](https://linear.app/cardstack/issue/CS-2407/update-wallet-tab)

Opened: 

image

Closed:

image

Todo:

- [x] Tidy up CSS, make it responsive, get rid of absolute units if possible
- [x] Add app store links
- [x] Add the bottom banner
- [x] Further rename files / css classes from "balances" to "wallet"


packages/web-client/app/templates/card-pay/wallet.hbs
```diff
@@ -0,0 +1,119 @@
...
+          <p class="dashboard-panel-section__desc">
+            {{svg-jar "card-wallet-app-icon" alt="" role="presentation" height="58"}}
+            <h3 class="dashboard-panel-section__title">Buy a Prepaid Card</h3>
+            <p class="dashboard-panel-section__desc">With Card Wallet you can buy a Prepaid Card directly in the mobile application.</p>
+            <div class="dashboard-panel-section__badges">
+              <div>
+                <a href={{this.testFlightLink}} target="_blank" rel="noopener noreferrer"><img src="/images/icons/download-on-the-app-store-badge.png" alt="Download on the App Store"></a>
+              </div>
+              <div>
+                <a href="#" onclick="return false"><img src="/images/icons/google-play-badge.png" class="dashboard-panel-section__badges-google-play-badge" alt="Get it on Google Play"></a>
```
Aierie, line 43: Should we omit the href and onclick here, and add `cursor: not-allowed`?

packages/web-client/app/components/card-pay/dashboard-panel/section/index.css
```diff
@@ -59,6 +59,70 @@
...
+  .dashboard-panel-section-wallet__cards {
+    width: calc(80% + (2 * 84px));
+  }
+}
+
+@media screen and (max-width: 1050px) {
+  .dashboard-panel-section-wallet__cards {
+    display: none;
+  }
+}
```
Aierie, line 124: A lot of this is specific to the Wallet route. What do you think about putting it in `wallet.css` instead and making the block of the BEM `card-pay-wallet-dashboard` or something similar?

Aierie (approved):
I think this looks good to merge, there is some room for styling improvements, but we can probably polish those things at a later date. The double footer thing doesn't seem ideal but given that the API only exposes one block, it seems ok for now.

packages/web-client/app/components/card-pay/dashboard-panel/index.css
```diff
@@ -56,6 +56,14 @@
   z-index: 1;
```
lukemelia, line 56:
```suggestion
  z-index: 1;
  cursor: pointer;
```

packages/web-client/app/components/card-pay/dashboard-panel/index.hbs
```diff
@@ -1,24 +1,41 @@
...
-    <header class="dashboard-panel__header">
-      <h2 class="dashboard-panel__header-title">{{@panel.title}}</h2>
-      <p class="dashboard-panel__header-desc">{{@panel.description}}</p>
-    </header>
-    <div class="dashboard-panel__hero" style={{css-url "background-image" @panel.heroImageUrl}} />
-    {{yield}}
-  </section>
-</details>
\ No newline at end of file
+{{#let (if @noCollapse true @open) as |isOpen|}}
```
lukemelia, line 1: We should probably remove this `let`. The original reason I introduced it didn't pan out, and we only use the result in one place.

packages/web-client/app/components/card-pay/dashboard-panel/index.hbs
```diff
@@ -1,24 +1,41 @@
...
+  <details
+    class={{cn
+      "dashboard-details-container"
```
lukemelia, line 4: The CSS class name pattern in this file is inconsistent. Sometimes it is `dashboard-details-container` and sometimes it is `dashboard-panel`. Based on the component name, I would expect it to consistently be `dashboard-panel`

packages/web-client/app/components/card-pay/dashboard-panel/section/index.css
```diff
@@ -59,6 +59,70 @@
...
+.dashboard-panel-section-wallet {
+  justify-content: flex-start;
+}
+
+.dashboard-panel-section-wallet__cards {
+  width: calc(100% + (2 * 84px));
+  position: absolute;
+  bottom: 0;
+  left: -84px;
+}
```
lukemelia, line 106: We should probably move all this wallet CSS to the wallet.css file and rename the classes accordingly.

packages/web-client/app/templates/card-pay/wallet.hbs
lukemelia: Sounds like a good idea!

lukemelia (changes requested):
Looks good. My suggestions are mostly requests to fix mistakes I introduced, lol.

lukemelia:
> The double footer thing doesn't seem ideal but given that the API only exposes one block, it seems ok for now

I would love to avoid rendering this twice. I couldn't figure out how to do it, though, within the confines of our use of details/summary.

packages/web-client/app/templates/card-pay/wallet.hbs
jurgenwerk: Or, we can just remove the link for now and only display the image

packages/web-client/app/components/card-pay/dashboard-panel/section/index.css
jurgenwerk: Makes sense!

packages/web-client/app/components/card-pay/dashboard-panel/index.css
jurgenwerk: 👍

packages/web-client/app/components/card-pay/dashboard-panel/index.hbs
jurgenwerk: Done!

packages/web-client/app/components/card-pay/dashboard-panel/index.hbs
jurgenwerk: Good suggestion - I cleaned that up.

packages/web-client/app/components/card-pay/dashboard-panel/section/index.css
jurgenwerk: Good idea.

PR #2369, opened 2021-11-09
web: Add Return to Dashboard button

backspace:
I’d prefer to be able to accomplish this in a higher-level
way vs having to update every workflow, but I couldn’t
figure out how.

This includes an override to Boxel CSS to add the thread
header button as that is not currently supported in Boxel.

### DIFF
--- a/packages/web-client/app/components/card-pay/create-merchant-workflow/index.hbs
+++ b/packages/web-client/app/components/card-pay/create-merchant-workflow/index.hbs
@@ -2,6 +2,7 @@
   <WorkflowThread
     class="create-merchant-workflow"
     @workflow={{this.workflow}}
+    @onClose={{@onClose}}
   />
 {{/if}}
 <Listener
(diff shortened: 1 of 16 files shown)


backspace:
I’ve looked at the Percy diffs and all seems well. I’m leaving it unapproved because it seems to me that should be something reviewers do vs the creator of a PR… does that make sense?

Aierie:
> I’ve looked at the Percy diffs and all seems well. I’m leaving it unapproved because it seems to me that should be something reviewers do vs the creator of a PR… does that make sense?

I think that makes sense. How do you feel about approving changes if they seem relatively minor, eg. in the decimal places PR?

Aierie (approved):
I agree it's tricky to provide a good way to centralize these controls. 

Just taking a stab at this - I'm imagining? this would involve changing the API of the workflow thread - generating the "before-content" stuff from markdown similar to a postable, binding listeners within the workflow class itself?, and making `Boxel::Modal` part of the `WorkflowThread` component directly - so that we could do this on the `deposit-withdrawal` route, for example:

```
<WorkflowThread
  @workflow={{this.depositWorkflow}}
  @isOpen={{eq this.flow 'deposit'}}
  @onClose={{this.resetQueryParams}}
  @initializing={{is-network-initialized}}
/>
```

But I don't know if that's what we should do? Also not entirely clear about how to handle the persistence id.

PR #2373, opened 2021-11-10
Boxel modal upstream fixes

burieberry:
Fixes CS-2507


packages/boxel/addon/components/boxel/modal/index.css
```diff
@@ -13,34 +13,44 @@
...
+  display: flex;
+  align-items: center;
+  justify-content: center;
   width: 100%;
   height: 100%;
-  max-width: var(--boxel-modal-max-width);
-  margin: 0 auto;
-  transition: max-width var(--boxel-transition);
+  margin: auto;
+  max-width: var(--boxel-modal-max-width, 65rem);
```
Aierie, line 50: Wondering if this is intentional? There is an escape hatch of sorts for modal sizing by omitting the size and setting the `--boxel-modal-max-width` variable. We could prevent this by setting a size default.

Aierie (approved):
Thanks, this fixes the modal nicely, and I like the new size arguments!

PR #2374, opened 2021-11-11
make token bridge eplorer link styles override boxel

Aierie:
I'm a bit hesitant to add yet another button variant before we know if we need it, so am making a judgement call to use an override for now. Also considered using `@kind="primary"`, but decided against it because it deviates from the spec, and feels like it's drawing too much attention to a detail, lmk if you think it's a better solution @burieberry? A bit unsure about how to approach this since we discussed respecting Boxel's constraints over the design spec.

### With href

image

### No href

image

### `@kind="primary"`

image

### DIFF
--- a/packages/web-client/app/components/card-pay/bridge/index.css
+++ b/packages/web-client/app/components/card-pay/bridge/index.css
@@ -17,7 +17,7 @@
   background: url('/images/backgrounds/bridge-background.svg');
 }
 
-/* 2 selectors because of specificity */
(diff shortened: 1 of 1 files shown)


lukemelia:
It seems like there could be a way to do this by conditionally setting the css variable (which is part of the API Boxel exposes) rather than overriding CSS.

burieberry (approved):
I think it's fine to do this until we figure out a better system. This won't be the only place where an outline button needs a background color the same as its surrounding.

PR #2375, opened 2021-11-11
update bridging finalization rates

Aierie:
Updated bridging finalization rate since https://docs.tokenbridge.net/eth-xdai-amb-bridge/about-the-eth-xdai-amb has recently changed.

re: these finalization rates, are they referencing [finality](https://ethereum.org/en/developers/docs/consensus-mechanisms/pow/#finality)?


habdelra (approved):
thanks!

PR #2380, opened 2021-11-12
web: Add Card Space button text chooser

backspace:
This adds a Card Space profile card button text chooser to the existing edit details card:

screencast 2021-11-18 14-24-59

This doesn’t attempt to address the duplication of [valid button text options](https://github.com/cardstack/cardstack/blob/15efcdf7f42b66c88ac2f4edc3ca75bc07bea459/packages/hub/services/validators/card-space.ts#L33) in Hub. Maybe a new directory in `packages` for sharing?


jurgenwerk (approved):
Looks nice! Regarding the duplicated validation options I think it's not a big deal and we can keep them in both places.

PR #2381, opened 2021-11-12
Demo cards view and edit modes

burieberry:
This PR adds CardSpaceLayout card and its instance, Emily's card space. The edit modes for all the contained cards are currently all in one place. There are some questions I added in the comments below. 

card-space-isolated

card-space-edit

card-space-emily

card-space-emily-edit


packages/cardhost/app/components/card-container/index.hbs
```diff
@@ -1,16 +1,19 @@
-<Boxel::CardContainer @displayBoundaries={{true}} style="margin:1vw 2vw;overflow:hidden;">
+<Boxel::CardContainer class="card-container" @displayBoundaries={{true}} ...attributes>
   <Boxel::Header @header="Layout">
```
burieberry, line 2: Added Boxel::Header component here because the edit buttons that need to be on the header are here. According to this, the isolated card container will always be called "Layout", but that may not always be the case.

packages/demo-cards/bio/embedded.js
```diff
@@ -3,11 +3,21 @@ import { precompileTemplate } from '@ember/template-compilation';
...
+import './embedded.css';
 
 export default setComponentTemplate(
-  precompileTemplate('<BoxelCardContainer><BoxelHeader @header="Bio" @editable={{true}}/></BoxelCardContainer>', {
-    strictMode: true,
-    scope: () => ({ BoxelCardContainer, BoxelHeader }),
-  }),
+  precompileTemplate(
+    `<BoxelCardContainer @displayBoundaries={{true}}>
+      <BoxelHeader>About Me</BoxelHeader>
```
burieberry, line 11: Not sure if the Boxel Card Container and Boxel Header for the contained cards should be on the template of each card, or if they should be on the CardSpaceLayout's isolated template.

packages/demo-cards/card-space-emily/card.json
```diff
@@ -1 +1,31 @@
-{"adoptsFrom":"../card-space-layout","data":{"profile":{"username":"Emily"}}}
+{
+  "adoptsFrom": "../card-space-layout",
+  "data": {
+    "background": {
+      "coverPhoto": "[COVER PHOTO]"
```
burieberry, line 5: The card space example has photo, link, and text area field types. I'm wondering if it's a good time to start creating those after this PR is merged, or if there's more work that needs to be done before we start adding more field types.

packages/demo-cards/card-space-layout/edit.js
```diff
@@ -0,0 +1,27 @@
...
+        <BoxelCardContainer @displayBoundaries={{true}}>
+          <:header>
+            {{name}}
+          </:header>
+          <:default>
+            <Field />
+          </:default>
+        </BoxelCardContainer>
+      {{/each-in}}
+    </section>`,
```
burieberry, line 20: For the edit templates, all the Boxel Card Container components are on the isolated edit.js file. They could also be on each card's edit.js file instead. Is there a preference for one approach over the other?

packages/demo-cards/bio/embedded.js
mattmcmanus: Hmmm. Have you tried doing on the isolated.js file of CardSpaceLayout? It seems like the visuals of the borders for these embedded views should really be controlled by the card that is inserting them.

packages/demo-cards/card-space-emily/card.json
mattmcmanus: I don't think there is any reason to wait. It sounds like you're work on the containsMany editing UI is also important, but that they're orthogonal

mattmcmanus (approved):
This is looking awesome!

PR #2383, opened 2021-11-15
Add ability to abandon (delete) incomplete workflows

jurgenwerk:
Ticket: [CS-2111](https://linear.app/cardstack/issue/CS-2111/abandon-workflow-modal)

This PR:

1. Adds a "danger" kind of button to Boxel

2. Adds a button + confirm modal for deleting active workflows

image

image

### DIFF
--- a/packages/web-client/app/components/card-pay/header/workflow-tracker/item/index.hbs
+++ b/packages/web-client/app/components/card-pay/header/workflow-tracker/item/index.hbs
@@ -1,4 +1,5 @@
 
+{{!-- template-lint-disable no-invalid-interactive --}}
 <div
   class={{cn "workflow-tracker-item" workflow-tracker-item--completed=this.isComplete}}
   data-test-workflow-tracker-item={{this.workflowId}}
--- a/packages/web-client/app/components/card-pay/header/workflow-tracker/item/index.hbs
+++ b/packages/web-client/app/components/card-pay/header/workflow-tracker/item/index.hbs
@@ -1,25 +1,75 @@
-<button
+
+<div
   class={{cn "workflow-tracker-item" workflow-tracker-item--completed=this.isComplete}}
-  type='button'
   data-test-workflow-tracker-item={{this.workflowId}}
-  {{on 'click' this.visit}}
+  {{on 'mouseover' this.showDeleteButton}}
+  {{on 'mouseleave' this.hideDeleteButton}}
   ...attributes
 >
-  <header class='workflow-tracker-item__heading'>
-    {{this.workflowDisplayName}}
-  </header>
-  <span class='workflow-tracker-item__milestone'>
-    {{#if this.isComplete}}
-      Complete
-    {{else}}
-      {{this.currentMilestoneTitle}}
-    {{/if}}
-  </span>
+  <div class="workflow-tracker-item__description">
+    <button
+      type='button'
+      {{on 'click' this.visit}}
+      data-test-visit-workflow-button
+    >
+      <header class='workflow-tracker-item__heading'>
+        {{this.workflowDisplayName}}
+      </header>
+      <div class='workflow-tracker-item__milestone'>
+        {{#if this.isComplete}}
+          Complete
+        {{else}}
+          {{this.currentMilestoneTitle}}
+        {{/if}}
+      </div>
+    </button>
+  </div>
 
-  <Boxel::ProgressIcon
-    class='workflow-tracker-item__icon'
-    @size="25"
-    @isComplete={{this.isComplete}}
-    @fractionComplete={{this.fractionComplete}}
-  />
-</button>
+  <div class="workflow-tracker-item__actions">
+    {{#if this.deleteButtonShown}}
+      <button
+        class="workflow-tracker-item__icon workflow-tracker-item__delete-icon"
+        type="button"
+        {{on 'click' this.showDeleteConfirmation}}
+        data-test-delete-workflow-button
+      >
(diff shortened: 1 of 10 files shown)


packages/web-client/app/components/card-pay/header/workflow-tracker/item/index.hbs
```diff
@@ -1,4 +1,5 @@
 
+{{!-- template-lint-disable no-invalid-interactive --}}
```
jurgenwerk, line 2: I'm struggling to conform to accessibility rules with this design 🤔

To satisfy the rule where buttons shouldn't be nested under each other, I refactored the workflow item component to have two buttons that aren't nested - one being the description part and the other one contains the delete icon.

However, this still isn't quite alright. The linter rule says `Interaction added to non-interactive element` - this is due to `{{on 'mouseover']]` and `{{on 'mouseleave'}}` handlers set on the parent `div` in this component. If I change the `div` to an interactive element, then I'm faced with the original accessibility issue again, described above.

I think this particular design (hover to unveil the buttons) isn't too accessible from the get-go. Any ideas? Can we ignore this issue in this case, or maybe adjust the design spec?

packages/web-client/public/images/icons/trash.svg
```diff
@@ -1,8 +1 @@
...
-  <g transform="translate(0.75 0.75)">
-    <path d="M3,6H17.4" transform="translate(-3 -2.8)" fill="none" stroke="#000" stroke-linecap="round" stroke-linejoin="round" stroke-width="1.5"/>
-    <path d="M15.844,5.2V16.4A1.575,1.575,0,0,1,14.3,18H6.549A1.575,1.575,0,0,1,5,16.4V5.2m2.324,0V3.6A1.575,1.575,0,0,1,8.873,2h3.1a1.575,1.575,0,0,1,1.549,1.6V5.2" transform="translate(-3.222 -2)" fill="none" stroke="#000" stroke-linecap="round" stroke-linejoin="round" stroke-width="1.5"/>
-    <path d="M10,11v5.333" transform="translate(-4.4 -4.143)" fill="none" stroke="#000" stroke-linecap="round" stroke-linejoin="round" stroke-width="1.5"/>
-    <path d="M14,11v5.333" transform="translate(-5.2 -4.143)" fill="none" stroke="#000" stroke-linecap="round" stroke-linejoin="round" stroke-width="1.5"/>
-  </g>
-</svg>
+<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24" fill="none" stroke="#fff" stroke-width="3" stroke-linecap="round" stroke-linejoin="round"><polyline points="3 6 5 6 21 6" /><path d="M19 6v14a2 2 0 0 1-2 2H7a2 2 0 0 1-2-2V6m3 0V4a2 2 0 0 1 2-2h4a2 2 0 0 1 2 2v2" /><line x1="10" y1="11" x2="10" y2="17" /><line x1="14" y1="11" x2="14" y2="17" /></svg>
```
jurgenwerk, line 1: @Aierie I took the liberty to change the trash icon to fit this feature, but this probably breaks the current upload component design. Given that that component is a prototype, are you ok if we adjust this icon when we return to the upload feature?

Aierie: Yeah, no worries on that front!

packages/web-client/app/components/card-pay/header/workflow-tracker/item/index.hbs
Aierie: It's possible to make the delete icon button `display: none`, and control its visibility via CSS's `:hover` pseudo-class on the parent:

```
.workflow-tracker-item:hover .workflow-tracker-item__delete-icon {
  display: flex;
}
```

thus circumventing this problem. An additional step we can take is to use a screenreader-friendly implementation instead of `display: none;`, see `.boxel-sr-only`, Boxel's screenreader-only utility class.

packages/web-client/tests/acceptance/persistence-view-restore-test.ts
```diff
@@ -305,14 +306,61 @@ module('Acceptance | persistence view and restore', function () {
...
+            milestonesCount: 3,
+          },
+        }),
+      });
+
+      await visit('/card-pay/');
+      await click('[data-test-workflow-tracker-toggle]');
+      await triggerEvent('[data-test-workflow-tracker-item]', 'mouseover');
+
+      assert.dom('[data-test-delete-workflow]').doesNotExist();
```
Aierie, line 361:
```suggestion
      assert.dom('[data-test-delete-workflow-button]').doesNotExist();
```

Aierie (approved):

packages/web-client/app/components/card-pay/header/workflow-tracker/item/index.hbs
Aierie: @jurgenwerk I think it's ok to ignore this issue for now. There are also issues around keyboard navigation - we don't show the delete button when. We can raise this in the design audit of Card Pay.

Aierie: Noted this in https://linear.app/cardstack/issue/CS-1745/visual-audit-of-card-pay-ui-on-production

jurgenwerk: I see, thank you @Aierie!

backspace:
I’m late getting to this but I tried it out on staging, much-needed 🤩

PR #2388, opened 2021-11-17
Adjust copy + add refresh page cta

jurgenwerk:
Ticket: [CS-2556](https://linear.app/cardstack/issue/CS-2556/add-a-refresh-button-to-the-disconnection-error-pop-up)

image


Aierie (commented):
Good copy improvement!

The task is for adding a second button, but I think replacing the existing button could work as well... do you think having a single button is better?

jurgenwerk:
@Aierie yeah I think the support button would also be useful! Something like this?

image

Aierie:
> @Aierie yeah I think the support button would also be useful! Something like this?
> 
> image

That looks good to me! 😄

Aierie (approved):

PR #2393, opened 2021-11-18
Use retry UI (over cancel) for deposit workflow

Aierie:
Modifies the deposit workflow to use a retry UI that allows us to keep track of every deposit signing request sent, until one returns with a transaction hash, and then we continue following only that one. The reason for this is that we don't reliably have a way to cancel transactions via nonce. 

image

Here's a loom of it in action (there have been some copy changes since - see above, but the retry mechanism is still the same): https://www.loom.com/share/779d508328b64ee78635cb6442a7346f

### Some stuff I'm not sure about
- Is "Resend Signing Request" the right button text? Is there a better way to phrase it? Should it be "Resend Signature Request"?

### Some work that could/should follow:
- Rename Action Chin's Cancel button to something that encompasses retry and cancel
- Add retry UI to the withdrawal flow's claim step

### Some edge cases that this does not cover
- User clicks deposit and refreshes the page, hence "disconnecting" our JS and the signing request in the user's layer 1 wallet
- Transactions that fail after a transaction hash is returned (to investigate in https://linear.app/cardstack/issue/CS-2502/investigate-how-the-dapp-handles-transactions-that-fail-after)

### DIFF
--- a/packages/web-client/app/components/card-pay/deposit-workflow/transaction-amount/index.hbs
+++ b/packages/web-client/app/components/card-pay/deposit-workflow/transaction-amount/index.hbs
@@ -57,17 +57,17 @@
       * The actual value depends on the current exchange rate and is determined at the time of authorization.
     </div>
     {{#if (and this.errorMessage this.depositTaskRetriable)}}
-      <CardPay::ErrorMessage as |supportURL|>
+      <CardPay::ErrorMessage data-test-deposit-error-message as |supportURL|>
         {{this.errorMessage}}
         Please try again if you want to continue with this workflow. If you haven't received a signing request, click on "Resend signing request" below, or contact <a href={{supportURL}} target="_blank" rel="noopener noreferrer">Cardstack Support</a>
       </CardPay::ErrorMessage>
     {{else if this.errorMessage}}
-      <CardPay::ErrorMessage as |supportURL|>
+      <CardPay::ErrorMessage data-test-deposit-error-message as |supportURL|>
         {{this.errorMessage}}
         Please try again if you want to continue with this workflow, or contact <a href={{supportURL}} target="_blank" rel="noopener noreferrer">Cardstack support</a>.
       </CardPay::ErrorMessage>
     {{else if this.depositTaskRetriable}}
-      <CardPay::ErrorMessage as |supportURL|>
+      <CardPay::ErrorMessage data-test-deposit-error-message as |supportURL|>
         If you haven't received a signing request, make sure to check your connected {{network-display-info "layer1" "fullName"}} wallet. You can also try to click on "Resend signing request" below, or contact <a href={{supportURL}} target="_blank" rel="noopener noreferrer">Cardstack Support</a>.
       </CardPay::ErrorMessage>
     {{/if}}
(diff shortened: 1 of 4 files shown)


packages/web-client/app/components/card-pay/deposit-workflow/transaction-amount/index.hbs
```diff
@@ -53,14 +53,24 @@
...
     </div>
-    {{#if this.errorMessage}}
-      <CardPay::ErrorMessage as |supportURL|>
-        {{this.errorMessage}} Please try again if you want to continue with this workflow, or contact <a href={{supportURL}} target="_blank" rel="noopener noreferrer">Cardstack support</a>.
-      </CardPay::ErrorMessage>
-    {{/if}}
     <div class="transaction-amount__footnote">
       * The actual value depends on the current exchange rate and is determined at the time of authorization.
     </div>
+    {{#if (and this.errorMessage this.depositTaskRetriable)}}
```
lukemelia, line 59: Although there is no authoritative correct spelling of retriable/retryable, I would go with the "y" version. Background reading: https://english.stackexchange.com/questions/305273/retriable-or-retryable

lukemelia (approved):

packages/web-client/tests/integration/components/card-pay/deposit-workflow/transaction-amount-test.ts
```diff
@@ -283,5 +298,124 @@ module(
...
+        await settled();
+
+        assert.ok(isComplete, 'onComplete was called');
+        assert.ok(
+          find('[data-test-deposit-etherscan-button]')
+            ?.getAttribute('href')
+            ?.includes('attempt 1 hash')
+        );
+      });
+    });
```
backspace, line 419: nice tests 🤩

packages/web-client/app/components/card-pay/deposit-workflow/transaction-amount/index.ts
```diff
@@ -213,43 +220,76 @@ class CardPayDepositWorkflowTransactionAmountComponent extends Component<Workflo
...
+            'layer2BlockHeightBeforeBridging',
+            layer2BlockHeightBeforeBridging
+          );
+          session.setValue('relayTokensTxnHash', txnHash);
+          this.errorMessage = '';
+          this.depositTaskRunningForAWhile = false;
+          this.cancelDepositsOtherThan(uid);
+        }
+      );
+      session.setValue('setting task', uid);
```
backspace, line 251: Is this leftover from debugging? 🤔

backspace (approved):
re your question about the button label, is “Retry” bad news? It seems suitably concise and accurate to me.

I noted a possible leftover from development so I’ll leave this unmerged but otherwise I’m 👍🏻 looks good!

packages/web-client/app/components/card-pay/deposit-workflow/transaction-amount/index.ts
Aierie: Yeah it is 😅

Aierie:
> re your question about the button label, is “Retry” bad news? It seems suitably concise and accurate to me.

Me and @jurgenwerk decided to use "Resend Signing Request" because retry didn't feel like it would be clear enough to a user who didn't have the understanding of the workflows that we do, originally it was "Retry"

backspace:
ya I guess “Retry” could suggest duplication, so the more verbose wording removes that concern

jurgenwerk (approved):
LGTM! And, as discussed, we're now using the "contact cardstack support" in several places and it would be helpful to extract that into a helper next time we need it 🙂

PR #2395, opened 2021-11-18
temporary fix for card space modal getting squeezed

Aierie:
used inline styles so the hacks are in one place.


backspace (approved):
much better, thanks!

image

PR #2397, opened 2021-11-19
Image upload endpoint

jurgenwerk:
This PR aims to add an /upload endpoint which will be used for image uploads in Card Space (profile image, background image). 

The uploading mechanism works like this:
1. Web client initiates a hub upload during the card space create or edit workflow from the image upload component
2. The hub uploads the image to [web3.storage](https://web3.storage/)
3. Web3 storage returns a CID which is used to construct an IPFS link, e.g https://bafybeifyqe334kqvxyswdwf5vi4ezpjmatrznq7a47fmq65uowls7pevzi.ipfs.dweb.link/branko.jpg
4. Then we use transform this URL to a CloudFlare IPFS CDN link, e.g. https://cloudflare-ipfs.com/ipfs/bafybeifyqe334kqvxyswdwf5vi4ezpjmatrznq7a47fmq65uowls7pevzi/branko.jpg
5. The hub creates a record in the uploads table, which we can use for keeping references that we can later use for housekeeping (deleting zombie uploads that aren't attached to card spaces) 
6. Endpoint responds with the image url
7. Web client can use this image url to create/edit card spaces 

To test this locally, register an account at [web3.storage](https://web3.storage/), generate a token and put it in your `.env` file: `WEB3_STORAGE_TOKEN=abc`. To test the request, I was using Postman configured with a `form-data` request.

Issues:

1. When uploading the same file several times, the url of the uploaded image will be the same. This means that if someone uploads the same image several times, we'll also have several same entries in the uploads table (apart from the id). I'm not sure if we should care 🤔

TODO:

- [x] More edge cases tests (file type, size)
- [x] Fix the build


jurgenwerk:
Getting an odd error in the discord bot after adding the web3 storage dependency to the app 🤨 https://github.com/cardstack/cardstack/runs/4263702208?check_suite_focus=true#step:9:312

packages/hub/routes/upload.ts
```diff
@@ -0,0 +1,114 @@
...
+  }
+
+  validate(byteSize: number, type: string): string | void {
+    // TODO: validate image dimensions?
+
+    let allowedTypes = ['image/jpeg', 'image/png'];
+
+    if (!allowedTypes.includes(type)) {
+      return `File type unsupported. Allowed types: JPG, JPEG, PNG`;
+    }
```
backspace, line 100: Should there be a test that exercises this? Like uploading a GIF

packages/hub/services/queries/upload.ts
```diff
@@ -0,0 +1,58 @@
...
+    });
+  }
+
+  async isAbusing(ownerAddress: string): Promise<boolean> {
+    let db = await this.databaseManager.getClient();
+
+    const query = `SELECT COUNT(*) FROM uploads WHERE owner_address = $1 AND created_at > now() - INTERVAL '10 min'`;
+    const queryResult = await db.query(query, [ownerAddress]);
+
+    return queryResult.rows[0]['count'] >= 10;
```
backspace, line 50: I wonder if these magic numbers (10 updates, 10 minutes) should eventually be extracted. Also, maybe we’ll want to obfuscate this at some point as abuse detection is sometimes worth hiding even in an otherwise-open source project, but I don’t know how that reconciles with the overall goals 🤔

backspace (approved):
I was able to exercise this:

```
curl -v -F upload=@/Users/b/Pictures/desert.png -H "Authorization: Bearer: $AUTH_TOKEN" http://localhost:3000/upload
```

The uploaded file showed up [here](https://cloudflare-ipfs.com/ipfs/bafybeigatmfwfqpcv6hbcq2pohonvkre4p2xjfiejftsyurlkcf7hobjfm/upload). Nice work!

I did have trouble initially with the `curl` because I needed `Bearer: TOKEN` where the standard [would be](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Authorization#syntax) `Bearer TOKEN`. It looks like the presence of the colon was already established outside this PR though, so I’ll address that elsewhere.

backspace:
Depending if/when #2398 is merged, this will need updating to remove the `Bearer: TOKEN` colon 🤓

packages/hub/config/structure.sql
```diff
@@ -693,6 +693,18 @@ CREATE TABLE public.card_spaces (
...
+--
+
+CREATE TABLE public.cards (
+    url text NOT NULL,
+    data jsonb
+);
+
+
+ALTER TABLE public.cards OWNER TO postgres;
+
```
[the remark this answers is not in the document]
jurgenwerk: Ideally this should go into another PR, but when I was updating structure.sql after adding a new table, I discovered this was missing so I committed it here. It looks like we forgot to add this when cards table was introduced to the app.

packages/hub/routes/upload.ts
jurgenwerk: Good point - will try to add that.

packages/hub/services/queries/upload.ts
jurgenwerk: Yeah that's a good point. Perhaps eventually the responsibility for the rate limiting should be moved out from the app and into the web server layer

jurgenwerk:
Accidentally closed the PR 🙈 Reopening.

packages/hub/routes/upload.ts
jurgenwerk: Added!

jurgenwerk:
Merging https://github.com/cardstack/cardstack/pull/2405 will fix the build failure.

backspace (approved):
Nice updates!

I’ve now merged #2398 so you could update to remove the `:` in the `Authorization` header value but the implementation is flexible enough to accept both with and without colon.

Aierie (approved):

lukemelia:
Is there a way we can avoid holding the contents of the image in memory in the hub instance? In AWS, we have pre-signed URLs that allow for uploading directly from the browser. This model is nice to protect against memory and state issues. e.g. what happens if the upload to the hub is successful but the upload to web3.storage fails?

packages/hub/config/structure.sql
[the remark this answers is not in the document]
habdelra: This schema change is actually necessary for the compiler work that @ef4 and team is working on--weird that is showed up here though

lukemelia:
After discussing with @habdelra, it sounds like there is likely no way to approximate the pre-signed URL approach with web3.storage currently. If anyone has ideas, let them fly, otherwise this seems good to go.

habdelra:
yeah, it's really unfortunate. it would be nice if there was a way for the server to sign the request (this is how wyre allows direct communication between clients and wyre servers using our wyre account). it seems like an obvious approach here that web3.storage could take so that we don't need to do server 2 server communication to hide our token (it still requires a central authority to perform the signatures--but would take us out of the critical path for uploads).

packages/hub/services/web3-storage.ts
```diff
@@ -0,0 +1,24 @@
...
+import { Filelike } from 'web3.storage/dist/src/lib/interface';
+import config from 'config';
+
+export default class Web3Storage {
+  async upload(path: string, name: string) {
+    const client = new Storage({
+      token: config.get('web3storage.token'),
+    });
+
+    let files = await getFilesFromPath(path);
```
Aierie, line 12: @jurgenwerk if this is meant to work on the hub, can `getFilesFromPath` actually get the files? It feels like it can't, and we might need to use `web3Storage.put`?

Aierie: Does koa/formidable do some work in the background that makes uploaded files have a path?

jurgenwerk: It looks like the files are uploaded to /tmp path. I didn't manage to find a way where we could stream the file upload directly to web3 storage, since the Web3.Storage client's put method only accepts [an array of File objects](https://docs.web3.storage/how-tos/store/#preparing-files-for-upload)

jurgenwerk:
@lukemelia @habdelra yeah unfortunately web3 storage doesn't support signed upload urls yet. I talked to the web3 product manager in the ipfs discord and he said web3 storage will eventually have something like that and that they are looking into it. 

> what happens if the upload to the hub is successful but the upload to web3.storage fails?

The file will be kept in /tmp until the next reboot. Actually now that I think about it I think it would be wise to ensure we delete the tmp file after the upload call - I'll look into it

PR #2398, opened 2021-11-19
hub: Remove colon in Authorization header

backspace:
I noticed while exercising #2397 that I needed a colon
between the authorisation scheme (Bearer) and the token,
which is counter to the [standard](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Authorization#syntax).

The only backward compatibility wrinkle I can think of with this is that a locally-installed Card Pay CLI will need updating for the changes in that package.


habdelra:
This looks good--but I think we also need to let the wallet team know about this change. because of the awkwardness of how mobile updates are pushed out I wonder i we need to support with and without ":" in the Authorization header

backspace:
ya I did wonder if this was a Postel’s Law situation; it would be easy enough to split on ` ` and have it work with and without colon, what do you think of that?

habdelra:
yeah, let's do that 👍

PR #2399, opened 2021-11-19
web: Update token icons with CPXD versions

backspace:
image

I was able to produce that locally by running against mainnet, as testnet doesn’t have the `.CPXD` suffixes on token symbols yet:

```
DEPLOY_TARGET=production ember s
```


Aierie (approved):

Aierie:
Might be good at some point in the future to have a package where token-related design assets/specs are shared between DApp and wallet?

backspace:
yes, agreed!

PR #2400, opened 2021-11-20
web: Add Card Space category chooser

backspace:
This adds an extracted `Common::RadioOption` component to encapsulate styling for radio button options. It then uses that in a new `CardSpace::EditDetails::Category` component patterned after other Card Space setup components.

screencast 2021-12-22 14-37-04

## Other field

The design specified an “Other” field but not how it should be implemented.

image

Rather than stacking yet another modal, I chose to use an inline `Common::ValidationStateInput` field backed by a new Hub validation endpoint for `profile-category` (patterned after `POST /api/card-spaces/validate-url`).

The implementation has these features that you can see in the tests:
* clicking the Other radio button autofocuses the text field
* typing anything in the field causes Other to become chosen (after validation success)

### Layout thinking

It looked fine in a valid state:

image

I thought it looked weird with a validation error showing:

image

Shouldn’t the label stay aligned with the field?

image

But then shouldn’t the radio button be aligned with the label?

image

That’s where I ended up!


packages/web-client/app/components/common/radio-option/index.css
```diff
@@ -0,0 +1,52 @@
+.radio-option {
+  --option-opacity: 1;
+  --option-input-border-color: var(--boxel-dark);
+
+  display: grid;
+  grid-template-columns: auto 1fr 1fr;
```
Aierie, line 6: Would it be better to have just 2 columns for now?

packages/hub/routes/card-spaces.ts
```diff
@@ -170,6 +170,21 @@ export default class CardSpacesRoute {
     ctx.type = 'application/vnd.api+json';
   }
 
+  async postProfileCategoryValidation(ctx: Koa.Context) {
+    if (!ensureLoggedIn(ctx)) {
+      return;
+    }
+
+    let profileCategory: string = ctx.request.body.data.attributes['profile-category'];
+    let errors = await this.cardSpaceValidator.validate({ profileCategory } as CardSpace);
```
Aierie, line 179: Do we need to add the profanity filter to profile category validation?

Aierie:
Not an effect of this PR but might be relevant: radio inputs trap escape key presses and lead to modal failing to close seemingly randomly. This doesn't play well with focus trap either, which returns focus to the context outside of the modal but keep the modal open.

packages/web-client/app/components/card-space/create-space-workflow/details/index.css
```diff
@@ -0,0 +1,5 @@
+/* FIXME without this, cannot see the whole form, but this causes a permanent scroll thumb */
+.create-space-workflow-details__modal .boxel-modal__inner {
+  align-items: flex-start;
+  overflow: scroll;
+}
```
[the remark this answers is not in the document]
backspace: I added a note in that issue since it seems directly related to making it “suitable for multi-layered usage”, thanks!

packages/web-client/app/components/common/radio-option/index.css
backspace: hmm yes, it does make sense that the shared CSS should be lowest common denominator-esque, so I’ve done that, thanks

packages/hub/routes/card-spaces.ts
backspace: probably… anywhere user-editable strings show up, really. I’m having build problems but will try this next week, thanks.

jurgenwerk:
A small UX suggestion - should we hide the "Other input" until Other option is actually selected? I think it's unnecessary to show it at all times.

Also, there are cases where this can be confusing, i.e. when user enters a custom category but then switches to a predefined one. If the input got nullified and hidden there would be less distraction IMO. 

image
image

backspace:
I tried to create a test that would exercise the early return in `validateCategoryTask` but I couldn’t get it to reliably fail when I commented out the early return.

```
    test('it ignores the validated other value if a preset value has been chosen in the meantime', async function (this: Context, assert) {
      let validationResolver: Function | undefined;
      let validationPromise = new Promise(function (resolve) {
        validationResolver = resolve;
      });

      this.server.post(
        '/card-spaces/validate-profile-category',
        async function () {
          console.log('going to await');
          await validationPromise;
          console.log('wated');
          return new MirageResponse(200, {}, { errors: [] });
        }
      );

      let workflowSession = new WorkflowSession();
      this.set('workflowSession', workflowSession);

      await render(hbs`
        <CardSpace::EditDetails::Category
          @workflowSession={{this.workflowSession}}
        />
      `);

      await click('[data-test-category-option-other-container]');
      console.log('going to fill');
      await fillIn('[data-test-category-option-other] input', 'Something');

      await click(`[data-test-category-option]:nth-child(2)`);

      assert.equal(
        workflowSession.getValue<string>('profileCategory'),
        OPTIONS[1]
      );
      assert.dom('[data-test-category-option]:nth-child(2) input').isChecked();

      if (validationResolver) {
        validationResolver(true);
      }

      await settled();
      await this.pauseTest();

      assert.equal(
        workflowSession.getValue<string>('profileCategory'),
        OPTIONS[1]
      );
      assert.dom('[data-test-category-option]:nth-child(2) input').isChecked();
    });
```

If you run it like this with a `pauseTest` and then manually `resumeTest` it fails, but the various delays I tried adding didn’t make it fail without that manual pause. Puzzling! I even tried `buildWaiter` from `@ember/test-waiters` (as well as the `@waitFor` decorator for the task) but those made the test hang indefinitely at the beginning, maybe not a surprise… 🧐

packages/hub/routes/card-spaces.ts
backspace: That’s done now, thanks.

backspace:
> should we hide the "Other input" until Other option is actually selected? I think it's unnecessary to show it at all times.

@Aierie and I paired on that yesterday, now it’s done, thanks!

PR #2404, opened 2021-11-22
Postcss maybe

Aierie:
Adds postcss to prepend a selector to css selectors in `web-client`'s components to increase specificity. Note that current implementation prepends an ID and makes pretty much all DApp CSS take priority over Boxel. It is possible to prepend a class, or a html element selector instead if we want to still allow Boxel to maintain some control over styling via specificity.

Am still in the middle of reading the cascade layers spec, but I'm drawing some parallels between this situation and CSS Cascade Layers + a suggested polyfill: https://github.com/w3c/csswg-drafts/issues/4985#issuecomment-758209624


packages/web-client/ember-cli-build.js
```diff
@@ -3,8 +3,37 @@
...
+const prependSelector = postcss.plugin(
+  'postcss-prepend-selector',
+  function (opts) {
+    opts = opts || {};
+    return function (css) {
+      css.walkRules(function (rule) {
+        rule.selectors = rule.selectors.map(function (selector) {
+          if (/(^|\b)(html|:root)\b/.test(selector)) {
+            return selector;
+          }
```
Aierie, line 19: The `#web-client-root` selector is on the html element, so we don't want to do anything with selectors that are targeting html or root.

packages/web-client/ember-cli-build.js
```diff
@@ -67,6 +108,24 @@ module.exports = function (defaults) {
...
+                      plugins: [
+                        prependSelector({
+                          selector: '#web-client-root ',
+                        }),
+                      ],
+                    },
+                  },
+                },
+              ],
+            },
```
Aierie, line 128: Necessary because some of web client's css is in `web-client/app/css`

packages/web-client/ember-cli-build.js
```diff
@@ -3,8 +3,37 @@
...
+            return selector;
+          }
+
+          return opts.selector + selector;
+        });
+      });
+    };
+  }
+);
+
```
backspace, line 36: I noticed this when running locally:

```
postcss-prepend-selector: postcss.plugin was deprecated. Migration guide:
https://evilmartians.com/chronicles/postcss-8-plugin-migration
```

So I adapted the relevant migration to the Ember CLI context:

```suggestion
const compileCSS = require('broccoli-postcss');
const webpack = require('webpack');

const prependSelector = (opts = {}) => {
  return {
    postcssPlugin: 'postcss-prepend-selector',
    Once(css) {
      css.walkRules(function (rule) {
        rule.selectors = rule.selectors.map(function (selector) {
          if (/(^|\b)(html|:root)\b/.test(selector)) {
            return selector;
          }
          // eslint-disable-next-line no-useless-escape
          if (/^([0-9]*[.])?[0-9]+\%$|^from$|^to$/.test(selector)) {
            // This is part of a keyframe
            return selector;
          }

          if (selector.startsWith(opts.selector.trim())) {
            return selector;
          }

          return opts.selector + selector;
        });
      });
    },
  };
};

```

It seems to have worked!

image

backspace (changes requested):
I tried this out and didn’t notice any difference in build times, which is excellent! I suggested an update to address a deprecation warning but otherwise looks good to me 🎉

packages/web-client/ember-cli-build.js
backspace: the [migration guide](https://evilmartians.com/chronicles/postcss-8-plugin-migration) has more after that actually, including changing `Once` to `AtRule`, might be worth looking more deeply into. I’m surprised `Once` didn’t need to be imported…???

backspace (approved):
total success! 🎉

jurgenwerk (approved):
Looks good!

PR #2406, opened 2021-11-22
fix issue prepaid card workflow not restoring

Aierie:
We renamed `balances` to `wallet`.


backspace (approved):
eep, how distressing 😳 thanks for fixing this, I wonder how we could ensure test coverage for this in the future 🤔

PR #2409, opened 2021-11-23
web: Add truncation for merchant names

backspace:
This is a brute force approach to addressing the unbounded text problem in merchant names. Maybe we should restrict length and get better design feedback, but in the interim, this will do, IMO.

Before and afters:

image

image


Aierie (approved):
Yeah, we should probably get length restrictions on merchant names (as we are going to do with spaces)

PR #2415, opened 2021-11-25
Fix balance & proof querying

tintinthong:
This is just fixing the balance and proof querying with new tally changes. Will fork-off this branch for the claiming


tintinthong:
@IanCal this is all moved to here https://github.com/cardstack/cardstack/pull/2421.

PR #2416, opened 2021-11-25
Pick username card prototype

Aierie:
This PR will be split into smaller PRs and is here for reference for how each smaller PR is to be used.

TODO:
- [x] PR for Hub route for validating user name (HUB2:) (https://github.com/cardstack/cardstack/pull/2424)
- [ ] Merge previous PRs then complete this card if API for image upload/edit in https://github.com/cardstack/cardstack/pull/2416/commits/577cc935c204fd917e1daa2bf116ac438acdb31a seems ok (maybe make the image uploader and editor more convenient to use together before that?)

Should do quickly after this is merged:
- [ ] Update Boxel's Modal for multi-layer use + fix other bugs
- [ ] PR for Boxel Button icon size variant, to replace rotation buttons + the trash button

PR #2417, opened 2021-11-25
Editor improvements

Aierie:
1. Make the editor accept arguments for a desired width and height, which will:
    - form the aspect ratio
    - determine the dimensions of the cropped image
2. Adds an error state (local) to the editor. 
3. Allows the editor to output the file type specified in the `fileType` argument.

Corresponds to commits with messages starting with EDITOR: in https://github.com/cardstack/cardstack/pull/2416


packages/web-client/app/components/image-editor/index.ts
```diff
@@ -1,69 +1,151 @@
...
+      }
+      this.errored = false;
+      let component = this;
+      this.cropper = new Cropper(image, {
+        autoCrop: true,
+        aspectRatio: this.args.width / this.args.height,
+        autoCropArea: 0.9,
+        viewMode: 1,
+        cropBoxMovable: false,
+        cropBoxResizable: false,
```
backspace, line 39: I was able to use scrolling on desktop to resize the image, but I think that’s somewhat obscure and wouldn’t work for certain pointer devices, maybe this should be `true`? I know this was already `false`, I just noticed it now 🤓

packages/web-client/tests/integration/components/image-editor-test.ts
```diff
@@ -65,6 +82,36 @@ module('Integration | Component | image-editor', function (hooks) {
 
     await click(SAVE_BUTTON);
 
-    assert.equal(result, rotatedAndCroppedImageDataUri);
+    // not sure why settled isn't picking up the async calls?
```
backspace, line 85: It might be because Ember is unaware of the functions, a [test waiter](https://github.com/emberjs/ember-test-waiters) might be of use to register the asynchronous calls so the testing infrastructure knows to wait.

backspace (approved):
🎉

packages/web-client/app/components/image-editor/index.ts
Aierie: Pinch zoom/resize is possible for pointer devices. This UI is emulating Twitter's image selection + cropping - not sure if it's achieving the same effect?

PR #2418, opened 2021-11-25
Validation improvements

Aierie:
Update image validation to have defaults closer to what we want to use it for + make the validation return a message that can be used in the UI.

Corresponds to commits with messages starting with VALIDATION: in #2416

### DIFF
--- a/packages/web-client/app/utils/image.ts
+++ b/packages/web-client/app/utils/image.ts
@@ -117,7 +117,7 @@ export class ImageValidation {
       } else {
         message = `Please upload a file between ${fileSize(
           minFileSize
-        )} and ${fileSize(maxFileSize)}.`;
+        )} and ${fileSize(maxFileSize)}`;
       }
     } else if (!imageSizeValid) {
       const { minWidth, maxWidth, minHeight, maxHeight } = this;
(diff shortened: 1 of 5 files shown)


jurgenwerk (approved):

packages/web-client/app/utils/image.ts
```diff
@@ -89,11 +91,59 @@ export class ImageValidation {
...
+        minWidth > 0 &&
+        minHeight > 0
+      ) {
+        message = `Please upload an image larger than ${minWidth}x${minHeight}, and smaller than ${maxWidth}x${maxHeight}`;
+      } else if (minWidth > 0 && minHeight > 0) {
+        message = `Please upload an image larger than ${minWidth}x${minHeight}`;
+      } else if (maxWidth !== Infinity && maxHeight !== Infinity) {
+        message = `Please upload an image smaller than ${maxWidth}x${maxHeight}`;
+      } else {
+        message = `Please upload an image larger than ${minWidth}x${minHeight}`;
```
backspace, line 137: maybe it doesn’t matter but isn’t it technically “at least” vs “larger than”? (same for file size)

I wonder whether the `boolean` nature of `imageSizeWithinBounds` could be decomposed so the calculation doesn’t need to happen twice-ish. But probs not important

backspace (approved):
thanks for breaking these up, made it way easier to review!

PR #2419, opened 2021-11-25
Uploader improvements

Aierie:
Adds error and loading states + introduces an avatar component that can be used separately from the image uploader, for use in memorialized cards.

Corresponds to the commits prefixed with UPLOADER: in #2416


backspace (approved):
A nice improvement for uploader consumers!

jurgenwerk (approved):
LGTM!

PR #2420, opened 2021-11-25
profane words check is no longer case sensitive

Aierie:
Corresponds to commit prefixed with HUB1: in #2416


jurgenwerk (commented):
Looks good!

backspace (approved):

PR #2424, opened 2021-11-29
card space profile name validation route

Aierie:
Adds a card space route for profile name validation, corresponds to commit prefixed `HUB2:` in #2416


jurgenwerk (approved):
Looks good!

PR #2426, opened 2021-11-29
Push notification registrations 

jurgenwerk:
Tickets: [CS-2613](https://linear.app/cardstack/issue/CS-2613/hub-api-endpoint-to-record-a-push-notification-registration-map-of-eoa), [CS-2614](https://linear.app/cardstack/issue/CS-2614/hub-api-endpoint-to-remove-a-push-notification-registration)

POST and DELETE api/push-notification-registrations endpoints were added so that the mobile app will be able to register and unregister push notifications coming from the hub.

Questions:
1. Should we also add a GET endpoint? I think it might be useful for the mobile app to first see if the registration already exists before attempting to register. 
2. Given `push-notification-registrations` routes don't actually accept an ID (because the registration is tied to `ctx.state.userAddress`, and there is only one registration per owner address), these route definitions don't quite conform to classic REST definitions. I'm thinking if it would be better to rename the routes from a plural to a singular resource naming (`POST push-notification-registration` and `DELETE push-notification-registration`)

### DIFF
--- a/packages/hub/config/structure.sql
+++ b/packages/hub/config/structure.sql
@@ -33,7 +33,7 @@ CREATE EXTENSION IF NOT EXISTS pgcrypto WITH SCHEMA public;
 
 
 --
--- Name: EXTENSION pgcrypto; Type: COMMENT; Schema: -; Owner: 
+-- Name: EXTENSION pgcrypto; Type: COMMENT; Schema: -; Owner:
 --
 
 COMMENT ON EXTENSION pgcrypto IS 'cryptographic functions';
@@ -834,6 +834,21 @@ CREATE TABLE public.prepaid_card_patterns (
 ALTER TABLE public.prepaid_card_patterns OWNER TO postgres;
 
 --
+-- Name: push_notification_registrations; Type: TABLE; Schema: public; Owner: postgres
+--
+
+CREATE TABLE public.push_notification_registrations (
+    id uuid NOT NULL,
+    owner_address text NOT NULL,
+    push_client_id text NOT NULL,
+    disabled_at timestamp without time zone,
+    created_at timestamp without time zone DEFAULT CURRENT_TIMESTAMP NOT NULL
+);
+
+
+ALTER TABLE public.push_notification_registrations OWNER TO postgres;
+
+--
 -- Name: reservations; Type: TABLE; Schema: public; Owner: postgres
 --
 
@@ -1039,6 +1054,14 @@ ALTER TABLE ONLY public.prepaid_card_patterns
 
 
 --
+-- Name: push_notification_registrations push_notification_registrations_pkey; Type: CONSTRAINT; Schema: public; Owner: postgres
+--
+
+ALTER TABLE ONLY public.push_notification_registrations
+    ADD CONSTRAINT push_notification_registrations_pkey PRIMARY KEY (id);
+
+
+--
 -- Name: reservations reservations_pkey; Type: CONSTRAINT; Schema: public; Owner: postgres
 --
 
@@ -1096,6 +1119,13 @@ CREATE INDEX discord_bots_bot_type_status_index ON public.discord_bots USING btr
 --
 
 CREATE UNIQUE INDEX merchant_infos_slug_unique_index ON public.merchant_infos USING btree (slug);
+
+
+--
+-- Name: push_notification_registrations_owner_address_push_client_id_un; Type: INDEX; Schema: public; Owner: postgres
+--
+
+CREATE UNIQUE INDEX push_notification_registrations_owner_address_push_client_id_un ON public.push_notification_registrations USING btree (owner_address, push_client_id);
 
 
 --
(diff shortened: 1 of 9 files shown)


packages/hub/config/structure.sql
```diff
@@ -1036,6 +1050,14 @@ ALTER TABLE ONLY public.prepaid_card_patterns
     ADD CONSTRAINT prepaid_card_patterns_pkey PRIMARY KEY (id);
 
 
+--
+-- Name: push_notification_registrations push_notification_registrations_pkey; Type: CONSTRAINT; Schema: public; Owner: postgres
+--
+
+ALTER TABLE ONLY public.push_notification_registrations
+    ADD CONSTRAINT push_notification_registrations_pkey PRIMARY KEY (owner_address);
```
lukemelia, line 1058: It is possible for an EOA to be connected on more than one device running Card Wallet, which will each register to receive push notifications (e.g. an iPad and an iPhone). Therefore owner address is not a suitable primary key. An alternative could be a compound primary key (owner address + push client id) or a surrogate key.

packages/hub/node-tests/routes/push-notification-registrations-test.ts
```diff
@@ -0,0 +1,134 @@
...
+    let dbManager = await getContainer().lookup('database-manager');
+    let db = await dbManager.getClient();
+
+    let queryResult = await db.query('SELECT * FROM push_notification_registrations WHERE owner_address = $1', [
+      stubUserAddress,
+    ]);
+
+    expect(queryResult.rows.length).to.equal(0);
+  });
+});
```
lukemelia, line 134: A spec to describe what happens if a registration already exists for this device/EOA pair should be added here. I would suggest this request should silently succeed and perhaps clear the disabled_at field if it has been set.

lukemelia (changes requested):
Two changes I think are important to make before merging. See my comments.

packages/hub/services/serializers/push-notification-serializer.ts
```diff
@@ -0,0 +1,29 @@
...
+      },
+    };
+
+    return result as JSONAPIDocument;
+  }
+}
+
+declare module '@cardstack/di' {
+  interface KnownServices {
+    'push-notification-registration-serializer': PushNotificationRegistrationSerializer;
```
backspace, line 27: Given that this has `registration` in it, should the filename have that as well?

backspace (commented):
Agreed with Luke re the plural-ness of this (which probably addresses your question about the RESTness of the endpoint names), but thanks for getting started on this, it helps me think about how the notifications worker might be structured

packages/hub/config/structure.sql
jurgenwerk: Ok, thanks for letting me know about this case! I made the adjustments and chose an uuid for the primary key.

packages/hub/node-tests/routes/push-notification-registrations-test.ts
jurgenwerk: Thanks for the suggestion. I added that case.

packages/hub/node-tests/routes/push-notification-registrations-test.ts
```diff
@@ -0,0 +1,134 @@
...
+          type: 'push-notification-registration',
+          attributes: {
+            'owner-address': stubUserAddress,
+            'push-client-id': 'FIREBASE_USER_ID',
+            'disabled-at': null,
+          },
+        },
+      })
+      .expect('Content-Type', 'application/vnd.api+json');
+  });
```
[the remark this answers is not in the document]
jurgenwerk: Added!

jurgenwerk:
@lukemelia @backspace this is now ready for another review. I changed the implementation so that there can be many (EOA, device) pairs.

packages/hub/node-tests/routes/push-notification-registrations-test.ts
```diff
@@ -0,0 +1,200 @@
...
+
+  it('does not fail when registration is already present + it reenables the existing one', async function () {
+    let pushNotificationRegistrationQueries = await getContainer().lookup('push-notification-registration-queries');
+
+    await pushNotificationRegistrationQueries.insert({
+      id: shortUuid.uuid(),
+      ownerAddress: stubUserAddress,
+      pushClientId: 'FIREBASE_USER_ID',
+      disabledAt: null,
```
backspace, line 105: Should be non-null if it’s meant to exercise re-enabling?

backspace (approved):
I left a minor suggestion but otherwise 👍🏻😀

packages/hub/node-tests/routes/push-notification-registrations-test.ts
jurgenwerk: Good catch! Fixed

packages/hub/node-tests/routes/push-notification-registrations-test.ts
```diff
@@ -0,0 +1,200 @@
...
+      .set('Accept', 'application/vnd.api+json')
+      .set('Content-Type', 'application/vnd.api+json')
+      .expect(200);
+
+    let records = await pushNotificationRegistrationQueries.query({
+      ownerAddress: stubUserAddress,
+      pushClientId: 'FIREBASE_USER_ID',
+    });
+
+    expect(records.length).to.equal(0);
```
lukemelia, line 198: I wonder if it would be better or worse to soft delete this with an deleted_at column? Not sure, but something to think about.

lukemelia (approved):

packages/hub/node-tests/routes/push-notification-registrations-test.ts
```diff
@@ -0,0 +1,200 @@
...
+    return stubAuthToken;
+  }
+  extractVerifiedTimestamp(_nonce: string) {
+    return stubTimestamp;
+  }
+
+  validateAuthToken(encryptedAuthToken: string) {
+    return handleValidateAuthToken(encryptedAuthToken);
+  }
+}
```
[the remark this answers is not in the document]
jurgenwerk: Agree! Will look into it.

packages/hub/node-tests/routes/push-notification-registrations-test.ts
jurgenwerk: That's a good idea. I think eventually we can introduce an ORM and define default scopes that will exclude the soft deleted ones.

PR #2429, opened 2021-11-29
Revert "Updated with new ankr RPC node"

backspace:
This reverts cardstack/cardstack#2427 for now until we can parameterise it.


habdelra:
@backspace @lukemelia the issue here is just the websocket URL. I think we should just rollback that. the HTTP URL's are actually working.

habdelra (approved):
looks good

PR #2431, opened 2021-11-29
Support to get versioned ABI from SDK

habdelra:
@backspace This PR will allow you to retrieve a versioned ABI from the SDK. To call it just ask for any of the contract names in the packages/cardpay-sdk/contracts/abi/<version> folder. for example:
```js
let abi = await getABI("price-oracle", web3); 
```
will return current price oracle contract ABI for the current protocol version


habdelra:
yeah, it looks like my dynamic  import usage needs some more compiler directives to work (and maybe embroider too?) I'll dig into it more.

habdelra:
@backspace do you know where else in the SDK we use `require`? I'd like for us to use `import` instead since we're using es6.

backspace:
I see now that I hadn’t restricted the search to `.ts` only so I was wrong to say `require` was being used!

image

PR #2434, opened 2021-11-29
hub: Add notifications

backspace:
This barely does anything yet, just committing as I’m reaching EOD so @Aierie has something to start on.

There’s a `web3-socket` service, should the existing service be `web3-http`? The `isAvailable` is patterned after the HTTP one but it’s always returning `true` even when I have deliberately broken it 🤔

This is based on #2431 for now as it will need to use that.


habdelra:
@backspace Yeah, let’s rename web3 service to web3-http

habdelra:
@backspace i would say the isAvailable is not actually super useful. I added because of the fickle nature of the RPC node, but actually when the RPC node starts returning 502 error, it doesn’t return the 502 for HTTP GET—rather it returns 502 for specific POST calls. So even the web3 one is not really doing the thing that I had hoped it would do. Ultimately I plan on removing it once we sort out the Ankr issues.

packages/hub/services/web3-socket.ts
```diff
@@ -0,0 +1,74 @@
...
+    }
+    return this.web3;
+  }
+
+  createWebsocketProvider(...args: ConstructorParameters<typeof Web3['providers']['WebsocketProvider']>) {
+    let [host, options] = args;
+    return new Web3.providers.WebsocketProvider(host, options);
+  }
+
+  async isAvailable(): Promise<boolean> {
```
habdelra, line 46: I was mentioning in linear that this is probably unecessary. For the original web3-http service this nerver really lived up to its promise since the 502 errors that the RPC node threw were not visible via HTTP GET--but rather specific HTTP POST for overloaded rpc calls. I think it's ok to remove this functionality.

packages/hub/node-tests/services/web3-socket-test.ts
```diff
@@ -0,0 +1,63 @@
...
+      throw new Error('mock web3 initialization failure');
+    };
+
+    expect(() => web3Socket.getInstance()).to.throw('mock web3 initialization failure');
+
+    await waitFor(async () => testkit.reports().length > 0);
+
+    expect(testkit.reports()[0].tags).to.deep.equal({
+      action: 'web3-socket-connection',
+    });
```
Aierie, line 61: Added tags, and testing for them because tags can be used to create alert rules:

https://docs.sentry.io/product/alerts/create-alerts/issue-alert-config/

We could also use the other metadata fields that Sentry's API provides, instead.

If/when socket disconnect handling is added, we could also add a similar Sentry event to that.

packages/hub/node-tests/services/web3-socket-test.ts
```diff
@@ -0,0 +1,63 @@
...
+    let web3Socket = new Web3SocketService();
+
+    web3Socket.initializeWeb3 = () => {
+      throw new Error('mock web3 initialization failure');
+    };
+
+    expect(() => web3Socket.getInstance()).to.throw('mock web3 initialization failure');
+
+    await waitFor(async () => testkit.reports().length > 0);
```
Aierie, line 57: `waitFor` is necessary because `Sentry.captureException` does some async stuff.

packages/hub/node-tests/services/web3-socket-test.ts
Aierie: https://linear.app/cardstack/issue/CS-2683/sentry-alert-for-web3-websocket-disconnectionfailure-to-initialize

packages/hub/main.ts
```diff
@@ -250,7 +260,7 @@ declare module '@cardstack/di' {
   }
 }
 
-function initSentry() {
+export async function initSentry() {
```
habdelra, line 263: is this really async? where's the `await`?

backspace: oops, Copilot mistake I think 🙃 removed now thanks!

backspace:
now realising maybe this needs a Waypoint change…?

packages/hub/tasks/notify-merchant-claim.ts
```diff
@@ -0,0 +1,42 @@
...
+    let { safe } = await Safes.viewSafe(merchantSafeAddress);
+
+    if (!safe) {
+      // Sufficient to log to Sentry because of on('job:error')…?
+      throw new Error('Safe not found');
+    }
+
+    // TODO: how should we resolve the token symbol from an address?
+    // let tokenAddress = event.returnValues.payableToken;
+    let token = 'DAI.CPXD';
```
habdelra, line 33: the subgraph has this information

habdelra: Here is an example illustrating how you might call this using a txn hash (which should be available from the event)
Screen Shot 2021-11-30 at 4 17 33 PM

habdelra: The SDK has an API for issuing direct graphQL queries

packages/hub/tasks/notify-customer-payment.ts
```diff
@@ -0,0 +1,34 @@
...
+}
+
+export default class NotifyCustomerPayment {
+  cardpay: CardpaySDKService = inject('cardpay');
+  web3: Web3HttpService = inject('web3-http', { as: 'web3' });
+  workerClient: WorkerClient = inject('worker-client', { as: 'workerClient' });
+
+  async perform(payload: CustomerPaymentSubscriptionEvent) {
+    let Safes = await this.cardpay.getSDK('Safes', this.web3.getInstance());
+    let { safe } = await Safes.viewSafe(payload.returnValues.merchantSafe);
```
habdelra, line 20: I think it's a lot easier to just query the subgraph for the payment event, that will include both the merchant, customer and amount paid in both the issuing token (useful for the merchant) and in SPEND.

packages/hub/tasks/notify-customer-payment.ts
```diff
@@ -0,0 +1,34 @@
...
+  };
+}
+
+export default class NotifyCustomerPayment {
+  cardpay: CardpaySDKService = inject('cardpay');
+  web3: Web3HttpService = inject('web3-http', { as: 'web3' });
+  workerClient: WorkerClient = inject('worker-client', { as: 'workerClient' });
+
+  async perform(payload: CustomerPaymentSubscriptionEvent) {
+    let Safes = await this.cardpay.getSDK('Safes', this.web3.getInstance());
```
habdelra, line 19: you'll need to wait for the subgraph to index the event. use `waitForSubgraphIndex` with the txn hash from the event

packages/hub/tasks/notify-merchant-claim.ts
```diff
@@ -0,0 +1,42 @@
...
+    amount: string;
+  };
+}
+
+export default class NotifyMerchantClaim {
+  cardpay: CardpaySDKService = inject('cardpay');
+  web3: Web3HttpService = inject('web3-http', { as: 'web3' });
+  workerClient: WorkerClient = inject('worker-client', { as: 'workerClient' });
+
+  async perform(payload: MerchantClaimSubscriptionEvent) {
```
habdelra, line 18: you'll need to wait for the subgraph to index the event. use `waitForSubgraphIndex` with the txn hash from the event

packages/hub/tasks/notify-merchant-claim.ts
```diff
@@ -0,0 +1,42 @@
...
+    let merchantSafeAddress = payload.returnValues.merchantSafe;
+    let amountInWei = payload.returnValues.amount;
+
+    let web3Instance = this.web3.getInstance();
+    let Safes = await this.cardpay.getSDK('Safes', web3Instance);
+    let { safe } = await Safes.viewSafe(merchantSafeAddress);
```
habdelra, line 24: let's not do this--its a very indirect way to get the information that you actually need

packages/hub/tasks/notify-customer-payment.ts
habdelra: Here is an example of a query that gets teh EOA for the prepaid card, the EOA of the merchant, issuing token amount, issuing token name, and SPEND amount: Screen Shot 2021-12-01 at 6 22 37 PM

habdelra (changes requested):
I think in general using the Safe.viewSafe SDK as a way to get the information that you need to send the notification is not a very good approach. Instead, use the subgraph to precisely ask for the thing that you need. each event the protocol fires has a subgraph entity for that specific event. So if you are interested in merchant claims and prepaid card payments, then query for those specific events in the subgraph--they will have all the info that you need for the notifications

habdelra:
@backspace I think there is already a waypoint app for workers--this should will live in that app. There should not need to be a new waypoint app for this, these are just worker tasks.

habdelra (approved):

PR #2435, opened 2021-11-30
add options for websocket autoreconnecting as used in cardwallet

Aierie:
see https://github.com/cardstack/cardwallet/blob/9c8dcb6c777f15de63938ce6a1cfc7aa6915bf04/cardstack/src/models/web3-provider.ts

I missed `reconnect.auto` when working on this the first time and tried this out after looking at card wallet's approach to use it in the notifications work. It works to reconnect for the main case of websocket disconnection we want to prevent, which is a disconnection after idling for a few minutes.


backspace (approved):
I wasn’t sure how to exercise this but it’s always nice to be able to use built-in facilities!

PR #2437, opened 2021-11-30
CS-2664 add prepaid card issuance to merchant revenue events subgraph

habdelra:
Hold off on merging this until we have deployed v0.8.6 of the protocol into xDai. There are some changes made to the subgraph that rely on v0.8.6, and we need a clear runway to xDai before we deploy this update.

Ran this in sokol via my own graph node and things index fine:
Screen Shot 2021-11-30 at 4 18 03 PM


habdelra:
- [ ] Need to add block number to the new subgraph entity

PR #2438, opened 2021-11-30
move realm config into container [CS-2659]

mattmcmanus:
- Move RealmsConfig into DI
- Tests now use DI system for setting up test realm
- Split up setupHub into several test helpers


mattmcmanus:
@ef4 in the last commit I split up the `setupHub` method. I'm not sure how in line with your vision of it being "composable" like you had mentioned, but it's my attempt. 

It's my understanding that you can't register things after the container is setup, so I had to create two helpers, to be used as follows:
```js
    let { realmURL, getRealmDir } = configureCompiler(this);
    let { getContainer } = setupHub(this);
    let { cards } = cardHelpers(this);
```

I'm open to suggestions. We could also pull the last commit into another branch and just get the first two merged.

packages/hub/services/realms-config.ts
```diff
@@ -0,0 +1,25 @@
+import { RealmConfig } from '@cardstack/core/src/interfaces';
+import { join } from 'path';
+
+export const BASE_REALM_CONFIG = {
+  url: 'https://cardstack.com/base/',
+  directory: join(__dirname, '..', '..', 'base-cards'),
```
habdelra, line 6: it might be jumping the gun a bit, but I wonder if we'll need to use the `require.resolve` approach here too once we merge the webpack build

habdelra (approved):
👍

packages/hub/services/realms-config.ts
mattmcmanus: That's a good question. I'll need to defer to @ef4 on that one, because I'm not sure.

PR #2390, opened 2021-11-17
upgrading deps in boxel

ef4:
Running ember-cli-update to update the addon blueprint & dependencies to ember 3.28.4.

PR #2440, opened 2021-11-30
Update ember-auto-import

mattmcmanus:
The ember upgrade PR for boxel is stalled out (https://github.com/cardstack/cardstack/pull/2390). This is an attempt to bring over some smaller commits and get this merged first. It involves updating ember-auto-import as well as handling several deprecations that will effect the ember 4 upgrade

### DIFF
--- a/packages/boxel/tests/integration/helpers/dayjs-format-test.js
+++ b/packages/boxel/tests/integration/helpers/dayjs-format-test.js
@@ -12,15 +12,15 @@ module('Integration | Helper | dayjs-format', function (hooks) {
   setupRenderingTest(hooks);
 
   test('one arg (date)', async function (assert) {
-    this.set('date', date(date(0)));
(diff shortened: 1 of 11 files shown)


packages/boxel/tests/integration/helpers/dayjs-format-test.js
```diff
@@ -12,15 +12,15 @@ module('Integration | Helper | dayjs-format', function (hooks) {
   setupRenderingTest(hooks);
 
   test('one arg (date)', async function (assert) {
-    this.set('date', date(date(0)));
+    this.set('date', '2020-01-02');
```
burieberry, line 15: Ha thanks for fixing this

burieberry (approved):
Thanks for porting over css-url, I had the same idea when I saw how outdated it was. That package should probably be redone in newer ember, I don't know how widely it's used though.

PR #2441, opened 2021-12-01
Transfer reward safe

tintinthong:
- I made this branch mainly to TEST the transfer function E2E. I suggest to put this in first as this in the sdk first anyway
  - [x] after transfer, reward safe can be claim rewards & withdraw out of it
- There has been a bit of discussion on where reward safe will get funds to pay for gas https://discord.com/channels/584043165066199050/830994718694441031/915197835304116225. Current implementation suggested by me is to use reward tokens claimed (this is what this pr does and deafults the gas payment token to cardCpxd). Other standing suggestion is to use prepaid card to pay for the gas but the protocol currently doesn't support it. As discussed we have decided to KIV the gas payment.


packages/cardpay-sdk/sdk/reward-manager/base.ts
```diff
@@ -520,6 +520,129 @@ The owner of reward safe ${safeAddress} is ${rewardSafeOwner}, but the signer is
...
+
+    // Temporarily, transfer function will use card as a the default gas token
+    // This means that the safe has to have card tokens within it before it can transfer
+    // This will change after we finalise the gas policy
+    let gasTokenAddress = await getAddress('cardCpxd', this.layer2Web3);
+    let gasToken = new this.layer2Web3.eth.Contract(ERC20ABI as AbiItem[], gasTokenAddress);
+
+    let safeBalance = new BN(await gasToken.methods.balanceOf(safeAddress).call());
+
+    let rewardSafeDelegateAddress = await getAddress('rewardSafeDelegate', this.layer2Web3);
```
alex-cardstack, line 562: same lookup issue here

alex-cardstack (approved):
Looks good to me except same address lookup issue

PR #2442, opened 2021-12-01
🔔 ⚙️ Notification preferences

jurgenwerk:
Ticket: [CS-2615](https://linear.app/cardstack/issue/CS-2615/hub-endpoint-to-update-push-notification-preferences-enabledisable)

This PR introduces the ability to fetch and modify push notification preferences for users of Card Wallet app. I believe there will be toggles where users will be able to enable or disable types of notificiations. 

There are two new tables, `notification_types` (need to run seeding), and `notification_preferences`:

```
                  id                  |    notification_type     | default_status |         created_at
--------------------------------------+--------------------------+----------------+----------------------------
 10b75b75-b855-42eb-893e-d223995b8872 | merchant_revenue_claimed | enabled        | 2021-11-30 14:59:09.179027
 1137c2b1-fb2e-45d2-9f62-d365b989d151 | merchant_payment         | enabled        | 2021-11-30 14:59:09.179027
 ```
 
 ```
  owner_address |         notification_type_id         |  status  |         created_at
---------------+--------------------------------------+----------+----------------------------
 0x0            | 10b75b75-b855-42eb-893e-d223995b8872 | disabled | 2021-12-02 12:57:18.766785
 ```
 
The way this works is that when there are no user preferences saved, the endpoint will return the default statuses from the 1st table. But when there is an override (a custom setting) defined in the 2nd table, the endpoint will respond with that (and not with the default value). The point of this is that we don't need to save preferences for each and every user separately. 

The endpoint is `/api/push-notification-registrations`. GET will return all notification preferences for the current wallet address, for example:

```
{
data: [
  {
    type: 'notification-preference',
    attributes: {
      'notification-type': 'merchant_revenue_claimed',
      'owner-address': '0x2f58630CA445Ab1a6DE2Bb9892AA2e1d60876C13',
      status: 'enabled',
    },
  },
  {
    type: 'notification-preference',
    attributes: {
      'notification-type': 'merchant_payment',
      'owner-address': '0x2f58630CA445Ab1a6DE2Bb9892AA2e1d60876C13',
      status: 'enabled',
    },
  },
],
}
```

To toggle a notification type, the Card Wallet app will issue a POST request with the following content where it specifies whether it should be enabled or disabled:

```
data: {
  type: 'notification-preference',
  attributes: {
    'notification-type': 'merchant_revenue_claimed',
    status: 'disabled',
  },
},
```

### CODE REVIEW

**Review by backspace — APPROVED** [review]
I like the design with the default values and handling of updates 🤩

I’m not sure whether this has been discussed elsewhere but having `owner_address` as the key for preferences makes me think that preferences will be controlled on a per-account basis vs per-app or even per-device. My instinct is that we want per-device: maybe I want to mute things on an iPad but not iPhone, and I don’t want to maintain preferences on a per-account basis within the app.

I also imagine that the web application might want to receive notifications in the future, I don’t know how that plays into this though.

File: packages/hub/db/seeds.ts [hunk 1]
```diff
@@ -117,3 +118,20 @@ async function loadPrepaidCardPatterns(db: Client) {
...
+  let query = `INSERT INTO notification_types(
+    id, notification_type, default_status
+  )
+  VALUES %L ON CONFLICT (id) DO UPDATE SET
+    notification_type = excluded.notification_type,
+    default_status = excluded.default_status;`;
+  let rows = [
+    ['10b75b75-b855-42eb-893e-d223995b8872', 'merchant_revenue_claimed', 'enabled'],
+    ['1137c2b1-fb2e-45d2-9f62-d365b989d151', 'merchant_payment', 'enabled'],
+  ];
```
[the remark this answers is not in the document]
jurgenwerk: Sure, can easily change the names so they will match.

jurgenwerk:
@backspace thanks for your review - now that I think about this I find your concern very valid. Having 2 devices ringing at the same time for the same thing would really be annoying 🙈. I'll go ahead and make the preference dependent on the device as well. Regarding web notifications, I believe FCM can run in a web worker and this setup would work for that case too.

packages/hub/config/structure.sql
```diff
@@ -754,6 +778,35 @@ CREATE TABLE public.merchant_infos (
...
+-- Name: notification_preferences; Type: TABLE; Schema: public; Owner: postgres
+--
+
+CREATE TABLE public.notification_preferences (
+    owner_address text NOT NULL,
+    notification_type_id uuid NOT NULL,
+    push_client_id text NOT NULL,
+    status public.notification_preferences_status_enum DEFAULT 'enabled'::public.notification_preferences_status_enum NOT NULL,
+    created_at timestamp without time zone DEFAULT CURRENT_TIMESTAMP NOT NULL
+);
```
lukemelia, line 791: Consider adding a unique index on the combination of owner_address, noification_type_id, and push_client_id

jurgenwerk: Yeah, it's already added: https://github.com/cardstack/cardstack/pull/2442/commits/8448d98ff942fdd77c156a297f7a8db3083e2cfa#diff-c33f7dc267c12ec716a2fc35c636dd8bcde8de8efecf7a54d0c171075b9e7fbdR16

backspace (approved):
I appreciate some of the extractions you’ve added 🤩

I still wonder whether it makes sense to manage notifications on an EOA basis vs client but we can see how it works in practice and change the approach if needed!

packages/hub/routes/notification-preferences.ts
```diff
@@ -0,0 +1,141 @@
...
+    let preferences: NotificationPreference[] = defaultPreferences.map((defaultPreference) => {
+      let preference = notificationPreferences.find(
+        (preference) => preference.notificationType === defaultPreference.notificationType
+      );
+      if (preference) {
+        return preference;
+      } else {
+        return defaultPreference;
+      }
+    });
```
lukemelia, line 68: This merging/fallback logic should probably be pushed into a model layer.

packages/hub/routes/notification-preferences.ts
```diff
@@ -0,0 +1,141 @@
...
+        status,
+      });
+    } else {
+      await this.notificationPreferenceQueries.insert({
+        ownerAddress: ctx.state.userAddress,
+        pushClientId,
+        notificationType,
+        status,
+      });
+    }
```
lukemelia, line 122: Consider making this an upsert instead. That will avoid errors from race conditions too.

packages/hub/services/api-router.ts
```diff
@@ -77,6 +81,8 @@ export default class APIRouter {
       parseBody,
       pushNotificationRegistrationsRoute.delete
     );
+    apiSubrouter.get('/notification-preferences/:push_client_id', parseBody, notificationPreferencesRoute.get);
+    apiSubrouter.post('/notification-preferences', parseBody, notificationPreferencesRoute.post);
```
lukemelia, line 85: The asymmetry here is weird. Perhaps we should have a PUT to /notification-preferences/:push_client_id instead of a POST?

jurgenwerk: Yes, I think that would be better. I added a fix.

packages/hub/routes/notification-preferences.ts
jurgenwerk: Good suggestion! Added.

packages/hub/routes/notification-preferences.ts
jurgenwerk: Agree. I'm planning on doing that in a separate PR.

PR #2445, opened 2021-12-01
updated SDK for 0.8.7 of protocol

habdelra:
note that I updated the subgraph for RewardPool so that it understands the new RewardeeClaim event. @tintinthong you'll still want to include the new `leaf` param in the subgraph though


tintinthong:
Cool got it. Thanks

PR #2447, opened 2021-12-01
hub: Add merchant name to notifications

backspace:
This follows #2434 to parse/fetch DIDs and insert the merchant name in the notification:

image

It ignores failures that might arise when parsing the DID and parsing and fetching its associated JSON and just produces a generic “your business” notification.

I extracted the `JSONAPIDocument` type that was scattered through various files.

### DIFF
--- a/packages/hub/node-tests/services/merchant-info-test.ts
+++ b/packages/hub/node-tests/services/merchant-info-test.ts
@@ -57,4 +57,22 @@ describe('MerchantInfoService', function () {
 
     expect(result).to.be.null;
   });
+
+  it('returns null when there is no DID', async function () {
+    registry(this).register('merchant-info', MerchantInfoService);
+
+    let subject = await getContainer().lookup('merchant-info');
+    let result = await subject.getMerchantInfo(undefined);
+
+    expect(result).to.be.null;
+  });
+
+  it('returns null the DID is invalid', async function () {
+    registry(this).register('merchant-info', MerchantInfoService);
+
+    let subject = await getContainer().lookup('merchant-info');
+    let result = await subject.getMerchantInfo('did:cardstack:hey');
+
+    expect(result).to.be.null;
+  });
 });
--- a/packages/hub/node-tests/services/merchant-info-test.ts
+++ b/packages/hub/node-tests/services/merchant-info-test.ts
@@ -42,4 +42,19 @@ describe('MerchantInfoService', function () {
       ownerAddress: '0x323B2318F35c6b31113342830204335Dac715AA8',
     });
   });
+
+  it('returns null when the fetch fails', async function () {
+    class PatchedMerchantInfoService extends MerchantInfoService {
+      async fetchMerchantInfo(): Promise<JSONAPIDocument> {
+        throw new Error('Simulated merchant info fetch failure');
+      }
+    }
+
+    registry(this).register('merchant-info', PatchedMerchantInfoService);
+
+    let subject = await getContainer().lookup('merchant-info');
+    let result = await subject.getMerchantInfo('did:cardstack:1m1C1LK4xoVSyybjNRcLB4APbc07954765987f62');
+
+    expect(result).to.be.null;
+  });
 });
--- a/packages/hub/node-tests/services/merchant-info-test.ts
+++ b/packages/hub/node-tests/services/merchant-info-test.ts
@@ -1,0 +1,49 @@
+import MerchantInfoService from '../../services/merchant-info';
+import { registry, setupHub } from '../helpers/server';
+
+interface JSONAPIDocument {
+  data: any;
+  included?: any[];
+}
+
+describe('MerchantInfoService', function () {
+  let { getContainer } = setupHub(this);
+
+  it('returns a MerchantInfo object extracted from a fetched document', async function () {
+    let mockMerchantInfoDocument = {
+      meta: { network: 'sokol' },
+      data: {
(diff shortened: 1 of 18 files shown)


lukemelia (approved):
I wonder if we will need/want to cache this in memory.

backspace:
yes, we might want to add caching here as we did with “off chain JSON” in web client

packages/hub/tasks/notify-customer-payment.ts
```diff
@@ -65,9 +69,23 @@ export default class NotifyCustomerPayment {
...
+    try {
+      if (result.merchantSafe?.infoDid) {
+        let merchantInfo = await this.merchantInfo.getMerchantInfo(result.merchantSafe.infoDid);
+
+        if (merchantInfo) {
+          merchantName = ` ${merchantInfo.name}`;
+        }
+      }
+    } catch (e) {
+      // Silently ignore and just show notification without merchant name
```
Aierie, line 83: Is it worth logging this to Sentry?

packages/hub/tasks/notify-merchant-claim.ts
```diff
@@ -53,10 +57,27 @@ export default class NotifyMerchantClaim {
...
+    }
+
     let token = result.token.symbol;
     let notifiedAddress = result.merchantSafe.merchant.id;
     let amountInWei = result.amount;
-    let message = `You just claimed ${Web3.utils.fromWei(amountInWei)} ${token} from your business account`;
+
+    let message = `You just claimed ${Web3.utils.fromWei(
+      amountInWei
+    )} ${token} from your${merchantName} business account`;
```
Aierie, line 80: I think this reads better than `Revenue deposited into account for [Merchant name].` in CS-2617. Will ask in the issue.

packages/hub/tasks/notify-customer-payment.ts
```diff
@@ -65,9 +69,23 @@ export default class NotifyCustomerPayment {
...
+    }
+
     let notifiedAddress = result.merchant.id;
     let spendAmount = result.spendAmount;
-    let message = `Your business received a payment of §${spendAmount}`;
+    let message = `Your business${merchantName} received a payment of §${spendAmount}`;
```
Aierie, line 88: Will also ask about this v `[Merchant name] received a new payment.` in CS-2616.

packages/hub/node-tests/services/merchant-info-test.ts
```diff
@@ -0,0 +1,78 @@
...
+
+  it('returns a MerchantInfo object extracted from a fetched document', async function () {
+    let mockMerchantInfoDocument = {
+      meta: { network: 'sokol' },
+      data: {
+        id: '05074049-be24-45fe-81ba-8e3c1a623aa7',
+        type: 'merchant-infos',
+        attributes: {
+          did: 'did:cardstack:1m1C1LK4xoVSyybjNRcLB4APbc07954765987f62',
+          name: 'a long long long long multi-word name that will cause problems because it is so long it really maybe shouldn’t even be permitted amirite maybe there should be (more of?) a limit',
```
Aierie, line 16: yes

packages/hub/tasks/notify-customer-payment.ts
```diff
@@ -65,9 +69,23 @@ export default class NotifyCustomerPayment {
       );
     }
 
+    let merchantName = '';
+
+    try {
+      if (result.merchantSafe?.infoDid) {
+        let merchantInfo = await this.merchantInfo.getMerchantInfo(result.merchantSafe.infoDid);
+
+        if (merchantInfo) {
```
Aierie, line 78: Maybe worth changing to `merchantInfo?.name`?

packages/hub/tasks/notify-merchant-claim.ts
```diff
@@ -53,10 +57,27 @@ export default class NotifyMerchantClaim {
       throw new Error(`Subgraph did not return information for merchant claim with transaction hash: "${payload}"`);
     }
 
+    let merchantName = '';
+
+    try {
+      if (result.merchantSafe.infoDid) {
+        let merchantInfo = await this.merchantInfo.getMerchantInfo(result.merchantSafe.infoDid);
+
+        if (merchantInfo) {
```
Aierie, line 66: Maybe worth changing to `merchantInfo?.name`?

Aierie (approved):
I think this looks good and it works for me locally.

I wonder if we should be introducing typescript-friendly utilities (types + (de)serialization) for JSON-API?

packages/hub/tasks/notify-customer-payment.ts
backspace: Maybe? I’m wary of more noisy logging, but perhaps an error here is worth knowing about. I’ll do the same in the other task, thanks.

backspace: done in c241690 🤓

packages/hub/tasks/notify-customer-payment.ts
backspace: Agreed, done in c91e5682a, thanks

packages/hub/tasks/notify-customer-payment.ts
backspace: Thanks for checking in, I was not thinking too much about wording pending concrete answers, but I’ve updated in c281ddd80 with the decision there.

backspace:
re Typescript and JSON:API, maybe, probably? haha

PR #2448, opened 2021-12-01
Publish hub via webpack

ef4:


mattmcmanus:
Ok. Writing out my current issue in the hopes that doing so will help me figure it out or give you plenty of context to continue on with. 

The webpack build is messing around with __dirname, so in a place where it should be the full path, it's literally just the directories name. Here is an example: 

https://github.com/cardstack/cardstack/blob/5c44abf05e9556fcb88cba7040472645bae2bc7a/packages/hub/services/card-cache-config.ts#L6-L8

`__dirname` there is coming back as simply "services". This is causing problems downstream when we make sure the card-cache's package.json is properly setup: 

https://github.com/cardstack/cardstack/blob/5c44abf05e9556fcb88cba7040472645bae2bc7a/packages/hub/services/card-cache.ts#L47-L62

Which is overwriting the existing package.json and causing our deps to be erased, causing the error in the cardhost tests you see.

habdelra:
@mattmcmanus I think this article about how webpack handles __dirname will help you: https://codeburst.io/use-webpack-with-dirname-correctly-4cad3b265a92

ef4:
No, I turned on `__dirname` support on purpose.

We need to refactor the card cache to not use this pattern -- it doesn't make sense to look for a location "relative to my source code" once your source code goes through a build process. There's nothing important about locating the cache relative to `__dirname`, it just happened to be convenient at the time.

habdelra:
@ef4 cool, can we just use the OS's tmp dir then?

mattmcmanus:
@habdelra @ef4 when running the development server we need the cache to go to `packages/compiled`. What is the better way to do that?

ef4:
I think we can resolve it as a dependency, the same way cardhost does. So something like 

```js
dirname(require.resolve('@cardstack/compiled/package.json'))
```

Except also you will need to spell that as:

```js
dirname(__non_webpack_require__.resolve('@cardstack/compiled/package.json'))
```

PR #2450, opened 2021-12-02
Recover reward tokens

tintinthong:
- allows rewardProgramAdmin to recover tokens from reward pool (into their single-owner safe).


tintinthong:
This may need a change in the relayer to handle the case for an empty depot

habdelra (approved):
👍

habdelra:
@tintinthong yeah, i could definitely see having to patch the relayer for the empty safe scenario on this one

PR #2451, opened 2021-12-02
Fix merchant claim guard (and perhaps also usage of dataGas)

tintinthong:
- I don't actually think this was the intent of the guard in `merchantClaim`. Currently, the guard is checking `unclaimedRevenue > claimAmount+gasCost`. But, when we transact, the `gasCost` will already be covered by the `claimAmount`. So we need only check that the `claimAmount > gasCost`( the check of `claimAmount < unclaimedRevenue` occurs on an earlier check).  
- `gasCost = (dataGas + baseGas)*gasPrice` was used -- not sure why.  I think by default we only need to calculate gasCost by `gasCost =(safeTxGas + baseGas)*gasPrice` so I made the change.
- Also, there are notes all over the relayer say `data_gas` is [deprecated](https://github.com/gnosis/safe-relay-service/blob/0829c7fec82f4db311bb2077fdd68b5a47687e10/safe_relay_service/relay/views.py#L325) and we should just use `base_gas`.  I messaged the gnosis team to double check. **If its ok,** I would like to remove places where `dataGas` are used. From my experience the relayer always returns the same value from `baseGas` and `dataGas` during estimation.These are the places which they appear. 
  - [here](https://github.com/cardstack/cardstack/blob/1c85c219a097aaadc3c287e60b2baf3336c95b65/packages/cardpay-sdk/sdk/token-bridge-home-side.ts#L106) 
  - [here](https://github.com/cardstack/cardstack/blob/1c85c219a097aaadc3c287e60b2baf3336c95b65/packages/cardpay-sdk/sdk/safes.ts#L343).
  -  the prepaid card base 
[here](https://github.com/cardstack/cardstack/blob/1c85c219a097aaadc3c287e60b2baf3336c95b65/packages/cardpay-sdk/sdk/prepaid-card/base.ts#L461)

### Testing
- [x] test safe transfer
- [x] test create prepaid card 
- [x] test relaying with bridge-to-l2 
- [x] test merchant claim


habdelra (approved):
Looks good.

For the dataGas, sure we can switch over to using baseGas

PR #2454, opened 2021-12-02
CS-2625 Add block numbers to subgraph

IanCal:
There are a lot of entities with a timestamp & a linked transaction, this adds block numbers alongside them consistently. This could (as with the timestamp) be determined by joining to the transaction table but for extracting & filtering it helps to have this in the tables themselves.

This could be limited to a smaller set of tables for the current use case but I thought it was better to be consistent across the board.

### DIFF
--- a/packages/cardpay-subgraph/schema.graphql
+++ b/packages/cardpay-subgraph/schema.graphql
@@ -63,6 +63,7 @@ type BridgeToLayer1Event @entity {
   transaction: Transaction!
   safe: Safe
   timestamp: BigInt!
+  blockNumber: BigInt!
   account: Account!
   token: Token!
   amount: BigInt!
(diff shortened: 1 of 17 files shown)


habdelra:
One of the things that we do for subgraph updates is to run a full index first using them. The penalties for an indexing error are severe--all subgraph indexing halts. Right now Justin and I both have our own hosted subgraph nodes (it's painful to run locally). Perhaps I can create a hosted one for you and you can perform a full index an verify that you are able to index to the current block head?

IanCal:
Sure, I've indexed it fully locally, but a remote one would be useful for a couple of reasons - testing it more in the environment it'll have to actually run in but also to do some of the other work extracting data out without breaking things for other people.

Not urgent though, with a local setup I'm not blocked.

habdelra:
cool--that works for me--I'll merge. I have a few other subgraph PR's as well that need to go in. once all those other ones are merged I'll deploy into sokol and xdai graph nodes.

PR #2460, opened 2021-12-02
yarn.lock was updated

habdelra:


backspace:
These changes keep coming up, it’s puzzling; what Yarn version are you on?

habdelra:
```sh
$ yarn --version                     
1.22.5
```

backspace:
hmm, maybe that’s why; we use Volta, which in the root `package.json` is locked to 1.22.10!

habdelra:
let me try that

habdelra:
@backspace so using the volta managed yarn I still end up with teh same lock file as is in this PR:
```sh
cardstack on yarn-lock-update-12-2-21 took 39s 
$ which yarn    
/Users/hassan/.volta/bin/yarn

cardstack on yarn-lock-update-12-2-21 
$ volta install yarn
success: installed and set yarn@1.22.15 as default
   note: you are using yarn@1.22.10 in the current project
```

backspace:
hmm, then the mystery continues, we keep flipping back between these two changes 😞

habdelra:
@backspace when you run `yarn install` in this branch, do you end up with the same lock file?

backspace:
yes. Hopefully this is the last time this needs to happen then 🤞🏻

PR #2463, opened 2021-12-03
Add leaf to subgraph

tintinthong:


tintinthong:
@habdelra this pr is good to go. Tested on my graph node and I have also pushed to the green node and its ready to be deployed. Just need to do a release and then rotate the nodes.

PR #2469, opened 2021-12-06
🔔 Add Firebase push notifications service

jurgenwerk:
Ticket: [CS-2701](https://linear.app/cardstack/issue/CS-2701/add-firebase-cloud-messaging-service)

This PR adds the Firebase client which the notification worker will use to send push notifications. 

We'll have to align with the mobile team to define the app link so that the user will get to the relevant screen when tapping the notification.

There are couple of new ENV vars: FIREBASE_PROJECT_ID, FIREBASE_CLIENT_EMAIL, FIREBASE_PRIVATE_KEY, FIREBASE_DATABASE_URL which should be added in .env for development. These can be obtained by going to [https://console.firebase.google.com](https://console.firebase.google.com), choosing Project Settings > Service accounts > Generate new private key.

### DIFF
--- a/packages/hub/services/push-notifications/firebase.ts
+++ b/packages/hub/services/push-notifications/firebase.ts
@@ -1,0 +1,24 @@
+import admin from 'firebase-admin';
+
+interface FCMPayload {
+  message: {
+    notification: {
+      body: string;
+      title: string;
+      data: any; // TODO: Define deep link
+    };
+  };
+  token: string;
+}
+
+export default class FirebasePushNotifications {
+  async send(payload: FCMPayload) {
+    await admin.messaging().send(payload);
+  }
+}
+
+declare module '@cardstack/di' {
+  interface KnownServices {
+    'firebase-push-notifications': FirebasePushNotifications;
+  }
+}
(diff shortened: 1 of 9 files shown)


packages/hub/services/push-notifications/firebase.ts
```diff
@@ -0,0 +1,24 @@
...
+
+interface FCMPayload {
+  message: {
+    notification: {
+      body: string;
+      title: string;
+      data: any; // TODO: Define deep link
+    };
+  };
+  token: string;
```
jurgenwerk, line 11: This is the push client id

packages/hub/services/push-notifications/firebase.ts
```diff
@@ -0,0 +1,24 @@
...
+}
+
+export default class FirebasePushNotifications {
+  async send(payload: FCMPayload) {
+    await admin.messaging().send(payload);
```
jurgenwerk, line 16: Error handling is missing here. But I think we can add this later when we start testing and see the possible edge cases

backspace (approved):
agreed re waiting to see how things actually unfold when it’s live 👍🏻😀

PR #2470, opened 2021-12-06
Nonusd to spend

Aierie:
Add utility function `convertToSpend` that converts amounts in native currencies to SPEND. Aims to solve CS-2641.


topwebtek7 (approved):
Looks good to me! 🚀

PR #2472, opened 2021-12-06
hub: Fix autotest script

backspace:
Since the built artefacts are no longer alongside the TS files, this runs the built tests from `dist/` instead.


habdelra:
Thanks!

PR #2474, opened 2021-12-07
Sent push notifications

Aierie:
add tracking of sent push notifications in order to deduplicate.

### DIFF
--- a/packages/hub/db/migrations/20211207151150639_sent-push-notifications.js
+++ b/packages/hub/db/migrations/20211207151150639_sent-push-notifications.js
@@ -11,7 +11,6 @@ exports.up = (pgm) => {
     notification_data: { type: 'json' },
     // This is firebase's id for the sent message
     message_id: { type: 'string' },
-    network: { type: 'string' },
     created_at: { type: 'timestamp', notNull: true, default: pgm.func('current_timestamp') },
   });
 };
--- a/packages/hub/db/migrations/20211207151150639_sent-push-notifications.js
+++ b/packages/hub/db/migrations/20211207151150639_sent-push-notifications.js
@@ -1,8 +1,6 @@
 const SENT_PUSH_NOTIFICATIONS_TABLE = 'sent_push_notifications';
 
 exports.up = (pgm) => {
-  // TODO: may be useful to add a message id field for debug purposes
-  // The message id is returned by firebase messaging
   pgm.createTable(SENT_PUSH_NOTIFICATIONS_TABLE, {
     transaction_hash: { type: 'string', notNull: true },
     owner_address: { type: 'string', notNull: true },
@@ -11,6 +9,8 @@ exports.up = (pgm) => {
     notification_title: { type: 'string' },
     notification_body: { type: 'string', notNull: true },
     notification_data: { type: 'json' },
+    message_id: { type: 'string', notNull: true },
+    network: { type: 'string' },
   });
 
   pgm.createIndex(SENT_PUSH_NOTIFICATIONS_TABLE, ['transaction_hash', 'owner_address', 'push_client_id'], {
--- a/packages/hub/db/migrations/20211207151150639_sent-push-notifications.js
+++ b/packages/hub/db/migrations/20211207151150639_sent-push-notifications.js
@@ -4,7 +4,8 @@ exports.up = (pgm) => {
   pgm.createTable(SENT_PUSH_NOTIFICATIONS_TABLE, {
     // This is our own id for the notification.
     notification_id: { type: 'string', primaryKey: true },
-    notification_type: { type: 'string', notNull: true },
+    push_client_id: { type: 'string' },
+    notification_type: { type: 'string' },
     notification_title: { type: 'string' },
     notification_body: { type: 'string', notNull: true },
     notification_data: { type: 'json' },
--- a/packages/hub/db/migrations/20211207151150639_sent-push-notifications.js
+++ b/packages/hub/db/migrations/20211207151150639_sent-push-notifications.js
@@ -11,6 +11,7 @@ exports.up = (pgm) => {
     notification_data: { type: 'json' },
     message_id: { type: 'string', notNull: true },
     network: { type: 'string' },
+    created_at: { type: 'timestamp', notNull: true, default: pgm.func('current_timestamp') },
   });
 
   pgm.createIndex(SENT_PUSH_NOTIFICATIONS_TABLE, ['transaction_hash', 'owner_address', 'push_client_id'], {
--- a/packages/hub/db/migrations/20211207151150639_sent-push-notifications.js
+++ b/packages/hub/db/migrations/20211207151150639_sent-push-notifications.js
@@ -13,6 +13,8 @@ exports.up = (pgm) => {
     message_id: { type: 'string' },
     created_at: { type: 'timestamp', notNull: true, default: pgm.func('current_timestamp') },
   });
+
+  pgm.createIndex(SENT_PUSH_NOTIFICATIONS_TABLE, 'created_at');
 };
 
 exports.down = (pgm) => {
--- a/packages/hub/db/migrations/20211207151150639_sent-push-notifications.js
+++ b/packages/hub/db/migrations/20211207151150639_sent-push-notifications.js
@@ -2,20 +2,16 @@ const SENT_PUSH_NOTIFICATIONS_TABLE = 'sent_push_notifications';
 
 exports.up = (pgm) => {
   pgm.createTable(SENT_PUSH_NOTIFICATIONS_TABLE, {
-    transaction_hash: { type: 'string', notNull: true },
-    owner_address: { type: 'string', notNull: true },
-    push_client_id: { type: 'string', notNull: true },
+    // This is our own id for the notification.
+    notification_id: { type: 'string', primaryKey: true },
     notification_type: { type: 'string', notNull: true },
     notification_title: { type: 'string' },
     notification_body: { type: 'string', notNull: true },
     notification_data: { type: 'json' },
+    // This is firebase's id for the sent message
     message_id: { type: 'string', notNull: true },
     network: { type: 'string' },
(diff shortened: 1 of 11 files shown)


packages/hub/db/migrations/20211207151150639_sent-push-notifications.js
```diff
@@ -0,0 +1,23 @@
+const SENT_PUSH_NOTIFICATIONS_TABLE = 'sent_push_notifications';
+
+exports.up = (pgm) => {
+  // TODO: may be useful to add a message id field for debug purposes
+  // The message id is returned by firebase messaging
```
Aierie, line 5: https://firebase.google.com/docs/cloud-messaging/send-message#send-messages-to-specific-devices

Aierie:
Another field I'm wondering about adding is a `network` field

packages/hub/db/migrations/20211207151150639_sent-push-notifications.js
```diff
@@ -0,0 +1,23 @@
...
+    transaction_hash: { type: 'string', notNull: true },
+    owner_address: { type: 'string', notNull: true },
+    push_client_id: { type: 'string', notNull: true },
+    notification_type: { type: 'string', notNull: true },
+    notification_title: { type: 'string' },
+    notification_body: { type: 'string', notNull: true },
+    notification_data: { type: 'json' },
+  });
+
+  pgm.createIndex(SENT_PUSH_NOTIFICATIONS_TABLE, ['transaction_hash', 'owner_address', 'push_client_id'], {
```
jurgenwerk, line 16: Do you know if we're going to have any notifications that aren't tied to a transaction_hash in the future?

packages/hub/services/queries/sent-push-notifications.ts
```diff
@@ -0,0 +1,54 @@
...
+        model.pushClientId,
+        model.notificationType,
+        model.notificationTitle,
+        model.notificationBody,
+        model.notificationData,
+      ]
+    );
+  }
+
+  async exists(index: SentPushNotificationsIndex): Promise<any> {
```
jurgenwerk, line 38: Should we name this SentPushNotificationsFilter - for consistency with other filtering queries?

packages/hub/services/queries/sent-push-notifications.ts
```diff
@@ -0,0 +1,54 @@
...
+    let db = await this.databaseManager.getClient();
+
+    const conditions = buildConditions(index);
+
+    const query = `SELECT transaction_hash FROM sent_push_notifications WHERE ${conditions.where}`;
```
jurgenwerk, line 43: Minor thing but I think for the most efficient "exists" lookup a combination of SELECT 1 and LIMIT 1 can be used. https://semaphoreci.com/blog/2017/03/14/faster-rails-how-to-check-if-a-record-exists.html

packages/hub/db/migrations/20211207151150639_sent-push-notifications.js
Aierie: Not that I know of, but maybe adding event type to the index and replacing transaction hash with `event_id` might be good just in case?

packages/hub/services/queries/sent-push-notifications.ts
Aierie: I was hoping to communicate that this was the unique index of the table and that these fields were necessary to identify a notification for an event. Reading this I realise it is probably better to use a comment to highlight this at the type and use the convention of naming the parameter `filter` for db reads.

packages/hub/node-tests/tasks/send-notifications-test.ts
```diff
@@ -0,0 +1,125 @@
...
+        existingNotification.pushClientId,
+        existingNotification.notificationType,
+        existingNotification.notificationTitle,
+        existingNotification.notificationBody,
+        existingNotification.notificationData,
+        'existing-message-id',
+        existingNotification.network,
+      ]
+    );
+    subject = (await getContainer().lookup('send-notifications')) as SendNotifications;
```
backspace, line 88: Could you use the queries class here?

packages/hub/tasks/notify-customer-payment.ts
```diff
@@ -88,13 +89,20 @@ export default class NotifyCustomerPayment {
...
     let spendAmount = result.spendAmount;
-    let message = `${merchantName} received a payment of §${spendAmount}`;
+    let notificationBody = `${merchantName} received a payment of §${spendAmount}`;
 
-    await this.workerClient.addJob('send-notifications', {
-      notifiedAddress,
-      message,
-    });
+    // TODO: find out which device-address combinations need notifying
+    // and add a job for each (CS-2726)?
```
backspace, line 97: I imagined this happening in `send-notifications` so it doesn’t have to be duplicated in every transaction-to-notification task 🤔

backspace (approved):

packages/hub/db/migrations/20211207151150639_sent-push-notifications.js
lukemelia: This part of the design does seem odd to me.

lukemelia (commented):
I missed the discussion that lead to this work, so I may be missing something. Is the purpose of tracking every sent push notification in the database so that we can avoid sending duplicate notifications? Or is there also another design goal?

packages/hub/node-tests/tasks/send-notifications-test.ts
Aierie: For database interactions? I was wondering about that... using the queries class is more readable, but went with this because it felt like the database manager would be a lot less likely to have additional logic introduced in the future/ is a lot more stable.

Aierie:
> I missed the discussion that lead to this work, so I may be missing something. Is the purpose of tracking every sent push notification in the database so that we can avoid sending duplicate notifications? Or is there also another design goal?

The primary purpose is to avoid sending duplicate notifications, another is to help with debugging push notifications

packages/hub/tasks/notify-customer-payment.ts
Aierie: I'm not too sure about this either. I think having it here makes it easier to handle partial failures in `send-notifications` where we send some notifications successfully but fail at others.

jurgenwerk: I think the logic should be here - @Aierie good point on separating the retryable worker concerns. I'm currently working on this part in another PR

jurgenwerk:
@lukemelia more context on why we think this is needed can be found here: https://linear.app/cardstack/issue/CS-2717/deduplication-of-contract-event-handling

jurgenwerk:
@Aierie I think there's two more things to address here (but maybe in a separate PR)?
1. Providing a job_key when adding a job
2. Periodically wiping out old notifications. Maybe with https://www.npmjs.com/package/node-cron?

packages/hub/db/migrations/20211207151150639_sent-push-notifications.js
```diff
@@ -0,0 +1,23 @@
+const SENT_PUSH_NOTIFICATIONS_TABLE = 'sent_push_notifications';
+
+exports.up = (pgm) => {
+  pgm.createTable(SENT_PUSH_NOTIFICATIONS_TABLE, {
```
jurgenwerk, line 4: created_at or sent_at missing

jurgenwerk (changes requested):

packages/hub/node-tests/tasks/send-notifications-test.ts
jurgenwerk: I would encourage to use queries services as much as possible, I think it will be easier to refactor later on if we decide to introduce an ORM.

Aierie:
> @Aierie I think there's two more things to address here (but maybe in a separate PR?)
> 
> 1. Providing a job_key when adding a job
> 2. Periodically wiping out old notifications. Maybe with https://www.npmjs.com/package/node-cron?

Yep, I was thinking those should be in separate PRs as well. Spent some time investigating the job_key and retry options earlier, will make a PR for them later.

packages/hub/node-tests/tasks/send-notifications-test.ts
Aierie: Database interactions (besides wiping the sent-push-notifications db each time) are now done with the query service

Aierie:
the notificationId generation is quite rough right now and generates a difficult-to-read id. There is probably room for improvement here

packages/hub/tasks/send-notifications.ts
```diff
@@ -1,7 +1,90 @@
...
 
 export default class SendNotificationsTask {
-  async perform(payload: any, helpers: Helpers) {
-    helpers.logger.info('Notification to send:', payload);
+  sentPushNotificationsQueries = inject('sent-push-notifications-queries', { as: 'sentPushNotificationsQueries' });
+  firebasePushNotifications = inject('firebase-push-notifications', { as: 'firebasePushNotifications' });
+
+  async perform(payload: PushNotificationData, helpers: Helpers) {
+    try {
+      // we do this in a separate block because failure to read should not affect
```
jurgenwerk, line 27: I think this is a bit too defensive - a failure to interact with the database would imply a much deeper problem and I think it shouldn't be a concern of the worker to handle that. Ignoring that the database is failing and still performing work can lead to undesirable states which are hard to debug. 

I think it would be better if we just let the worker fail and let it be retried.

jurgenwerk (changes requested):

packages/hub/tasks/send-notifications.ts
Aierie: Currently not separating the try catch block since db services + the worker are both reliant on the db.. and the hub's subscriptions also rely on writing to the db succeeding. 

Maybe we can re-examine this at some point, in https://linear.app/cardstack/issue/CS-2806/review-push-notifications-dependency-on-database-services?

PR #2475, opened 2021-12-07
hub: Add tests for contract subscription handler

backspace:
These are tests for the `ContractSubscriptionEventHandler`, which imports contracts and maps subscribed contract events to notification-generating tasks. This is a precursor to persisting the latest-seen block number and resuming from it if there’s a disruption.


Aierie (approved):
Nice, hope the Contracts service keeps things tidy as we add more blockchain interactions to the hub!

PR #2478, opened 2021-12-08
Add migration script to hub CLI

habdelra:
To deal with the fact that everything is build via webpack now, I added node-pg-migrate as part of our hub CLI. I tested this locally and I was able to perform @jurgenwerk 's DB migration via the webpack build.

### DIFF
--- a/packages/hub/DEPLOYMENT.md
+++ b/packages/hub/DEPLOYMENT.md
@@ -71,12 +71,12 @@ Usage:
 
 Connect to the instance where the app is deployed:
 ```sh
-waypoint exec -app=hub bash
+waypoint exec -app=hub sh
 ```
 
 Then, add node to the PATH and run the migrations:
 ```sh
 heroku@ip-10-91-1-8:/ cd /workspace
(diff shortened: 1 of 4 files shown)


packages/hub/DEPLOYMENT.md
```diff
@@ -71,12 +71,12 @@ Usage:
...
 Connect to the instance where the app is deployed:
 ```sh
-waypoint exec -app=hub bash
+waypoint exec -app=hub sh
 ```
 
 Then, add node to the PATH and run the migrations:
 ```sh
 heroku@ip-10-91-1-8:/ cd /workspace
 heroku@ip-10-91-1-8:/workspace$ PATH=$PATH:/layers/heroku_nodejs-engine/nodejs/bin/;
```
lukemelia, line 80: Is the line above still correct? It looks like it is buildpack-related

lukemelia (approved):

packages/hub/DEPLOYMENT.md
habdelra: Probably not. I’ll fix this

jurgenwerk (approved):
@habdelra thanks so much for making this! 🙏 I ran this locally and works great. Will merge this and test on staging too.

PR #2480, opened 2021-12-09
include node-pg-migrate as external pkg in webpack build

habdelra:
Fixes #CS-2770 This works locally for me, but I'll test it as well in waypoint too. at the very least this shields the rest of the hub from the migration script logic since we are dynamically requiring it now.


habdelra:
argh, it looks like node-pg-migrate has a peer dep on `pg`, which means that has to be installed... Perhaps just doing a yarn install on pg is ok?

habdelra:
works now--tested it in staging and successfully ran the migrations in staging

PR #2483, opened 2021-12-09
[CS-2744] Cleanup global ts compile 

mattmcmanus:


habdelra:
This is looking really good but we still need:
- [ ] solve the heap out of memory issues in CI when we lint
- [ ] to have a `tsc` run in each package in order to typecheck the entire repo. 
- [ ] also I need to redo the cardpay CLI packaging. 
 
so don't merge yet 😉

PR #2484, opened 2021-12-09
hub: Add preliminary status endpoint

backspace:
This is a simple `GET /api/status` endpoint that reports degraded status if the subgraph is ≥10 blocks behind the RPC node. It has opportunities for improvement like adding status messages from a table, checking on the subgraph’s `hasIndexingErrors` value, and caching.

### DIFF
--- a/packages/hub/routes/status.ts
+++ b/packages/hub/routes/status.ts
@@ -50,9 +50,11 @@ export default class StatusRoute {
       data: {
         type: 'status',
         attributes: {
-          status,
-          subgraphBlockNumber,
-          rpcBlockNumber,
+          subgraph: {
+            status,
+            subgraphBlockNumber,
+            rpcBlockNumber,
+          },
         },
       },
     };
--- a/packages/hub/routes/status.ts
+++ b/packages/hub/routes/status.ts
@@ -1,0 +1,42 @@
+import Koa from 'koa';
+import autoBind from 'auto-bind';
+import { inject } from '@cardstack/di';
+
+const DEGRADED_THRESHOLD = 10;
+
+export default class StatusRoute {
+  subgraph = inject('subgraph');
+  web3 = inject('web3-http', { as: 'web3' });
+
+  constructor() {
+    autoBind(this);
+  }
+
+  async get(ctx: Koa.Context) {
+    let subgraphMeta = await this.subgraph.getMeta();
+
+    let subgraphBlockNumber = subgraphMeta.data._meta.block.number;
+    let rpcBlockNumber = await this.web3.getInstance().eth.getBlockNumber();
+
+    let status = rpcBlockNumber - subgraphBlockNumber >= DEGRADED_THRESHOLD ? 'degraded' : 'healthy';
+
+    ctx.status = 200;
+    ctx.body = {
+      data: {
+        type: 'status',
+        attributes: {
+          status,
+          subgraphBlockNumber,
+          rpcBlockNumber,
+        },
+      },
+    };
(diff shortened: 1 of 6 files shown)


packages/hub/routes/status.ts
```diff
@@ -0,0 +1,67 @@
...
+    if (rpcBlockNumber && subgraphBlockNumber && rpcBlockNumber - subgraphBlockNumber < DEGRADED_THRESHOLD) {
+      status = 'operational';
+    }
+
+    ctx.status = 200;
+    ctx.body = {
+      data: {
+        type: 'status',
+        attributes: {
+          status,
```
jurgenwerk, line 53: Would it be good to include the degradation notice title and message? Otherwise I imagine it would take some duplicated logic to form a meaningful degradation message across the apps that will consume this endpoint (dApp, Card Wallet).

Am I right to anticipate the /status endpoint could contain more metrics in the future? In this case I think it would be good to include info on which metric is the culprit for the degradation.

jurgenwerk: There's also a task where we'd like to have an externally hosted status page. I think those kinds of pages usually show the status of different services. Would it be good to model our status endpoint to include the status of individual services as well? Maybe something like:

```
status: {
 subgraph: {
   status: "degraded",
   reason: "",
  subgraphBlockNumber,
  rpcBlockNumber
 },
 someOtherService: {
   status: "operational",
   ... 
 }, 
 {...}
}
```

backspace: Great suggestions, thanks. I’ve updated to move all subgraph status-related information into a `subgraph` container. I’ve also added an `errors` top-level container when there’s an error that has a relevant `title` and a `source`.

The known failure of the subgraph being behind seems self-evident:

```
{
        data: {
          type: 'status',
          attributes: {
            subgraph: {
              rpcBlockNumber: 19492430,
              status: 'degraded',
              subgraphBlockNumber: 19492419,
            },
          },
        },
        errors: [
          {
            id: 'subgraph',
            title: 'Experiencing slow service',
            source: {
              pointer: '/data/attributes/subgraph/subgraphBlockNumber',
            },
          },
        ],
      }
```

However, representing a failure to query the subgraph or the RPC node is more challenging as the [error object specification](https://jsonapi.org/format/#error-objects) only supports a `parameter` or `pointer` under `source`; I used `service` but maybe the `title` alone is sufficient?
```
{
    data: {
        type: 'status',
        attributes: {
        subgraph: {
            rpcBlockNumber: null,
            status: 'degraded',
            subgraphBlockNumber: 19492428,
        },
        },
    },
    errors: [
        {
        id: 'subgraph',
        title: 'Error checking status',
        source: {
            service: 'web3-http',
        },
        },
    ],
}
```

jurgenwerk: Cool, thanks for considering this!. Do you think we can drop the `errors` part and move the explanation to the `attributes`? I feel like the `errors` is more suitable for model validation errors while here we're just serving some information. My suggestion:

```
data: {
  type: 'status',
  attributes: {
    subgraph: {
      status: 'degraded',
      rpcBlockNumber: 19492430,
      subgraphBlockNumber: 19492419,
      details: "Experiencing slow service"
    },
  },
},
```

And in cases when there is a failure to query, I'd consider setting the status to `unknown` 🤷‍♂️

backspace: ya, makes sense that it isn’t really fitting with errors in JSON:API. I’ve done both of those now, thanks!

jurgenwerk (approved):

PR #2494, opened 2021-12-13
update getABI in SDK to import ABIs as .ts files

habdelra:
@Aierie This fixed the bug on my local after first running `yarn clean`


Aierie:
Thanks! Didn't realise the solution would be so simple.

PR #2500, opened 2021-12-13
Move hub serialization into core

mattmcmanus:
- [x] Serialization done in Hub into core
- [x] CardModel serializing/deserializing: Can it be consolidated with the rawcard de/serializers?
- [x] What can be reused between raw-card deserializer and serializer?
- [x] Types around json payloads need a lot of work


packages/core/src/interfaces.ts
```diff
@@ -159,45 +165,37 @@ export interface RealmConfig {
...
-export interface CardJSONResponse {
-  data: {
-    id: string;
-    type: string;
-    attributes?: { [name: string]: any };
-    meta?: {
-      componentModule: string;
-    };
-  };
+export interface JSONAPIDocument<Identity extends Saved | Unsaved = Saved> {
```
habdelra, line 168: nice!

habdelra (approved):
LGTM

PR #2501, opened 2021-12-14
prevent dynamic ts imports from breaking hub rebuilds

Aierie:
Webpack was trying to compile the `.d.ts` files - happens when `yarn rebuild` is run or sequential `yarn build`s are run without a `yarn clean` in between. This change excludes `.d.ts`. This works for consuming this package via hub's webpack config.

### DIFF
--- a/packages/cardpay-sdk/sdk/version-resolver.ts
+++ b/packages/cardpay-sdk/sdk/version-resolver.ts
@@ -102,12 +102,7 @@ export async function getABI(contractName: string, web3: Web3): Promise<AbiItem[
   for (let version of protocolVersions) {
     // we have to exclude .d.ts files because webpack tries to build
     // these for the dynamic import case.
-    versionMap[version.replace('v', '')] = (
-      await import(
-        /* webpackExclude: /\.d\.ts$/ */
-        `../contracts/abi/${version}/${contractName}.ts`
-      )
-    ).default;
+    versionMap[version.replace('v', '')] = (await import(`../contracts/abi/${version}/${contractName}.ts`)).default;
   }
   let abi = getAPIVersion(versionMap, protocolVersion);
   return abi;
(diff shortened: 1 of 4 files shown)


Aierie:
Something seems off here...? This is not causing hub CI to run.

habdelra (changes requested):
this seems very wrong. we should address this kind of thing in tsconfig. not in the literal code

habdelra:
Actually this stack overflow answer looks really good: https://stackoverflow.com/questions/38231645/typescriptwebpack-typescript-emmited-no-output-for-index-d-ts:

> In this particular question the content of index.d.ts was not a definition but a module and should have been moved into index.ts file. I've also ran into error "typescript emmited no output for index.d.ts" but with valid declaration files. 
It seems ts-loader tries to add .d.ts files to final bundle but finds nothing to add since they contain only declarations needed for type-checking during build.
Working solution for me is not to pass .d.ts files to ts-loader but to some loader that does nothing, e. g. ignore-loader. Corresponding rules in my webpack.config.js are:
```
{
    test: /\.tsx?$/,
    loader: 'ts-loader',
    exclude: /node_modules|\.d\.ts$/
},
{
    test: /\.d\.ts$/,
    loader: 'ignore-loader'
},
```
> ts-loader can be configured slightly differently if you use ES2018, where negative lookbehind for regular expressions was added:
```
{
    test: /(?<!\.d)\.tsx?$/,
    loader: 'ts-loader',
    exclude: /node_modules/
},
//same ignore-loader config here
```

habdelra (approved):

habdelra:
@Aierie I used the stack overflow example above and was able to get the webpack build to ignore .d.ts files when processing the dynamic ts imports. This solution will work a lot better since it works for any ts dynamic imports instead of the specific scenario in the ABI loading

Aierie:
Cool, thanks! This does seem a lot more robust.

PR #2502, opened 2021-12-14
add cron job to delete old notifications every day at NY midnight

Aierie:
https://github.com/graphile/worker#recurring-tasks-crontab

If there is an error, this task will fail and send a message to Sentry, retrying up to 5x. There is no backfill for this task, so if the worker is dead at midnight, it won't run. Given its frequency and purpose, it seems ok to not introduce backfill for now?


jurgenwerk (approved):
Looks good. Can we add a test?

habdelra:
@Aierie I made the discord tests more stable. if you merge upstream changes it should fix the broken tests

PR #2509, opened 2021-12-14
Fix/incorrect time comparison test

Aierie:
Providing UTC strings seems to have been the problem. Tested by setting `process.env.TZ="America/New_York"` before running tests.


backspace (approved):
This works for me in `America/Winnipeg` time zone 🎉

PR #2510, opened 2021-12-14
[CS-1806] working toward compile error handling in indexers

ef4:
- [x] Error serialization for cards that can't be compiled
- [x] Next test: Fix index error and see that invalidation system kicks in
  - Write dependencies to index column when error is saved
  - Call create to make the missing thing in the test
  - Keep track of all changed cards in an indexerRun, recompile dependents
  - See that recompile happens
- [ ] ~Remove old cache support~ https://linear.app/cardstack/issue/CS-2840/remove-storage-of-compiled-cards-from-card-cache


mattmcmanus:
Moved last TODO here to it's own card: https://linear.app/cardstack/issue/CS-2840/remove-storage-of-compiled-cards-from-card-cache

PR #2515, opened 2021-12-14
Updating main in other packages to point to .ts file rather than .js

lukemelia:


habdelra (approved):
👍

habdelra:
Actually i think the firebase functions could be fixed up too

PR #2516, opened 2021-12-14
web: Add log exclusion of expected cancelation

backspace:
This is fix akin to #2330 to reduce log noise, as we don’t
care about cancelation that happens because the user
dismissed the workflow. It would be nice to solve this
in a general way but I’m unsure how.

I exercised this in a similar fashion to #2330. I was motivated to fix this because of [this noise today](https://discord.com/channels/584043165066199050/855105564643819530/920428699273265163).


Aierie:
I guess the general case is solved by making `perform task` ala http://ember-concurrency.com/docs/task-cancelation-help a convention?

PR #2519, opened 2021-12-15
A minimal degraded service banner

Aierie:
CS-2829: Minimal banner to use until/if we get design updates. The examples below are done by adding the component to the top of the `Boxel::Dashboard` component's `<:body>` named block in `card-pay.hbs`.

image

image


Aierie:
Design will be modified, closing for now.

backspace (approved):
I think this will work well with #2505 🤩

jurgenwerk (approved):

PR #2524, opened 2021-12-16
👂 📲 Separate process for hub contract event listener for notifications 

jurgenwerk:
Ticket: [CS-2721](https://linear.app/cardstack/issue/CS-2721/run-subscriptions-in-separate-unique-process)

Introduces a separate app based on the hub which runs the contract event subscription handler, which makes new jobs for the notification sender worker. 

I deployed this to staging and it looks like it works: https://waypoint-staging.stack.cards/default/cardstack/app/hub-event-listener/logs

TODO:

- [x] Add production config


packages/hub/process-controllers/hub-event-listener-controller.ts
```diff
@@ -0,0 +1,54 @@
...
+
+    try {
+      let handler = await container.instantiate(
+        ContractSubscriptionEventHandler,
+        web3,
+        workerClient,
+        contracts,
+        latestEventBlockQueries
+      );
+      await handler.setupContractEventSubscriptions();
```
backspace, line 37: The reason we made `ContractSubscriptionEventHandler` not a service is that we were afraid it might be instantiated elsewhere, so making it not easily injectable was a way to hopefully prevent that. But it wasn’t really that sensible maybe. Now that you’ve made it a service, it can just use `inject` like other services, vs having the dependencies passed in via the constructor.

backspace (changes requested):
Thanks for working through this and somewhat reducing the hugeness of `main.ts` 😆 I think the listener should be fully service-ified but otherwise this looks good to me

packages/hub/process-controllers/hub-event-listener-controller.ts
jurgenwerk: Thanks for explaining - makes sense. Will adjust it.

jurgenwerk:
@buck I made the thing a service now. I think it should be safe - we're only calling it in the `HubEventListenerController`. Can you check if it looks good to you now?

packages/hub/process-controllers/hub-bot-controller.ts
```diff
@@ -0,0 +1,47 @@
+import { runInitializers, createRegistry } from '../main';
+import HubBot from '../services/discord-bots/hub-bot';
+import * as Sentry from '@sentry/node';
+import { Registry, Container } from '@cardstack/di';
+import { botLog } from '../utils/logger';
+
+export class HubBotController {
```
habdelra, line 7: So the way the hub bot and worker controllers do thier DI is actually inverted and we have tickets to fix them. that being said we should stop repeating this pattern. The idea is that the controller should not create the container, rather the controller lives in the container. The correct pattern to follow is the HubServer. Where it is dependency injected into the container (rather than creating the container). You can refer to tickets #CS-2707 and #CS-2606

packages/hub/process-controllers/hub-event-listener-controller.ts
```diff
@@ -0,0 +1,43 @@
+import { ContractSubscriptionEventHandler } from '../services/contract-subscription-event-handler';
+import { runInitializers, createRegistry } from '../main';
+import * as Sentry from '@sentry/node';
+import { Registry, Container } from '@cardstack/di';
+import { eventListenerLog } from '../utils/logger';
+
+export class HubEventListenerController {
```
habdelra, line 7: So the way the hub bot and worker controllers do their DI is actually inverted and we have tickets to fix them. that being said we should stop repeating this pattern. The idea is that these controllers should not create the container, rather the controllers live in the container. The correct pattern to follow is the HubServer. Where it is dependency injected into the container (rather than creating the container)--note how the HubServer is registered in the registry, and the CLI creates the container, and then grabs the HubServer out of the container and starts it. The Hub EventListener should follow this pattern too. it should have a CLI command that creates the container and then starts the hub event listener. You can refer to tickets #CS-2707 and #CS-2606

habdelra (changes requested):
invert the DI for the HubEventListener as mentioned in the comments

packages/hub/process-controllers/hub-event-listener-controller.ts
```diff
@@ -0,0 +1,43 @@
...
+    registryCallback?: (r: Registry) => void;
+  }): Promise<HubEventListenerController> {
+    this.logger.info(`booting pid:${process.pid}`);
+    runInitializers();
+
+    let registry = createRegistry();
+    if (serverConfig?.registryCallback) {
+      serverConfig.registryCallback(registry);
+    }
+    let container = new Container(registry);
```
habdelra, line 21: this is a smell--we should not create containers at this level. the hob bot controller and the worker controller are actually incorrect in this regards

jurgenwerk:
@habdelra I followed your suggestion and adjusted the handler to do lookup instead. I think the controller might be unnecessary now and the cli handler should suffice. Can you check if it looks ok?

habdelra:
yup this looks great--thanks!

habdelra (approved):

backspace:
> @buck I made the thing a service now. I think it should be safe - we're only calling it in the `HubEventListenerController`. Can you check if it looks good to you now?

great, thanks!

PR #2528, opened 2021-12-20
Show degradation banner by reading from the statuspage

jurgenwerk:
Ticket: [CS-2752](https://linear.app/cardstack/issue/CS-2752/service-degradation-notice-in-dapp)

image

This PR shows the degradation banner in Card Pay route in case there is an open incident in Statuspage. 

Incident severity can be categorized as "none", "minor", "major", "critical" in statuspage. For "major" and "critical" we show the red background, and yellow for the rest.

### DIFF
--- a/packages/web-client/app/components/common/degraded-service-banner/index.css
+++ b/packages/web-client/app/components/common/degraded-service-banner/index.css
@@ -15,6 +15,27 @@
   color: var(--degraded-service-banner-text-color);
 }
 
+.degraded-service-banner__container {
+  max-width: calc(var(--degraded-service-banner-icon-area-width) + 80ch);
+  display: flex;
+}
+
+.degraded-service-banner__icon-area {
+  --degraded-service-banner-icon-area-width: 1.7rem;
+
+  display: flex;
+  align-items: center;
+  width: var(--degraded-service-banner-icon-area-width);
+}
+
+.degraded-service-banner__container a {
+  text-decoration: underline;
+}
+
+.degraded-service-banner__container a:hover {
+  color: inherit;
+}
+
 .degraded-service-banner.degraded-service-banner--severe {
   --degraded-service-banner-text-color: var(--boxel-light);
   --degraded-service-banner-background-color: var(--boxel-red);
--- a/packages/web-client/app/components/common/degraded-service-banner/index.ts
+++ b/packages/web-client/app/components/common/degraded-service-banner/index.ts
@@ -1,0 +1,17 @@
+import config from '@cardstack/web-client/config/environment';
+import { inject as service } from '@ember/service';
+import DegradedServiceDetector from '@cardstack/web-client/services/degraded-service-detector';
+import { reads } from 'macro-decorators';
+import Component from '@glimmer/component';
+
+export default class DegradedServiceBannerComponent extends Component {
+  statusPageUrl = config.urls.statusPageUrl;
+  @service declare degradedServiceDetector: DegradedServiceDetector;
+
+  @reads('degradedServiceDetector.notificationShown')
+  declare notificationShown: boolean;
+  @reads('degradedServiceDetector.notificationBody') declare notificationBody:
+    | string
+    | null;
+  @reads('degradedServiceDetector.isSevere') declare isSevere: boolean;
+}
--- a/packages/web-client/app/services/degraded-service-detector.ts
+++ b/packages/web-client/app/services/degraded-service-detector.ts
@@ -1,0 +1,98 @@
+import Service from '@ember/service';
+import config from '../config/environment';
+import { taskFor } from 'ember-concurrency-ts';
+import { task, TaskGenerator, timeout } from 'ember-concurrency';
+import { tracked } from '@glimmer/tracking';
+import * as Sentry from '@sentry/browser';
+
+export default class DegradedServiceDetector extends Service {
+  @tracked notificationShown: boolean = false;
+  @tracked notificationBody: string | null = null;
+  @tracked impact: 'none' | 'minor' | 'major' | 'critical' | null = null;
+
+  statusPageUrl = config.urls.statusPageUrl;
+
+  constructor() {
+    super(...arguments);
+    taskFor(this.pollForStatusTask).perform();
+  }
+
+  get isSevere() {
+    return this.impact === 'major' || this.impact === 'critical';
+  }
+
+  @task *pollForStatusTask(): TaskGenerator<void> {
+    while (true) {
+      let statusData = yield this.getDegradationStatusData();
+
+      if (statusData) {
+        this.notificationShown = true;
+        this.impact = statusData.impact;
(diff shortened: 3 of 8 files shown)


packages/web-client/config/environment.js
```diff
@@ -82,6 +82,7 @@ module.exports = function (environment) {
       testFlightLink: 'https://testflight.apple.com/join/OgFq1EZ0',
       discordSupportChannelUrl:
         'https://discord.com/channels/584043165066199050/899645340746141806',
+      statusPageUrl: 'https://vandelayindustries1.statuspage.io',
```
jurgenwerk, line 85: My own statuspage for testing purposes, will add Cardstack's status page once it's fully set up

packages/web-client/app/components/common/degraded-service-banner/index.ts
```diff
@@ -0,0 +1,71 @@
...
+  }
+
+  async getDegradationStatusData(): Promise<any> {
+    let statusPageUrl = `${this.statusPageUrl}/api/v2/incidents/unresolved.json`;
+
+    try {
+      let response = await fetch(statusPageUrl);
+      let data = await response.json();
+
+      let lastUnresolvedIncident = data.incidents[0];
```
jurgenwerk, line 53: Need to confirm if the incidents are sorted by most recent first

packages/web-client/app/components/common/degraded-service-banner/index.css
```diff
@@ -15,11 +15,35 @@
   color: var(--degraded-service-banner-text-color);
 }
 
+.degraded-service-banner__container {
+  max-width: 80%;
```
Aierie, line 19: Minor since design updates may be incoming: `max-width: calc(var(--icon-space) + 40ch);`? This would make it work across most screen sizes without the need for a media query, and directly address the intention of restricting text length for readability purposes.

(requires setting variable(s) that represents length - icon width + margin)

packages/web-client/app/components/common/degraded-service-banner/index.ts
```diff
@@ -0,0 +1,71 @@
...
+        this.notificationBody = statusData.name;
+      } else {
+        this.notificationShown = false;
+        this.notificationBody = null;
+        this.impact = null;
+      }
+
+      yield timeout(1000 * 60);
+    }
+  }
```
Aierie, line 44: There might be 2 of these banners in the design because we display one in a workflow, so this might be better as a service?

packages/web-client/app/components/common/degraded-service-banner/index.ts
```diff
@@ -0,0 +1,71 @@
...
+          impact: lastUnresolvedIncident.impact,
+        };
+      } else {
+        return null;
+      }
+    } catch (_) {
+      Sentry.captureException(
+        `Unable to fetch service status from ${statusPageUrl}`
+      );
+      return {};
```
Aierie, line 68: Should this return null?

backspace (approved):
This seems like a good start! Some tests would be great 😀

packages/web-client/app/components/common/degraded-service-banner/index.ts
jurgenwerk: Sounds good, will make a service

packages/web-client/app/components/common/degraded-service-banner/index.ts
jurgenwerk: Good catch!

packages/web-client/app/components/common/degraded-service-banner/index.ts
```diff
@@ -0,0 +1,71 @@
...
+
+      let lastUnresolvedIncident = data.incidents[0];
+
+      if (lastUnresolvedIncident) {
+        return {
+          status: lastUnresolvedIncident.status,
+          name: lastUnresolvedIncident.name,
+          impact: lastUnresolvedIncident.impact,
+        };
+      } else {
```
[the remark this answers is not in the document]
jurgenwerk: Good idea! Will adjust it

packages/web-client/app/services/degraded-service-detector.ts
```diff
@@ -0,0 +1,98 @@
...
+        this.impact = statusData.impact;
+        this.notificationBody = statusData.name;
+      } else {
+        this.notificationShown = false;
+        this.notificationBody = null;
+        this.impact = null;
+      }
+
+      if (config.environment === 'test') {
+        return;
```
jurgenwerk, line 39: Need this to stop looping in tests otherwise the tests never finish

packages/web-client/app/services/degraded-service-detector.ts
```diff
@@ -0,0 +1,98 @@
...
+        return -1;
+      }
+      if (order.indexOf(a.impact) < order.indexOf(b.impact)) {
+        return 1;
+      }
+
+      return 0;
+    })[0].impact;
+
+    let incident = data.incidents
```
jurgenwerk, line 78: Most recent incident with the highest impact

jurgenwerk:
@Aierie @backspace this is now ready for another review. I suppose it would be good if we come up with a service degradation playbook (a process to follow when we detect issues where we'd include a step to update https://status.cardstack.com/)

packages/web-client/tests/integration/components/degraded-service-banner-test.ts
```diff
@@ -0,0 +1,194 @@
...
+
+interface Context extends MirageTestContext {}
+
+module('Integration | Component | degraded-service-banner', function (hooks) {
+  setupRenderingTest(hooks);
+  setupMirage(hooks);
+
+  test('It doesn not display a banner when there are no incidents', async function (this: Context, assert) {
+    this.server.get(
+      'https://status.cardstack.com/api/v2/incidents/unresolved.json',
```
backspace, line 17: Not that this is likely to change, but would it make sense to use the host from the config in the tests too?

packages/web-client/tests/integration/components/degraded-service-banner-test.ts
```diff
@@ -0,0 +1,194 @@
...
+    );
+
+    await render(hbs`<Common::DegradedServiceBanner />`);
+    await waitFor('[data-test-degraded-service-banner]');
+    assert
+      .dom('[data-test-degraded-service-banner]')
+      .containsText('World down. For more details, check our status page');
+    assert
+      .dom('[data-test-degraded-service-banner]')
+      .hasClass('degraded-service-banner--severe');
```
backspace, line 164: Nice test 🤩

backspace (approved):

packages/web-client/tests/integration/components/degraded-service-banner-test.ts
jurgenwerk: Yup! Will clean it up.

PR #2529, opened 2021-12-20
silence walletconnect and metamask connection cancelation in sentry

Aierie:
CS-2845

### DIFF
--- a/packages/web-client/app/utils/web3-strategies/layer-1-connection-manager.ts
+++ b/packages/web-client/app/utils/web3-strategies/layer-1-connection-manager.ts
@@ -149,8 +149,10 @@ export class ConnectionManager {
 
   async connect(web3: Web3, providerId: WalletProviderId) {
     await this.setup(providerId);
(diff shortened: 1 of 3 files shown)


packages/web-client/app/utils/web3-strategies/layer-1-connection-manager.ts
```diff
@@ -149,8 +149,10 @@ export class ConnectionManager {
 
   async connect(web3: Web3, providerId: WalletProviderId) {
     await this.setup(providerId);
+    if (!this.strategy)
+      throw new Error('Failed to setup strategy in layer 1 connection manager');
     web3.setProvider(this.provider);
-    await this.strategy?.connect();
+    return await this.strategy.connect();
```
Aierie, line 155: This change was to make this function consistently return a boolean if setup works correctly.

backspace (approved):
Thanks for doing this, the fewer spurious issues in Sentry the more focused we can be on true problems! 🤩

PR #2530, opened 2021-12-20
Deep dependency invalidations and Update support

habdelra:
This PR supports deep dependency invalidations. Previous we only tracked deps one level deep. In this PR we track deps as deep as they go. The unit test introduced in this PR has a grandparent card that is invalidated. The grandparent invalidation triggers reindexing of the grandchild card. 

Within the grandchild card unit test, the following is the contents of the `deps` column after this change:
Screen Shot 2021-12-20 at 1 43 13 PM

For this change, the current `TrackedBuilder` was insufficient, as the context that we use to append the dependencies needs to recurse with compiler. In order to support this I resurrected the `CompileContext` class, and added a single field `dependencies` which was the `Set<string>` we were using in the `TrackedBuilder`. The `context` is then passed into the recursing compilers so that it can have a complete view of the deeply nested deps for a card.


ef4:
I think this is relying on the recursive-compilation that we're about to eliminate.

The idea is that most of the time you *won't* call `getCompiledCard` recursively for each of the inner dependencies that you need, because each CompiledCard already rolls up everything that it depended on. Rather, you'd want `TrackedBuilder` to notice the `deps` associated with each of the `CompiledCards` that it is returning and add those deps to its own dependencies set.

habdelra:
@ef4 So a challenge with that approach is that sometimes the compile will fail (like in the new unit test I added for this scenario) because the grandparent card doesn't exist when the grandchild is first indexed. As a result, there will be no compiled card to inspect in the `TrackedBuilder` wrapper. How should we handle that kind of situation?

ef4:
In that case there's no dependency (yet) that needs to be tracked. We still get the dependency on the missing grandparent itself, so we will recompile when it becomes available and then we will gather its inner dependencies.

habdelra:
@ef4 I thought that in order for our invalidation system to work without having to discover new items to invalidate during the invalidation sweep required that we deeply record all the deps. When the grandparent is created, how is the grandchild going to know to be invalidated since there is no reference to it? do we add a new rule that any card that with a compilation error is volatile?

habdelra:
one thing that might help is that we do see in the error the dependency to the missing card--maybe the TrackedBuilder needs to also take as input 404 compilation errors?

habdelra:
@ef4 here is another take where I'm using the 404s as a signal that there is a dependency in the `TrackedBuilder` so that we can do our best effort to make the deps that we add to the cards table as deep as possible so that we can avoid having to worry about discovering more invalidations during an invalidation sweep.

packages/core/src/compiler.ts
```diff
@@ -452,12 +452,40 @@ export function makeGloballyAddressable(
...
     this.dependencies.add(url);
-    return this.realBuilder.getCompiledCard(url);
+    let card: CompiledCard | undefined;
+    await this.captureMissingCardDependencies(async () => {
+      card = await this.realBuilder.getCompiledCard(url);
+    });
+    if (!card) {
+      throw new Error('Should never get here');
+    }
+    // TODO capture additional deps from the compiled card
```
mattmcmanus, line 465: This todo isn't super clear. What additional deps?

habdelra:
I think we need to get the card service update working before I can write a test that can explore the deps returned from the `CompiledCard` in the `TrackedBuilder`.

packages/core/src/compiler.ts
habdelra: @mattmcmanus so this would entail recursing through the `adoptsFrom` and the `fields` properties of the `CompiledCard` to record all the cards that we encounter. For example, if we have an `author` field that uses a `Person` card and the `Person` card itself adopts from a `User` card, then `User` is a dependency of our card. Like that kind of stuff...

habdelra: This is basically @ef4 's first comment on the PR.

habdelra: I made a note above that I think we really need to have a card service update in place before I can really write this kind of test. it would be the sort of thing were we _update_ one of our deps and we should observe that triggers an invalidation of our card. we can really only push so far with purely a _create_ mutator in the card service and see deep invalidations, i think this PR is the extent of that. So I think we'll have to circle back and write those tests after the update is ready.

habdelra:
@ef4 @mattmcmanus i've merged Matt's PR which cleans up the `builder.getCompiledCard` so that we really are no longer recursively compiling. I've updated the `card-service.update()` to perform indexing on update. Also, I had to create a `card-service.updateData()` for card data only updates (the unit tests revealed this need). To support this, I also made a simpler mechanism to upsert a data-only entry into the index so that we don't have to compile the card. As well as we are now recursively discovering all the dependencies in the `TrackedBuilder` for `adoptsFrom` and `fields` simultaneously, and our tests show that, where we can alter a granparent card, and see a new field added into the grandchild as a result of the index invalidations. Same with altering a field of a field of a field, and seeing the index invalidate that graph correctly.

packages/hub/services/search-index.ts
```diff
@@ -222,6 +232,19 @@ class IndexerRun implements IndexerHandle {
     });
     return out!.code!;
   }
+  private async writeCardDataToIndex(card: RawCard, compiledCard: CompiledCard): Promise<CompiledCard> {
```
mattmcmanus, line 235: What did the unit tests reveal as to why was this separate method necessary? it seems like writeToIndex would be sufficient. Regardless, you'll also need to update the `raw` key as well, since it also includes the data

mattmcmanus (commented):
I have two questions that might require changes...

packages/hub/services/search-index.ts
habdelra: There was a unit test for updating a data only card. We simply can’t compile a card that is data only.

packages/hub/node-tests/services/search-index-test.ts
```diff
@@ -89,13 +90,171 @@ if (process.env.COMPILER) {
       expect(grandChild.data?.title).to.eq('Hello World');
     });
 
-    it(`can invalidate a card via the update of an adoptsFrom card`);
-    it(`can invalidate a card via the update of a field card`);
-    it(`can invalidate a card via the update of a grandparent adoptsFrom card`);
-    it(`can invalidate a card via the update of a "grandparent" field card`); // add a new field to a field of a field
+    it(`can invalidate a card via the update of an adoptsFrom card`, async function () {
+      let si = await getContainer().lookup('searchIndex');
+      await si.indexAllRealms();
```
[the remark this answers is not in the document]
habdelra: We’ll I was seeing that a new generation wasn’t being created without this and index state in this module that didn’t have to is were bleeding into the different tests

packages/hub/services/search-index.ts
habdelra: I wonder if that means we need to merge the data only update into the existing raw card?

packages/hub/node-tests/services/search-index-test.ts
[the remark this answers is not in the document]
habdelra: I’d be happy to pair on this with you tomorrow @mattmcmanus

packages/hub/services/search-index.ts
```diff
@@ -143,6 +149,19 @@ class IndexerRun implements IndexerHandle {
...
+      // whose parent, and grandchild were invalidated. We need to compile the
+      // parent card before the grandchild, as when the grandchild is compiled,
+      // it will use the parent's compiled card from the index. if we don't
+      // compile the parent card first, then the grandchild card will not
+      // contain the updated source from the original invalidation. In order to
+      // accomplish this consider converting the invalidated cards into a
+      // consumption graph, and then traverse it in such a way that we compiled
+      // the consumed cards before the consumer cards. (this is a pattern we
+      // used in CatalogJS that might be wroth repurposing)
+
```
habdelra, line 164: @ef4 @mattmcmanus so right now we have some flaky tests because of this which comes down to the order of reindexing. This is what the tests looks like when they pass:
``` 
  hub/search-index starting to index realm https://cardstack.local/
  hub/search-index reindexing https://cardstack.local/example because [
  hub/search-index   'https://cardstack.local/post',
  hub/search-index   'https://cardstack.com/base/base',
  hub/search-index   'https://cardstack.com/base/string'
  hub/search-index ]
  hub/search-index reindexing https://cardstack.local/grandchild because [
  hub/search-index   'https://cardstack.local/example',
  hub/search-index   'https://cardstack.local/post',
  hub/search-index   'https://cardstack.com/base/base',
  hub/search-index   'https://cardstack.com/base/string'
  hub/search-index ]
  hub/search-index finished indexing realm https://cardstack.local/
    ✓ can invalidate a card via the update of a grandparent adoptsFrom card (176ms)
```
this is what they look like when they fail: 
```
  hub/search-index starting to index realm https://cardstack.local/
  hub/search-index reindexing https://cardstack.local/grandchild because [
  hub/search-index   'https://cardstack.local/example',
  hub/search-index   'https://cardstack.local/post',
  hub/search-index   'https://cardstack.com/base/base',
  hub/search-index   'https://cardstack.com/base/string'
  hub/search-index ]
  hub/search-index reindexing https://cardstack.local/example because [
  hub/search-index   'https://cardstack.local/post',
  hub/search-index   'https://cardstack.com/base/base',
  hub/search-index   'https://cardstack.com/base/string'
  hub/search-index ]
  hub/search-index finished indexing realm https://cardstack.local/
    1) can invalidate a card via the update of a grandparent adoptsFrom card
```
note that in the case where it passes, the parent card is reindexed before the grandchild card
in the case where it fails, the grandchild card is reindexed before the parent card
i think it comes down to the fact that the grandchild is looking at the index for the compiled card of it's parent
in the case we index the parent after the grandchild, the parent does not yet have the new field
and hence the granchild does not yet have the new field

packages/core/src/compiler.ts
```diff
@@ -452,12 +452,53 @@ export function makeGloballyAddressable(
 class TrackedBuilder implements Builder {
   readonly dependencies = new Set<string>();
   constructor(private realBuilder: Builder) {}
-  getCompiledCard(url: string): Promise<CompiledCard> {
+
+  async getCompiledCard(url: string): Promise<CompiledCard> {
     this.dependencies.add(url);
```
habdelra, line 457: look at the row for `url` and use its deps as your deps.

packages/core/src/compiler.ts
```diff
@@ -452,12 +452,53 @@ export function makeGloballyAddressable(
...
-    return this.realBuilder.getCompiledCard(url);
+    let card: CompiledCard | undefined;
+    await this.captureMissingCardDependencies(async () => {
```
habdelra, line 459: no need to do this if you are checking for deps above

packages/core/src/compiler.ts
```diff
@@ -452,12 +452,53 @@ export function makeGloballyAddressable(
...
+      card = await this.realBuilder.getCompiledCard(url);
+    });
+    if (!card) {
+      throw new Error('Should never get here');
+    }
+    this.discoverDeps(card);
```
habdelra, line 465: likely we don't need to do this if we added the recursive deps from above

habdelra:
This work has moved into https://github.com/cardstack/cardstack/pull/2547

PR #2531, opened 2021-12-20
Boxel radio component upstream changes

burieberry:
#CS-2847

Changes:
- Implemented in boxel the radio button styles used in card pay and renamed the boxel component
- Add new args to the API
- Used component in Card Pay

radio-input


Aierie (approved):
Thanks, I think this works well and should reduce mental overhead + boilerplate writing when we're implementing common radio functionality! 

With the radio buttons unified, it should be easier to add a visible focus state that is distinguishable from the checked state, and not masked by the hover state. May be a good time to prompt design about this?

packages/boxel/addon/components/boxel/radio-input/usage.ts
```diff
@@ -2,9 +2,8 @@ import Component from '@glimmer/component';
 import { tracked } from '@glimmer/tracking';
 import { A } from '@ember/array';
 import { action } from '@ember/object';
-import './usage.css';
 
-export default class CustomRadioUsage extends Component {
+export default class RadioInputUsage extends Component {
```
backspace, line 6: I don’t know what it is but something’s broken with `items` in the documentation; it doesn’t show any contents and clicking the button to add causes an error 😳

image

packages/boxel/addon/components/boxel/radio-input/usage.hbs
```diff
@@ -0,0 +1,99 @@
...
+      @onInput={{fn (mut this.gridColumnWidth)}}
+      @value={{this.gridColumnWidth}}
+    />
+    <Args.String
+      @name="containerWidth"
+      @description="Width for the container. ie. `max-content`, `50%`, `400px`..."
+      @defaultValue="100%"
+      @onInput={{fn (mut this.containerWidth)}}
+      @value={{this.containerWidth}}
+    />
```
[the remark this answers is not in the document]
burieberry: This was the trickiest part, since I don't know how well it will work going forward with other examples. We want boxel components to have styling and layout out-of-the-box, so I set these attributes for it. If it doesn't work well for our needs, I'm open to reconsidering including these.

packages/boxel/addon/components/boxel/radio-input/usage.ts
burieberry: Thanks for noticing, fixing now.

PR #2533, opened 2021-12-21
Use more focused loggers

mattmcmanus:
Reusing the `hub/server` logger is an anti-pattern. Using more focused loggers allows us to get access to the specific logging we care about: 
```
LOG_LEVELS='*=none,hub/fs-realm*=trace' yarn start:server
```


habdelra (approved):
looks good 👍

PR #2534, opened 2021-12-21
[CS-2707] Fix inverted worker di

mattmcmanus:
- Fix inverted DI setup for hub worker
- Simplify task setup for the worker


lukemelia (approved):
Looks solid to me @mattmcmanus 👍

PR #2535, opened 2021-12-21
Fixes inverted DI pattern for HubBot

habdelra:
I tested this against my personal discord server and it's working fine


packages/hub/node-tests/services/discord-bots/hub-discord-bots-db-gateway-test.ts
```diff
@@ -21,10 +21,9 @@ describe('HubDiscordBotsDatabaseGateway', function () {
   describe('updateStatus', function () {
     it('adds a database row if none present', async function () {
       let client = await dbManager.getClient();
-      let rows = (await client.query('SELECT * FROM discord_bots')).rows;
-      expect(rows.length).to.equal(0);
+      await client.query(`DELETE FROM discord_bots`);
```
habdelra, line 24: There was state leakage after we fixed up the DI--it looked like this test assumes the DB is clean, so i made sure by removing all the rows first.

packages/hub/node-tests/services/discord-bots/hub-discord-bots-db-gateway-test.ts
```diff
@@ -34,6 +33,7 @@ describe('HubDiscordBotsDatabaseGateway', function () {
 
     it('updates the database row if already present', async function () {
       let client = await dbManager.getClient();
+      await client.query(`DELETE FROM discord_bots`);
```
habdelra, line 36: This test also needed protection from state leakage

packages/hub/node-tests/services/discord-bots/hub-discord-bots-db-gateway-test.ts
mattmcmanus: Is there something you could put in the bot service that automatically cleans up on tear down? The container will automatically call teardown() on the service if it's there (Example: https://github.com/cardstack/cardstack/blob/main/packages/db/index.ts#L37-L45)

I'm not sure though what the state is and whether you want it to be long lived....

mattmcmanus: or is this something you could put in the `beforeEach` of your test helper?

mattmcmanus (approved):
I left some comments but I don't necessarily think they're blockers.

packages/hub/node-tests/helpers/server.ts
```diff
@@ -48,3 +50,41 @@ export function setupHub(mochaContext: Mocha.Suite) {
     },
   };
 }
+
+export function setupBot(mochaContext: Mocha.Suite, setupType: TestSetupType = 'beforeEach') {
+  let container: Container;
+  let bot: HubBot;
+
+  if (setupType === 'beforeEach') {
```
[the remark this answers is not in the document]
habdelra: good idea 👍

packages/hub/node-tests/services/discord-bots/hub-discord-bots-db-gateway-test.ts
habdelra: well we are--the HubBot has a `teardown()` which hooks into the DI teardown. I did explore clearing this DB table in all the bot related tests, but that did not stop the state leaking. I'm not exactly sure what test is causing this leak.

habdelra: but in general i think it's a good pattern that any test that depends on DB state needs to initialize the DB state first. Even though the container is new for each test--that doesn't change what is in the DB.

habdelra: I did try using the DB clear in a before each, and that caused the non-transactional tests in this module to blow up because the test cleanup triggered trigger pg notifications on that the non-transactional tests are sensitive to

PR #2537, opened 2021-12-22
Footer

Aierie:
CS-2846

This footer appears on every page of the DApp except for Card Space - since there has been no clear indication of how it should appear there yet. Layout breakpoint at 560px.

## <= 560px
image

## > 560px
image


backspace:
CSS sadness perhaps:

image

but… maybe it doesn’t matter because that’s an artefact of the different rendering for Percy? 🤔🤔🤔

packages/web-client/public/images/logos/github-logo.svg
```diff
@@ -0,0 +1,13 @@
+<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 18.693 18.086">
+  <defs>
+    <clipPath id="github-logo-clippath">
+      <path data-name="Path 13101"
```
backspace, line 4: Maybe it doesn’t matter because it’s built via SVG Jar/SVGO, but I’ve run SVGs through ImageOptim (which also uses SVGO I believe) to strip these kinds of things. But it actually makes it kind of annoying to edit if styles or the like need to be changed in the future. Undecided 🤔

backspace (approved):
I feel like the Percy thing is indeed irrelevant. I approved the build there.

packages/web-client/app/components/common/footer/index.hbs
```diff
@@ -0,0 +1,48 @@
+<footer class="cardstack-footer" ...attributes>
+  <div class="cardstack-footer__section cardstack-footer__section--top">
+    <div class="cardstack-footer__important-links">
+      <a href="https://support.cardstack.com" target="_blank" rel="noopener noreferrer" >
```
jurgenwerk, line 4: Too bad Opera and Android WebView don't support it, but all major browsers [imply noopener](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#browser_compatibility) when `target="_blank"` is given. 

Additionally, it [looks that](https://html.spec.whatwg.org/multipage/links.html#link-type-noreferrer) `noreferrer` implies `noopener`.

packages/web-client/public/images/logos/github-logo.svg
Aierie: Removed these, I hope we can treat svg assets as... atomic? and not bother about the internal details.

jurgenwerk (approved):
Tested it - looks good! 👍

packages/web-client/app/components/common/footer/index.hbs
Aierie: Yeah, this is soon becoming respecifying a browser default, but was thinking we could leave rel="noopener noreferrer" in until `ember-template-lint` decides that it's out of date.

jurgenwerk: Yeah I agree. Hope it soon becomes archaic to add these attributes to external links. 😀

PR #2540, opened 2021-12-23
[CS-2849] add indexing to card service delete

mattmcmanus:
Right now we just nuke the children when you delete a card. Maybe this is not what we want in the long run, but it's consistent with how we do things when creating/updating

### DIFF
--- a/packages/hub/node-tests/routes/cards/delete-test.ts
+++ b/packages/hub/node-tests/routes/cards/delete-test.ts
@@ -28,10 +28,8 @@ if (process.env.COMPILER) {
             import { contains } from "@cardstack/types";
             import string from "https://cardstack.com/base/string";
             export default class Post {
-              @contains(string)
-              title;
-              @contains(string)
-              body;
+              @contains(string) title;
+              @contains(string) body;
             }
           `,
           'isolated.js': templateOnlyComponentTemplate('<h1><@fields.title/></h1><article><@fields.body/></article>'),
@@ -53,7 +51,7 @@ if (process.env.COMPILER) {
       await deleteCard(`${realmURL}car0`).expect(404);
     });
 
-    it.skip('can delete an existing card that has no children', async function () {
+    it('can delete an existing card that has no children', async function () {
       await getCard(`${realmURL}post0`).expect(200);
 
       await deleteCard(`${realmURL}post0`).expect(204);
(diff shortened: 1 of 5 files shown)


packages/hub/node-tests/routes/cards/delete-test.ts
```diff
@@ -66,5 +64,19 @@ if (process.env.COMPILER) {
 
       expect(existsSync(join(getRealmDir(), 'post0')), 'card is deleted from realm').to.be.false;
     });
+
+    it('can delete an existing card that has children', async function () {
+      await getCard(`${realmURL}post`).expect(200);
+      await deleteCard(`${realmURL}post`).expect(204);
+      await getCard(`${realmURL}post`).expect(404); // Deleted card is deleted!
+      await getCard(`${realmURL}post0`).expect(422); // card is in an error state
```
habdelra, line 72: yeah this was what i was mentioning while we were pairing--it seems weird that a card that is not under your control can invalidate your card, especially if it came from a different realm, given when the card being removed is included in your compiled card. this seems worth a discussion with @ef4 for how we want the system to behave.

like with the card updates, there might be an opt-in/out policy on accepting an upstream change based on the changed card being in your realm or not, or if you are in "dev" mode or not, or some other heuristic

ef4: Fixing this will probably go with the work to do finer-grained invalidation & recompilation.

Each CompiledCard has within it *all* the details of its dependencies, which means you can in theory recompile just fine even if your parent or one of your fields has been deleted. It comes down to deciding which parts of the card needs to be recompiled and which should be preserved from the previous iteration.

mattmcmanus: Agreed. Definitely needs to be discussed.

ef4 (approved):

packages/hub/node-tests/routes/cards/delete-test.ts
mattmcmanus: @habdelra We decide to keep this aggressive for now. The scenario where we want the durable index can be added onto this when we flesh out what those realm contracts should look like. Right now, we're always just in dev, so we want to be "noisy." But once we have realms on different machines where you don't have control, that's when we'll want the durability.

PR #2541, opened 2021-12-23
Fixing the standalone CLI by making ESM and CJS builds for the SDK

habdelra:
apologies for all the npm pushes--i needed to do a full circle tests of the build process which was initially not working for the standalone CLI consuming the SDK from npm.

The main takeaway from this PR is that we are using [tsup](https://tsup.egoist.sh/#what-can-it-bundle) to create ESM and CJS bundles (as well as the `index.d.ts` file) coupled with the `exports` property in the `package.json` to be able to describe to the different types of bundles. Confirmed that this works with the standalone CLI. This, in theory, should work with the metro bundler in the wallet app too (the SDK package.json `main` is pointing to the ESM entry point still to make things really simple for metro in case it doesn't understand the package.json `exports`).

I pushed a pkg for the CLI into AWS, so that should be happy now. we can hold off, though, on merging this until after the holidays so that we can test together with metro.

EDIT: it looks like according to this issue metro does not currently understand package.json "exports" https://github.com/facebook/metro/issues/670, so it will fallback to using the package.json "main" instead, which should be fine.


packages/cardpay-sdk/package.json
```diff
@@ -32,14 +39,15 @@
...
+    "tsup": "^5.11.7",
     "typescript": "4.2.3"
   },
   "files": [
     "dist"
   ],
   "scripts": {
-    "postinstall": "node bin/codegen.js",
-    "prepack": "tsc"
+    "postinstall": "if [ -f ./bin/codegen.js ]; then node bin/codegen.js; fi",
```
habdelra, line 49: @lukemelia this is an interesting gotcha--the package in npm doesn't actually include `bin/`. this script only works for maintainers of this source code, not for pkgs downloaded from npm. This actually blows up if you yarn install this package because of this file not being present

packages/cardpay-sdk/package.json
```diff
@@ -1,10 +1,17 @@
...
-  "main": "./dist/index.js",
+  "main": "./dist/esm/index.js",
   "types": "./dist/index.d.ts",
+  "exports": {
+    ".": {
+      "require": "./dist/index.js",
+      "import": "./dist/esm/index.js",
+      "types": "./dist/index.d.ts"
+    }
+  },
```
habdelra, line 14: this is the new shiny thing that node recognizes that will allow the CLI to use the CJS entrypoint. Metro hopefully can understand this, otherwise, it should be able to fallback to the `main` property which defaults to the ESM entrypoint.

habdelra: Arg--it looks like TS is mad now, because it's trying to use ./dist/esm/index.js. i mean we can allow that to happen, but that's kind of annoying because it's yet another build you have to have running. i wonder if there is a rule that will work for TS but not for metro...

habdelra: According to this, metro does not understand package.json `exports`, so this approach probably works in our favor https://github.com/facebook/metro/issues/670

lukemelia (approved):
tsup tsup and AWAY!!!! 🦸

PR #2542, opened 2021-12-23
web: Add link preview meta tags for pay routes

backspace:
This is a step toward payment request links showing appropriate
previews when shared on social media.

I tried getting Fastboot working to actually expose these for link previews and failed… but it seems worth having them here anyway?


packages/web-client/app/components/card-pay/merchant-payment-request-header/index.hbs
```diff
@@ -1,3 +1,7 @@
+<InHead>
+  {{!-- template-lint-disable no-forbidden-elements --}}
+  <meta property='og:image' content={{this.cardPayLogo}} />
```
Aierie, line 3: OG protocol doesn't support SVG, I will change this to a PNG.

Aierie: Also needs to be an absolute URL.

Aierie (commented):
Thanks for starting this, it should make payment sharing more user-friendly.

packages/web-client/app/components/card-pay/merchant-payment-request-header/index.hbs
backspace: Thanks for these fixes!

PR #2545, opened 2022-01-04
Create Statuspage incidents when Checkly sends an alert

jurgenwerk:
Ticket: [CS-2885](https://linear.app/cardstack/issue/CS-2885/automatically-create-statuspage-incidents-when-subgraph-lag-is)

This PR aims to automatically create incidents on https://status.cardstack.com/ when Checkly reports a failed check. When there is an active incident on Statuspage, the web client will display it in the banner.

Currently, this is only configured for the ["hub-prod subgraph / RPC node block number diff within threshold"](https://app.checklyhq.com/checks/5add1b2a-89c8-4f4c-8921-638b4b1643fd/) check, but we can add more.

When the above check performed by Checkly fails, it will send a webhook (`ALERT_FAILURE` alert type) to the hub. There, the hub endpoint will create an incident on Statuspage using the Statuspage API. When Checkly detects the degradation is over, it will send another webhook (`ALERT_RECOVERY` alert type) to the hub endpoint, which will resolve the incident.

There is an issue though - this whole mechanism relies on the hub being up and running. Meaning the automated incident reporting will not work in case the hub is not reachable. The most basic case where this falls apart is the "hub health check", where if the hub is down, Checkly can't use it to create an incident on Statuspage.

I'm thinking it would be great if we could host this somewhere separately from the hub. Perhaps this can be a Lambda function exposed via AWS API Gateway which Checkly can send alerts to directly, effectively removing the hub's point of failure. What do you think?

TODO:
- [ ] Configure the webhook URL in Checkly 
- [x] Implement webhook validation (for security)

### DIFF
--- a/packages/hub/node-tests/routes/checkly-webhook-test.ts
+++ b/packages/hub/node-tests/routes/checkly-webhook-test.ts
@@ -18,9 +18,9 @@ describe('POST /api/checkly-webhook', async function () {
 
   it('returns 200 when processing a check', async function () {
     await request()
-      .post('/api/checkly-webhook')
-      .set('Accept', 'application/vnd.api+json')
-      .set('Content-Type', 'application/vnd.api+json')
+      .post('/callbacks/checkly')
+      .set('Accept', 'application/json')
+      .set('Content-Type', 'application/json')
       .send({
         check_name: 'hub-prod subgraph / RPC node block number diff within threshold',
         alert_type: 'ALERT_FAILURE',
@@ -30,9 +30,9 @@ describe('POST /api/checkly-webhook', async function () {
 
   it('returns 422 when processing an unrecognized check', async function () {
     await request()
-      .post('/api/checkly-webhook')
-      .set('Accept', 'application/vnd.api+json')
-      .set('Content-Type', 'application/vnd.api+json')
+      .post('/callbacks/checkly')
+      .set('Accept', 'application/json')
+      .set('Content-Type', 'application/json')
       .send({
         check_name: 'abc123',
         alert_type: 'ALERT_FAILURE',
--- a/packages/hub/node-tests/routes/checkly-webhook-test.ts
+++ b/packages/hub/node-tests/routes/checkly-webhook-test.ts
@@ -1,0 +1,42 @@
+import { registry, setupHub } from '../helpers/server';
+
+class StubStatuspageApi {
+  async createIncident(_componentName: string, _name: string) {
+    return true;
+  }
+  async resolveIncident(_componentName: string) {
+    return true;
+  }
+}
+
+describe('POST /api/checkly-webhook', async function () {
+  this.beforeEach(function () {
+    registry(this).register('statuspage-api', StubStatuspageApi);
+  });
+
+  let { request } = setupHub(this);
+
+  it('returns 200 when processing a check', async function () {
+    await request()
+      .post('/api/checkly-webhook')
+      .set('Accept', 'application/vnd.api+json')
+      .set('Content-Type', 'application/vnd.api+json')
+      .send({
+        check_name: 'hub-prod subgraph / RPC node block number diff within threshold',
+        alert_type: 'ALERT_FAILURE',
+      })
+      .expect(200);
+  });
+
+  it('returns 422 when processing an unrecognized check', async function () {
+    await request()
+      .post('/api/checkly-webhook')
+      .set('Accept', 'application/vnd.api+json')
+      .set('Content-Type', 'application/vnd.api+json')
+      .send({
+        check_name: 'abc123',
+        alert_type: 'ALERT_FAILURE',
+      })
+      .expect(422);
+  });
+});
--- a/packages/hub/node-tests/routes/checkly-webhook-test.ts
+++ b/packages/hub/node-tests/routes/checkly-webhook-test.ts
@@ -1,22 +1,33 @@
 import { registry, setupHub } from '../helpers/server';
 
+let createdIncident: { componentName: string; message: string } | null,
+  resolvedIncident: { componentName: string } | null;
+
 class StubStatuspageApi {
-  async createIncident(_componentName: string, _name: string) {
-    return true;
+  async createIncident(componentName: string, message: string): Promise<void> {
+    createdIncident = {
+      componentName,
+      message,
+    };
   }
-  async resolveIncident(_componentName: string) {
-    return true;
+  async resolveIncident(componentName: string): Promise<void> {
+    resolvedIncident = {
+      componentName,
+    };
   }
 }
 
 describe('POST /api/checkly-webhook', async function () {
   this.beforeEach(function () {
+    createdIncident = null;
+    resolvedIncident = null;
+
     registry(this).register('statuspage-api', StubStatuspageApi);
   });
 
   let { request } = setupHub(this);
 
-  it('returns 200 when processing a check', async function () {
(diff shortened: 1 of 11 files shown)


lukemelia:
Implementing this as a service that is independent of the hub seems like a good idea. I am sensitive to over-engineering this status effort before living with the basic tools we have put in place for a little while. WDYT?

packages/hub/routes/checkly-webhook.ts
```diff
@@ -0,0 +1,52 @@
...
+  };
+};
+
+export default class ChecklyWebhookRoute {
+  statuspageApi: StatuspageApi = inject('statuspage-api', { as: 'statuspageApi' });
+
+  checks: Checks = {
+    'hub-prod subgraph / RPC node block number diff within threshold': {
+      componentName: 'Subgraph',
+      incidentMessage: 'Subgraph block number is behind RPC block number by more than 10 blocks',
```
lukemelia, line 20: Should this `10` come from config or a constant?

habdelra: i think it already does, but that was part of a different PR that monitors the subgraph block height delta. this PR is more focused on opening and closing incidents in status page.

packages/hub/services/statuspage-api.ts
```diff
@@ -0,0 +1,124 @@
...
+        components: { [componentId]: 'partial_outage' }, // Will update individual components' status
+        component_ids: [componentId], // Will associate incident with the component
+      },
+    });
+  }
+
+  async resolveIncident(componentName: string) {
+    let componentId = await this.getComponentId(componentName);
+    let unresolvedIncidents = await this.getUnresolvedIncidents();
+    let relatedIncident = this.findIncidentByComponentId(unresolvedIncidents, componentId);
```
Aierie, line 43: How should this interact with manually created incidents for the same component? It seems like right now there is a possibility that the manually created incidents could get resolved

Aierie: The case I'm wondering about is: 
1. There is a manually created incident for the "Subgraph" component, and also an automatically created one
2. The subgraph is no longer slow so hub tries to resolve the incident. The manually created incident is the one that gets resolved, while the automatically created one remains.

habdelra: i think the compromise here is that if we want to have automatically resolved incidents we need to accept the fact that we can't manually create incidents in the same component.

habdelra: But it also make me wonder what kind of incident we would create manually that wouldn't be detected by the block height delta do you forsee that we need to create

packages/hub/routes/checkly-webhook.ts
jurgenwerk: Should we adjust this message to be a bit more user friendly, and explain the possible consequences? 

The following is probably not the best way to describe this situation, but I think something like that would be more suitable. Can you help to form a better message?

For example: "We're experiencing issues with lag in the sync between the subgraph and the chain. This can result in increased time needed for transactions."

packages/hub/services/statuspage-api.ts
Aierie: That is an interesting question... I guess something like https://thegraph.com/docs/en/developer/create-subgraph-hosted/#non-fatal-errors

habdelra: hmm, we are not actually supporting non-fatal errors though. right now any kind of error will halt the indexing and trigger the block delta check.

habdelra: it seems like a large penalty to pay--but continuing to index with errors results in bad data, and that bad data can feed into other index data which will slowly taint all the data after the bad data was introduced. so failing fast and loud is an easier way for us to find and fix the problems

jurgenwerk: I just noticed there's a `metadata` field we can use to tag automatically created ones, and then only allow automatically resolving these. However this sounds like an edge case we can live with for now and see how this automatisation plays out.

backspace (approved):
I agree with @lukemelia that we can accept the known limitation of having it be dependent on Hub uptime for now and extend as needed.

Is there any possibility of having tests for this? 🤞🏻

jurgenwerk:
Also I just realized this can be abused without authorization 🙈. Will implement [webhook secrets](https://www.checklyhq.com/docs/alerting/webhooks/#webhook-secrets).

packages/hub/services/api-router.ts
```diff
@@ -38,6 +38,7 @@ export default class APIRouter {
   reservationsRoute = inject('reservations-route', { as: 'reservationsRoute' });
   inventoryRoute = inject('inventory-route', { as: 'inventoryRoute' });
   wyrePricesRoute = inject('wyre-prices-route', { as: 'wyrePricesRoute' });
+  checklyWebhookRoute = inject('checkly-webhook-route', { as: 'checklyWebhookRoute' });
```
habdelra, line 41: we actually have a route in our server for `application/json` content types specifically to support webhooks (because it is rare that webhooks are actually called with a JSON-API content type). it's the `packages/hub/services/callbacks-router.ts`. I suggest using that router instead.

packages/hub/node-tests/routes/checkly-webhook-test.ts
```diff
@@ -0,0 +1,42 @@
...
+    registry(this).register('statuspage-api', StubStatuspageApi);
+  });
+
+  let { request } = setupHub(this);
+
+  it('returns 200 when processing a check', async function () {
+    await request()
+      .post('/api/checkly-webhook')
+      .set('Accept', 'application/vnd.api+json')
+      .set('Content-Type', 'application/vnd.api+json')
```
habdelra, line 23: it would be really strange if checkly actually called us with a JSON-API content type. I'm pretty sure they use `application/json`.

packages/hub/services/api-router.ts
jurgenwerk: Thanks! Will use that.

packages/hub/routes/checkly-webhook.ts
habdelra: Sure, may something like "Experiencing blockchain indexing delay: the blockchain index delayed by at least 10 blocks". But like @lukemelia was saying instead of hard coding `10`, let's export `DEGRADED_THRESHOLD` from the `StatusRoute` and use that variable in the message.

jurgenwerk:
@habdelra this is now ready for another review - I used the callbacks router and improved the degradation message.

packages/hub/services/api-router.ts
habdelra, line 41: we can remove this route from the API router now that we added it to the callback router

habdelra (approved):
There is a minor nit were you forgot to remove the route from the API router (since we've added it to the callback router). other than that it looks good

jurgenwerk:
@habdelra thanks, forgot to commit that removal 🙈! Cool, will set the secrets in our prod deployment and get this merged and deployed.

jurgenwerk:
I set the STATUSPAGE_API_KEY, STATUSPAGE_PAGE_ID, and CHECKLY_WEBHOOK_SECRET in production.

Will get this merged and deployed now.

There is one missing step I'll be working on - configuring Checkly webhook in infra ([ticket here](https://linear.app/cardstack/issue/CS-2926/set-up-checkly-webhook-url-in-infra)).

PR #2546, opened 2022-01-04
tests for wc-provider

Aierie:
Basic tests for wc-provider, based on  https://github.com/WalletConnect/walletconnect-monorepo/blob/v1.0/packages/providers/web3-provider/test/index.spec.ts. 

The tests for ethers are currently not working - they freeze at the `.wait()` method calls (waiting for block confirmations: https://docs.ethers.io/v5/api/providers/types/#providers-TransactionResponse , and it looks like state (wallet balances etc) is not updated either... looks like ethers is not being notified of changes?

Since we don't use ethers with WalletConnectProvider at the moment, wondering if it would be ok to omit these tests and leave a note about it? 

Note: FIXMEs are from original authors

TODO:
- [ ] add CI


habdelra:
so for the ethers tests, I think you need to spin up a private blockchain for that to actually work (it's the same for web3). this is actually what we are doing in the protocol repo.

Aierie:
I had a closer look at `ethereum-test-network` - it does actually mine blocks, and under the hood it instantiates HardHat classes that create an Ethereum VM (https://github.com/nomiclabs/ethereumjs-vm)  and passes calls to it. 

However, there might be an issue with this version of HardHat + websockets:

- Websocket fix: https://github.com/nomiclabs/hardhat/pull/1780
- Applied in 2.6.1: https://github.com/nomiclabs/hardhat/releases/tag/hardhat-core-v2.6.1

Will attempt fixing tests by updating ethers as a dev dependency.

Aierie:
https://github.com/cardstack/cardstack/pull/2546/commits/0d54bdb4842bbdfacb42b6eb52c5e62325a2ebbf shows the changes necessary to get `ethereum-test-network` to work properly with websockets - tests all pass on that commit. `ethereum-test-network` is Apache-2.0. It might be good to make a fork and install from there + submit a PR to the original repo since websocket support seems pretty fundamental.

packages/wc-provider/tests/utils/ethereum-test-network.ts
```diff
@@ -0,0 +1,97 @@
...
+    };
+    const logger = new HardhatNetworkFakeModuleLogger(false);
+    const hardhatNetwork = new HardhatNetworkProvider(
+      'istanbul',
+      'WalletConnectTest',
+      params.chainId,
+      params.networkId,
+      params.blockGasLimit,
+      undefined,
+      new BN(1),
```
Aierie, line 62: The change (besides vendoring code) was just adding these 2 newly-introduced arguments to make sure that we don't mess up the arguments provided.

habdelra:
that's great to hear that the web socket updates pass on that commit. Yeah i agree, let's fork it and use that commit SHA in our package.json, as well as submitting a PR back to the upstream project

PR #2548, opened 2022-01-04
hub: Add documentation about required plugin

backspace:
I was getting a silent failure running the tunnel command until
I ran it with verbose logging, then I saw this:

```
SessionManagerPlugin is not found. Please refer to SessionManager Documentation here: http://docs.aws.amazon.com/console/systems-manager/session-manager-plugin-not-found
```

However, even with that I’m still unable to connect, so I’ll keep this a draft in case I find anything else.


backspace:
The problem I was having was that I didn’t use the private key I created for Cardstack, I’m able to connect now 🎉

lukemelia (approved):

PR #2552, opened 2022-01-05
Fix/intercept statuspage calls

Aierie:
Intercept degraded-service-detector's calls to the statuspage API to make sure the degraded service banner doesn't show and cause test failures randomly.

Am not sure if modifying the behaviour/url used for testing in the degraded-service-detector is better. Would prefer not to do that... but using mirage does mean that we should add mirage to almost all acceptance tests to prevent these sporadic failures.

### DIFF
--- a/packages/web-client/app/services/degraded-service-detector.ts
+++ b/packages/web-client/app/services/degraded-service-detector.ts
@@ -52,7 +52,7 @@ export default class DegradedServiceDetector extends Service {
       let response = await fetch(statusPageUrl);
       data = await response.json();
     } catch (e) {
-      console.error('Failed to fetch exchange rates');
+      console.error('Failed to fetch unresolved status page incidents');
       Sentry.captureException(e);
 
       return null;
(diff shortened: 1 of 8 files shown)


jurgenwerk (commented):
I think this looks ok but on the other hand we can easily forget to setupMirage in new acceptance tests 🤔 perhaps adding a conditional test environment behaviour in the degraded service detector isn't a bad option - how come you think it's not that great?

Aierie:
> I think this looks ok but on the other hand we can easily forget to setupMirage in new acceptance tests 🤔 perhaps adding a conditional test environment behaviour in the degraded service detector isn't a bad option - how come you think it's not that great?

I think if we forget to add `setupMirage` and our tests fail, that affects devs and is more easily detected and corrected with minimal impact to the user. But if we add a conditional test environment behaviour but make a mistake when editing it at some point (eg. if requirements change, or in certain situations we want to test showing the banner in an acceptance test), then that affects users and is less easily detected.

PR #2553, opened 2022-01-05
hub: Add Firebase config for deployments

backspace:
Assumption: there isn’t separate configuration for staging
and production.

I haven’t yet found and entered the values for these in the repository settings 🔍


danibonilha (approved):
Yeah, I think in order to split into stag/prod we would need to have two diff firebase projects

backspace:
maybe it doesn’t matter? I’m unsure of the implications of this. There are some environment variables that have `STAGING_` and `PRODUCTION_` prefixes, others don’t.

backspace:
@jurgenwerk do you know what the configuration values are for these? I found the project id but I don’t know where to find the others 😯

backspace:
maybe we figured it out 🤔

PR #2554, opened 2022-01-06
Prevent multiple firebase client initialization

jurgenwerk:
Been getting this error when running tests (while having `FIREBASE_PROJECT_ID` configured in the .env file). 

```
Error: The default Firebase app already exists. This means you called initializeApp() more
than once without providing an app name as the second argument. In most cases you 
only need to call initializeApp() once. But if you do want to initialize multiple apps, 
pass a second argument to initializeApp() to give each app a unique name.
```

### DIFF
--- a/packages/hub/initializers/firebase.ts
+++ b/packages/hub/initializers/firebase.ts
@@ -3,7 +3,8 @@ import admin from 'firebase-admin';
 import config from 'config';
 
 export default function initFirebase() {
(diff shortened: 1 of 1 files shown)


jurgenwerk:
It looks like this doesn't only affect tests, but also the deployment. I set the `FIREBASE_PROJECT_ID` in prod and the hub went down, I suspect because of this same issue.

Aierie (approved):

PR #2558, opened 2022-01-06
Don't close workflow modal when restarting a workflow

lukemelia:
Fixes reported bug of "odd UI when you click restart" CS-2791

Before (may need to slow down to see the modal close and immediately reopen):

https://user-images.githubusercontent.com/353/148439011-5fd1a289-0f81-4ac2-a740-c1fd9c4b9332.mov

After (the workflow restarts without the modal closing):

https://user-images.githubusercontent.com/353/148439390-c031df70-71cb-4dbb-93ae-6a5fcaee42a2.mov

### DIFF
--- a/packages/web-client/app/components/card-pay/restorable-workflow-component/index.ts
+++ b/packages/web-client/app/components/card-pay/restorable-workflow-component/index.ts
@@ -1,0 +1,41 @@
+import { Workflow } from '@cardstack/web-client/models/workflow';
+import WorkflowPersistence from '@cardstack/web-client/services/workflow-persistence';
+import { getOwner } from '@ember/application';
+import RouterService from '@ember/routing/router-service';
+import Component from '@glimmer/component';
+import { cached, tracked } from '@glimmer/tracking';
+import { inject as service } from '@ember/service';
+
+interface WorkflowConstructor<TWorkflow> {
+  new (owner?: any, workflowPersistenceId?: string): TWorkflow;
+}
+
+export default abstract class RestorableWorkflowComponent<
+  TWorkflow extends Workflow
+> extends Component {
+  @service declare workflowPersistence: WorkflowPersistence;
+  @service declare router: RouterService;
+  @tracked isInitializing = true;
+
+  get workflowPersistenceId() {
+    return this.router.currentRoute.queryParams['flow-id']!;
+  }
+
+  abstract get workflowClass(): WorkflowConstructor<TWorkflow>;
+
+  @cached
+  get workflow() {
+    const WorkflowClass = this.workflowClass;
+    return new WorkflowClass(getOwner(this), this.workflowPersistenceId);
+  }
+
+  constructor(owner: unknown, args: {}) {
+    super(owner, args);
+    this.restore();
+  }
+
+  async restore() {
+    await this.workflow.restore();
+    this.isInitializing = false;
+  }
+}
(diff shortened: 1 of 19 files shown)


packages/web-client/app/components/card-pay/restorable-workflow-component/index.ts
```diff
@@ -0,0 +1,41 @@
...
+> extends Component {
+  @service declare workflowPersistence: WorkflowPersistence;
+  @service declare router: RouterService;
+  @tracked isInitializing = true;
+
+  get workflowPersistenceId() {
+    return this.router.currentRoute.queryParams['flow-id']!;
+  }
+
+  abstract get workflowClass(): WorkflowConstructor<TWorkflow>;
```
Aierie, line 24: Just an observation:

There's an interesting (Ember?) bug here, I think? In normal typescript we shouldn't need to declare that an abstract property is a getter for it to work as a getter in components that extend this class, but here the subclasses throw errors if this is replaced with `abstract workflowClass: WorkflowConstructor<TWorkflow>`. I guess extending glimmer's `Component` has some side effects. 🤔

packages/web-client/app/services/card-customization.ts
```diff
@@ -81,8 +81,15 @@ export default class CardCustomization extends Service {
...
-      return taskFor(this.fetchCustomizationOptionsTask).perform();
+      try {
+        await taskFor(this.fetchCustomizationOptionsTask).perform();
+      } catch (e) {
+        if (!didCancel(e)) {
+          throw e;
+        }
+      }
     }
+    return Promise.resolve();
```
Aierie, line 92: We can omit this since `async` functions will return `Promise<void>` if there is no return statement.

packages/web-client/app/components/workflow-thread/index.ts
```diff
@@ -40,14 +47,15 @@ export default class WorkflowThread extends Component<WorkflowThreadArgs> {
       // if prefers-reduced-motion, don't autoscroll, and release postables as soon as available
       // this puts control on when and how to scroll in users' hands
       this.autoscroll = false;
-      this.workflow.interval = 0;
-      this.threadAnimationInterval = `0ms`;
+      this.threadAnimationInterval = 0;
+      this.cssThreadAnimationInterval = '0ms';
```
Aierie, line 51: `cssThreadAnimationInterval` seems like it can become a getter that tacks on `"ms"` to `threadAnimationInterval`, or even be removed in favour of simply concatenating within the template?

Aierie (approved):
I think this looks good, will submit a PR with some possible improvements as mentioned in comments!

packages/web-client/app/components/workflow-thread/index.ts
```diff
@@ -12,12 +12,19 @@ interface WorkflowThreadArgs {
 
 export default class WorkflowThread extends Component<WorkflowThreadArgs> {
   threadEl: HTMLElement | undefined;
-  workflow = new AnimatedWorkflow(this.args.workflow);
+  @cached
```
Aierie, line 15: Why does this getter generate an infinite loop without `@cached`?

Aierie, line 15: I didn't realise how many times this property was being read until I removed `@cached`. Did you happen to look into why it is sometimes read incessantly, @lukemelia ?

packages/web-client/app/components/card-pay/restorable-workflow-component/index.ts
lukemelia: Yes, I wrestled with this quite a bit yesterday. I don't think it is Ember-related, but could be wrong.

packages/web-client/app/components/workflow-thread/index.ts
lukemelia: I did not. I just knew that the logic wouldn't work if we were constructing new AnimatedWorkflow instances that maintain their own state.

lukemelia:
Thanks for the review and tweaks @Aierie!

PR #2559, opened 2022-01-06
web: Update Card Space next step to link to space

backspace:
This is a narrowly-focused PR to wire up the Next Steps button
after the Create Card Space workflow to forward to the newly-
created space.

Some arbitrary decisions that are lightweight enough to change
easily if/when needed:
* the workflow property name is `username`
* the route is `/card-space/username`
* the template is a placeholder that shows the username

### DIFF
--- a/packages/web-client/app/components/card-space/create-space-workflow/next-steps/index.ts
+++ b/packages/web-client/app/components/card-space/create-space-workflow/next-steps/index.ts
@@ -2,13 +2,23 @@ import Component from '@glimmer/component';
 import { action } from '@ember/object';
 import { inject as service } from '@ember/service';
 import RouterService from '@ember/routing/router-service';
+import { next } from '@ember/runloop';
 import { WorkflowCardComponentArgs } from '@cardstack/web-client/models/workflow';
 
 class CreateSpaceWorkflowNextStepsComponent extends Component<WorkflowCardComponentArgs> {
   @service declare router: RouterService;
 
   @action async visitSpace() {
-    // TODO
+    let username = this.args.workflowSession.getValue('username');
+
+    await this.router.transitionTo({
+      queryParams: { flow: null, 'flow-id': null },
+    });
+
+    next(this, () => {
+      this.router.transitionTo('card-space.userspace', { username });
+    });
+
     return;
   }
 }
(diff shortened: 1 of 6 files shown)


packages/web-client/app/components/card-space/create-space-workflow/next-steps/index.ts
```diff
@@ -2,13 +2,23 @@ import Component from '@glimmer/component';
...
 class CreateSpaceWorkflowNextStepsComponent extends Component<WorkflowCardComponentArgs> {
   @service declare router: RouterService;
 
   @action async visitSpace() {
-    // TODO
+    let username = this.args.workflowSession.getValue('username');
+
+    await this.router.transitionTo({
+      queryParams: { flow: null, 'flow-id': null },
+    });
```
lukemelia, line 16: Why do we need to do this transition before the card-space.username transition?

Aierie: The workflow will still be open if we don't have this transition.

image

Card Space's creation workflow, and its trigger button, currently live in `card-space.hbs`. I think Card Space's creation workflow probably could be more cleanly separated from a single user's page (by creating a separate template and putting the creation workflow + trigger button there, in a separate PR?) so that this does not happen. Another effect of this is shown below - user space is shown next to the creation button:

image

Aierie (approved):

packages/web-client/app/components/card-space/create-space-workflow/next-steps/index.ts
backspace: maybe this narrow focus is _too_ narrow, since the ultimate goal would have the button lead to `username.card.space`. Which I don’t believe can be accomplished in an acceptance test. I may change it to be a link instead of a button with an action.

backspace: but now I’m wondering if some kind of wild URL-rewriting would be helpful for this, then there wouldn’t have to be multiple deployments. I’ll check that out when I get access to Cloudflare.

lukemelia: Sounds like the modal/workflow should be in the card space index route (card-space/index.hbs)

PR #2567, opened 2022-01-10
Consolidate business/balances cards in Business and Deposit&Withdrawal routes

jurgenwerk:
Ticket: [CS-2890](https://linear.app/cardstack/issue/CS-2890/leverage-business-account-form-from-depositwithdrawal-page-on)

This PR consolidates the "business&balances" components from both routes, and uses the same component to display depot and merchant balances info. 

Business route:

image

Deposit & Withdrawal: 

image


packages/web-client/app/templates/card-pay/payments.hbs
```diff
@@ -69,11 +69,7 @@
       </header>
       <section class="card-pay-dashboard__payments-cards">
         {{#each this.merchantSafes as |merchantSafe|}}
-          <CardPay::MerchantSafe
-            @safe={{merchantSafe}}
-            @waitForCustomization={{true}}
-            data-test-merchant-dashboard-card
-          />
+          <CardPay::SafeBalances @safe={{merchantSafe}} />
```
jurgenwerk, line 72: This component is now used in both /payments and /deposit-withdrawal routes. 

Now that this component contains the safe info, together with balances, would it make sense to rename it to `CardPay::Safe` (and all the CSS that goes with it)?

jurgenwerk:
What I'm not sure about is if we should be displaying the safe footer in the Depot ... Will ask Kyle

image

Aierie (changes requested):
I think this works! Besides the question you've raised earlier, I think it might be good to clarify if the "footer" is meant to be placed below the business' logo with this new change, since that is the order that it appears in the spec and it seems to make sense to put a business' identifiers together. 

Requested change: `CardPay::MerchantSafe` is no longer used + references the removed component `CardPay::MerchantCard` and should be removed.

Aierie:
> What I'm not sure about is if we should be displaying the safe footer in the Depot ... I asked Kyle
> 
> image

Is it useful to the user to show "1 Manager" for either of these? It feels like it isn't.

packages/web-client/app/templates/card-pay/payments.hbs
lukemelia: I like that idea!

lukemelia:
> Is it useful to the user to show "1 Manager" for either of these? It feels like it isn't.

Not useful in either case right now, but in the case of the Depot it doesn't make any sense period, and in the case of the Business, it may later when we support multiple managers.

jurgenwerk:
@Aierie thanks for pointing out the dead code, I removed it. Regarding whether the footer should really be a footer vs. being in the middle (as in the original spec) - I thought that placing it on the bottom looks better, and Kyle commented in the ticket it looks ok. So I think we can go with this.

@lukemelia thanks for the tip, I removed that line from the depot safe but kept it in others. Let me know if that isn't ok.

Aierie (approved):
Looks good! Thanks for checking w design 😄

PR #2569, opened 2022-01-10
web: Add preliminary Card Space subdomain routing

backspace:
This updates the Create Card Space next step button to link to a subdomain instead of using Ember routing. It also hacks the index route to render the Card Space userspace template when the hostname matches a Card Space suffix, which seems impossible to cover within Ember’s testing facilities.

The Card Space rendered is bare bones but can be exercised with an edit to `/etc/hosts`:

image

The goal is to get something minimal in place to then exercise with the necessary DNS setup.


Aierie (approved):
Is it worth it testing the conditional rendering by having a service determine whether the condition is true, making it mockable?

backspace:
yes, I thought of this last night, twinsies! 😆

PR #2572, opened 2022-01-11
Reduce webpack warnings

Aierie:
CS-2823


packages/cardpay-sdk/.eslintrc.cjs
```diff
@@ -4,4 +4,12 @@ module.exports = {
...
     tsconfigRootDir: __dirname,
   },
+  overrides: [
+    {
+      files: ['*.ts'],
+      rules: {
+        '@typescript-eslint/consistent-type-exports': 'error',
+      },
+    },
+  ],
```
Aierie, line 14: Am not sure about using this rule vs ignoring warnings. If we choose this fix, should this be a repo-wide change?

Aierie: This rule makes sure that type exports are specified using the type-only syntax, introduced in Typescript 3.8.

habdelra: I don't have any strong feelings on this change. Maybe @ef4 would like to weigh in

packages/web-client/ember-cli-build.js
```diff
@@ -80,6 +81,49 @@ module.exports = function (defaults) {
...
+          },
+          {
+            /**
+             * ember-cli-mirage made a mistake here I think?
+             * _DbCollection exists, but not _dbCollection
+             */
+            module: /node_modules\/ember-cli-mirage/,
+            message:
+              /export '_dbCollection' \(reexported as 'default'\) was not found in /,
+          },
```
ef4, line 125: This and the ones above it are real bugs in these libraries. Because the classic ember-cli build doesn't really follow the ES module spec strictly, people don't notice until they switch to Embroider. 

Technically in Javascript, reexporting a thing that doesn't exist is supposed to be a SyntaxError. Webpack is letting these slip by as errors because they're unused, so it doesn't actually break anything.

ef4 (approved):

packages/cardpay-sdk/.eslintrc.cjs
Aierie: > I think it's a good rule, we could upstream it into the shared config too.

Cool, will follow up with a PR for that. Thanks!

PR #2573, opened 2022-01-11
Working on computed & LinkedTo

ef4:
- [x] make the split implementations of load vs loadData, etc, actually optimized so the data ones aren't doing unnecessary work
 - [x] since CardModel isn't going to incorporate user-authored code from schema.js, it makes sense to not compiled `extends CardModel` into the component modules, instead they can just `export const serializationMap = ...`.
 - [ ] we're moving toward replacing CardContent with CardModel, which will involve some refactoring of CardModel because today it's very client-oriented.
 - [ ] we plan to run computeds in the compiled schemaModule (which exists once per card) not CardModel (which exists per format and doesn't always even need to load the computed runner code for a card.


ef4:
Next step refactors:

 - make the split implementations of load vs loadData, etc, actually optimized so the data ones aren't doing unnecessary work
 - we plan to run computeds in the compiled schemaModule (which exists once per card) not CardModel (which exists per format and doesn't always even need to load the computed runner code for a card.
 - since CardModel isn't going to incorporate user-authored code from schema.js, it makes sense to not compiled `extends CardModel` into the component modules, instead they can just `export const serializationMap = ...`.
 - we're moving toward replacing CardContent with CardModel, which will involve some refactoring of CardModel because today it's very client-oriented.

ef4:
Things we've identified about the API we need for CardModel (unified with CardContent):

 - serialize and deserialize to database
 - serialize and deserialize to network (this needs to be aware of the saved vs unsaved distinction so that a missing field in the PATCH is interpreted as "not changed" vs in a POST where it means "undefined")
 - CardContent doesn't have an unsaved state right now, but CardModel does and that's important to keep
 - for data create, hub/card-routes would use network deserialize to get CardModel<Unsaved>, and the hub card service would have `createData(card: CardModel<Unsaved>): Promise<CardModel<Saved>` (this would unify with the pattern that already exists on the client)
 - for data update, hub/card-routes would use network deserialize to get a  `CardModel<Saved>` that knows its a patch and that it doesn't imply anything about not-included fields
 - when we're running computeds for a CardModel, if we encounter a request for a field that wasn't loaded with the CardModel, at that point we load *all* remaining fields (being careful not to stomp the fields you had already loaded with newer state that might exists on the server)

ef4:
Next step is compiling schema.js into runnable code. Sketch of how that will work:

```js
// before
import { contains } from "@cardstack/types";
import string from "https://cardstack.com/base/string";
import bio from "https://demo/cards/bio";

export default class Person {
  @contains(string) lastName;
  @contains(bio) aboutMe;

  @contains(string)
  async fullName() {
    return "Mr or Mrs " + (await this.lastName());
  }
}
```


```js
// after
import BioClass from "@cardstack/compiled/https-demo/cards/bio/schema.js"

export default class Person {
  // types for illustration purposes only, this won't be TypeScript
  constructor(get: { getRawField: (name: string) => Promise<any> }) {
    this.#getRawField = get;
   }

   // primitive field implementation
   async lastName() {
     return await this.#getRawField('lastName');
   }

    // composite field implementation
   async aboutMe() {
      return new BioClass((innerField) => this.#getRawField("aboutMe." + innerField) )
   }

   // computed field implementation
   async fullName() {
      return "Mr or Mrs " + (await this.lastName());
   }
}

```

habdelra:
Also, the items identified in Ed's review commit are also good things to work on too in terms of separable tasks: https://github.com/cardstack/cardstack/pull/2573/commits/5d543dbf4c2208e4ea2e45f01d6586a84dff9082

packages/core/src/babel-plugin-card-schema-transform.ts
```diff
@@ -131,6 +131,9 @@ export default function main(babel: typeof Babel) {
         // @babel/plugin-proposal-class-properties creates a "Class" visitor
         // that you need to hook into.
         for (let bodyItem of path.get('body').get('body')) {
+          if (type === 'primitive') {
+            throw error(bodyItem, `primitive cards cannot have fields`);
+          }
```
habdelra, line 136: we should perform this check before we try to transpile the schema, and the error that we throw should describe the problem. i think packages/core/src/compiler.getParentCard() is probably the best place to check for this scenario and throw a meaningful error

habdelra:
As we discussed, since the tests are green and we are current with master, merging this branch to prevent it from getting any larger. We'll continue the work in `links-to-one-cont` branch

PR #2575, opened 2022-01-11
Improve documentation around using waypoint exec

lukemelia:


packages/hub/README.md
```diff
@@ -155,6 +155,15 @@ Hub > const workerClient = await container.lookup('worker-client');
 Hub > await workerClient.addJob('persist-off-chain-merchant-info', { id: 1 });
 ```
 
+To connect to a console in a remote environment (staging/production), use waypoint:
+
+```sh
+waypoint exec
```
backspace, line 161: Is `-app hub` needed? Without I got an error that I needed an app because there’s more than one 🤔

lukemelia: Whoops, good catch!

PR #2578, opened 2022-01-11
Adding container health checks in order to prevent 503 errors during hub rollover

habdelra:
According to the documentation here https://garbe.io/blog/2020/03/04/deep-dive-ecs-deployment/ a healthcheck should exist in the container along with the ELB. We currently have a healthcheck in our ELB, but we don't have container healthchecks. The hope is that this will prevent us from seeing 503's when waypoint performs the hub rollover


packages/hub/package.json
```diff
@@ -15,9 +15,9 @@
     "rebuild": "webpack --watch --progress",
     "start": "run-p -c start:*",
     "start:server": "node --no-deprecation -r source-map-support/register dist/hub.js server",
-    "start:worker": "node --no-deprecation -r source-map-support/register dist/hub.js worker",
-    "start:bot": "node --no-deprecation -r source-map-support/register dist/hub.js bot",
-    "start:event-listener": "node --no-deprecation -r source-map-support/register dist/hub.js event-listener",
+    "start:worker": "HEALTH_CHECK_PORT=3001 node --no-deprecation -r source-map-support/register dist/hub.js worker",
+    "start:bot": "HEALTH_CHECK_PORT=3002 node --no-deprecation -r source-map-support/register dist/hub.js bot",
+    "start:event-listener": "HEALTH_CHECK_PORT=3002 node --no-deprecation -r source-map-support/register dist/hub.js event-listener",
```
habdelra, line 20: we set different ports for the health checks so they don't collide if you want to run all the things at once

lukemelia (approved):
👍  Very nice. This makes sense and I hope it does the trick!

PR #2579, opened 2022-01-12
Json api upload

Aierie:
CS-2713

### DIFF
--- a/packages/hub/routes/upload.ts
+++ b/packages/hub/routes/upload.ts
@@ -84,8 +84,18 @@ export default class UploadRouter {
           ownerAddress: '0x0',
         });
 
-        ctx.body = url;
-        ctx.status = 200;
+        ctx.body = {
+          data: {
+            type: 'uploaded-asset',
+            id: cid,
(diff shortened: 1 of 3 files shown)


packages/hub/routes/upload.ts
```diff
@@ -84,8 +84,18 @@ export default class UploadRouter {
...
+            type: 'uploaded-asset',
+            id: cid,
+            attributes: {
+              url,
+            },
+          },
+        };
+        ctx.append('Location', url);
+        ctx.type = 'application/vnd.api+json';
+        ctx.status = 201;
```
Aierie, line 98: This follows: https://jsonapi.org/format/#crud-creating

jurgenwerk: I'm fine with it being the CID. We can also just drop the id, like we do here: https://github.com/cardstack/cardstack/blob/main/packages/hub/services/serializers/notification-preference-serializer.ts#L9. I'm fine with both.

jurgenwerk (approved):

PR #2580, opened 2022-01-12
report web3-socket events to sentry

Aierie:
CS-2683

This lets us know if our websocket fails/disconnects for some reason. The tested service failing currently means that the contract subscription handler will not work and no push notifications for merchant payments + claims will be sent until it is back up, which probably needs manual intervention right now.

Will also add an alert to Sentry.

### DIFF
--- a/packages/hub/node-tests/services/web3-socket-test.ts
+++ b/packages/hub/node-tests/services/web3-socket-test.ts
@@ -7,25 +7,40 @@ const { testkit, sentryTransport } = sentryTestkit();
 const DUMMY_DSN = 'https://acacaeaccacacacabcaacdacdacadaca@sentry.io/000001';
 
 describe('Web3Socket', function () {
-  this.beforeEach(function () {
+  let subject: Web3SocketService;
+  this.beforeAll(function () {
     Sentry.init({
       dsn: DUMMY_DSN,
       release: 'test',
       tracesSampleRate: 1,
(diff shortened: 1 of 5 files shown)


packages/hub/node-tests/services/web3-socket-test.ts
```diff
@@ -2,30 +2,115 @@ import Web3SocketService from '../../services/web3-socket';
...
+    httpServer.listen(8545);
+    wsServer = new WebsocketServer({
+      httpServer,
+      autoAcceptConnections: true,
+    });
+    wsServer.on('request', function (request: any) {
+      request.accept('echo-protocol', request.origin);
+    });
+    subject = new Web3SocketService();
+    subject.rpcURL = 'ws://localhost:8545';
```
backspace, line 29: is there any possibility this port will be taken? I’ve sometimes seen test suites ask for a random unassigned port vs hard-coding one 🤔

backspace (approved):
It’ll be good to actually be alerted to this, hopefully we can figure out how to automate reacting to it!

packages/hub/node-tests/services/web3-socket-test.ts
Aierie: Oh, that might be good! though I'm not very sure how to do that. I picked this because it's the default hardhat (private blockchain, with js implementations: https://hardhat.org/) ws url.

PR #2584, opened 2022-01-13
Pick username card

Aierie:
Adds Card Space username (display name) card. Video shows me entering a display name into the card, and adding a profile image ("display name" field label is not true for this PR, it reads "user name" as of latest commit. there are substantial changes to make to change username -> display name, so that will be in a separate PR):

https://user-images.githubusercontent.com/39579264/149521419-7358377a-30fb-4fc6-8706-877e5f24acca.mov

### DIFF
--- a/packages/web-client/tests/acceptance/create-card-space-test.ts
+++ b/packages/web-client/tests/acceptance/create-card-space-test.ts
@@ -3,12 +3,12 @@ import { setupApplicationTest } from 'ember-qunit';
 import {
   click,
   currentURL,
+  fillIn,
   settled,
   visit,
   waitFor,
 } from '@ember/test-helpers';
 import percySnapshot from '@percy/ember';
-import config from '@cardstack/web-client/config/environment';
 import Layer2TestWeb3Strategy from '@cardstack/web-client/utils/web3-strategies/test-layer2';
 import { currentNetworkDisplayInfo as c } from '@cardstack/web-client/utils/web3-strategies/network-display-info';
 import { setupHubAuthenticationToken } from '../helpers/setup';
@@ -18,8 +18,6 @@ import {
   createSafeToken,
 } from '@cardstack/web-client/utils/test-factories';
 import { setupMirage } from 'ember-cli-mirage/test-support';
-import { LocationService } from '@cardstack/web-client/services/location';
-import Service from '@ember/service';
 
 function postableSel(milestoneIndex: number, postableIndex: number): string {
   return `[data-test-milestone="${milestoneIndex}"][data-test-postable="${postableIndex}"]`;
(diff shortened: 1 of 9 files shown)


packages/web-client/tests/acceptance/create-card-space-test.ts
```diff
@@ -177,10 +171,12 @@ module('Acceptance | create card space', function (hooks) {
...
-    let spaceHostname = `usernametodo.${config.cardSpaceHostnameSuffix}`;
-    assert
-      .dom('[data-test-card-space-next-step="visit-space"]')
-      .hasAttribute('href', new RegExp(spaceHostname.replace(/\./g, '\\.')));
+    // TODO: fix this assertion after we have fixed the subdomain/business ID
+    // things
+    // let spaceHostname = `usernametodo.${config.cardSpaceHostnameSuffix}`;
+    // assert
+    //   .dom('[data-test-card-space-next-step="visit-space"]')
+    //   .hasAttribute('href', new RegExp(spaceHostname.replace(/\./g, '\\.')));
```
Aierie, line 179: Left this out for now, to be restored after we decide on how we deal with subdomain.

jurgenwerk:
A couple of comments:

1. We should make "User name" and "username" consistent
2. I think we discussed this before but "user name" sounds like some kind of an ID, something that is not easily changeable, while in this case, it's used for display purposes. I think we should rename this to "Display name" to avoid any confusion
3. Not sure why, but the 1Password icon shows up in the username input. I guess this is another problem due to the name of the attribute
image
 4. Looks like there are issues with the upload. I'm getting this:
 
image
image

... while there are no errors raised in the hub, and the image was actually uploaded successfully in my web3 storage account. 

5. Is it easy to add the zoom functionality to the upload preview component? I think it would be very convenient. I see cropper has zoom buttons too, maybe we can add those?

Aierie:
> We should make "User name" and "username" consistent
I think we discussed this before but "user name" sounds like some kind of an ID, something that is not easily changeable, while in this case, it's used for display purposes. I think we should rename this to "Display name" to avoid any confusion

Yep, am leaving changing copy to another PR, since the term "user name" and variable names derived from it are widespread in Card Space. 

> Looks like there are issues with the upload. 

I think that might be an issue with hub build. If you didn't rebuild, you might not get JSON https://github.com/cardstack/cardstack/pull/2579, and instead would be getting a string `http://....`

> Not sure why, but the 1Password icon shows up in the username input. I guess this is another problem due to the name of the attribute

Thanks! Will modify some HTML to try and avoid this case.

> Is it easy to add the zoom functionality to the upload preview component? I think it would be very convenient. I see cropper has zoom buttons too, maybe we can add those?

I don't think it's too hard, and I agree it's probably quite convenient for users. I can open a separate PR for this, since I'm guessing that it's not too hard, but am not certain.

packages/web-client/tests/integration/components/card-space/create-space-workflow/username-test.ts
```diff
@@ -0,0 +1,249 @@
...
+          @workflowSession={{this.session}}
+          @onComplete={{this.onComplete}}
+          @onIncomplete={{this.onIncomplete}}
+          @isComplete={{this.isComplete}}
+        />
+      `);
+
+      // this opens the image editor. users would normally trigger this
+      // by clicking the image uploader
+      await mockPngUpload(sampleImage, IMAGE_UPLOADER);
```
backspace, line 196: I appreciate that you put in the significant-seeming effort to make this look simple in the test! 🤩

backspace (approved):
This worked for me locally 🥳

PR #2586, opened 2022-01-13
hub: Add GET /api/card-spaces/:slug

backspace:
This adds a route to fetch Card Space information by the slug of
the related merchant. It also returns the merchant details in
the response.


backspace:
This is perhaps ready BUT there’s an ambiguity we should decide about; maybe d67a75a, which changes it from `GET /api/card-spaces/user` to `GET /api/card-spaces/user.card.space`, can be removed.

When a Card Space is created, where will the domain used in the `url` field come from? Does Hub need to know the Card Space domain? At the moment, `POST /api/card-spaces` allows the client to set the `url`, which includes a domain, but I don’t think we should allow that. Maybe we should change it from `url` to `username` and ignore domains altogether?

Thoughts, @cardstack/web-client?

jurgenwerk:
@backspace now that we're leaning towards using the merchant's slug to act as the card space subdomain, I think it makes sense that we should drop the `url` in `POST /api/card-spaces`, and include the merchant's slug instead? That would be the  `username` you're mentioning, right?

backspace:
ya… what do you think of not having the slug stored in two places, though? Maybe it’s annoying to have to check two tables: `username.card.space` ➡️ `0x1234` (merchant address from merchants table) ➡️ Card Space details for `0x1234` (from Card Spaces table)

If we have the slug stored in both the merchants and Card Spaces tables, we would have to keep it in sync if we end up allowing people to change their slug.

But… that’s not currently possible. So maybe having it in both places is fine! 😆 That’s what I’ll proceed with, we can change later if we need to.

jurgenwerk:
@backspace I'm fine with making a join when doing a lookup. However keeping the same slugs in both tables makes me a bit uneasy ☹️ Maybe we can consider dropping the url field from card spaces table, and add a foreign key to the merchant record?

backspace:
okay, join it is then! I do have the uneasiness re duplicate slugs too.

backspace:
Does [this commit](https://github.com/cardstack/cardstack/pull/2595/commits/ef608ac431a7d895e0ee6d5d51941813cdf540ef) look suitable for the database changes, @jurgenwerk? I’ll address the various test and code changes now.

jurgenwerk:
Yes that looks ok to me! 👍

backspace:
this is p outdated based on new thoughts but something noteworthy is that this should return the owner address for the merchant, or perhaps embed the entire merchant? then web client can know whether the current user is the owner and has edit privileges

Aierie (approved):

PR #2587, opened 2022-01-13
Update hub start script to build first if the dist dir is not present

lukemelia:
I ran into a situation where running `yarn start:hub` was failing because I had not previously run `yarn build` in the hub dir. This PR avoids this failure.


Aierie (approved):
Thanks, this seems convenient!

PR #2594, opened 2022-01-18
Add business account selector to card space workflow

jurgenwerk:
Ticket: [CS-1981](https://linear.app/cardstack/issue/CS-1981/updates-to-card-space-setup-workflow-if-user-already-has-a-merchant)

Adds a business account selection step in the create card space workflow.

Edit:
image

Materialized:
image

User didn't create a merchant:
image

### DIFF
--- a/packages/hub/services/queries/merchant-info.ts
+++ b/packages/hub/services/queries/merchant-info.ts
@@ -3,9 +3,13 @@ import { inject } from '@cardstack/di';
 import { MerchantInfo } from '../../routes/merchant-infos';
 import { buildConditions } from '../../utils/queries';
 
-interface MerchantInfoQueriesFilter {
+export interface MerchantInfoQueriesFilter {
   id?: string;
   slug?: string;
+  ownerAddress?: string;
+  customFilter?: {
+    availableForCardSpace?: boolean;
+  };
 }
 
 export default class MerchantInfoQueries {
@@ -14,10 +18,15 @@ export default class MerchantInfoQueries {
   async fetch(filter: MerchantInfoQueriesFilter): Promise<MerchantInfo[]> {
     let db = await this.databaseManager.getClient();
 
-    const conditions = buildConditions(filter);
+    let conditions = buildConditions(filter);
+    let query = `SELECT id, name, slug, color, text_color, owner_address, created_at from merchant_infos WHERE ${conditions.where}`;
 
-    const query = `SELECT id, name, slug, color, text_color, owner_address, created_at from merchant_infos WHERE ${conditions.where}`;
-    const queryResult = await db.query(query, conditions.values);
+    if (filter.customFilter?.availableForCardSpace) {
+      // Merchants that haven't been used in a card space yet
+      query += ' AND merchant_infos.id NOT IN (SELECT merchant_id FROM card_spaces WHERE merchant_id IS NOT NULL)';
+    }
+
+    let queryResult = await db.query(query, conditions.values);
 
     return queryResult.rows.map((row) => {
       return {
--- a/packages/hub/utils/queries.ts
+++ b/packages/hub/utils/queries.ts
@@ -9,10 +9,17 @@ const camelToSnakeCase = (str: string) => str.replace(/[A-Z]/g, (letter) => `_${
 // => { where: 'name=$1 AND age=$2', values: [ 'Phil', 45 ] }
 
 export function buildConditions(params: any) {
-  let nonNullParams = pickBy(params, function (value) {
+  // Only allow nulls, strings and numbers in the params
+  let filteredParams = Object.fromEntries(
+    Object.entries(params).filter(
+      ([_key, value]) => value == null || typeof value === 'number' || typeof value === 'string'
+    )
+  );
+
+  let nonNullParams = pickBy(filteredParams, function (value) {
     return value !== null;
   });
-  let nullParams = pickBy(params, function (value) {
+  let nullParams = pickBy(filteredParams, function (value) {
     return value === null;
   });
 
--- a/packages/web-client/app/components/card-pay/hub-authentication/index.hbs
+++ b/packages/web-client/app/components/card-pay/hub-authentication/index.hbs
@@ -54,7 +54,7 @@
       </i.ActionStatusArea>
     </:in-progress>
     <:memorialized as |m|>
-      <m.ActionStatusArea data-test-memorialized>
+      <m.ActionStatusArea data-test-authentication-memorialized>
         Authenticated with Hub
       </m.ActionStatusArea>
     </:memorialized>
--- a/packages/web-client/app/components/card-pay/hub-authentication/index.hbs
+++ b/packages/web-client/app/components/card-pay/hub-authentication/index.hbs
@@ -1,5 +1,6 @@
 <ActionCardContainer
   @isComplete={{@isComplete}}
+  {{did-insert this.checkIfAuthenticated}}
 >
   <Boxel::ActionChin
     @state={{this.authState}}
@@ -53,7 +54,7 @@
       </i.ActionStatusArea>
     </:in-progress>
     <:memorialized as |m|>
-      <m.ActionStatusArea>
+      <m.ActionStatusArea data-test-memorialized>
         Authenticated with Hub
       </m.ActionStatusArea>
     </:memorialized>
--- a/packages/web-client/app/components/card-pay/safe-chooser-dropdown/safe-option/index.hbs
+++ b/packages/web-client/app/components/card-pay/safe-chooser-dropdown/safe-option/index.hbs
@@ -9,11 +9,18 @@
       @logoTextColor={{this.data.info.textColor}}
     />
   {{/if}}
-  <div class='safe-option__name-and-type'>
-    <span class='safe-option__name'>{{this.data.info.name}}</span>
-    {{#if this.data.type}}
-      <span class='safe-option__type'>{{this.data.type}}</span>
-    {{/if}}
-  </div>
-  <div class='safe-option__address'>{{@safe.address}}</div>
+  {{#if @showSafeId}}
+    <div class='safe-option__name-and-type'>
+      <span class='safe-option__name'>{{this.data.info.name}}</span>
+    </div>
+    <div class='safe-option__id'>{{this.data.info.id}}</div>
+  {{else}}
+    <div class='safe-option__name-and-type'>
+      <span class='safe-option__name'>{{this.data.info.name}}</span>
+      {{#if this.data.type}}
+        <span class='safe-option__type'>{{this.data.type}}</span>
+      {{/if}}
+    </div>
+    <div class='safe-option__address'>{{@safe.address}}</div>
+  {{/if}}
 </div>
--- a/packages/web-client/app/components/card-space/create-space-workflow/index.ts
+++ b/packages/web-client/app/components/card-space/create-space-workflow/index.ts
@@ -18,6 +18,8 @@ import {
 } from '@cardstack/web-client/models/workflow';
 import { standardCancelationPostables } from '@cardstack/web-client/models/workflow/cancelation-helpers';
 import RestorableWorkflowComponent from '../../card-pay/restorable-workflow-component';
+import MerchantInfoService from '@cardstack/web-client/services/merchant-info';
+import { taskFor } from 'ember-concurrency-ts';
 
 const FAILURE_REASONS = {
   L2_DISCONNECTED: 'L2_DISCONNECTED',
@@ -28,6 +30,9 @@ const FAILURE_REASONS = {
   RESTORATION_UNAUTHENTICATED: 'RESTORATION_UNAUTHENTICATED',
   NO_BUSINESS_ACCOUNT: 'NO_BUSINESS_ACCOUNT',
   RESTORATION_NO_BUSINESS_ACCOUNT: 'RESTORATION_NO_BUSINESS_ACCOUNT',
+  ALL_BUSINESS_ACCOUNTS_TAKEN: 'ALL_BUSINESS_ACCOUNTS_TAKEN',
+  RESTORATION_ALL_BUSINESS_ACCOUNTS_TAKEN:
+    'RESTORATION_ALL_BUSINESS_ACCOUNTS_TAKEN',
 } as const;
 
 export const MILESTONE_TITLES = [
@@ -44,6 +49,7 @@ class CreateSpaceWorkflow extends Workflow {
   @service declare router: RouterService;
   @service declare layer2Network: Layer2Network;
   @service declare hubAuthentication: HubAuthentication;
+  @service declare merchantInfo: MerchantInfoService;
 
   name: WorkflowName = 'CARD_SPACE_CREATION';
   version = WORKFLOW_VERSION;
@@ -55,6 +61,7 @@ class CreateSpaceWorkflow extends Workflow {
         new WorkflowMessage({
           message: `Hello, welcome to Card Space, we're happy to see you!`,
         }),
+
         new NetworkAwareWorkflowMessage({
           message: `Looks like you’ve already connected your ${c.layer2.fullName} wallet, which you can see below.
           Please continue with the next step of this workflow.`,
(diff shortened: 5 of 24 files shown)


packages/web-client/app/components/card-pay/safe-chooser-dropdown/safe-option/index.hbs
```diff
@@ -9,11 +9,18 @@
...
-  <div class='safe-option__name-and-type'>
-    <span class='safe-option__name'>{{this.data.info.name}}</span>
-    {{#if this.data.type}}
-      <span class='safe-option__type'>{{this.data.type}}</span>
-    {{/if}}
-  </div>
-  <div class='safe-option__address'>{{@safe.address}}</div>
-</div>
\ No newline at end of file
+  {{#if @useUrl}}
```
jurgenwerk, line 12: I'm unsure about this but I think it could be convenient to show the card space url here maybe?

Default:
image

With url option:
image

packages/web-client/app/components/common/forms/index.css
```diff
@@ -0,0 +1,6 @@
+.form-input-info {
```
jurgenwerk, line 1: I feel like this is something that can be reusable so I put it under common. If boxel had a dropdown then I guess it would make sense to add "info text" support there but since it doesn't I think we can have this here? 

image

packages/web-client/app/components/card-space/create-space-workflow/select-business-account/index.ts
```diff
@@ -0,0 +1,72 @@
...
+import OffChainJsonService from '@cardstack/web-client/services/off-chain-json';
+import { MerchantInfo } from '@cardstack/web-client/resources/merchant-info';
+import { useResource } from 'ember-resources';
+
+class CreateSpaceWorkflowSelectBusiness extends Component<WorkflowCardComponentArgs> {
+  @service declare layer2Network: Layer2Network;
+  @tracked merchantSafes: MerchantSafe[] | null = null;
+  @tracked selectedSafe: MerchantSafe | null = null;
+  @service declare router: RouterService;
+  @service declare offChainJson: OffChainJsonService;
```
jurgenwerk, line 18: Note to myself - delete this, it's not needed anymore

packages/web-client/app/components/card-space/create-space-workflow/select-business-account/index.hbs
```diff
@@ -0,0 +1,92 @@
+<div {{did-insert this.setupBusinessAccountSelection}}>
+  {{#if (eq this.merchantSafes.length 0)}}
+    <ActionCardContainer
+      @header="Business account"
+      @isComplete={{@isComplete}}
+    >
+      <ActionCardContainer::Section>
+        <p>In order to create your Card Space, you must first create your first a business account. This is required
```
jurgenwerk, line 8: Should this be a `WorkflowMessage` message instead? 🤔

packages/web-client/app/components/card-pay/safe-chooser-dropdown/safe-option/index.hbs
backspace: agreed! Much more intelligible

packages/web-client/app/components/card-space/create-space-workflow/select-business-account/index.hbs
backspace: I imagined this working akin to how merchant creation [bounces you](https://github.com/cardstack/cardstack/blob/09351f488a9112357583191bcc7f0a23374e071b/packages/web-client/app/components/card-pay/create-merchant-workflow/index.ts#L224-L238) if you don’t have a prepaid card or prepaid card issuance [bounces you](https://github.com/cardstack/cardstack/blob/09351f488a9112357583191bcc7f0a23374e071b/packages/web-client/app/components/card-pay/issue-prepaid-card-workflow/index.ts#L232-L241) if you don’t have a sufficient balance, maybe that pattern would work here too?

backspace:
This is looking promising so far! 🤩

packages/web-client/app/components/card-space/create-space-workflow/select-business-account/index.hbs
jurgenwerk: Ah, I sensed there is a way where we already do similar things but couldn't remember where. Thanks!

packages/web-client/app/components/card-pay/safe-chooser-dropdown/safe-option/index.hbs
Aierie: I think showing the business ID makes sense, over showing the safe address. I'm not sure if we can or should show the card space url, since users who create business accounts may not have created card spaces yet, and may not understand the concept. The identity of a merchant seems somewhat separate from a card space it owns.

Aierie: Tying a business' display name to a card space URL might confuse a user since a Card Space has its own changeable display name.

jurgenwerk: Thanks @Aierie. This solution does bring additional confusion (ie. "are all my businesses already accessible with this URL"?)

I'm thinking about this design - just the name and id, and the url below:

image

Thoughts?

Aierie: I think that does look good, and I think these changes could make the withdrawal dropdown more user-friendly as well!

backspace: agreed… I wonder if it would be good to add a faint truncated address anyway…??? But I don’t really know how to imagine actual usage 🤔

jurgenwerk: Possibly! I'll ask in the design channel about that, but for now I think we can go with this

jurgenwerk:
@backspace @Aierie I think this is now ready for another review. I also asked design if they have any comments regarding this.

packages/web-client/app/components/card-space/create-space-workflow/select-business-account/index.ts
```diff
@@ -0,0 +1,63 @@
...
+      }
+    }
+  }
+
+  @action chooseSafe(safe: MerchantSafe) {
+    this.selectedSafe = safe;
+    this.args.workflowSession.setValue(
+      'merchantSafeAddress',
+      this.selectedSafe!.address
+    );
```
Aierie, line 59: Could `selectedSafe` be a getter that is based on `merchantSafeAddress`?

packages/web-client/tests/acceptance/create-card-space-test.ts
```diff
@@ -107,70 +128,61 @@ module('Acceptance | create card space', function (hooks) {
...
-    await waitFor(postableSel(3, 1));
+    // Reservation badge
+    await waitFor(postableSel(4, 1));
     assert
-      .dom(postableSel(3, 0))
+      .dom(postableSel(4, 0))
       .containsText(`We have sent your URL reservation badge`);
-    // // Badge
-    assert.dom(postableSel(3, 1)).containsText(`URL reservation`);
+    assert.dom(postableSel(4, 1)).containsText(`URL reservation`);
```
Aierie, line 159: Not related to this PR: Does saying we've sent the reservation badge make sense here? I don't think it does, the user hasn't actually submitted their card space for creation, so we have no reason to send a URL reservation badge, and we don't actually have a URL reservation badge to display... I think this should be taken out of the workflow later until we actually have badges.

Aierie (changes requested):
I think we should filter out merchants that already have a card space, other than that this looks good to go!

packages/web-client/app/components/card-space/create-space-workflow/select-business-account/index.ts
```diff
@@ -0,0 +1,63 @@
...
+    }
+  }
+
+  @action async setupBusinessAccountSelection() {
+    await this.layer2Network.waitForAccount;
+
+    let merchantSafes = this.layer2Network.safes.value.filterBy(
+      'type',
+      'merchant'
+    ) as MerchantSafe[];
```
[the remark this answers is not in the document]
jurgenwerk: Wow, good catch! Yeah sure - maybe even a better option would be to disable them in the dropdown. I think this is going to need an additional endpoint in the hub which would respond with user's created card spaces. I think @backspace is doing card space <-> merchant adjustments in the backend, so I think this new endpoint would depend on that work (but I can mock the response for now).

packages/web-client/tests/acceptance/create-card-space-test.ts
jurgenwerk: Yeah good point! I think we can remove it for now. Since I already did some workflow cleanup in this PR I guess I can include the removal here as well.

packages/web-client/app/components/card-space/create-space-workflow/select-business-account/index.ts
```diff
@@ -0,0 +1,63 @@
...
+      return useResource(this, MerchantInfo, () => ({
+        infoDID: this.selectedSafe!.infoDID,
+      }));
+    } else {
+      return null;
+    }
+  }
+
+  @action async setupBusinessAccountSelection() {
+    await this.layer2Network.waitForAccount;
```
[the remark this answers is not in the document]
jurgenwerk: I actually have no real reason for this, I think it's just a defensive measure taken from other workflows. Will check if this is actually needed.

packages/web-client/app/components/card-space/create-space-workflow/select-business-account/index.ts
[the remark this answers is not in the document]
jurgenwerk: Perhaps the easiest would be to add a new endpoint `/card-spaces/available-merchant-ids`. This way we wouldn't have to fetch all card spaces and all merchant infos in the web client to determine which ids are available 💡

lukemelia:
Nice work here, folks!

packages/hub/utils/queries.ts
```diff
@@ -9,10 +9,17 @@ const camelToSnakeCase = (str: string) => str.replace(/[A-Z]/g, (letter) => `_${
 // => { where: 'name=$1 AND age=$2', values: [ 'Phil', 45 ] }
 
 export function buildConditions(params: any) {
-  let nonNullParams = pickBy(params, function (value) {
+  // Only allow nulls, strings and numbers in the params
+  let filteredParams = Object.fromEntries(
```
jurgenwerk, line 13: Preparation for https://github.com/cardstack/cardstack/pull/2594/commits/d8abeab3f06c534f46d99cbc0a55800d8a9efaeb#diff-4dc024baa8514b72c704540caefc15706cd8e19273678717f4d573381aa7b6c0R49

jurgenwerk:
@Aierie @backspace can you take a look at the new commits, please? This evolved into something much bigger than I hoped for so extra pairs of careful eyes would be great!

Most notable changes are:

1. Adding an endpoint in the hub that returns merchant infos which weren't used in a card space yet. In the selector, we show only merchant safes that match what was returned from the endpoint. This is so that users can't select a business account which is already associated with a card space

2. The hub authentication ticket is now always shown in this workflow. This is so that we can put a `check` on it which checks if we can proceed to the business account selector. This is how it looks like when entering the workflow while already authenticated:
image

3. In addition to "did the user create a merchant yet?" check,  the `check()` function now also checks if there's any business accounts that are still free to associate with a card space

Aierie (commented):
I'm halfway through this, will continue tomorrow - yet to test the cancelation behaviour locally. This has indeed become quite huge!

Aierie (approved):
I think this looks good, thanks for pushing this through! Would be good to either remove or the `RESTORATION_` failure reasons or trigger them during restoration, + maybe use the `conditionalCancelationMessage` utility (https://github.com/cardstack/cardstack/blob/main/packages/web-client/app/models/workflow/cancelation-helpers.ts).

Aierie:
> The hub authentication card is now always shown in this workflow (includeIf was removed). This is so that we can put a check on it which checks if we can proceed to the business account selector. This is how it looks like when entering the workflow while already authenticated:

@backspace I remember you mentioning before that we could improve the messaging around the hub auth card ("only need to do this once per device/browser pair"), do you think it would be good to follow up this change with always showing the hub auth card + changing the message on other workflows to be the same as this workflow?

packages/web-client/app/components/card-space/create-space-workflow/index.ts
```diff
@@ -185,6 +244,43 @@ class CreateSpaceWorkflow extends Workflow {
...
+      },
+    }),
+    new WorkflowMessage({
+      message: `It looks like you all your business accounts have already been used to create a Card Space. In order to create your Card Space, you must first create a new business account.`,
+      includeIf() {
+        return (
+          this.workflow?.cancelationReason ===
+            FAILURE_REASONS.ALL_BUSINESS_ACCOUNTS_TAKEN ||
+          this.workflow?.cancelationReason ===
+            FAILURE_REASONS.RESTORATION_ALL_BUSINESS_ACCOUNTS_TAKEN
```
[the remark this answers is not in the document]
Aierie: I think a reason why we have `RESTORATION_` counterparts of the `FAILURE_REASONS` is because of different copy. If we expect to be handling it the same way, do we need a `RESTORATION_` prefix? Could we use the `ALL_BUSINESS_ACCOUNTS_TAKEN` cancelation reason during restoration?

jurgenwerk:
@Aierie @backspace Hub auth copy issue captured in [CS-3015](https://linear.app/cardstack/issue/CS-3015/adjust-hub-connect-card-copy-in-workflows-always-show-it).

packages/hub/services/queries/merchant-info.ts
```diff
@@ -14,10 +18,15 @@ export default class MerchantInfoQueries {
   async fetch(filter: MerchantInfoQueriesFilter): Promise<MerchantInfo[]> {
     let db = await this.databaseManager.getClient();
 
-    const conditions = buildConditions(filter);
+    let conditions = buildConditions(filter);
+    let query = `SELECT id, name, slug, color, text_color, owner_address, created_at from merchant_infos WHERE ${conditions.where}`;
 
-    const query = `SELECT id, name, slug, color, text_color, owner_address, created_at from merchant_infos WHERE ${conditions.where}`;
-    const queryResult = await db.query(query, conditions.values);
+    if (filter.customFilter?.availableForCardSpace) {
```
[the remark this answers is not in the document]
jurgenwerk: I was worried we'd end up with many arguments for different custom scopes down the road. That's why I propose to have a filter object even for custom scopes.

packages/web-client/app/components/card-pay/hub-authentication/index.hbs
```diff
@@ -53,7 +54,7 @@
       </i.ActionStatusArea>
     </:in-progress>
     <:memorialized as |m|>
-      <m.ActionStatusArea>
+      <m.ActionStatusArea data-test-memorialized>
```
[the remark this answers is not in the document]
jurgenwerk: Good point, will fix.

packages/web-client/app/components/card-space/create-space-workflow/index.ts
```diff
@@ -185,6 +244,43 @@ class CreateSpaceWorkflow extends Workflow {
...
     }),
+    new WorkflowMessage({
+      message: `It looks like you haven't created a business account yet. In order to create your Card Space, you must first create your first business account. This is required
+          because your Card Space URL will depend on your business account ID.`,
+      includeIf() {
+        return (
+          this.workflow?.cancelationReason ===
+            FAILURE_REASONS.NO_BUSINESS_ACCOUNT ||
+          this.workflow?.cancelationReason ===
+            FAILURE_REASONS.RESTORATION_NO_BUSINESS_ACCOUNT
```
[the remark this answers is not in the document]
jurgenwerk: Good catch, I think we don't need it. Removed.

packages/web-client/app/components/card-space/create-space-workflow/index.ts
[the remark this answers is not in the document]
jurgenwerk: I made an attempt to add the check for this during restoration, but it introduced many rendering side effects  😔 I think this needs a more thorough look so I created a [separate ticket](https://linear.app/cardstack/issue/CS-3016/support-async-restoration-checks-for-merchant-availability) for that. I think this is good to go without the merchant restoration checks since this is more of an edge case.

PR #2595, opened 2022-01-18
hub: Change spaces to belong to merchants

backspace:
This sets up a `has_one`-type relationship between `merchant_infos` and `card_spaces`, where a `card_spaces` cannot exist without linking to a `merchant_infos`. It also removes the `url` and `ownerAddress` fields in `card_spaces`. `merchant_infos.slug` is the unique identifier that will be used in #2586 to fetch the details of a Card Space.


backspace:
hmm maybe it should be `NOT NULL` for `card_spaces.owner_address` 🤔

Aierie:
How does this change handle a user having / creating multiple business accounts? `owner_address` seems to be the EOA address, not the safe address.

Aierie:
It feels like this change introduces a 1-1 relationship between business and card space that is not easily changed. This means that a merchant of a Card Space cannot be swapped out after creation, eg. if a user wants to have all payments to their spaces directed to a particular merchant. Is my understanding correct?

backspace:
ah, thanks for the correction, it’s true that this shouldn’t relate a Card Space to an EOA. It should relate to a merchant safe. I think it’s okay to have a 1:1 from merchant account to Card Space though, as that’s what the current Card Space creation workflow suggests: you choose a merchant account that the Card Space connects to, right? I feel like for now we can treat it as an unchangeable relationship and revisit that as needed in the future, it seems lightweight enough to make changeable.

backspace:
In that vein, new plan: a traditional `belongs_to`-style setup, with `card_spaces.merchant_info_id` being a key into `merchant_infos.id`. But I’m confused… when someone creates a merchant account, Hub doesn’t even know the safe address, right? So when someone creates a Card Space, Hub can follow `merchant_info_id` to the `merchant_infos` to get the `owner_address` and produce a signing request to confirm ownership of the merchant safe…?

packages/hub/config/structure.sql
```diff
@@ -597,17 +597,6 @@ $$;
...
---
--- Name: add(integer, integer); Type: FUNCTION; Schema: public; Owner: postgres
---
-
-CREATE FUNCTION public.add(integer, integer) RETURNS integer
-    LANGUAGE sql IMMUTABLE STRICT
-    AS $_$select $1 + $2;$_$;
-
-
-ALTER FUNCTION public.add(integer, integer) OWNER TO postgres;
```
backspace, line 609: Is this because [this migration](https://github.com/cardstack/cardstack/pull/2547/files#diff-930150a7f3faea990b45d2cdfaad9876332f6fc13cec22a67cfb034969b11302R5), which includes the other `lt` `eq` and `gt` functions doesn’t have `add`? Current theory is that it was in a migration but then removed and `structure.sql` wasn’t updated.

backspace: It seems like it’s [safe to remove](https://discord.com/channels/584043165066199050/900021750367129641/936726752259096587).

jurgenwerk: Nice, thanks for checking!

packages/hub/routes/card-spaces.ts
```diff
@@ -54,36 +59,74 @@ export default class CardSpacesRoute {
...
       profileImageUrl: this.sanitizeText(ctx.request.body.data.attributes['profile-image-url']),
       profileCoverImageUrl: this.sanitizeText(ctx.request.body.data.attributes['profile-cover-image-url']),
-      ownerAddress: ctx.state.userAddress,
     };
 
-    let errors = await this.cardSpaceValidator.validate(cardSpace);
-    let hasErrors = Object.values(errors).flatMap((i) => i).length > 0;
+    let merchantInfoError;
+
+    if (cardSpace.merchantId) {
```
jurgenwerk, line 79: Shouldn't this attribute always be present? Maybe we can pass the `cardSpace` into the validator at this place and let it figure out whether merchantId is present or not and return an error if it's not?

packages/hub/node-tests/services/card-space-validator-test.ts
```diff
@@ -4,40 +4,16 @@ import { setupHub } from '../helpers/server';
 describe('CardSpaceValidator', function () {
   let { getContainer } = setupHub(this);
 
-  let invalidUrls = [
-    'www.card.space',
```
jurgenwerk, line 8: Should we add a check for these the subdomains to the slug validator in the merchant info?

jurgenwerk (commented):
Looks great! But have a couple of comments regarding validation. I also think it would be a good idea to add the following checks to the card space validator:

1. Is the merchant id present
2. Does the merchant id in card space really belong to the current user in context

I think it would also make the route leaner this way.

packages/hub/routes/card-spaces.ts
backspace: `validateRequiredFields` will indeed actually catch this, removed now, thanks.

packages/hub/routes/card-spaces.ts
```diff
@@ -54,36 +59,74 @@ export default class CardSpacesRoute {
...
+      let merchantId = cardSpace.merchantId;
+      let merchant = await this.merchantInfoQueries.fetch({ id: merchantId });
```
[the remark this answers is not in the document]
backspace: hah yes that was a particularly awkward naming/arrangement of things, fixed in 30fdddf47, thanks

packages/hub/node-tests/services/card-space-validator-test.ts
backspace: the merchant info slug tests check for [forbidden words](https://github.com/cardstack/cardstack/blob/aaffdbc4142fba53032ca1662b137a4c741673a9/packages/hub/node-tests/routes/merchant-infos-test.ts#L181) and [only lowercase letters](https://github.com/cardstack/cardstack/blob/aaffdbc4142fba53032ca1662b137a4c741673a9/packages/hub/node-tests/routes/merchant-infos-test.ts#L123) which I think covers everything?
*

PR #2596, opened 2022-01-19
hub: Update documentation for running components

backspace:
I also removed the “tip” that referenced the now-outdated `/bin` reference.


lukemelia (approved):
👍

PR #2597, opened 2022-01-19
Add roots and payment cycle to subgraph

tintinthong:
**Description**

Tally server (the one which writes merkle roots) has a hard time deciding which payment cycle a reward program is on. Currently, tally literally relies on its database to tell the state of the blockchain; we intend to rely on the subgraph indexer as the single source of truth (which payment cycles have roots written already). 

This is made worse when developers use card pay staging but their own local development databases. Meaning that the roots written will get rejected because "the root has already been written". It is quite a pain trying to write to the blockchain over and over until success.

From discussion with @IanCal, we've identified that with distributed reward system, we may also want roots to be written non-sequentially. For example,
- the root to payment cycle 2 gets written before payment cycle 1 
- payment cycle 2 to be written after payment cycle 1 and payment cycle 3 is written. 

This feature enables non-sequential roots and allows consumer of data to decide a strategy on how to write merkle roots.

**Test**

- [x] Justin Private node


packages/cardpay-subgraph/src/mappings/reward-pool.ts
```diff
@@ -72,3 +76,23 @@ export function handleRewardTokensAdded(event: RewardTokensAdded): void {
...
+  if (rewardProgramEntity == null) {
+    log.warning(
+      'Cannot process tokens added txn {}: Reward program entity does not exist for safe address {}. This is likely due to the subgraph having a startBlock that is higher than the block the safe was created in.',
+      [txnHash, rewardProgramID]
+    );
+    return;
+  }
+
+  let entity = new MerkleRoot(txnHash);
+  entity.rewardProgram = rewardProgramID;
```
IanCal, line 95: Should this be 

    entity.rewardProgram = rewardProgramEntity

?

tintinthong: No I don't think so. You can see multiple examples 
- [here](https://github.com/cardstack/cardstack/blob/73e6b58d4b862229097c941be53b9edfdfda36ab/packages/cardpay-subgraph/src/mappings/prepaid-card.ts#L63)
- [here](https://github.com/cardstack/cardstack/blob/73e6b58d4b862229097c941be53b9edfdfda36ab/packages/cardpay-subgraph/src/mappings/merchant-manager.ts#L26)

This is a pattern I've followed. But there was never convincing documentation to point me to the correctness of this pattern.

tintinthong:
Ok deployed to my hosted graph and it seems to work just fine. I will pair this pr with another pr in tally (of updating the payment cycle). Assuming all is well, I will merge this when that is done (just in case I miss out some data).

habdelra:
@tintinthong what did you do to get it to deploy?

tintinthong:
> @tintinthong what did you do to get it to deploy?

Good question. I'm not really sure what affected it. I just deployed to the instance. I didn't taint it or anything. Its probably some ephemeral network thing.

BUT I did one thing before that, which I do not think affected it. I did a terraform plan and apply before. Maybe u might have a thought. There was some inconsistent state here which I ran `terraform apply` but I don't think it affected anything. 

Screenshot 2022-01-20 at 9 03 13 PM

habdelra:
Since the apply didn’t change anything I don’t think it would have effected any of the services. Yeah must have been an ephemeral network thing 🤷🏽‍♂️

tintinthong:
The corresponding pr is up here https://github.com/cardstack/tally-service/pull/45

tintinthong:
> Ok deployed to my hosted graph and it seems to work just fine. I will pair this pr with another pr in tally (of updating the payment cycle). Assuming all is well, I will merge this when that pr is done (just in case I miss out some data).

The corresponding pr is up here cardstack/tally-service#45

IanCal:
@tintinthong Looks like it was just a change to the outputted variables, 

https://github.com/cardstack/infra/commit/616179d5d3a1652141d6062ae81a2900a4e062fd is the commit they were removed in.

Probably would come down to the order apply has been run while that was being developed.

tintinthong:
I would like to merge this in. Don't think there are any big issues from this pr https://github.com/cardstack/tally-service/pull/45

IanCal:
@tintinthong I have a small request, I'd like to add the block number where the root was submitted. It'll make pulling this data back out for the indexing a lot easier.

tintinthong:
> @tintinthong I have a small request, I'd like to add the block number where the root was submitted. It'll make pulling this data back out for the indexing a lot easier.

darn! right I forgot. either, u can add it in or ill fix it up tmz.

tintinthong:
> @tintinthong I have a small request, I'd like to add the block number where the root was submitted. It'll make pulling this data back out for the indexing a lot easier.

Fixed. And have tested on my graph node

PR #2599, opened 2022-01-19
footer fix for safari 15 bug

Aierie:
Safari 15 bug leads to footer layout being distorted. Bug: the gap property is applied by adding a gap on top of the top element in column-reverse.

Bug in question: the gap property is applied by adding a gap on top of the top element in column-reverse (should be bottom). 

Fix: don't use gap. apply a margin directly, either on the bottom or the left, depending on the size of the screen.

### DIFF
--- a/packages/web-client/app/components/common/footer/index.css
+++ b/packages/web-client/app/components/common/footer/index.css
@@ -4,7 +4,6 @@
 
 .cardstack-footer__section {
   display: flex;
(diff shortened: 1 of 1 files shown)


jurgenwerk (approved):
Looks good. A note to self to start testing in other browsers as well (usually I check Chrome only). I think we can also set cross browser testing in Percy. I just enabled Safari (was disabled before): 

image

PR #2600, opened 2022-01-19
dependency-injection system can dynamically import

mattmcmanus:
- [x] Simple top level service imports
- [x] Support for both pascal and camelcase names
- [x] Arbitrary nesting of services
- [x] Routes 
- [ ] Validators and Queries
  - [ ] Move both out of services into the top level of the app
  - [ ] Setup config and prefixes


ef4:
@habdelra and I have started an alternative approach to this here: https://github.com/cardstack/cardstack/pull/2608

There were a few reasons why. 

 - using walkSync like this won't work in production (there are no separate files present after webpack has built)
 - this re-introduces a subtle race condition that we had eliminated from the system (async lookups need to get into the cache synchronously, so that multiple lookups will share work)
 - there's no need to support a microsyntax with ":" inside strings, we can make `type` a first-class argument to register and inject. This allows some nice ergonomic patterns, like defining type-specific injection helpers.
 - `rootDir`, `importFactory`, and `importConfig` are really not three separate things, they can be unified into one thing that maps from types to `import()` callbacks.

PR #2602, opened 2022-01-20
remove unnecessary messages from card space

Aierie:
These are no longer relevant since we are not requiring payment for card space.


jurgenwerk (approved):
Looks good. It's probably gonna conflict a bit with my changes in https://github.com/cardstack/cardstack/pull/2594 (where I did a similar cleanup) but it won't be a problem at all.

PR #2603, opened 2022-01-20
svg id gen for web client

Aierie:
attempt at CS-2826 to automate id generation for svgs in web client to prevent collision of ids causing bugs. 

TODO:
- [ ] better way to specify ember-svg-jar options, eg. inheriting from app that requires boxel?


ef4:
The SVG handling in boxel and web-client is done two different ways, and this is only going to influence one of them.

We have ember-svg-jar, used via `{{svg-jar ...}}`.

And we have webpack 5's built-in asset loader, used via `import url from "./path/to/something.svg"`.

I think we should refactor to only use one of those. But while we have both, you'll also need to configure svgo for webpack.  Probably with https://github.com/hyperbola/svgo-loader

packages/web-client/package.json
```diff
@@ -39,6 +39,7 @@
     "@embroider/webpack": "^0.49.0",
     "@glimmer/component": "^1.0.3",
     "@glimmer/tracking": "^1.0.3",
+    "@hyperbola/svgo-loader": "0.1.2",
```
Aierie, line 42: Realised later that 0.1.2 supported svgo 1

Aierie:
@ef4 I've updated the webpack svg loading to include the same optimization

ef4 (approved):

PR #2605, opened 2022-01-20
hub: Change incidents to have shorter names

backspace:
This shifts what was the `name` to the `body` when creating a Statuspage incident to allow the `name` to be a shorter string (`Experiencing degraded service` only, for now). When you [create a new incident](https://developer.statuspage.io/#operation/postPagesPageIdIncidents), `body` is the “initial message, created as the first incident update.”

The desire is to have a more concise message to show in Card Wallet:

image

Since web client’s design is more able to facilitate longer messages, it ignores the `name` and chooses the `body` of the last `incident_updates` entry.

### DIFF
--- a/packages/hub/node-tests/routes/checkly-webhook-test.ts
+++ b/packages/hub/node-tests/routes/checkly-webhook-test.ts
@@ -1,12 +1,13 @@
 import { registry, setupHub } from '../helpers/server';
 
-let createdIncident: { componentName: string; message: string } | null,
+let createdIncident: { componentName: string; name: string, message: string } | null,
   resolvedIncident: { componentName: string } | null;
 
 class StubStatuspageApi {
-  async createIncident(componentName: string, message: string): Promise<void> {
+  async createIncident(componentName: string, name: string, message: string): Promise<void> {
     createdIncident = {
       componentName,
+      name,
       message,
     };
   }
--- a/packages/hub/node-tests/routes/checkly-webhook-test.ts
+++ b/packages/hub/node-tests/routes/checkly-webhook-test.ts
@@ -41,7 +41,7 @@ describe('POST /api/checkly-webhook', async function () {
 
     expect(createdIncident).to.deep.equal({
       componentName: 'Subgraph',
-      name: 'Experiencing degraded service',
+      name: 'Transactions delayed',
       message:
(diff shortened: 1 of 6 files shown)


backspace:
When you first exercised this, @jurgenwerk, did you create a fake incident? But that wouldn’t work for the Checkly part… 🤔

Aierie:
Having this behaviour (of the wallet reading the title and the DApp reading the latest status) imposes some constraints on copy:

- Titles of incidents should be short
- Incident updates should be self-contained and meaningful in isolation. This means we can't say things like "We are investigating this issue and will keep you informed" as a single update.

Is this something we want/should commit to, and if so, should this be added to our internal documentation about incident handling too?

There are probably alternative approaches to handling this... eg. getting a collapsible/expandable design for incidents on the wallet (not sure how difficult this is?). This should probably still impose the "titles of incidents should be short" constraint, but will leave room for whoever is manually posting incidents to not have to craft status updates around the DApp reading them.

jurgenwerk:
@backspace yeah, at first I was creating fake incidents on statuspage manually. Maybe for testing in development a better option would be to create your own statuspage account and wire it up with your local setup?

I pretty much agree with @Aierie - If we decide to go with having two attributes with incident info, one could be the short and the other one the long description of the incident, but I feel like both names should be concrete and not just "Experiencing degraded service". I feel like users deserve a good explanation right when they see it.

Also, does anyone know if it's intentional the mobile version doesn't have the "Check our status page for more info?" link?

One other alternative I've been thinking about is to adjust the wallet app to support larger texts by truncating it and allow users to reveal more if they want to. Something like this:

image

Aierie:
I think pressing the mobile app banner is meant to bring you to status page

jurgenwerk:
Following up after our call: 
1. Suggesting a change from "Experiencing degraded service" to "Transactions delayed" (incident name). Will give more info to the user
2. Use this format in the web client: "${name}: ${description}"

Other things from this PR can probably stay the same? @backspace I can help with testing once the changes here are final.

backspace:
Successful in production:

image

image

jurgenwerk:
@backspace looking great! I wish the card wallet warning pill had some kind of an indicator I can tap on it and find out more

backspace:
agreed, maybe something conventional like [this icon](https://fontawesome.com/v5.15/icons/external-link-alt) would make sense. @lukemelia?

PR #2606, opened 2022-01-20
Remove id check during adoptInToRealm and add test to confirm

burieberry:
If `id` is sent into `adoptIntoRealm` method, a model will be returned but it's not possible to save it into the db because the realm manager is already doing its job. I couldn't find anything to change, unless we want to call realm manager's `create` method earlier, such as inside the `adoptFromRealm` and not wait until a `save`.


habdelra:
sweet, the issue I saw must have been addressed by other refactors that i did. thanks!!

habdelra (approved):

PR #2607, opened 2022-01-21
Update merchant claim notification data

topwebtek7:


packages/hub/node-tests/tasks/notify-merchant-claim-test.ts
```diff
@@ -110,6 +110,9 @@ describe('NotifyMerchantClaimTask', function () {
         notificationId: 'sokol::a::123::eoa-address',
         notificationData: {
           notificationType: 'merchant_claim',
+          transactionInformation: JSON.stringify({
```
jurgenwerk, line 113: Why is there a need to stringify? I see we use the same here so I guess it is necessary: https://github.com/cardstack/cardstack/blob/main/packages/hub/tasks/notify-customer-payment.ts#L141

@Aierie Is this because of the worker args serialization or does Firebase client expect a string there?

jurgenwerk (approved):
This looks ok to me but I'm unsure how to test this locally. However it doesn't seem like it wouldn't break anything so I'm fine if we deploy directly.

packages/hub/node-tests/tasks/notify-merchant-claim-test.ts
topwebtek7: Right, firebase expect string for all values inside notificationData object

PR #2608, opened 2022-01-21
Refactoring DI system to support types & dynamic lookup

ef4:
First commit here adds an explicit type concept to the di system.

Next we should implement registerType.

### DIFF
--- a/packages/di/dependency-injection.ts
+++ b/packages/di/dependency-injection.ts
@@ -10,21 +10,24 @@ export class Container implements ContainerInterface {
 
   constructor(private registry: Registry) {}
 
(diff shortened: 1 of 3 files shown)


ef4:
I had an idea to simplify the problem of adding the async code loading.

We can have a phase that doesn't allow *arbitrary* asynchrony. 

The code loading and object instantiation phase will still need to move from synchronous to asynchronous, but the only asynchrony allowed during that phase is our own dynamic lookups. No user async code runs there, so it can be kept separate from the later phase (the one that handles all the existing asynchrony).

mattmcmanus:
Updated the description of the PR to include info on how to migrate

PR #2612, opened 2022-01-24
Fix/missing dropdown highlights

Aierie:
CS-2991

https://www.loom.com/share/174a2d01f5d042a5b35da28f66e45295

After recording the video, I changed the highlight color to a lighter one to avoid contrast issues:

dropdown with new highlight color, a light cyan


Aierie:
Note: the dropdown scrolling to highlighted items is dependent on setting a fixed height for the dropdown.

jurgenwerk (approved):
Looking good! After watching the video I wanted to suggest a lighter color but I see you already changed that - it looks good!

PR #2615, opened 2022-01-25
add eoa address + network to all current push notifications' data fields

Aierie:
Works locally with a dry run (`admin.messaging().send(notification, true)`).


topwebtek7 (approved):
Looks good to me! 🚀

PR #2616, opened 2022-01-25
Add tally reward rules

tintinthong:
- move tally rules into mono repo
- add test for rules 
- ~~rules made such that they can be "distributed" as tasks~~
- create streamlit that allows analysis (moved it into its own folder)
- fixed streamlit issue where `selectbox` was nested
- configure dev environment
- tweaks to rule configuration
- ~tweaks to merkle tree function st payment_cycle is an argument passed in~
- ~adding of idea of rule_hash. I foresee that in the runner, we can validate results. something like hash(rule_hash, payment_cycle, reward_program) will yield same payment list~

Best way to test is to use vscode launch debug "Explore using Streamlit"

I have some changes in the runner that wud like to implement but want to see feedback of this pr first


tintinthong:
Should put ineritEnv to false to point to the correct python. Otherwise will use system python which is outdated. See [here](https://code.visualstudio.com/updates/v1_36#_launch-terminals-with-clean-environments)

packages/cardpay-reward-programs/cardpay_reward_programs/payment_tree.py
```diff
@@ -0,0 +1,130 @@
...
+
+
+def group_by(data_array, callback):
+    # remember: itertools.groupby requires that keys are sorted; it only groups common keys which are next to each other
+    sorted_data = sorted(data_array, key=callback)
+    return itertools.groupby(sorted_data, callback)
+
+
+class PaymentTree:
+    def __init__(self, payment_list: List[Payment], payment_cycle:int) -> None:
```
IanCal, line 33: I'm not following why payment cycle is now passed into the constructor of the tree rather than being part of the payment. It is also then re-passed in into the as_arrow function.

packages/cardpay-reward-programs/cardpay_reward_programs/payment_tree.py
```diff
@@ -0,0 +1,130 @@
...
+        self.payment_nodes = self.aggregate_payments(payment_list)
+        self.data = [self.encode_payment(payment_node, payment_cycle) for payment_node in self.payment_nodes]
+        self.tree = MerkleTree(self.data, hashfunc)
+
+    def aggregate_payments(self, payments: Payment) -> List[Payment]:
```
IanCal, line 38: I copied this over, but I think it should just be down to the rules to create valid lists. I don't think anything breaks if there's more than one payment per person, and this function would fail when we start to consider non-claimable rewards (there's no amount). I think we can just delete this function.

packages/cardpay-reward-programs/cardpay_reward_programs/rule.py
```diff
@@ -0,0 +1,90 @@
...
+        self.set_user_defined_parameters(**user_defined_parameters)
+
+    def set_core_parmaters(
+        self, subgraph_config_location, payment_cycle_length, valid_from, valid_to, token
+    ):
+        self.subgraph_config_location = subgraph_config_location
+        self.payment_cycle_length = payment_cycle_length
+        self.valid_from = valid_from
+        self.valid_to = valid_to
+        self.token = token
```
IanCal, line 26: I don't think token should be part of the core params, that would be down to the rule which token(s) it should use.

packages/cardpay-reward-programs/cardpay_reward_programs/rule.py
```diff
@@ -0,0 +1,90 @@
...
+    """
+
+    def __init__(self, core_parameters, user_defined_parameters):
+        self.set_core_parmaters(**core_parameters)
+        self.set_user_defined_parameters(**user_defined_parameters)
+
+    def set_core_parmaters(
+        self, subgraph_config_location, payment_cycle_length, valid_from, valid_to, token
+    ):
+        self.subgraph_config_location = subgraph_config_location
```
IanCal, line 22: I've been thinking on this and I think the config location is also not a core param. You could read from more than one.

packages/cardpay-reward-programs/cardpay_reward_programs/rule.py
```diff
@@ -0,0 +1,90 @@
...
+    def get_core_hash(self):
+        core_parameters = {
+            "subgraph_config_location": self.subgraph_config_location,
+            "payment_cycle_length": self.payment_cycle_length,
+            "valid_from": self.valid_from,
+            "valid_to": self.valid_to,
+            "token": self.token,
+        }
+        o = json.dumps(core_parameters, sort_keys=True)
+        return hashlib.md5(o.encode("utf-8")).hexdigest()
```
IanCal, line 37: The hashing here is to get a unique value that represents the config for the rules right? Is there a reason to split it into core & user defined? We could just hash the params being passed in in total.

IanCal (commented):
I added some comments but mostly looking good. The run_mock will be pulled out pretty much entirely imo as a separate running task.

packages/cardpay-reward-programs/cardpay_reward_programs/main.py
```diff
@@ -0,0 +1,68 @@
...
+
+    run_parameters = parameters["run"]
+    name, core_parameters, user_defined_parameters = get_parameters(parameters)
+    rule = select_rule(name, core_parameters, user_defined_parameters)
+
+    results_df = rule.run(run_parameters["payment_cycle"])
+    payment_list = rule.df_to_payment_list(results_df, reward_program_id)
+    tree = PaymentTree(payment_list.to_dict("records"), payment_cycle)
+    table = tree.as_arrow(payment_cycle)
+    pq.write_table(table, Path(output_location) / "results.parquet")
```
IanCal, line 47: We might need to do a little more here to get it writing to S3. Cloudpathlib is mostly an abstraction but doesn't play as nicely with writing these files, here's some code that handles local & remote: 

https://github.com/cardstack/subgraph-extractor/blob/main/subgraph_extractor/cli.py#L262

packages/cardpay-reward-programs/cardpay_reward_programs/rule.py
tintinthong: ok sure i'll change it!

packages/cardpay-reward-programs/cardpay_reward_programs/rule.py
tintinthong: agreed. this was actually my thought until I reverted my changes from the module pr. Thanks.

packages/cardpay-reward-programs/cardpay_reward_programs/rule.py
tintinthong: I agree with you. I was really wanting to make the hash a staticmethod type. But there is a few refactoring questions around it. 

Question: should the `core_params` and `user_defined_params` be variables that are set only upon instantiating of the class? Think the implications around it is around how an analytical miner interacts with the docker container. If a rule is updated with params, I presume the docker container will be restarted. Available for call to clarify this.

packages/cardpay-reward-programs/cardpay_reward_programs/main.py
tintinthong: Cool. Will fix it.

packages/cardpay-reward-programs/cardpay_reward_programs/rule.py
IanCal: Not sure if this was before/after we talked the other day, so this may have been clarified already but

* Can set them on instantiating the class - neither will be built into the image
* I think it's reasonable that once a miner starts working on something they don't need to check for rule changes
* * That is - if a calculation can be triggered (progam not locked, data available and the right payment cycle/etc), it's fair game.
* * My expectation is that most calculations will be fast - these aggregations & duckdb based queries are (once data is downloaded) sub-second pieces of work so the window for timing issues around changing rules is very small
* Each piece of work is independent - the containers will be started, the task run, then cleared away rather than having long running containers. Miners would have their own scheduler/runner (to look for things that can be run, perhaps work out the most valuable ones, and then trigger the tasks) which would be long running

packages/cardpay-reward-programs/cardpay_reward_programs/payment_tree.py
tintinthong: I resolved this.

tintinthong:
@IanCal I made follow-up changes based on our conversation about payment cycle and config

tintinthong:
@IanCal do you want me to edit the"scheduler" as well. I figured because there are quite a few decisions to be made there, I left it?

IanCal (approved):

PR #2617, opened 2022-01-25
web: Check maintenances for degraded service

backspace:
This changes the URL fetched for determining whether to show the degraded service banner from `…/unresolved.json` to `…/summary.json`, as the former did not included scheduled maintenance notices, which we do want to display. This is a minor intervention that merges `incidents[]` and `scheduled_maintenances[]` and then uses the same logic to decide which to display: the highest-impact most-recent one.

I noticed that the `/api/v2/summary.json` was duplicated everywhere with the root URL never being used without it, so I moved the path into the config variable. If we do end up using multiple paths from Statuspage, we can break it into multiple variables.


jurgenwerk (approved):
Looks great! Thanks for adding this.

backspace:
I created a maintenance and here it is showing locally:

image

I’ll verify on staging and then production.

backspace:
On staging:

image

On production:

image

PR #2620, opened 2022-01-26
Render user into card space template

Aierie:
1. Separates UI for card space index route (where you can create a card space) from  the user page so that users cannot open the creation workflow from a user's page via the query params of the url
3. Renders the user page with some UI shared with the card space index route, such as the left edge.
4. Prevents a user from visiting `someone.card.space/card-pay/wallet` or other non-index routes on `someone.card.space`, with a caveat of not preventing transitions initiated within the app (clicking on a `<LinkTo>` in Card Space).

### DIFF
--- a/packages/web-client/app/routes/application.ts
+++ b/packages/web-client/app/routes/application.ts
@@ -2,8 +2,9 @@ import Route from '@ember/routing/route';
 import RouterService from '@ember/routing/router-service';
 import { LocationService } from '../services/location';
 import { inject as service } from '@ember/service';
-import '@cardstack/web-client/css/variables.css';
 import config from '@cardstack/web-client/config/environment';
+import '@cardstack/web-client/css/variables.css';
+import '@cardstack/web-client/css/card-space.css';
 
 export default class ApplicationRoute extends Route {
   @service declare router: RouterService;
--- a/packages/web-client/app/routes/application.ts
+++ b/packages/web-client/app/routes/application.ts
@@ -1,5 +1,25 @@
 import Route from '@ember/routing/route';
+import RouterService from '@ember/routing/router-service';
+import { LocationService } from '../services/location';
+import { inject as service } from '@ember/service';
+import '@cardstack/web-client/css/variables.css';
+import config from '@cardstack/web-client/config/environment';
 
-import '../css/variables.css';
+export default class ApplicationRoute extends Route {
+  @service declare router: RouterService;
+  @service declare location: LocationService;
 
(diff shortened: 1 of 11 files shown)


packages/web-client/app/routes/application.ts
```diff
@@ -2,8 +2,9 @@ import Route from '@ember/routing/route';
 import RouterService from '@ember/routing/router-service';
 import { LocationService } from '../services/location';
 import { inject as service } from '@ember/service';
-import '@cardstack/web-client/css/variables.css';
 import config from '@cardstack/web-client/config/environment';
+import '@cardstack/web-client/css/variables.css';
+import '@cardstack/web-client/css/card-space.css';
```
Aierie, line 7: I wonder if there's a good way to dynamically load CSS when we dynamically render a route template, so we don't have to load this on the application route.

packages/web-client/app/routes/application.ts
```diff
@@ -1,5 +1,25 @@
...
+    // some important functionality will be messed up
+    // eg. layer 1 infura calls via WalletConnect
+
+    // NOTE: there is a caveat to this. If the user uses a LinkTo from this page
+    // that leads to, say, card-pay.wallet, this interception will not run
+    // it may be better to create a base class for all routes that has this functionality
+    if (this.location.hostname.endsWith(config.cardSpaceHostnameSuffix)) {
+      this.router.replaceWith('index');
+    }
+  }
```
Aierie, line 25: The redirect implemented here prevents folks from visiting child routes when visiting x.card.space.... Not sure how to handle this caveat, or if we should think about this caveat for now? Would lean towards not thinking about it.

There is a known way to prevent transitions:

https://guides.emberjs.com/v3.28.0/routing/preventing-and-retrying-transitions/

But I'm not very confident in handling this event correctly.

Aierie:
If this approach seems ok, I should probably also implement route restrictions on the Wallet route.

backspace (changes requested):
It looks like this removes the ability to view a user’s card space locally without editing `/etc/hosts`, which is something I’d like to avoid… I don’t mind it being at an ugly path but I think we should make it easy on us when we actually start rendering Card Spaces. Maybe I’m missing it though 🤔

Aierie:
@backspace tried this locally, it did persist the query params through refresh.

packages/web-client/app/routes/application.ts
backspace: yes, let’s not think about it for now 😆

backspace (approved):

PR #2621, opened 2022-01-26
make session keys match hub CardSpace type

Aierie:
Things like `displayName` don't match hub and so have a mismatch with `profileName`... but they match our UI. However the items stored in workflow session (that need to be retrieved to be posted to hub, eg. `profileName`) do match the way hub refers to this data.

Referred to https://github.com/cardstack/cardstack/blob/c7df8cabd565ace3637225a97daa1efc0797dcf5/packages/hub/routes/card-spaces.ts#L13-L33 for the keys.


Aierie:
Am not entirely sure about this change since because the UI and data model are competing as sources of truth for naming some of these things.

backspace (approved):
I’m happy to change Hub/database to match whatever we prefer here, there’s no content yet that’ll break

jurgenwerk (approved):
Thanks for making this consistent! I also like what @backspace suggested. We could take this a step further and rename the attributes in the backend as well so they all would match.

PR #2623, opened 2022-01-26
move compiler services to new di pattern

mattmcmanus:
- Migrate kebob cased compiler services to new DI system
- Dynamic import handles camelCased injection strings for services

### DIFF
--- a/packages/hub/node-tests/helpers/cards.ts
+++ b/packages/hub/node-tests/helpers/cards.ts
@@ -1,7 +1,7 @@
 import Mocha from 'mocha';
 import tmp from 'tmp';
 import { join } from 'path';
-import { FileCacheConfig } from '../../services/file-cache-config';
+import FileCacheConfig from '../../services/file-cache-config';
 import { contextFor, registry, setupHub } from './server';
 import CardServiceFactory, { CardService, INSECURE_CONTEXT } from '../../services/card-service';
 import FileCache from '../../services/file-cache';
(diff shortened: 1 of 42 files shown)


packages/hub/node-tests/helpers/cards.ts
```diff
@@ -105,11 +106,11 @@ export function cardHelpers(mochaContext: Mocha.Suite) {
     if (!container) {
       throw new Error('Make sure cardHelpers are run after setupHub. It needs a configured container');
     }
-    currentCardService = await container.lookup('card-service');
-    fileCacheConfig = (await container.lookup('file-cache-config')) as TestFileCacheConfig;
-    fileCache = await container.lookup('file-cache');
+    currentCardService = await container.lookup('card-service', { type: 'service' });
+    fileCacheConfig = (await container.lookup('file-cache-config', { type: 'service' })) as TestFileCacheConfig;
+    fileCache = await container.lookup('file-cache', { type: 'service' });
```
mattmcmanus, line 111: @ef4 @habdelra This pattern, along with it's counterpart in `register`, is a pain. Can we:
1. Change the APIs to first class "types" so it's like: `container.lookup('service', 'file-cache')`
1. If we dont want to change the api, at least change the default type from `default` to `service` so this isn't necessary in the default cause?

habdelra: yeah we were conflicted over this ordering as well. we set it like this because the type is optional and optional params are added to the end. i don't have any strong feelings for preserving that though...

mattmcmanus:
@habdelra It looks like the bot tests are failing and my guess is that it's because discord is down? Do we really hit discord in our test suite? Have you ever used ruby VCR style request stubbing for stuff like this?

habdelra:
I'm seeing this same this--no in these tests we have mocked out discord. I think this is actually a DI related issue. I'm digging into this too.

PR #2625, opened 2022-01-26
Don't login to discord when under test (also disable corde during discord outage)

habdelra:


habdelra:
this was included in PR https://github.com/cardstack/cardstack/pull/2623

PR #2631, opened 2022-01-28
Show Card Space URL in the preview

jurgenwerk:
Ticket: [CS-3022](https://linear.app/cardstack/issue/CS-3022/update-url-on-merchant-select-in-the-preview)

image

### DIFF
--- a/packages/web-client/app/components/card-space/create-space-workflow/index.ts
+++ b/packages/web-client/app/components/card-space/create-space-workflow/index.ts
@@ -20,6 +20,8 @@ import { standardCancelationPostables } from '@cardstack/web-client/models/workf
 import RestorableWorkflowComponent from '../../card-pay/restorable-workflow-component';
 import MerchantInfoService from '@cardstack/web-client/services/merchant-info';
 import { taskFor } from 'ember-concurrency-ts';
+import { MerchantInfo } from '@cardstack/web-client/resources/merchant-info';
+import { useResource } from 'ember-resources';
 
 const FAILURE_REASONS = {
   L2_DISCONNECTED: 'L2_DISCONNECTED',
@@ -319,6 +321,10 @@ export default class CreateSpaceWorkflowComponent extends RestorableWorkflowComp
 
   @tracked detailsEditFormShown: boolean = true;
 
+  merchantInfo = useResource(this, MerchantInfo, () => ({
+    infoDID: this.workflow.session.getValue('merchantInfoDID'),
+  }));
+
   get workflowClass() {
     return CreateSpaceWorkflow;
   }
@@ -328,7 +334,7 @@ export default class CreateSpaceWorkflowComponent extends RestorableWorkflowComp
       profilePhoto: this.workflow.session.getValue('profileImageUrl'),
       coverPhoto: this.workflow.session.getValue('profileCoverImageUrl'),
       name: this.workflow.session.getValue('profileName'),
-      host: this.workflow.session.getValue('url'),
+      host: this.merchantInfo.id ? `${this.merchantInfo.id}.card.space` : null,
       category: this.workflow.session.getValue('profileCategory'),
       description: this.workflow.session.getValue('profileDescription'),
       buttonText: this.workflow.session.getValue('profileButtonText'),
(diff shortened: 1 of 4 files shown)


packages/web-client/app/services/merchant-info.ts
```diff
@@ -21,6 +26,22 @@ interface CheckMerchantSlugUniquenessTaskParams {
 
 export default class MerchantInfoService extends Service {
   @service declare hubAuthentication: HubAuthentication;
+  @service declare offChainJson: OffChainJsonService;
+
+  @task *fetchOffChainMerchantInfo(
```
jurgenwerk, line 31: Moved this here so I can use it in the business selector card to fetch merchant info - the slug. And then write it in the workflow session.

packages/web-client/app/components/card-space/create-space-workflow/select-business-account/index.ts
```diff
@@ -62,14 +62,24 @@ class CreateSpaceWorkflowSelectBusiness extends Component<WorkflowCardComponentA
...
     this.args.workflowSession.setValue(
       'merchantSafeAddress',
       this.selectedSafe!.address
     );
+
+    let merchantInfo = await taskFor(
+      this.merchantInfo.fetchOffChainMerchantInfo
+    ).perform(safe.infoDID!);
+
+    this.args.workflowSession.setValue('merchantId', merchantInfo.id);
```
jurgenwerk, line 82: This is for the preview so it can show the card space url

packages/web-client/app/components/card-space/create-space-workflow/index.ts
```diff
@@ -328,7 +328,9 @@ export default class CreateSpaceWorkflowComponent extends RestorableWorkflowComp
       profilePhoto: this.workflow.session.getValue('profileImageUrl'),
       coverPhoto: this.workflow.session.getValue('coverPhotoUrl'),
       name: this.workflow.session.getValue('displayName'),
-      host: this.workflow.session.getValue('url'),
+      host: this.workflow.session.getValue('merchantId')
```
jurgenwerk, line 331: I tried using `useResource` here (as an alternative to the fetching operations in the previous commits) like this:

```
get merchantData() {
    let did = this.workflow.session.getValue('merchantInfoDID');
    
    if (did) {
      return useResource(this, MerchantInfo, () => ({
        infoDID: did,
      }));
    } else {
      return null;
    }
  }
  
  get currentCardSpaceDetails() {
    return {
      ...
      host: this.merchantData ? `${this.merchantData.id}.card.space` : null,
      ...
    };
  }
  ```
  
  ... but this ends up in in an endless loop (the merchantData is called endlessly)  😵‍💫 Any ideas why that would happen?

backspace:
I’m not sure if this is a bug, and if it is maybe it’s out of the scope of this PR, but in preparing to exercise this, I first got the “all your business accounts have already been used” message (though actually the database was just empty). So I added a `merchant_infos` record and refreshed the workflow; the same message showed but so did the “select a business” card:

image

Starting a fresh workflow didn’t exhibit this problem.

backspace:
Once I got everything set up it worked well:

screencast 2022-01-31 15-43-13

This is maybe in another PR but this should be corrected:

> Only business accounts that you haven't yet been used for Card Space are shown.

I’ll now look at the actual code! 🤓

packages/web-client/app/components/card-space/create-space-workflow/index.ts
backspace: I think the problem is that `useResource` is meant to be like a service injection or other property, not something used within a function. I tried this and it worked:

image

As an aside, `ember-resources` now has a [`@use` decorator](https://github.com/NullVoxPopuli/ember-resources#use) that we should probably start `use`ing 😆

backspace (changes requested):
This works well, though I think it would be beneficial to update it with `useResource` (or `@use` if you like)

packages/web-client/app/components/card-space/create-space-workflow/index.ts
jurgenwerk: Thanks! I used the `useResource` which made this PR much simpler! 🎉

jurgenwerk:
Thanks @backspace for your `useResource` example and for pointing out the bug. I captured it in [CS-3033](https://linear.app/cardstack/issue/CS-3033/cancellation-bug-when-there-are-no-merchant-infos-present) so that we can fix it in another PR.

PR #2632, opened 2022-01-28
CS-3023 Fix delegation of command line cardpay-cli args in standalone installed version

lukemelia:
- [x] arg parsing
- [x] ENV var support
- [x] walletconnect support


tintinthong (approved):
Tested MNEMONIC_PHRASE and --mnemonic option. And works like a charm.

PR #2637, opened 2022-01-31
Base components for card space user page

Aierie:
Base components for card space user page. These components provide basic UI styling + behaviour, and some conveniences for editing. For examples of how these components are used and some other card space UI behaviour, see https://github.com/cardstack/cardstack/pull/2636, which includes some HAX! (see the description for other caveats).

re: `@sufficientContent` in `CardContainer`, this is here so that cards that use the `CardContainer` can decide on their own whether to display themselves (in view-only mode), and display their placeholder content (in editable mode). I'm not sure whether this is good - having sufficient content to not display a placeholder in editable mode might not necessarily mean having sufficient content to display a card in view-only mode, but it makes it easy to specify things like placeholders for these cards at the moment. It may be good to introduce a separate way to decide whether or not to display anything for empty cards in view-only mode (or simply use conditionally render in the consuming component), and leave `@sufficientContent` to help with deciding whether to render placeholder content.


backspace:
Sorry for the delay, things have been a lot, but I’ll finally review tomorrow! 😯

packages/web-client/tests/integration/components/card-space/user-page/layout-card-container-test.ts
```diff
@@ -0,0 +1,175 @@
...
+  '[data-test-card-space-card-edit-modal-title]';
+
+module(
+  'Integration | Component | card-space/user-page/layout-card-container',
+  function (hooks) {
+    setupRenderingTest(hooks);
+
+    hooks.beforeEach(function () {
+      this.setProperties({
+        editable: false,
```
backspace, line 29: minuscule importance but this is set to `true` in all but one test, could make that the default 🤓

backspace (approved):
This seems well-factored and thoroughly exercised with tests 🤩 I tried it out with the HAX branch, looks good! It can be hard to know with this kind of thing whether it’ll be a good abstraction in practice but that branch helps show that it probably is, and we can always adjust if needed.

I wonder if there’s a more clear name that `sufficientContent`… `shouldRender`? 🤔

Aierie:
Closing for now. It looks like these components may not be used for the MVP.

PR #2639, opened 2022-01-31
Use stylelint directly, plus stylelint upgrade

lukemelia:
- removes ember-cli-stylelint
- upgrades stylelint to 14
- defines regex for selector-class-pattern to batch BEM syntax


backspace (approved):
🥳

burieberry (approved):

PR #2640, opened 2022-01-31
CardModel.getField support for contained card field

habdelra:
This adds support for getting a contained card data via `CardModel.getField`

### DIFF
--- a/packages/hub/node-tests/@core/compiler-test.ts
+++ b/packages/hub/node-tests/@core/compiler-test.ts
@@ -484,6 +484,31 @@ if (process.env.COMPILER) {
         `);
       });
 
+      it('can compile schema class promise functions', async function () {
+        let { compiled } = await cards.load(`${realm}person`);
+        let source = getFileCache().getModule(compiled.schemaModule.global, 'browser');
+        expect(source).to.containsSource(`
+          then(resolve) {
+            let fields = ["lastName", "aboutMe", "fullName"];
+            return Promise.all(fields.map(field => this[field])).then(values => resolve(Object.fromEntries(fields.map((field, i) => [field, values[i]]))));
+          }
+        `);
+        expect(source).to.containsSource(`
+          catch(err) {
+            if (err) {
+              err();
+            }
+          }
+        `);
+        expect(source).to.containsSource(`
+          finally(cb) {
+            if (cb) {
+              cb();
+            }
+          }
+        `);
+      });
+
       it('can compile primitive field implementation in schema.js module', async function () {
         let { compiled } = await cards.load(`${realm}bio`);
         let source = getFileCache().getModule(compiled.schemaModule.global, 'browser');
--- a/packages/hub/node-tests/@core/compiler-test.ts
+++ b/packages/hub/node-tests/@core/compiler-test.ts
@@ -488,12 +488,19 @@ if (process.env.COMPILER) {
         let { compiled } = await cards.load(`${realm}person`);
         let source = getFileCache().getModule(compiled.schemaModule.global, 'browser');
         expect(source).to.containsSource(`
+          #allValuesPromise;
+
           get allFields() {
             return ["lastName", "aboutMe", "fullName"];
           }
+
           get allValues() {
-            return Promise.all(this.allFields.map(field => this[field]));
+            if (!this.#allValuesPromise) {
+              this.#allValuesPromise = Promise.all(this.allFields.map(field => this[field]));
+            }
+            return this.#allValuesPromise;
           }
+
           then(resolve) {
             return this.allValues.then(values => resolve(Object.fromEntries(this.allFields.map((field, i) => [field, values[i]]))));
           }
--- a/packages/hub/node-tests/@core/compiler-test.ts
+++ b/packages/hub/node-tests/@core/compiler-test.ts
@@ -488,31 +488,28 @@ if (process.env.COMPILER) {
         let { compiled } = await cards.load(`${realm}person`);
         let source = getFileCache().getModule(compiled.schemaModule.global, 'browser');
         expect(source).to.containsSource(`
-          #allValuesPromise;
-
-          get allFields() {
+          import { getAllValues } from "@cardstack/core/src/utils/schema-class";
+        `);
+        expect(source).to.containsSource(`
+          static get allFields() {
             return ["lastName", "aboutMe", "fullName"];
           }
 
-          get allValues() {
-            if (!this.#allValuesPromise) {
-              this.#allValuesPromise = Promise.all(this.allFields.map(field => this[field]));
-            }
-            return this.#allValuesPromise;
-          }
-
           then(resolve) {
-            return this.allValues.then(values => resolve(Object.fromEntries(this.allFields.map((field, i) => [field, values[i]]))));
+            let allValues = getAllValues(this);
+            return allValues.then(values => resolve(Object.fromEntries(Person.allFields.map((field, i) => [field, values[i]]))));
           }
         `);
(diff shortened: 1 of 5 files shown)


packages/hub/node-tests/@core/compiler-test.ts
```diff
@@ -484,6 +484,31 @@ if (process.env.COMPILER) {
...
+            return Promise.all(fields.map(field => this[field])).then(values => resolve(Object.fromEntries(fields.map((field, i) => [field, values[i]]))));
+          }
+        `);
+        expect(source).to.containsSource(`
+          catch(err) {
+            if (err) {
+              err();
+            }
+          }
+        `);
```
mattmcmanus, line 502: This doesn't seem correct...why would we call the error? Also, whats the plan for this?

packages/hub/node-tests/@core/compiler-test.ts
```diff
@@ -484,6 +484,31 @@ if (process.env.COMPILER) {
...
+        expect(source).to.containsSource(`
+          finally(cb) {
+            if (cb) {
+              cb();
+            }
+          }
+        `);
+      });
```
mattmcmanus, line 510: Same here, it's not clear to me the intended use for this

mattmcmanus (commented):
This makes sense to me based on the conversation earlier today. 

I'm a little fuzzy on where this fits into the near term picture. My understanding is that, writing a computed property, it may be necessary for a user to fetch all of the data for a particular field, to do with as they see fit. Is this correct? Is there some other use case for this API that I'm missing?

packages/hub/node-tests/@core/compiler-test.ts
habdelra: so when you want to use a finally for a promise you provide a callback. So a consumer of this promise that wants to use the finally would do this:

```js
promise.then(() => {
  // doSomething();
})
.finally(() => {
  doMyCustomTearDown();
});
```
so the finally needs to take as input a callback which is provided from outside the promise, `cb`. If finally is called with a callback, then we'll execute it.

packages/hub/node-tests/@core/compiler-test.ts
habdelra: This is for the consumer of the promise to provide an error callback that is evaluated when our promise rejects. I think the way this is really implemented, according to MDN, is that it is just an alias for a `.then(undefined, reject)`. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/catch. I wasn't really fleshing this out. but i can do that.

packages/hub/node-tests/@core/compiler-test.ts
```diff
@@ -484,6 +484,31 @@ if (process.env.COMPILER) {
         `);
       });
 
+      it('can compile schema class promise functions', async function () {
+        let { compiled } = await cards.load(`${realm}person`);
+        let source = getFileCache().getModule(compiled.schemaModule.global, 'browser');
+        expect(source).to.containsSource(`
+          then(resolve) {
```
habdelra, line 491: we can set the params for this to `resolve, reject` and thread the error that is raised from the Promise.all.catch thru to the reject

habdelra:
@mattmcmanus i don't think this really fits into a computed property.  I think computeds really wanna bottom out on primitive fields. If you have a computed that needs to consume all the fields of a contained card field, you still need to write the code that indicates what you are doing for each field (like sum all the fields of a card):
```js
async sum() {
  return await this.blah.a + await this.blah.b + await this.blah.c;
}
```
I don't think you would ever get `this.blah` _and then_ pull off `a`, `b`, `c` to sum it within the computed--feels like that is working against the grain. 

Based on the discussion today, my understanding is that this is how we could support passing a contained card's data to a component within a template like:
```hbs
<Bio @model={{@model.aboutMe}}>
```

packages/hub/node-tests/@core/compiler-test.ts
habdelra: Something like this might work out nicely: 
```js
          get #allValues() {
            return Promise.all(this.#allFields.map(field => this[field]))
          }
          get #allFields() {
            return ['lastName', 'aboutMe', 'fullName'];
          }
          then(resolve, reject) {
            this.#allValues
              .then(values => {
                if (resolve) {
                  resolve(Object.fromEntries(this.#allFields.map((field, i) => [field, values[i]])))
                }
              })
              .catch(err => { 
                if( reject) { reject(err) } 
              });
          }
          catch(err) {
            this.then(undefined, err);    
          }
```

habdelra: in terms of your question "why would we call the error". It's not that "we" as in our schema class would ever call the error--but rather if we want to behave like a promise we need to provide a catch that allows consumers of the promise to handle errors. For instance, the process of getting a value from one of our fields (the `Promise.all`) may trigger an error in a computed (it is user code after all--as well as our own hub code may encounter an error. i believe for `async` functions, when you wrap a `try/catch` around an `await`, that is another way to provide a `.catch` error callback to the promise. Node should know how to go between these 2 different patterns for error handling. In terms of being good citizens, if we are going to say that our Schema classes implement a `Promise` interface, it would behoove us to provide all the promise capabilities. We will also need to do this for our `FieldGetters` as well.

packages/hub/node-tests/@core/compiler-test.ts
habdelra: looking thru the docs on MDN, it looks like the `cb` param is required, so the `if` that I'm doing is unnecessary. and It is probably better to use the idea I was sketching out above with the `this.#allValues` getter that is returning the `Promise.all()` for getting all our values. if we leverage that, then we can have a finally that looks like:
```js
finally(cb) {
  this.#allValues.finally(cb);
}
```

packages/hub/node-tests/@core/compiler-test.ts
habdelra: turns out private methods pull in lots of babel custom code and significantly rewrites our transpiled schema. In the interest of nice clean schema sources, I'm gonna just use normal class methods....

mattmcmanus:
@habdelra Thanks for the explanations. I hadn't fully understood all the implications of making the schema itself thennable.

packages/hub/node-tests/@core/compiler-test.ts
```diff
@@ -488,12 +488,19 @@ if (process.env.COMPILER) {
...
         expect(source).to.containsSource(`
+          #allValuesPromise;
+
           get allFields() {
             return ["lastName", "aboutMe", "fullName"];
           }
+
           get allValues() {
-            return Promise.all(this.allFields.map(field => this[field]));
+            if (!this.#allValuesPromise) {
```
mattmcmanus, line 498: I'm not sure if this is something we need to consider now, but what if we need to rerun this? Do we need to wire-up clearing of the promise when it's done so that can happen?

habdelra: there is an open question in the CardModel around how long we hang on to the schema class. right now we just make a new one each time we get a field. but when we get into caching and mutating we'll need to think more deeply about this.

habdelra:
From ed:
- [ ] the allFields could be a static method on the class
- [ ] the allValues can be removed, and we can use a weakmap in a core util to share the promise state
- [ ] use babel.template() instead of the giant nested babel node mess for adding code

habdelra:
@ef4 i've incorporated your review comments around not polluting the schema class namespace and to use babel.template() to construct the promise functions in the schema class

ef4 (approved):

PR #2642, opened 2022-02-01
Boxel field refactor

burieberry:
#CS-2764
#CS-2766

field

form-field


backspace (approved):
This looks good to me, thanks for the refactoring. I’ve added @Aierie as a reviewer, as the one who requested this.

Aierie (approved):
This seems good, thanks, and also thanks for clearing the unused components 😄 !  I noticed 2 edge cases that I don't think need immediate attention:

1. The field component does not respect its vertical argument when there is a validation error message in a child input. This means a vertical input might suddenly switch to horizontal, which seems like odd behaviour. This is because of `grid-column` being used in `Boxel::Input`, which is supplying this styling because it assumes that `Boxel::Field` behaves in a certain way. We do not have any vertical field inputs at the moment.

The Boxel field component does not respect its vertical argument when there is a validation error message in a child input

2. I think that we could treat this as an exception, there is also a vertical + centered + bottom case that the field component does not cover:

a prepaid card with a boxel field component where the text is left-aligned where it should be centered

burieberry:
> This seems good, thanks, and also thanks for clearing the unused components 😄 ! I noticed 2 edge cases that I don't think need immediate attention:
> 
> 1. The field component does not respect its vertical argument when there is a validation error message in a child input. This means a vertical input might suddenly switch to horizontal, which seems like odd behaviour. This is because of `grid-column` being used in `Boxel::Input`, which is supplying this styling because it assumes that `Boxel::Field` behaves in a certain way. We do not have any vertical field inputs at the moment.
> 
> The Boxel field component does not respect its vertical argument when there is a validation error message in a child input
> 
> 2. I think that we could treat this as an exception, there is also a vertical + centered + bottom case that the field component does not cover. May be good to fix (prepaid card workflow):
> 
> a prepaid card with a boxel field component where the text is left-aligned where it should be centered

Thanks, good point! I'll merge this and take a look at them this afternoon

PR #2643, opened 2022-02-01
Card space details card 

jurgenwerk:
Ticket: [CS-2990](https://linear.app/cardstack/issue/CS-2990/add-card-space-details-card)

image

image

TODO:
- [x] Tests (fix existing ones and add new `fillIn`s in the acceptance tests)
- [ ] Adjust cover image thumbnail dimensions (currently it's a square but would be better if it was a rectangle, like in the preview)
- [ ] Consider DRYing up the profile & cover image uploading logic


packages/web-client/app/components/card-space/create-space-workflow/index.hbs
```diff
@@ -24,21 +24,22 @@
 
 
 {{#unless this.workflow.isCanceled}}
+  {{!-- CardSpace::ProfileCard --}}
```
jurgenwerk, line 26: Added this comment so it's easier to find the `CardSpace::ProfileCard` component when searching the codebase

jurgenwerk:
@backspace @Aierie can someone please continue with this ticket while I'm on vacation? 🙏 The functionality is mostly done but there's a couple of TODOs left (in the description). Thanks!

Aierie:
Leaving image upload and edit related refactors and improvements for a separate PR:

> - Adjust cover image thumbnail dimensions (currently it's a square but would be better if it was a rectangle, like in the preview)
> - Consider DRYing up the profile & cover image uploading logic

Aierie:
Made cover image optional because profile image is optional too, seems to make sense that cover image is.

For the future, if it sounds ok:
I think the way the form fields are split (display name v details) at the moment is somewhat arbitrary. We could split it into required fields and non-required fields, and have clear messaging that users can ignore the optional fields.

backspace:
This isn’t specific to your work here, which I’m exercising now locally and looks good, but when this card is added to the workflow, it’s scrolled to the bottom, so to fill it out I have to scroll back up to its beginning. Maybe this is something that needs to be fixed in `WorkflowThread`? But something we’ll want to polish before release.

packages/web-client/app/components/card-space/edit-details/description/index.ts
```diff
@@ -0,0 +1,43 @@
+import Component from '@glimmer/component';
+import { WorkflowCardComponentArgs } from '@cardstack/web-client/models/workflow';
+import { action } from '@ember/object';
+import { tracked } from '@glimmer/tracking';
+
+export default class CardSpaceEditDetailsDescriptionComponent extends Component<WorkflowCardComponentArgs> {
+  @tracked validationMessage: string | null = null;
+  @tracked description: string | null = null;
```
backspace, line 8: This is quite minor but is there any value in using `null` for the types instead of `undefined`? Then it can be left as the default.

backspace (approved):
👍🏻, agreed that DRYing would be good, but also agreed with @Aierie that it can be accomplished in a new PR 🎉

PR #2644, opened 2022-02-01
Card Space: Remove URL reservation badge and copy mentioning payment

jurgenwerk:
Removes payment related messages, and the URL reservation badge. Because we don't support this yet.

image


backspace (approved):
It’s unfortunate that the Percy diff is unhelpful but I’m not sure what we can do about it 🤔

Aierie:
When we confirm how the create card looks, we should also add a message above it to follow the UI pattern of having explanatory copy be in a message above a card. I'm merging this for now, have added that detail to CS-1939.

PR #2645, opened 2022-02-01
web: Change Card Space subdomains to use config

backspace:
I hate having to create a component file just to import config 😢


jurgenwerk:
What about introducing a config helper? Something like [this](https://github.com/nvdk/ember-config-helper/blob/103202547400b6138a38720fd5de04b9bce8ff4c/addon/helpers/config.js). This could remove the need for creating new component files just to import config

Aierie (approved):

PR #2651, opened 2022-02-02
Computing fields in cardhost

mattmcmanus:


habdelra:
TODOs:
- [ ] field level deserialization in CardModel
- [ ] make the component module in CardModel be ID and load it when we need it (this requires an async data, and also makes the serializer async)
- [ ]cardhost test for setting id on new card
- [ ]handle scenario where we import primitive in schema but never consume it

habdelra:
All the TODO's have been converted into linear issues

PR #2652, opened 2022-02-03
Replace Services/Queries/PushNotificationRegistration#insert with upsert

lukemelia:
- Use postgres syntax to improve reliability of handling multiple registration attempts and reduce code


backspace (approved):
I appreciate this opportunity to “use the platform” 🤩

PR #2653, opened 2022-02-03
Rename xDai Chain to Gnosis Chain

lukemelia:


backspace:
I approved the Percy diff though it seems spurious, hopefully this doesn’t keep happening

image

backspace (approved):
I appreciate how minimal this is 🎉

PR #2654, opened 2022-02-03
web: Add config helper

backspace:
As suggested by @jurgenwerk here:
https://github.com/cardstack/cardstack/pull/2645#issuecomment-1028371355

This throws an error when an unknown config property is used, such as here:

image

It also adds an inline linting plugin to [catch unknown keys](https://github.com/cardstack/cardstack/runs/5113853994?check_suite_focus=true#step:6:45):

image


packages/web-client/app/helpers/config.ts
```diff
@@ -0,0 +1,18 @@
...
+
+  constructor() {
+    super(...arguments);
+
+    this.config = getOwner(this).resolveRegistration('config:environment');
+  }
+
+  compute([path]: [string]) {
+    // @ts-ignore FIXME, can’t figure it out right now
+    return get(this.config, path);
```
jurgenwerk, line 16: I would be fine with `return get(this.config as any, path);`. Not sure if typescript would help us here given it's just a getter to be consumed in templates 🤔

jurgenwerk (approved):
Nice cleanup!

Aierie:
Nice! This should help with CS-2589 too. I feel like there's not much point in Typescript for the helper right now because it is used in the template, so we don't really get the typechecking benefits until we use something like Glint. Would it be ok to use `config: typeof config`, where `config` is `import config from '@cardstack/web-client/config/environment';`, or simply `config: any`?  

What do you think about adding a Sentry breadcrumb, or event for config helper calls that return undefined?

packages/web-client/app/helpers/config.ts
```diff
@@ -0,0 +1,18 @@
+import { getOwner } from '@ember/application';
+import Helper from '@ember/component/helper';
+import { get } from '@ember/object';
+
+export default class extends Helper {
+  config: Config | undefined;
```
Aierie, line 6: Config is pointing to qunit.Config here. The union type with undefined makes any property access invalid to Typescript, since undefined has no keys.

packages/web-client/app/helpers/config.ts
```diff
@@ -0,0 +1,24 @@
...
+  constructor() {
+    super(...arguments);
+
+    this.config = getOwner(this)
+      .resolveRegistration('config:environment')
+      .clone();
+
+    this.config.unknownProperty = (path: string) => {
+      throw new Error(`Unknown config property: ${path}`);
+    };
```
backspace, line 18: I added the `clone` above to avoid attaching `unknownProperty` to the global-esque `config:environment`. Is it even worth it? It just seemed dirty to me to mutate that upon instantiation of this helper.

backspace: lol `clone` doesn’t exist… now I have removed that, but if mutation-avoidance seems important, `JSON.parse(JSON.stringify` is one option, sadly hideous though.

Aierie: `unknownProperty` doesn't seem to be called with missing, nested properties. Maybe the nested object that is missing the property needs to also implement `unknownProperty`? 

I think it's probs better to throw in the compute function if it's unknown, given this limitation.

packages/web-client/app/components/card-pay/create-merchant-workflow/prepaid-card-choice/index.hbs
```diff
@@ -64,15 +64,15 @@
       </Boxel::Field>
     {{/if}}
     {{#if this.error}}
-      <CardPay::ErrorMessage data-test-prepaid-card-choice-error-message as |mailToSupportUrl|>
+      <CardPay::ErrorMessage data-test-prepaid-card-choice-error-message>
         {{#if (eq this.error.message "INSUFFICIENT_FUNDS")}}
           It looks like your prepaid card doesn't have enough funds to pay the {{format-amount this.merchantRegistrationFee 0}} SPEND ({{format-usd (spend-to-usd this.merchantRegistrationFee)}}) business account creation fee. Please try another prepaid card, or buy one in Card Wallet.
         {{else if (eq this.error.message "TIMEOUT")}}
-          There was a problem creating your business account. Please contact <a href={{mailToSupportUrl}} target="_blank" rel="noopener noreferrer">Cardstack support</a> to find out the status of your transaction.
+          There was a problem creating your business account. Please contact <a href={{config 'urls.malToSupportUrl'}} target="_blank" rel="noopener noreferrer">Cardstack support</a> to find out the status of your transaction.
```
Aierie, line 71: Typo, `malToSupportUrl`

packages/web-client/app/components/card-pay/create-merchant-workflow/prepaid-card-choice/index.hbs
```diff
@@ -64,15 +64,15 @@
...
         {{else if (eq this.error.message "USER_REJECTION")}}
           It looks like you have canceled the request in your wallet. Please try again if you want to continue with this workflow.
         {{else}}
-          There was a problem creating your business account. This may be due to a network issue, or perhaps you canceled the request in your wallet. Please try again if you want to continue with this workflow, or contact <a href={{mailToSupportUrl}} target="_blank" rel="noopener noreferrer">Cardstack support</a>.
+          There was a problem creating your business account. This may be due to a network issue, or perhaps you canceled the request in your wallet. Please try again if you want to continue with this workflow, or contact <a href={{config 'urls.malToSupportUrl'}} target="_blank" rel="noopener noreferrer">Cardstack support</a>.
```
Aierie, line 75: Typo, `malToSupportUrl`

packages/web-client/app/helpers/config.ts
backspace: wow, huge oversight, thanks for catching that. Do you mean that it should throw if the `get` returns `null`?

Aierie: > Now I’m wondering if linting might work here, to check against properties known to exist in config 🤔

Oh that sounds even better for non-dynamic (I hope it's always static, dynamic strings for this sounds messy) cases!

PR #2657, opened 2022-02-03
Replace searchOptimizedData with function from compiler

burieberry:
Fixes #cs-2996


burieberry:
@habdelra The change I made is still using the compiled card, since it still needs to follow the adoptsFrom to get ancestor card fields. I tried a couple ideas but didn't seem right. Let me know if I'm missing something simple, or maybe we can take a look together when you have time.

habdelra:
so this is not quite what we had in mind. rather the idea was that the compiler would manufacture this function--via a babel transform (perhaps in the schema module). and that the search index would load the module that has this new function and then call it with the data that it has for the card.

PR #2659, opened 2022-02-04
Update statuspage incident copy

Aierie:
I've updated incident name to be more focused on the user since this is shown in Card Wallet + the DApp's degraded service notices.

### DIFF
--- a/packages/hub/routes/checkly-webhook.ts
+++ b/packages/hub/routes/checkly-webhook.ts
@@ -40,7 +40,7 @@ export default class ChecklyWebhookRoute {
       componentName: 'Archive RPC node',
       incidentName: 'RPC Node Degradation',
       incidentMessage:
-        'We are experiencing degraded service with our archive RPC node. This will result in reduced transaction processing times.',
+        'We are experiencing degraded service with our archive RPC node. This will result in increased transaction processing times.',
     },
     'xdai non-archive health check - late-cold-smoke (eth_blockNumber)': {
       componentName: 'RPC node (non-archive)',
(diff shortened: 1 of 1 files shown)


packages/hub/routes/checkly-webhook.ts
```diff
@@ -38,19 +38,19 @@ export default class ChecklyWebhookRoute {
...
       incidentMessage:
-        'We are experiencing degraded service with our archive RPC node. This will result in reduced transaction processing times.',
+        'We are experiencing degraded service with our archive RPC node. This will result in increased transaction processing times.',
     },
     'xdai non-archive health check - late-cold-smoke (eth_blockNumber)': {
       componentName: 'RPC node (non-archive)',
-      incidentName: 'RPC Node Degradation',
+      incidentName: 'Transactions delayed',
       incidentMessage:
         'We are experiencing degraded service with our non-archive RPC node. This will result in reduced application performance.',
```
Aierie, line 49: Unsure about the implications of "reduced application performance" here, outside of transactions. I think the DApp still might try to request some data from the layer 2 chain directly, but for the most part we are getting our data from the subgraph.

habdelra: Yeah, but the subgraph gets its information from the rpc node that the warning is about.

habdelra: As well as the relay server uses the same rpc node the warning is about. Basically when this node goes down, so does the relay server, which means the apps will not be able to send any txns to layer 2

habdelra (approved):

backspace (approved):
I appreciate the framing of the language being about the effects for the user, as they’ll often be unaware of the precise implications of particular subsystem downtime 😀

PR #2660, opened 2022-02-04
web: Remove TestFlight, add Google store links

backspace:
This adds a `CardWalletBadges` component to consistently include the iOS and Android store badges wherever. It also updates the Card Wallet app-on-phone image and the About URL.


packages/web-client/app/components/card-pay/layer-connect-modal/index.hbs
```diff
@@ -55,7 +55,6 @@
         <CardPay::LayerTwoConnectCard
           class="layer-connect-modal__wrapped-card"
           @suppressHeader={{true}}
-          @includeExplanations={{true}}
```
backspace, line 58: This conditional was hard-coded and the explanation isn’t shown in the design, so I’ve removed it.

packages/web-client/app/components/card-pay/layer-two-connect-card/index.hbs
```diff
@@ -72,11 +72,6 @@
         class="layer-two-connect-card__section layer-two-connect-card__section--one"
         data-test-layer-2-connect-prompt
       >
-        {{#if @includeExplanations}}
-          <p class="layer-two-connect-card__explanation">
-            Please install the Card Wallet app on your mobile phone and create/add an account.
-          </p>
-        {{/if}}
```
Aierie, line 79: I think this text is for use outside of the workflows. Within workflows, this message precedes the card:

> Once you have installed the app, open the app and add an existing wallet/account or create a
 new wallet/account. Use your account to scan this QR code, which will connect your account with Card Pay.
 
Outside workflows, there is no other copy that mentions the step of creating/adding an account. This copy may still be relevant because of that.

Aierie: Also if we remove this text, we should probably also remove the `@includeExplanations` for the section below.

Aierie (changes requested):
Spec for the pay page mobile view is here:

Spec for links to google and apple app stores on the merchant payment request page, the links are vertically stacked to prevent horizontal crowding

I think this makes the links more readable, they currently get squished.

This PR's version of the google and apple app stores on the merchant payment request page, the links are squeezed onto the same line as the header, and become smaller and less readable as a result

PR #2661, opened 2022-02-04
web: Fix missed supportURL conversion

backspace:
This should have been in #2658, but also this CI error should
have failed the build! I’ll examine that separately:
https://github.com/cardstack/cardstack/runs/5059503542?check_suite_focus=true#step:6:44


lukemelia (approved):
☺️

PR #2662, opened 2022-02-04
ci: Fix commands that mask linting errors

backspace:
This removes trailing `cd -` in Yarn scripts that had no effect other than swallowing errors and producing false-positive CI runs in favour of the `--cwd` flag for Yarn, as [suggested by @lukemelia](https://github.com/cardstack/cardstack/pull/2662#discussion_r799868940).

This is meant to fix things like this where linting failed but CI passed:
https://github.com/cardstack/cardstack/runs/5059503542?check_suite_focus=true#step:6:44

I included that problem again and [the build failed](https://github.com/cardstack/cardstack/runs/5073817709?check_suite_focus=true) as desired.


package.json
```diff
@@ -32,15 +32,15 @@
   },
   "scripts": {
     "clean": "git clean -x -f --exclude=packages/*/.env --exclude=cardstack.code-workspace && lerna run clean",
-    "postinstall": "cd ./packages/cardpay-sdk && yarn codegen; cd -",
+    "postinstall": "cd ./packages/cardpay-sdk && yarn codegen",
```
lukemelia, line 35: Consider using the [cwd option](https://stackoverflow.com/questions/46891622/run-yarn-in-a-different-path):

```
"postinstall": "yarn --cwd packages/cardpay-sdk codegen",
```

lukemelia (approved):

PR #2665, opened 2022-02-07
web: Add test failure annotations

backspace:
This adds [`EnricoMi/publish-unit-test-results-action`](https://github.com/marketplace/actions/publish-unit-test-results) for the web client CI jobs. The action adds a comment showing an overview of test results:

image

The comment gets updated upon subsequent runs, which should reduce annoyance. (In this PR it shows twice because I tweaked the default `check_name`.)

If there’s a failure, it links to the “check” that shows it/them ([example](https://github.com/cardstack/cardstack/runs/5097323381)):

image

The ideal arrangement would exercise the action’s ability to directly annotate the failing test, but we couldn’t get that to work on first pass, a possible future improvement.

This works by hacking Testem (copied from [here](https://github.com/hashicorp/nomad/pull/7585)) to produce xUnit output alongside the default [TAP output](https://en.wikipedia.org/wiki/Test_Anything_Protocol). The xUnit output is written to a `test-results` directory in the repository root.

### DIFF
--- a/.github/workflows/pr-web-client.yml
+++ b/.github/workflows/pr-web-client.yml
@@ -46,3 +46,8 @@ jobs:
         if: always()
         with:
           files: test-results/**/*.xml
+      - name: FIXME archive test results
+        uses: actions/upload-artifact@v2
+        with:
+          name: results
+          path: test-results/ui.xml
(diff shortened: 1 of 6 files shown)


Aierie (approved):
Nice!

PR #2667, opened 2022-02-08
Fixes #CS-3085 add support for creating a local card with presupplied ID

habdelra:
Also fixed the issue where we are not able to load modules when you create a card in the local realm from a card in the remote realm: all module leading needs to go thru the card service which knows how to load modules from the right place. Sadly this is not testable, because we cant create cards on the server from our cardhost tests--but you can see this when playing with the demo app.

### DIFF
--- a/packages/cardhost/app/lib/card-model-for-browser.ts
+++ b/packages/cardhost/app/lib/card-model-for-browser.ts
@@ -95,6 +95,11 @@ export default class CardModelForBrowser implements CardModel {
     if (this.state.type !== 'loaded') {
       throw new Error(`tried to adopt from an unsaved card`);
     }
+    let builder = await this.cards.builder();
+    let localRealm;
+    if (realm === builder.realmURL) {
+      localRealm = builder;
+    }
     return new (this.constructor as typeof CardModelForBrowser)(
       this.cards,
       {
@@ -105,7 +110,7 @@ export default class CardModelForBrowser implements CardModel {
         schemaModuleId: this.state.schemaModuleId,
         format: this.state.format,
       },
-      this.localRealm
+      localRealm ?? this.localRealm
     );
   }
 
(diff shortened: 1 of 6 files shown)


packages/cardhost/app/lib/card-model-for-browser.ts
```diff
@@ -262,7 +261,11 @@ export default class CardModelForBrowser implements CardModel {
   serialize(): ResourceObject<Saved | Unsaved> {
     return serializeResource(
       'card',
-      this.state.type === 'loaded' ? this.state.url : undefined,
+      this.state.type === 'loaded'
+        ? this.state.url
+        : this.state.id
+        ? cardURL({ realm: this.state.realm, id: this.state.id })
+        : undefined,
```
mattmcmanus, line 268: This is really unclear

packages/cardhost/app/lib/card-model-for-browser.ts
```diff
@@ -90,22 +92,27 @@ export default class CardModelForBrowser implements CardModel {
     }
   }
 
-  // TODO: add failing test for `_id` being unimplemented here and then fix
-  async adoptIntoRealm(realm: string, _id?: string): Promise<CardModel> {
+  async adoptIntoRealm(realm: string, id?: string): Promise<CardModel> {
     if (this.state.type !== 'loaded') {
       throw new Error(`tried to adopt from an unsaved card`);
     }
+    let builder = await this.cards.builder();
```
mattmcmanus, line 99: This does a lot of work that I don't think we need to do necessarily. The realmURL is a constant exported by the builder that is what your checking when you do builder.realmURL.

packages/cardhost/app/lib/card-model-for-browser.ts
habdelra: sure, i can turn this into if statements instead

PR #2669, opened 2022-02-09
Clean up image uploading and editing

Aierie:
This prepares image uploading components for usage in and out of workflows by separating UI state/trigger concerns from the logic of uploading and editing + hiding the messy and fragile state management of image editing. The image uploading functionality is handled by `common/image-upload-action` and the UI is handled by `common/image-upload-interface`.


backspace (approved):
This looks great, a nice cleanup/rationalisation of the chaos that was growing, and a good use of composability and named blocks 🥳

PR #2671, opened 2022-02-09
add deserialization to CardModel.getField

mattmcmanus:
- [x] Refactor data computing to include field serialization
- [x] Refactor serializerMap to be `Record<string, PrimitiveSerializer>`


habdelra (approved):
After our discussion today I’m good with this. We basically concluded that we should stop fighting against the eager data loading for a format when a card model is loaded.

ef4 (approved):

PR #2672, opened 2022-02-09
hub: Add Prisma experiments

backspace:
This was a struggle to get working but I finally could get the changed test passing locally with this environment variable override:

```
PRISMA_QUERY_ENGINE_BINARY=$(pwd)/node_modules/.prisma/client/query-engine-darwin yarn autotest
```

This will therefore presumably fail in CI, though I wouldn’t be surprised if it magically passed 😬

Select todos:
* [ ] service-ify it in a way that can provide test environment transactions with rollback
  * would probably also permit the database to be configured when instantiating the client vs the current requirement of `DATABASE_URL` in `.env`, which isn’t test-friendly
* [ ] expand it to more models
* [ ] eliminate the weird environment variable workaround


backspace:
Another finding: if I remove the `engineType = "binary"` from `schema.prisma` and use this environment variable, the test also passes:
```
PRISMA_QUERY_ENGINE_LIBRARY=$(pwd)/node_modules/.prisma/client/libquery_engine-darwin.dylib.node
```

There’s probably some Webpack tweak we can do to fix this in a deeper way.

jurgenwerk:
@backspace that looks good to me! The compound unique key querying does look awkward but I'm ok with it.

PR #2677, opened 2022-02-10
Do not send user transaction signing rejection errors to Sentry.

lukemelia:


backspace (approved):
Thanks for this, every little bit helps to reduce noise in Sentry that obscures true problems 🎉

lukemelia:
>The interception here is happening in the hub though, and we need it in web client

lol, good catch, thanks!

PR #2678, opened 2022-02-11
Remove spend display in prepaid cards, show usd only

jurgenwerk:
Ticket: [CS-3078](https://linear.app/cardstack/issue/CS-3078/update-prepaid-card-component-web)

Looks like this now:

image

image


jurgenwerk:
Hold off with the review for a bit please, I started observing floating point display errors, also need to fix the tests.

Aierie:
This should be paired with CS-3089, I think.

packages/web-client/app/helpers/format-amount.ts
```diff
@@ -14,9 +14,18 @@ export function formatAmount(
     return '';
   }
 
+  let amountWithoutFloatingPointRoundingError = parseFloat(
```
jurgenwerk, line 17: Looks like `spend-to-usd` sometimes produces floats like `23.240000000000002` which will mess up the currency display and show "$23.240000000000000 USD" in this case. However I'm not sure this is the right place to handle this, maybe a better place would be a function in the SDK from where all number/currency related helpers could import.

jurgenwerk:
Ok, fixed the tests and addressed the floating point issue. @Aierie thanks for pointing out that ticket, agree it would make sense to do that at the same time.

jurgenwerk:
I removed SPEND in all other places I could find. Here are the comparisons before - after:

image

image

image

jurgenwerk:
Does it still make sense to have `spendFaceValue` and other spend based calculations in the code now that we don't show SPEND anymore? Is it likely we will reintroduce it later on?

Aierie (approved):
Looks good, have added a PR for removing SPEND from the merchant payment request page as well (https://github.com/cardstack/cardstack/pull/2688).

Aierie:
> Does it still make sense to have spendFaceValue and other spend based calculations in the code now that we don't show SPEND anymore?

I think so, since the prepaid card balance data is read as SPEND. It also feels better to have a whole number being used for calculations as opposed to floating point USD. Not sure if we will reintroduce SPEND in the UI.

PR #2679, opened 2022-02-11
configure prepaid card asset compression

Aierie:
This configures `ember-cli-deploy-compress` to use gzip for the prepaid card patterns.

### DIFF
--- a/packages/web-client/config/deploy.js
+++ b/packages/web-client/config/deploy.js
@@ -18,6 +18,9 @@ module.exports = function (deployTarget) {
         s3: {
           as: ['s3Pages', 's3WellKnown', 's3Assets'],
         },
+        compress: {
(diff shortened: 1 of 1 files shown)


Aierie:
The resulting prepaid card customization svgs will have a `content-encoding` of `gzip` in S3. We don't configure cloudfront to compress things right now, and adding configuration for compression there will still respect the provided `content-encoding` from s3.

oidoug:
Great, thanks for the quick response :)
Is there a way to test this before merging? Maybe a staging bucket?

Aierie:
> Great, thanks for the quick response :)
> 
> Is there a way to test this before merging? Maybe a staging bucket?

There is a dev bucket up, will send you the url in a bit.

lukemelia (commented):
👍 👍 👍

PR #2682, opened 2022-02-14
web: Change Card Space hostname suffix

backspace:
This shouldn’t be merged until card.xyz is in Cloudflare.


backspace:
This is passé with the creation of `ssr-web`, I’ll recreate it there.

PR #2684, opened 2022-02-14
CS-3106 refactor computeds: synchronous computed support

habdelra:
This PR refactors computeds and provides synchronous computed support. The async computed support will arrive in the next PR


mattmcmanus (approved):
I'm still not super clear on the full scope of the computeds refactor, so I can't comment on whether this is in the right direction or not. But from an isolated-to-this-pr view, this looks good.

ef4 (approved):

PR #2690, opened 2022-02-15
[CS-3174] Dedicated module for Component meta information

mattmcmanus:
My main question is whether ComponentMeta is too similar to ComponentInfo


ef4:
I don't think ComponentMeta needs to be a thing. There's no reason to export one object with three things on it, export the three things separately. That collapses the ComponentMeta type into the ComponetMetaModule type.

You can still do the reexport, using "*":

```diff
-export { ComponentMeta } from %%metaModulePath%%;
+export * from %%metaModulePath%%;
```

mattmcmanus:
Closing in favor of https://github.com/cardstack/cardstack/pull/2699

PR #2691, opened 2022-02-15
Async Computed Fields

habdelra:
~~This is basically the prep work for the computed async. I thought i'd break this down so the final PR isn't super huge. This includes some new FieldMeta for the async computeds (`computeVia`), as well as the transpiled schema source and new `NotReady` error. the next step will be to harness the updated Schema and `NotReady` errors for the async computeds. This source code follows the sketch in https://linear.app/cardstack/issue/CS-3106/refactor-computeds~~

UPDATE: it turns out the extra work for handling the `NotReady` errors and loading the async is actually not that much more. I'll just include into this PR too. This PR includes the remaining async computed support for both hub and browser.


habdelra:
Oh also, i'll will out of the office on thurs and friday. If there are no review comments by wed afternoon, I'll optimistically merge this so I can get as much done on the followup PR as I can before I take off. but still feel free to leave comments after the merge and I'll address that in followup PR's

mattmcmanus (approved):
This did turn out well!

PR #2694, opened 2022-02-17
Further ssr web cleanup

Aierie:


Aierie:
If we are going to try running a server to test for returned html, we might need to keep the test-support CI?

backspace (approved):
Thanks for these 🎉

PR #2695, opened 2022-02-17
Throw on reverted receipt

Aierie:
Make the SDK throw when a transaction receipt is for a reverted transaction, prevent returning reverted transaction receipts by banning the `TransactionReceipt` type in favour of `SuccessfulTransactionReceipt`.

Part of CS-2502

### DIFF
--- a/packages/cardpay-sdk/sdk/token-bridge-home-side.ts
+++ b/packages/cardpay-sdk/sdk/token-bridge-home-side.ts
@@ -1,5 +1,5 @@
 import Web3 from 'web3';
-import { TransactionReceipt } from 'web3-core';
+import type { SuccessfulTransactionReceipt } from './utils/successful-transaction-receipt';
 import { ContractOptions, EventData } from 'web3-eth-contract';
 import ERC677ABI from '../contracts/abi/erc-677';
 import HomeAMBABI from '../contracts/abi/home-amb';
@@ -17,7 +17,7 @@ import { TransactionOptions, waitForSubgraphIndexWithTxnReceipt, isTransactionHa
 // The Foreign network can be any chain, but generally refers to the Ethereum mainnet.
 
 export interface ITokenBridgeHomeSide {
-  waitForBridgingToLayer2Completed(recipientAddress: string, fromBlock: string): Promise<TransactionReceipt>;
+  waitForBridgingToLayer2Completed(recipientAddress: string, fromBlock: string): Promise<SuccessfulTransactionReceipt>;
 }
 
 export interface BridgeValidationResult {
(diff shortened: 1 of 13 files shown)


packages/cardpay-sdk/sdk/token-bridge-home-side.ts
```diff
@@ -229,11 +232,16 @@ export default class TokenBridgeHomeSide implements ITokenBridgeHomeSide {
         `Timed out waiting for tokens to be bridged to layer 2 for safe owned by ${recipientAddress} after block ${fromBlock}`
       );
     }
+    // TODO: can this use waitForSubgraphIndex?
```
lukemelia, line 235: Can we answer or remove this question?

packages/cardpay-sdk/sdk/token-bridge-home-side.ts
```diff
@@ -229,11 +232,16 @@ export default class TokenBridgeHomeSide implements ITokenBridgeHomeSide {
...
     let {
       transaction: { id: txnHash },
     } = receivedBridgedTokens[0];
     let receipt = await this.layer2Web3.eth.getTransactionReceipt(txnHash);
-    return receipt;
+    if (receipt.status === true) {
+      return receipt as SuccessfulTransactionReceipt;
+    } else {
+      throw new Error(`Transaction with hash "${txnHash}" was reverted`);
```
lukemelia, line 243: Is there any other information available in this circumstance that would be useful to include as part of the error?

lukemelia (approved):

packages/cardpay-sdk/sdk/token-bridge-home-side.ts
Aierie: I'm not very clear what revert reasons we should expect for this scenario, do you know any that we need special cases for @habdelra?

packages/cardpay-sdk/sdk/token-bridge-home-side.ts
Aierie: I'm understanding this transaction to be the "minting" transaction on the layer 2 side, in the bridging process. It isn't directly user-initiated, so we don't currently have an easy way to get the transaction hash and wait for the subgraph to index that transaction. 

I've removed the comment without changing the existing implementation, please do correct me if I've misunderstood!

packages/cardpay-sdk/sdk/token-bridge-home-side.ts
habdelra: I don't think these txn's will actually ever revert since they are created programmatically via bridge validators. These types of txn's are use an ethereum `call()` API which allows the client to perform a dry run of the txn before the real txn. If the `call()` reverts, then the bridge validator will never actually send the txn for real (our relay server works the same way).

habdelra (commented):
This looks great! thx @Aierie

PR #2697, opened 2022-02-17
ssr: Add mobile Percy snapshots

backspace:
This is actually [the default](https://docs.percy.io/docs/cli-configuration#snapshot) so it could be removed entirely but I’m ambivalent about that. Thoughts?


Aierie:
I think explicitly specifying it seems good, makes for less guessing/shifting when defaults changing.

Aierie (approved):
Thanks!

PR #2699, opened 2022-02-17
[CS-3087] Primitive cards can provide a custom serializer

mattmcmanus:
Ok, this is a substantial change. From a high level: 

- It includes the commits from https://github.com/cardstack/cardstack/pull/2690, which cause us to generate a dedicated module for meta information about components
- Set's up the APIs for card to provide their own serializers

There are two major loose ends that needed to require the dedicated Component Meta module has brought about:

1. Now that there are generated files that are needed at runtime, our previous optimizations need to be applied to the file-cache in some way. I've documented my current approach, which works for now. If we're fine with keeping the file-cache around longer, we would need to come up with scenarios where we would want the cache cleared and how that would be triggered
2. Our current file-cache.`loadModule` approach has no obvious way to support npm packages. The serializerMap, exported by the Component Meta module, includes direct references to the serializer module code itself. So when we require it, we necessarily need to require the serializers, which have package deps. Is there a smart way to do this?

### DIFF
--- a/packages/core/src/babel-plugin-card-component-meta.ts
+++ b/packages/core/src/babel-plugin-card-component-meta.ts
@@ -1,0 +1,94 @@
+// import ETC from 'ember-source/dist/ember-template-compiler';
+// const { preprocess, print } = ETC._GlimmerSyntax;
+
+import { transformSync } from '@babel/core';
+import { NodePath } from '@babel/traverse';
+import type * as Babel from '@babel/core';
+import type { types as t } from '@babel/core';
+
+import { CompiledCard, SerializerName, SerializerMap } from './interfaces';
+
+import { buildSerializerMapFromUsedFields } from './utils/fields';
+import { augmentBadRequest } from './utils/errors';
+
+export interface CardComponentMetaPluginOptions {
+  debugPath: string;
+  fields: CompiledCard['fields'];
+  // these are for gathering output
+  usedFields: string[];
+  serializerMap: SerializerMap;
+}
+
+interface State {
+  opts: CardComponentMetaPluginOptions;
+}
+
+export default function (options: CardComponentMetaPluginOptions): { source: string; ast: t.File } {
+  try {
+    let out = transformSync('', {
+      ast: true,
+      plugins: [[babelPluginComponentMeta, options]],
+      // HACK: The / resets the relative path setup, removing the cwd of the hub.
+      // This allows the error module to look a lot more like the card URL.
+      filename: '/' + options.debugPath.replace(/^\//, ''),
+    });
+    return { source: out!.code!, ast: out!.ast! };
+  } catch (e: any) {
+    throw augmentBadRequest(e);
+  }
+}
+
+export function babelPluginComponentMeta(babel: typeof Babel) {
+  return {
+    visitor: {
+      Program: {
+        exit(path: NodePath<t.Program>, state: State) {
+          addComponentMetaExport(path, state, babel);
+        },
+      },
+    },
+  };
+}
+
+function addComponentMetaExport(path: NodePath<t.Program>, state: State, babel: typeof Babel) {
+  let t = babel.types;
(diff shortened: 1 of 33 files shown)


packages/hub/node-tests/@core/card-model-test.ts
```diff
@@ -182,7 +182,9 @@ if (process.env.COMPILER) {
       delete result.meta;
 
       let expectedAttributes = cloneDeep(attributes);
-      set(expectedAttributes, 'address.zip', null);
+      // TODO: We currently do not store enough info in the database
+      // to know this should be her
+      // set(expectedAttributes, 'address.zip', null);
```
mattmcmanus, line 187: The changes I've made to the serializer seemed to have removed the ability to maintain the proper data shape. I'm not sure why.

packages/hub/node-tests/helpers/cards.ts
```diff
@@ -7,12 +7,14 @@ import { CardService, INSECURE_CONTEXT } from '../../services/card-service';
 import FileCache from '../../services/file-cache';
 import { TEST_REALM } from '@cardstack/core/tests/helpers';
 import { RealmConfig } from '@cardstack/core/src/interfaces';
-import { BASE_REALM_CONFIG, DEMO_REALM_CONFIG } from '../../services/realms-config';
+import { BASE_REALM_CONFIG } from '../../services/realms-config';
 
 tmp.setGracefulCleanup();
 
+let fileCacheTmpDir = tmp.dirSync();
```
mattmcmanus, line 14: This change, along with the removal of the cache-dir cleanup on Container teardown allows us to use previously generated ComponentMeta files. This is because our file-cache has optimizations so that we do not re-index a realm if the files haven't changed. The problem is, we clear the cache between each test, and the newly generated meta modules are not being recreated.

packages/core/src/babel-plugin-card-component-meta.ts
```diff
@@ -0,0 +1,148 @@
...
+    let { fields } = field.card;
+    for (const name in fields) {
+      addFieldToSerializerMap(map, fields[name], `${usedPath}.${name}`, path, state);
+    }
+  } else {
+    if (!field.card.serializerModule) {
+      return;
+    }
+    // TEMP: Would be nice to have the Proper CardID on the compiled card
+    let cardId = field.card.url.split('/')[field.card.url.split('/').length - 1];
```
mattmcmanus, line 140: This is gross, but with the info we have here this is the best we can do. Basically, we have a Date fields under the key birthdate, I want to add an import for it's serializer with the name "DateSerializer". The `Field` definition and the compiled card have know easy way to say that it's a date field. An ideal solution would be to make a similar change that we made to RawCard and have a dedicated ID field.

habdelra: i don't know if its a same assumption to use the last path segment as the ID. the realm itself is responsible for its own ID's and it might decide to create an ID with a "/" in it. but yeah, a dedicated "realm ID" field is probably not a bad idea

packages/hub/services/file-cache.ts
```diff
@@ -160,7 +165,19 @@ export default class FileCache {
       }
       let context = {
         exports: {},
-        require: (specifier: string) => this.loadModule(specifier),
+        require: (specifier: string) => {
+          // TODO: We need to figure out a pattern for importing npm code
```
habdelra, line 169: i  can only see this growing. perhaps we want to create an object whose keys are the specifier values are the exported functions, and have a separate module whose whole job is to reexport allowed npm deps into a specifier based POJO. and then in this require we would check to see if the specifier matches a key in that POJO and return it's value, or otherwise call `this.loadModule(specifier)` if the specifier key doesn't exist in the allowed npm deps POJO

mattmcmanus: Yeah, some basic structure like that seems like a good idea. We currently have another place where we need to do that module tracking manually, the compiled package.json.

PR #2700, opened 2022-02-17
CS-3098: Boxel::Field w/ error state input and special display case 

burieberry:
- Added an example of Boxel::Field to the docs with an error state validation input:
validation-field

- Added new API parameter: `centeredDisplay`. It centers the field contents and places the label under the value:
centered-display-option

- Used the new param `centeredDisplay` in web-client. It's used in various places where prepaid-card and merchant previews are displayed:
special-centered-display-2
special-centered-display-1

### DIFF
--- a/packages/boxel/addon/components/boxel/input/index.hbs
+++ b/packages/boxel/addon/components/boxel/input/index.hbs
@@ -19,7 +19,7 @@
       ...attributes
     />
     {{#if shouldShowErrorMessage}}
-      <div id={{concat "error-message-" helperId}} class="boxel-input__error-message" data-test-boxel-input-error-message>{{@errorMessage}}</div>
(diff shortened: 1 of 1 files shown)


Aierie (approved):
Thanks, I think this looks good - would go with merging after rebasing and confirming that there are no percy issues. 

The a11y error is about `aria-errormessage` possibly not being able to be handled by assistive tech. The original author explains here: https://github.com/dequelabs/axe-core/issues/499#issuecomment-325383562.

One solution is making `aria-describedby` point to the error message element (`error-message-${id}`) while there is an error. This means that browsers/screen readers that don't support `aria-errormessage` will still be able to understand what element is providing the context. 

Alternative solution is adding one of these attributes to the error message div (I'm guessing so it announces itself?):

```
role='alert'
aria-live='assertive'
aria-live='polite'
```

See the source that was generating this error here:

https://github.com/dequelabs/axe-core/blob/93a765c372b29a0764872f50f044bfbabed68207/lib/checks/aria/aria-errormessage-evaluate.js#L50-L53

PR #2702, opened 2022-02-18
Fix/clear transaction hash on error

Aierie:
Clear transaction hashes when a transaction errors so we don't end up retrying a transaction that should fail. Handling of resumed transactions should come with https://github.com/cardstack/cardstack/pull/2695


jurgenwerk (approved):
Nice!

PR #2705, opened 2022-02-21
Add a tooltip modifier

Aierie:
## What
Use a `{{tippy}}` modifier to replace the `title` attribute. 

## Why
This means less of a delay for users to get the tooltip, more readable text with styling that better fits the DApp, and better touch support.  The original task (CS-2889) was only for business accounts, but replacing only one `title` while leaving the rest meant that we would have 2 different ways of showing tooltip-ish stuff, so all `title`s were replaced.

Using a modifier + limiting options for styling while using sensible library-provided defaults feels better than using a full component for this, since we should not have to support functionality/UI that is too complex within a tooltip.

## Usage
Plain text content: `{{tippy "Some string"}}`
HTML content: `{{tippy (html-safe "<div style=\"color:red;\">Some string</div>")}}`
Options:  `{{tippy "Some string" ...tippy options here}}`

#### Theme
Following tippy's API, use themes for custom, targeted styling of groups of tooltips. Add a theme by providing the CSS (see the new `tippy.css` file, we basically just match `data-theme`, which is set by tippy), and by specifying the theme in the tippy modifier. 

```
{{tippy "Content" theme="my-theme"}}
{{tippy "Content" theme="light-border address"}} // multiple themes separated by space
```

The default theme, if unspecified, is the tippy-provided `light-border`. I've added an `address` theme as well so that the font difference is not too jarring when hovering an address, and also for text wrapping styles.

## Shots
Note that all tooltips are shown here only for ease of viewing changes, these will only show one-by-one as their targets are interacted with.

Prepaid card with tooltips showing its full address, which was truncated, and its issuer name, which potentially can get truncated

A business account with its name truncated to two lines. A tooltip shows its full name. A separate tooltip shows its id, which is not truncated but potentially could be.

The workflow tracker with a tooltip shown on a button with a trash can icon, the tooltip's text reads 'Abandon workflow'.

## Drawbacks
We are largely adding this to non-interactive content which means we either: 
- add non-interactive items to the tab order to show tooltips for plain text, OR
- don't show tooltips for plain text for keyboard-exclusive users

Neither of these are very nice, though I am not very clear what the impact is of choosing either just yet. I've opted to not show tooltips for keyboard-exclusive users for now, as this involves less change than adding many random items to the tab order and potentially making someone have to sift through a lot of noninteractive plain text to get to important things. 

While this use of a tooltip is an improvement over `title`, I hope that we can provide design solutions that don't truncate static text in the future to replace this solution.

## Some changes I'm considering
- Don't allow HTML by default, make the user have to specify `allowHTML=true` in the modifier arguments
- Tooltips are currently left aligned (if space allows) to make the association between them and their target elements clearer. Design spec appeared to specify center. 
- Typography tweaks. 
- Turning off tooltips in specific situations (a prepaid card in selection)
Prepaid cards in a prepaid card selector still can show their tooltips for address and name.

### DIFF
--- a/packages/boxel/addon/components/boxel/card-catalog-tray-item/index.css
+++ b/packages/boxel/addon/components/boxel/card-catalog-tray-item/index.css
@@ -68,7 +68,7 @@
   grid-area: description;
   height: 1.875rem;
   margin: 0;
-  display: box;
+  display: -webkit-box;
   -webkit-box-orient: vertical;
   -webkit-line-clamp: 2;
   overflow: hidden;
--- a/packages/boxel/tests/dummy/app/css/components/queue-card.css
+++ b/packages/boxel/tests/dummy/app/css/components/queue-card.css
@@ -79,7 +79,7 @@
   font: var(--boxel-font-sm);
   font-weight: 700;
   letter-spacing: var(--boxel-lsp);
-  display: box;
+  display: -webkit-box;
   -webkit-line-clamp: 2;
   -webkit-box-orient: vertical;
   overflow: hidden;
--- a/packages/web-client/app/components/card-pay/header/workflow-tracker/item/index.hbs
+++ b/packages/web-client/app/components/card-pay/header/workflow-tracker/item/index.hbs
@@ -33,8 +33,9 @@
         type="button"
         {{on 'click' this.showDeleteConfirmation}}
         data-test-delete-workflow-button
+        {{tippy "Abandon workflow"}}
       >
-        {{svg-jar 'trash' width="16" title="Abandon workflow"}}
+        {{svg-jar 'trash' width="16"}}
       </button>
     {{else}}
       <Boxel::ProgressIcon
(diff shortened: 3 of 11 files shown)


backspace (approved):
This works well locally. Do you think we’ll want this in `ssr-web`? 😬

re accessibility, maybe this opens things toward an ARIA attribute that can operate similarly to `title`. But I’m not deeply familiar

Aierie:
> This works well locally. Do you think we’ll want this in `ssr-web`? 😬
> 
> re accessibility, maybe this opens things toward an ARIA attribute that can operate similarly to `title`. But I’m not deeply familiar

Tippy does add `aria-describedby` to its target when the tooltip is shown!

Aierie:
I don't think we need this in SSR Web for truncation purposes yet, but one place where this could be useful is to alert a user that they haven't installed Card Wallet when they click a payment link but nothing happens because there is no protocol handler for `cardwallet://`.

PR #2706, opened 2022-02-22
Update safe signing to support 1.3 safes

alex-cardstack:
1.3 safes need the chain ID as part of the EIP712Domain, which prevents cross-chain replay


habdelra:
i think merging upstream into this PR will fix the percy failure

alex-cardstack:
I merged upstream and you were right, percy now passing

PR #2708, opened 2022-02-22
SAVEPOINT only works when there is an existing wrapper transaction

jurgenwerk:
Using `SAVEPOINT` results in an error when there is no existing transaction. This worked in tests but not in other environments where this query is not inside another transaction. So the solution is to add a transaction and commit, but not in test environment.


Aierie (approved):
Exercised this locally, creating businesses works with this change. Thanks!

PR #2709, opened 2022-02-22
waypoint-vars script to get waypoint variables

Aierie:
Waypoint variable-getting is often hard to read and often happens in the course of node backend tasks. This adds a QOL script to help with that.

With no arguments specified, this prints a table, truncating to keep values within one line.

Table of sample keys and values from the w-var script. A long value is truncated.

With arguments, this checks for keys that contain those strings and prints the key-value pairs out in full as JSON:
```
// yarn waypoint-vars 2
{
  sampleKey2: 'sampleLongValue sampleLongValue sampleLongValue sampleLongValue sampleLongValue sampleLongValue sampleLongValue sampleLongValue'
}

// yarn waypoint-vars sample
{
      sampleKey: 'sampleValue',
      sampleKey2:
        'sampleLongValue sampleLongValue sampleLongValue sampleLongValue sampleLongValue sampleLongValue sampleLongValue sampleLongValue',
}
```


backspace (approved):
I tried this locally and it worked nicely. Discoverability is a perpetual problem with this kind of ergonomic improvement, maybe a README mention is worth it?

jurgenwerk:
Thanks for this! My preference is to inspect the variables scoped to the app. I wanted to add this feature to this PR, but it seems that I can't get it to work 😵‍💫 The [docs say](https://www.waypointproject.io/commands/config-get) there can be a flag provided, i.e. `waypoint config get -json -app=hub` but the `-app` flag does nothing when I try it - the output still contains all apps (hub, hub-worker, hub-bot, relay-web,...).

package.json
```diff
@@ -76,7 +77,8 @@
     "status:subgraph-sokol-green": "yarn --cwd packages/cardpay-subgraph sokol-green-status",
     "status:subgraph-sokol-blue": "yarn --cwd packages/cardpay-subgraph sokol-blue-status",
     "status:subgraph-xdai-green": "yarn --cwd packages/cardpay-subgraph xdai-green-status",
-    "status:subgraph-xdai-blue": "yarn --cwd packages/cardpay-subgraph xdai-blue-status"
+    "status:subgraph-xdai-blue": "yarn --cwd packages/cardpay-subgraph xdai-blue-status",
+    "w-var": "node ./scripts/w-var.js"
```
jurgenwerk, line 81: I'm also ok with naming this command `waypoint-vars` to make it clearer when glancing over package.json commands

Aierie: That seems good, I wasn't sure if that would be too verbose but I think it makes sense to make it clear rather than save on typing a few letters.

Aierie:
> Thanks for this! To reduce noise my preference is to inspect the variables scoped to the app that interests me. I wanted to add this feature to this PR, but it seems that I can't get it to work 😵‍💫 The [docs say](https://www.waypointproject.io/commands/config-get) there can be a flag provided, i.e. `waypoint config get -json -app=hub` but the `-app` flag does nothing when I try it - the output still contains all apps (hub, hub-worker, hub-bot, relay-web,...).
> 
> There's another person experiencing the same issue in [hashicorp/waypoint#3031](https://github.com/hashicorp/waypoint/issues/3031). Looks like a bug in waypoint?

I have the same issue with the app flag. It does nothing for me.

PR #2710, opened 2022-02-22
Business account name char limit

Aierie:
CS-3219 - Limit business names to 50 characters.

### DIFF
--- a/packages/hub/node-tests/routes/merchant-infos-test.ts
+++ b/packages/hub/node-tests/routes/merchant-infos-test.ts
@@ -181,6 +181,35 @@ describe('POST /api/merchant-infos', function () {
       .expect('Content-Type', 'application/vnd.api+json');
   });
 
+  it('validates name for length', async function () {
+    const payload = {
+      data: {
+        type: 'merchant-infos',
+        attributes: {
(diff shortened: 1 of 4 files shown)


Aierie:
I wonder whether there is a good way to keep web client and hub in sync on their validations?

backspace (approved):
I sometimes see this kind of restriction expressed in the database vs enforced before writing to the database, but this way seems fine.

re sharing, I’m not clear on the build implications, but could we have a package to host this kind of config that both Hub and web client (and SSR) could consume?

Aierie:
> I sometimes see this kind of restriction expressed in the database vs enforced before writing to the database, but this way seems fine.

Ya for now it seems easier than introducing a second way of validation, even validating in db is probably more robust.

> re sharing, I’m not clear on the build implications, but could we have a package to host this kind of config that both Hub and web client (and SSR) could consume?

I think a new package probably wouldn't have much build implications if all it did was export utilities or just static data, it sounds like an export of the constraints would be good. Especially as the number of places where we handle business data increase (currently Hub + Web Client, soon SSR Web (Card Space) + Wallet?).

PR #2711, opened 2022-02-22
Support for gnosis 1.3 in subgraph

habdelra:
- [x] Perform full sokol subgraph index and confirm no errors that break indexing
- [ ] deploy to staging
- [ ] deploy to production


habdelra:
The original bug reported that the subgraph had no entry for the safe `0x6616A0f9700c4a2ce2272374c6d385eF53832f96` on sokol. After doing a full index of sokol with this branch, I now see this safe in our subgraph:
Screen Shot 2022-02-22 at 3 23 39 PM

PR #2698, opened 2022-02-17
FastBoot SSR deployment

jurgenwerk:
In the current state, docker container works, but need to test and configure the CI build step and waypoint deployment.

TODO: 
- [x] build infra for this (https://github.com/cardstack/infra/pull/87)
- [x] node-fetch in deployment/package.json version is outdated and should be upgraded
- [x] working action in the CI (need to test)
- [x] working waypoint "ssr-web" deployment (need to fill in the config, currently most of the config is copied from the hub app in waypoint)

PR #2713, opened 2022-02-22
ssr: Add meta tags for link previews

backspace:
This is extracted from #2698, where it has been exercised (in concert with the in-progress work). For [this payment request](https://ssr-wallet-staging.stack.cards/pay/sokol/0xA458a7C13B33188b4F71aA47A1067f651284361E?amount=1&currency=CAD), here are previews as rendered by Twitter and Facebook developer tools:

image

image

The wording has since been changed since I inserted those screenshots. This is easily tweaked in the future, as is the image. The goal here is to make the necessary changes so that tags show at all.


jurgenwerk (approved):
Looking good!

PR #2716, opened 2022-02-23
Caching successful txn receipts to workaround web3.eth.getTransactionReciept issue

habdelra:
- [x] Test in sokol via CLI
- [x] Test in xdai via CLI

### DIFF
--- a/packages/cardpay-sdk/sdk/token-bridge-home-side.ts
+++ b/packages/cardpay-sdk/sdk/token-bridge-home-side.ts
@@ -10,7 +10,12 @@ import { executeTransaction, gasEstimate, getNextNonceFromEstimate, Operation } 
 import { AbiItem, fromWei, toBN } from 'web3-utils';
 import { signSafeTx } from './utils/signing-utils';
 import { query } from './utils/graphql';
-import { TransactionOptions, waitForSubgraphIndexWithTxnReceipt, isTransactionHash } from './utils/general-utils';
+import {
+  TransactionOptions,
+  waitForSubgraphIndexWithTxnReceipt,
+  isTransactionHash,
+  waitUntilTransactionMined,
+} from './utils/general-utils';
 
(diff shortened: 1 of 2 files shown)


habdelra:
@Aierie I confirmed in sokol that the `web3.eth.getTransactionReceipt` is only ever called once now after support for caching these values (we call it more than once as sometimes we actually need the txn receipt to parse useful information out of the event logs, like the address of a newly created prepaid card, or newly created merchant safe).

Do you wanna try testing in xdai via the CLI with this change and if it looks good then I think we can merge it.

habdelra:
I'm not sure why percy is failing--it looks unrelated to this change, and this PR has the latest upstream changes...

Aierie:
> I'm not sure why percy is failing--it looks unrelated to this change, and this PR has the latest upstream changes...

Percy keeps getting false positives. I'll test this PR out now, thanks!

habdelra:
@Aierie don't forget to run `yarn prepack` in the `cardpay-sdk` before testing.

Aierie (approved):

Aierie:
Do we ever want to clear the receipt cache?

habdelra:
I don’t think it’s really necessary, the txn receipt is immutable. In terms of memory, I can’t imagine it would ever be a considerable resource constraint. It would take millions of txn’s happening for a particular EOA for it to begin to have an impact, which seems very far fetched.

backspace:
Does the change in 228d04a look good to you, @habdelra? I opened an incident to reflect that prepaid card issuance and business creation are currently unavailable, would be great to merge and deploy ASAP 🤓

habdelra:
@backspace yeah that's really smart. I always forget to do an immediate return for promises, thanks for catching that

backspace:
It was @Aierie! I’m just following through. Thank you 🎉

habdelra:
I revved the npm version (for the sake of the react native app), and pushed a new CLI package too with this fix

PR #2717, opened 2022-02-23
ssr: Add Percy run on main

backspace:
This should establish a baseline build for PR changes to
be compared against so they won’t all say “New snapshot”.


Aierie (approved):
Thanks! Will try this out now.

PR #2718, opened 2022-02-23
ssr: Add client- and server-side Sentry

backspace:
This is currently based on the branch from #2698 so I could deploy and produce errors. Here’s a [client-side error](https://sentry.io/organizations/cardstack/issues/3042210782/?project=6229132&query=&sort=freq&statsPeriod=14d) from the new `/boom-client` route. Here’s a [server-side error](https://sentry.io/organizations/cardstack/issues/3042496613/?project=6223279&query=&statsPeriod=14d) from `/boom`.

This was comparatively easier to get working with source maps because there’s no Brotli compression in the building of the Ember assets.


packages/ssr-web/deployment/fastboot-server.js
```diff
@@ -18,6 +21,14 @@ let server = new FastBootAppServer({
   },
   // This should be false for Twitter/Linkedin according to https://github.com/ember-fastboot/ember-cli-fastboot/tree/master/packages/fastboot-app-server#twitter-and-linkedin
```
Aierie, line 22: Not related to this PR, but I wonder if it's good to somehow provide `buildSandboxGlobals` to the local dev + the deployment server from the same source, so we don't have mysterious bugs show up if something changes.

packages/ssr-web/deployment/fastboot-server.js
```diff
@@ -1,6 +1,9 @@
 /* eslint-disable no-undef */
 const FastBootAppServer = require('fastboot-app-server');
 const fetch = require('node-fetch');
+const Sentry = require('@sentry/node');
+
+Sentry.init({ dsn: process.env.SSR_WEB_SERVER_SENTRY_DSN });
```
Aierie, line 6: Can we add an `environment` property, or the corresponding Sentry environment variable? The deployment step doesn't seem like it will get the `SENTRY_ENVIRONMENT` variable from the build step.

jurgenwerk: Added!

jurgenwerk:
I added SSR_WEB_ENVIRONMENT and SSR_WEB_SERVER_SENTRY_DSN env vars to staging and production Waypoint.

PR #2719, opened 2022-02-23
Lazy load queries

mattmcmanus:
This PR setups a new DI type for queries. Queries, which have been living in a folder inside of services, can now be added as in a sibling directory to services/routes/etc.

This change also enables it so that you do not need to append queries to the service name. 

Let me know what you think about this and I can move the rest of the queries around


backspace (approved):
Nice, thanks! Maybe we can escape all this in the future with Prisma but who knows when 😆

mattmcmanus:
@backspace Ok super! I've moved all the rest of the queries over.

jurgenwerk (approved):
Looks good!

PR #2720, opened 2022-02-23
ssr: Add server-side Sentry

backspace:
This is extracted from #2718, where client-side works and
server-side doesn’t, so will continue separately…


backspace:
Actually I had the wrong DSN so I think it works there!

PR #2721, opened 2022-02-24
Fix/ci xml reporter

Aierie:
Attempt to fix failing CI https://github.com/cardstack/cardstack/runs/5314992207?check_suite_focus=true :
- no need to check for existence of a folder and dynamically create it
- each package gets separate reporting


jurgenwerk (approved):
Thanks! I was just wondering why the main build [fails](https://github.com/cardstack/cardstack/runs/5317752151?check_suite_focus=true) after merging the ssr branch but it looks like this will fix it! 🎉

PR #2722, opened 2022-02-24
[CS-3218] Update convertToSpend to fix a bug with decimal conversion

topwebtek7:
`convertToSpend(0.56, 'USD', 1)`($0.56USD to spend) returns 57 but should be 56. Investigated and looks it's cause of divide by decimal number and then use Math.ceil. Tests are passed on CardWallet side.

Ref: [CS-3218](https://linear.app/cardstack/issue/CS-3218/bug-as-a-customer-if-i-enter-56-usd-or-56-spend-into-a-payment-request)


Aierie (approved):
Thanks for this fix!

PR #2724, opened 2022-02-24
Add github action & workflow for subgraph extraction package

IanCal:
This is a copy & cut down version of the hub setup.


alex-cardstack:
This looks good to me but I haven't used waypoint much, so I'll leave to @lukemelia to double check

PR #2725, opened 2022-02-24
FastBoot environment uses SSR_WEB_ENVIRONMENT variable to distinguish between staging/prod links

jurgenwerk:
In the production ssr-web deploy I noticed the links lead to staging:

image

This PR fixes the logic so that FastBoot is able to see which environment it is in.

### DIFF
--- a/packages/ssr-web/config/environment.js
+++ b/packages/ssr-web/config/environment.js
@@ -22,9 +22,8 @@ module.exports = function (environment) {
     locationType: 'auto',
     hubURL: process.env.HUB_URL,
     universalLinkDomain:
-      universalLinkHostnamesByTarget[
-        process.env.DEPLOY_TARGET || process.env.SSR_WEB_ENVIRONMENT // front-end uses DEPLOY_TARGET, fastboot uses SSR_WEB_ENVIRONMENT
-      ] ?? MERCHANT_PAYMENT_UNIVERSAL_LINK_STAGING_HOSTNAME,
+      universalLinkHostnamesByTarget[process.env.SSR_WEB_ENVIRONMENT] ??
+      MERCHANT_PAYMENT_UNIVERSAL_LINK_STAGING_HOSTNAME,
     version: pkg.version,
     sentryDsn: process.env.SENTRY_DSN,
     '@sentry/ember': {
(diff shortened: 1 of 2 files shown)


packages/ssr-web/config/environment.js
```diff
@@ -22,8 +22,9 @@ module.exports = function (environment) {
     locationType: 'auto',
     hubURL: process.env.HUB_URL,
     universalLinkDomain:
-      universalLinkHostnamesByTarget[process.env.DEPLOY_TARGET] ??
-      MERCHANT_PAYMENT_UNIVERSAL_LINK_STAGING_HOSTNAME,
+      universalLinkHostnamesByTarget[
+        process.env.DEPLOY_TARGET || process.env.SSR_WEB_ENVIRONMENT // front-end uses DEPLOY_TARGET, fastboot uses SSR_WEB_ENVIRONMENT
+      ] ?? MERCHANT_PAYMENT_UNIVERSAL_LINK_STAGING_HOSTNAME,
```
backspace, line 27: I think `DEPLOY_TARGET` actually isn’t used at all in this application because that’s an Ember CLI Deploy thing.

backspace (changes requested):
Thanks for catching this. I’m not sure how we should make this distinction, maybe what I called `SENTRY_ENVIRONMENT` can actually be broadened to the name you use here of `SSR_WEB_ENVIRONMENT`?

packages/ssr-web/config/environment.js
Aierie: Is it worth keeping `DEPLOY_TARGET` and setting it in our CI for consistent language with `web-client`?

jurgenwerk: I think we can remove `DEPLOY_TARGET`. Thanks @backspace for the tip, I think we can use `SSR_WEB_ENVIRONMENT` for both Sentry and app logic for links. Will adjust this

jurgenwerk:
@Aierie @backspace I removed `DEPLOY_TARGET` and used `SSR_WEB_ENVIRONMENT` for both Sentry and app logic.

Aierie (approved):
I'm ok with this. Though it is odd for me that the property that answers the question "Is this build for staging, production, or neither" is `DEPLOY_TARGET` in web-client, but `SSR_WEB_ENVIRONMENT` in ssr-web.

PR #2730, opened 2022-02-25
Basic user space page

Aierie:
Adds:
- A user page with the basic info, no styling
- An error page
- A service to decide which app we are on, based on the host (works for window + fastboot)


packages/ssr-web/app/services/app-context.ts
```diff
@@ -0,0 +1,28 @@
...
+
+  get currentApp(): 'card-space' | 'wallet' {
+    // return 'wallet';
+    return 'card-space';
+  }
+
+  get cardSpaceId() {
+    return 'nosuchmerchant';
+  }
+}
```
Aierie, line 21: Intention is for this to centralize the knowledge of which app we are serving, and if it is card space, provide the id.

packages/ssr-web/config/environment.js
```diff
@@ -74,6 +81,11 @@ module.exports = function (environment) {
     'ember-cli-mirage': {
       enabled: false,
     },
+    fastboot: {
+      hostWhitelist: hostWhitelistByTarget[process.env.DEPLOY_TARGET] ?? [
+        '/^localhost:\\d+$/',
+      ],
+    },
```
Aierie, line 88: Whitelist config is necessary for us to get the host

packages/ssr-web/config/environment.js
```diff
@@ -86,6 +86,9 @@ module.exports = function (environment) {
         '/^localhost:\\d+$/',
       ],
     },
+    chains: {
+      layer2: process.env.DEPLOY_TARGET === 'production' ? 'xdai' : 'sokol',
```
Aierie, line 90: DEPLOY_TARGET doesn't exist yet so this won't work.

packages/ssr-web/app/routes/index.ts
```diff
@@ -0,0 +1,82 @@
...
+        }`,
+          {
+            did: merchant.attributes.did,
+          }
+        );
+        let address = queryResult.data.merchantSafes[0].id;
+
+        if (!merchant || !address) {
+          // TODO: replace with proper 404 somehow
+          // this 404 is for card space
```
Aierie, line 57: The app context service could be used to inform the 404 about what to show on the 404 page.

packages/ssr-web/app/routes/index.ts
```diff
@@ -0,0 +1,82 @@
...
+          `query($did: String!) {
+          merchantSafes(where: { infoDid: $did }) {
+            id
+          }
+        }`,
+          {
+            did: merchant.attributes.did,
+          }
+        );
+        let address = queryResult.data.merchantSafes[0].id;
```
Aierie, line 53: This could move to the hub for now. maybe we can run a script to update the db (whether in the merchant info table or a new one) so each merchant has a record of its safe address, and update the merchant creation flow to have the safe address stored in the db after the merchant is created.

packages/ssr-web/app/services/app-context.ts
Aierie: This should also allow us to hide the local-dev specific hacks.

packages/ssr-web/app/routes/index.ts
```diff
@@ -0,0 +1,82 @@
...
+  @service('app-context') declare appContext: AppContextService;
+
+  async model(): Promise<UserSpaceRouteModel> {
+    if (this.appContext.currentApp === 'card-space') {
+      try {
+        const cardSpaceResult: {
+          included: any[];
+        } = await (
+          await fetch(
+            `${config.hubURL ?? 'http://localhost:3000'}/api/card-spaces/${
```
Aierie, line 27: hubURL in local dev is not set. We can't use `environment === 'development'`, so we need to find something else to conditionally set hubURL to `localhost:3000`.

packages/ssr-web/app/services/app-context.ts
Aierie: I've figured out how to get local DNS hax in so we don't have to use a different way to get the id for dev, will post the method here when I've ironed out the details

Aierie: @backspace

packages/ssr-web/app/routes/index.ts
Aierie: Discussed w @jurgenwerk  - probably will move this to after the model hook for perf reasons.

Aierie: TODO: decide how to display the page when we fail to fetch the address - edge case might occur where we can't find an address.

packages/ssr-web/config/environment.js
jurgenwerk: Should probably use `SSR_WEB_ENVIRONMENT` when https://github.com/cardstack/cardstack/pull/2725 is merged?

packages/ssr-web/app/routes/index.ts
Aierie: `environment.js` should be modified so that the absence of `SSR_WEB_ENVIRONMENT` or `SSR_WEB_ENVIRONMENT` not being staging or production will replace `environment === 'development'`.

Aierie:
Sorry, I didn't include the test environment when making changes to set hubURL for dev, included in https://github.com/cardstack/cardstack/pull/2730/commits/7b69f548f17700c961e6c11aa0a612369ad2be2b

packages/ssr-web/app/routes/index.ts
Aierie: Will do this in a separate PR.

Aierie:
## Follow-ups
CS-3240: Prevent card space from accessing wallet routes
CS-3241: Real 404s. We can't return these at the moment, setting the status code doesn't seem to work
CS-3242: more detailed tests, might involve mocking the subgraph query for merchant safe address

packages/ssr-web/app/routes/index.ts
Aierie: CS-3243

packages/ssr-web/app/routes/index.ts
```diff
@@ -0,0 +1,103 @@
...
+          }),
+        };
+      } catch (e) {
+        Sentry.captureException(e);
+        throw e;
+      }
+    } else {
+      if (this.fastboot.isFastBoot) {
+        this.fastboot.response.statusCode = 404;
+      }
```
backspace, line 99: What do you think about encapsulating this as something like `this.fastboot.setStatus(404)`, including the conditional?

packages/ssr-web/tests/integration/services/app-context-test.ts
```diff
@@ -0,0 +1,102 @@
...
+
+      fastboot.request.host = `oops.two${config.cardSpaceHostnameSuffix}`;
+      assert.equal(appContext.cardSpaceId, 'oops.two');
+
+      fastboot.request.host = 'oops.wallet';
+      assert.equal(appContext.cardSpaceId, '');
+    });
+  });
+
+  module('window', function (hooks) {
```
backspace, line 68: is this more `browser` than `window`?

backspace (approved):
I appreciate that you added the `AppContext` service tests 🎉

packages/ssr-web/app/routes/index.ts
```diff
@@ -0,0 +1,103 @@
...
+interface UserSpaceRouteModel {
+  id: string;
+  name: string;
+  backgroundColor: string;
+  textColor: string;
+  paymentURL: string;
+  deepLinkPaymentURL: string;
+}
+
+export default class UserSpaceRoute extends Route {
```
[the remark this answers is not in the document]
Aierie: Nah it's from me i think

packages/ssr-web/tests/integration/services/app-context-test.ts
Aierie: ya

packages/ssr-web/app/routes/index.ts
Aierie: ambivalent, I think the meaning is clear enough either way, though the current form is more verbose. Would this be `this.fastboot.setStatus`, or `this.setFastbootStatus` though? Fastboot is imported, so we would be extending the fastboot service with our own custom version I think.

PR #2731, opened 2022-02-25
Protocol version v0.9.0 bump

alex-cardstack:
I think I did all the things from the previous commit except the subgraph stuff, let me know if I missed anything


habdelra (approved):
This looks perfect! now we need to wait until this is live in the app store as well as in the production dapp before updating the protocol

PR #2732, opened 2022-02-25
This adds support for a card's computeds to run live while editing a card

habdelra:
The general approach here is that when a setter is triggered, the setter will hand off the updated value to an ember concurrency task. The ember concurrency task creates a new schema instance for the card and recomputes the values of the card's fields with the new schema instance (which is an async process). When the computation is done, the EC task sets the new card data in one fell swoop using the glimmer tracked property that represents the card's data.

A couple nuances here:
1. it looks like previous we were only setting the value of the card's data in the wrapped component manufactured from the card model once in its constructor. In this PR we change that to a getter so it's always just returning the live data throughout the component's life cycle.
2. our schema instance becomes the card model's data that it shares with the outside (including templates that consume the card model)
3. the setters kick off an inherently async re-rendering process. To help manage the lifecycle, we use `@ember/destroyable` to register the EC task that does the re-render in the destructor, as well as to allow the `Modal` component responsible for rendering the cards to manage the lifecycle of the card models, so that we destroy a card model when we are done with it.


packages/cardhost/app/lib/card-model-for-browser.ts
```diff
@@ -265,6 +262,32 @@ export default class CardModelForBrowser implements CardModel {
     return s;
   }
 
+  @restartableTask async rerenderData(
+    data: Record<string, any>
+  ): Promise<void> {
+    this.state.rawData = merge({}, this.state.rawData, {
+      attributes: data,
+    });
+    let newSchemaInstance = await this.createSchemaInstance();
```
mattmcmanus, line 271: I'm confused why we need to instantiate a new instance every time we rerender. Why is this necessary?

habdelra: we are instantiating a new instance because the `this._schemaInstance` is a tracked property. We want to have the rerenders of the template present consistent data for all the fields, such that we don't see a field consumed by a computed being rerendered at a different time than the computed field that consumes it. all the field updates need to atomic which includes any computeds (anyc or sync) that consume the fields. In order to do this we create a "working" schema instance that we use for the intermediate states, and when the schema instance has settled, we then set it to the tracked tracked property for a consistent rendering of the template.

PR #2734, opened 2022-02-28
extract payment link

Aierie:
For use in Card Space. Also corrects the text under a deep link to the deep link URL, and allows custom text for the payment link as opposed to "Pay Business".

## mode = "link"
Payment link component showing its 'link' state. A link styled as a Boxel Button, text showing the link url, and a transparent button with text 'Show as QR Code'

## mode = "qr"
Payment link component showing its 'qr' state. A QR code, text showing the url that the QR code will lead you to once scanned, and a transparent button with text 'Show Payment Link'

## mode = "qr-non-mobile"
Payment link component showing its 'qr-non-mobile' state. A QR code and text showing the url that the QR code will lead you to once scanned. There is no toggle to change to 'link' state as links will not work on a non-mobile device.


jurgenwerk (approved):
Looking good! The code duplications across web-client and ssr-web do feel a bit unfortunate though for now. Maybe add comments to the web-client code which is to be deleted once we get rid of the pay route in web-client?

PR #2735, opened 2022-02-28
allow android deep link payments

Aierie:


jurgenwerk (approved):

backspace (commented):
👍🏻 maybe we don’t need a Percy snapshot for this? 🤔

PR #2741, opened 2022-02-28
support for ?allFields query param to get all of a card's fields

habdelra:
This adds support for the browser to ask for all of a card's fields, such that in a future PR the browser's card model has the ability to ask for more fields than were originally provided from the requested format.


mattmcmanus (approved):

packages/hub/lib/card-model-for-hub.ts
```diff
@@ -234,6 +240,7 @@ export default class CardModelForHub implements CardModel {
       componentMeta: this.state.componentMeta,
       original: undefined,
       deserialized: false,
+      allFields: false,
```
habdelra, line 243: we can just leave this untouched

PR #2742, opened 2022-02-28
[CS-3232] Break up code analyzing and validation

mattmcmanus:
This is a rough first pass just for the serializer. The next step would be to move the schema analyzer into the base analyzer and improve the types


packages/cardhost/app/lib/builder.ts
```diff
@@ -216,7 +216,8 @@ export default class LocalRealm implements Builder {
     assertDocumentDataIsResource(resource);
     let data = resource.attributes;
     let id = resource.id
-      ? this.parseOwnRealmURL(resource.id)!.id
+      ? // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
```
habdelra, line 219: @ef4 started seeing these too. we should just add this to our eslint-config package

packages/core/src/compiler.ts
```diff
@@ -23,31 +29,29 @@ import {
...
+};
+export const BASE_CARD_URL = cardURL(BASE_CARD_ID);
 
 export class Compiler<Identity extends Saved | Unsaved = Saved> {
   private builder: TrackedBuilder;
   private cardSource: RawCard<Identity>;
-  private modules: CompiledCard<Unsaved, LocalRef>['modules'];
+
+  private sourceModules: CompiledCard<Unsaved, LocalRef>['modules'];
+  private globalModules: CompiledCard<Unsaved, LocalRef>['modules'];
```
ef4, line 48: 1. I find it surprising that sourceModules and globalModules should really have the same type. 
2. I don't think "global" is a good name here. The distinction is more input vs output. Even the output are still local refs, not global refs.

mattmcmanus: @ef4 Yeah. I wondered about it but didn't really push the thought as far as I could. I'll ponder.

mattmcmanus: @ef4 I just did another pass. It's a subtle split up (at least for now, until we consolidate the rest of the analyze work)

PR #2743, opened 2022-03-01
ssr: Add Card Space redirects

backspace:
This adds redirects when in a Card Space context: `/*` → `/`. It also adds a 404 page that will apply in a payment request context.

It’s set to merge into #2730 at the moment as it uses work from there.


Aierie:
re: isCardSpace, Typescript is preventing us from using arbitrary strings because of the definition of `currentApp` as `"card-space"|"wallet"`. Would this be enough, or do you think a dedicated boolean is still better?

VSCode showing a hint for a Typescript error. The error says that `currentApp === 'carad-space'` will always return `false` because there is no overlap between 'carad-space' and `'card-space' | 'wallet'`

Aierie:
I'm not getting the `node-fetch` error, I'm getting successful redirects:

Terminal logging from local development fastboot shows redirects from the pay route to index

Aierie:
It's very odd that tests fail with the `isCardSpace` boolean, happens on local for me too.

backspace:
> ~It's very odd that tests fail with the `isCardSpace` boolean, happens on local for me too.~
> 
> Fixed in [92d0b48](https://github.com/cardstack/cardstack/commit/92d0b485b62f7dfccf09ca49175d12e5e6e2aeb8)

I failed to push this last night:

image

What do you think? I think it’s nice to be able to extend the mocked service to be able to not have to reimplement every piece of it, just override the parts we need to.

Aierie:
> What do you think? I think it’s nice to be able to extend the mocked service to be able to not have to reimplement every piece of it, just override the parts we need to.

Is it possible to avoid weird intermediate states (some properties manually set, some properties not) that cause weird bugs?

An alternative: https://github.com/cardstack/cardstack/blob/e0076b1b97ecc7891dce90476072204b025c3e71/packages/ssr-web/app/services/subgraph.ts#L4-L15

Aierie:
Also see the `Partial` type: https://www.typescriptlang.org/docs/handbook/utility-types.html#partialtype, which is kind of the mirror of the `Required` approach

backspace:
I don’t feel that worried about weird bugs! I like less having to replicate the definition of `isCardSpace`, it stops being a convenience method then 😢

packages/ssr-web/app/routes/not-found.ts
```diff
@@ -0,0 +1,15 @@
+import Route from '@ember/routing/route';
+import { inject as service } from '@ember/service';
+import AppContextService from '@cardstack/ssr-web/services/app-context';
+
+export default class PayRoute extends Route {
```
Aierie, line 5: NotFoundRoute

PR #2744, opened 2022-03-01
Create card spaces for merchants

jurgenwerk:
This creates card spaces for merchants created before we added automatic card space creation.

This could probably be just one SQL statement if our DB had the `uuid-ossp` extension installed which would allow us to use  the `uuid_generate_v4()` function for creating the ids. But without it we need to use `shortUuid.uuid()`. 

Currently there's 260 merchants in production which don't have a card space so I think this migration shouldn't eat too much of resources.


Aierie (approved):
Tried this locally, works nicely!

backspace (approved):

PR #2751, opened 2022-03-02
Card services unification

habdelra:
This PR takes steps towards unifying the card services on the hub and browser by using a single interface for these, as well as to takes steps towards unifying the card model for the hub and browser. Also, we're moving fields that are not state specific out of the card model state, as well as updating the builder to use CardModels for mutation of the local realm.

both the `getField()` and the `save()` are very close to being the same. I think with a bit more effort we could make them the same. The unifying the setting of the fields I'll tackle in a subsequent PR

### DIFF
--- a/packages/hub/services/search-index.ts
+++ b/packages/hub/services/search-index.ts
@@ -16,6 +16,7 @@ import { service } from '@cardstack/hub/services';
 import { transformToCommonJS } from '../utils/transforms';
 import flatMap from 'lodash/flatMap';
 import CardModelForHub from '../lib/card-model-for-hub';
+import { INSECURE_CONTEXT } from './card-service';
 
 const log = logger('hub/search-index');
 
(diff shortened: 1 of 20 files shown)


packages/hub/services/search-index.ts
```diff
@@ -259,17 +261,23 @@ class IndexerRun implements IndexerHandle {
...
-      componentMeta,
-      allFields: false,
-    });
+    let cardModel = await getOwner(this).instantiate(
+      CardModelForHub,
+      await this.cardService.as(INSECURE_CONTEXT),
+      {
+        type: 'loaded',
+        id: rawCard.id,
+        componentModule: definedCard.componentInfos[format].componentModule.global,
```
habdelra, line 270: I wonder if we really need to set this, does it ever change? if we don't need to set it, it would make things simpler since there would be no need to use `compiled` and the indexing could get pushed into the card service

PR #2755, opened 2022-03-03
Disable MetaMask when there is no extension

Aierie:
Browsers that don't have an extension, + people who don't have the extension should not be able to select this option because nothing will happen. Not hiding the option, because it seems better to inform people that they have this choice if they do install the extension.

# If browser has MetaMask extension
Layer one connection wallet selection showing both MetaMask and WalletConnect enabled. MetaMask is selected as the method of connection.

# If browser does not have MetaMask extension
Layer one connection wallet selection showing WalletConnect enabled, and MetaMask disabled. MetaMask is placed 2nd since it is disabled. WalletConnect is selected as the method of connection.

### DIFF
--- a/packages/web-client/app/components/card-pay/layer-one-connect-card/index.ts
+++ b/packages/web-client/app/components/card-pay/layer-one-connect-card/index.ts
@@ -48,9 +48,7 @@ class CardPayDepositWorkflowConnectLayer1Component extends Component<CardPayDepo
        , otherwise focus leaves the page
      - selecting a radio makes the connect button enabled and focusable.
    */
-  // pick the first enabled option
-  @tracked radioWalletProviderId: WalletProvider['id'] =
-    this.walletProviders[0].id;
+  @tracked radioWalletProviderId: WalletProvider['id'] = 'wallet-connect';
 
   constructor(
     owner: unknown,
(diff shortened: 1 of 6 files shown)


Aierie:
For mobile, we may want to remove the MetaMask option altogether?

packages/web-client/app/components/card-pay/layer-one-connect-card/index.ts
```diff
@@ -23,7 +23,20 @@ interface CardPayDepositWorkflowConnectLayer1ComponentArgs
...
+            ...w,
+            enabled: !!window.ethereum?.isMetaMask,
+            explanation: window.ethereum?.isMetaMask
+              ? ''
+              : 'MetaMask extension not detected',
+          }
+        : { ...w, enabled: true, explanation: '' }
+    )
+    // sort stuff so that the enabled options always appear first
+    .sort((a, b) => Number(b.enabled) - Number(a.enabled));
```
jurgenwerk, line 39: Would `sortBy('enabled')` work?

jurgenwerk (approved):

packages/web-client/app/components/card-pay/layer-one-connect-card/index.ts
Aierie: It works with `desc`!

backspace (approved):
I tried this in browsers with and without Metamask installed, nice improvement 🎉

PR #2756, opened 2022-03-03
Redirect cardpay -> card-pay

Aierie:
We have an incorrect link in mobile that is leading users to cardpay.

For this PR, if approved, would be good to merge and move to prod asap


backspace (approved):
I think we should have a 404 for unknown routes at least, but this makes sense given the incorrect incoming link.

PR #2757, opened 2022-03-04
ssr: Update for Card Space domains

backspace:
This updates the host list to serve `*.card.xyz` and changes `card.space` in favour of that new suffix. It includes a Waypoint tweak from Terraform changes.

### DIFF
--- a/packages/ssr-web/config/environment.js
+++ b/packages/ssr-web/config/environment.js
@@ -23,6 +23,7 @@ const hostWhitelistByTarget = {
   staging: [
     MERCHANT_PAYMENT_UNIVERSAL_LINK_STAGING_HOSTNAME,
     /.+\.pouty\.pizza$/,
+    /^10.91.\d+.\d+:4000$/, // FIXME if this is needed, it’s in infra 🤔
   ],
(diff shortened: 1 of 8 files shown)


packages/ssr-web/config/environment.js
```diff
@@ -11,20 +11,28 @@ const universalLinkHostnamesByTarget = {
...
 const CARD_SPACE_STAGING_HOSTNAME_SUFFIX = '.pouty.pizza';
 const CARD_SPACE_HOSTNAME_SUFFIX = '.card.xyz';
 
 const cardSpaceHostnameSuffixByTarget = {
   staging: CARD_SPACE_STAGING_HOSTNAME_SUFFIX,
   production: CARD_SPACE_HOSTNAME_SUFFIX,
 };
+
+function convertHostnameSuffixToRegex(suffix) {
+  return `/^.+\\${suffix}$/`;
```
Aierie, line 24: this doesn't escape the full stops, which should still work but could technically match a set of other domains.

backspace: true 😯 fixed in b0be65a, thank you!

PR #2758, opened 2022-03-04
styled qr code loading state

Aierie:
Part of CS-3170

Previously this qr code would appear as a small square in ssr-web while it was being generated. 

The ssr-web card space user page for merchie in its initial loading state. It has a small square that is many times smaller than the QR code is supposed to be, while the QR code is being generated.

This adds a loading indicator and also sets its size beforehand. VSCode will warn about aspect ratio but caniuse suggests that it is actually pretty well supported.

Video of the responsive QR code loading state below, on different small screens:

https://user-images.githubusercontent.com/39579264/156735042-196806c1-5a20-456d-86cd-5045d7cd7eb8.mov


Aierie:
styled-qr-code may be a good component to extract to Boxel.

backspace (approved):
Great improvement 🤩

PR #2759, opened 2022-03-04
Wrap card space in a component and load address async

jurgenwerk:
Ticket: [CS-3243](https://linear.app/cardstack/issue/CS-3243/move-query-for-merchant-safe-address-to-controller)

In contrast to what the ticket says (moving logic to controller) I opted into making a new component because that seemed simpler, and do the subgraph fetch there.

Fetching address from the subgraph enables faster `model()` resolve and doesn't cause the route to go into an error state in case there are issues with the subgraph, or the safe cannot be found.


packages/ssr-web/app/components/common/card-space-user-page/index.ts
```diff
@@ -0,0 +1,94 @@
...
+      address = queryResult?.data?.merchantSafes[0]?.id;
+    } catch (e) {
+      this.addressFetchingError = this.defaultAddressFetchingErrorMsg;
+      Sentry.captureException(e);
+    }
+
+    if (address) {
+      this.address = address;
+    } else {
+      this.addressFetchingError = this.defaultAddressFetchingErrorMsg;
```
jurgenwerk, line 72: This message is likely not very appropriate for this case. To tune this, we need to know what are the possible values the subgraph can return when there is no merchant safe id included.

1. Subgraph returns empty merchants 
- Odd edge case where safe is deleted or can't be associated for some reason? Should we handle this case?

2. Subgraph returns something like `indexing_error` (as we observed recently)
- We can probably keep that error message

3. Anything else?

backspace: I can’t think of anything else at least, but this is easily revisited if we find other failures.

backspace (approved):
👍🏻

routeable components, someday? 😆

packages/ssr-web/app/components/common/card-space-user-page/index.ts
Aierie: I think this is ok, is it useful to have this logged to sentry?

packages/ssr-web/app/components/common/card-space-user-page/index.hbs
```diff
@@ -0,0 +1,20 @@
...
+    @logoTextColor={{@model.textColor}}
+    @vertical={{true}}
+    @size="large"
+    @text={{@model.id}}
+  />
+  <Common::PaymentLink
+    @mode={{this.paymentLinkMode}}
+    @switchMode={{this.setPaymentLinkMode}}
+    @paymentURL={{this.paymentURL}}
+    @addressFetchingError={{this.addressFetchingError}}
```
Aierie, line 17: Why have the payment link handle the address fetching error state rather than have it in a separate component that is rendered instead of the payment link? 🤔

packages/ssr-web/app/components/common/card-space-user-page/index.ts
jurgenwerk: I think it would be useful. Will add that!

packages/ssr-web/app/components/common/card-space-user-page/index.hbs
jurgenwerk: Yeah that does sound more appropriate! Will change it

jurgenwerk: Done!

packages/ssr-web/app/components/common/card-space-user-page/index.ts
jurgenwerk: Added!

PR #2763, opened 2022-03-07
Comments on why we have focus states for platters

Aierie:
@burieberry did some thoughtful work to help keyboard/non-pointer-device users in web client, so I did not apply any changes to address CS-3198, I believe that keyboard users will still want these focus states, and focus-visible by default is coming to Safari soon, making these states only appear to users who would like to see them. Some work could be done on the design side to make them more aesthetically pleasing. This PR just adds comments to explain why these focus states are necessary.

## Focus for modal UX
The focus on the workflow thread body (first image) is to ensure that focus starts at the right place, and that tabbing will bring the users to a region at the top of the modal. 

Freshly opened workflow with a blue focus outline on the outside of the thread

## Focus to allow scrolling via keyboard
The focus on these 2 areas is to ensure that keyboard users can get there and scroll. Because of the generic nature of the workflow thread scrollable areas, there is no guarantee when a focusable element will show up so it seems good to allow focusing of the scrollable areas for now. 

Workflow scrollable content with focus received, showing a blue focus outline

Workflow sidebar scrollable content with focus received, showing a blue focus outline

### DIFF
--- a/packages/boxel/addon/components/boxel/thread/index.hbs
+++ b/packages/boxel/addon/components/boxel/thread/index.hbs
@@ -4,6 +4,7 @@
   </div>
 
   <div class="boxel-thread__content-wrapper">
+    {{!-- Note that tabindex=0 is applied to ensure that this scroll context can be reached and operated by keyboard --}}
     <div class="boxel-thread__scroll-wrapper" tabindex="0" {{autoscroll enabled=@autoscroll}}>
       <div class="boxel-thread__sticky-container">
         {{yield to="taskbar"}}
@@ -13,6 +14,7 @@
       </section>
     </div>
 
+    {{!-- Note that tabindex=0 is applied to ensure that this scroll context can be reached and operated by keyboard --}}
     <div class="boxel-thread__scroll-wrapper" tabindex="0">
       <Boxel::Sidebar class="boxel-thread__sidebar" as |SidebarSection|>
         {{yield SidebarSection to="sidebar"}}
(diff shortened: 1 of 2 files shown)


packages/web-client/app/components/workflow-thread/index.hbs
```diff
@@ -9,7 +9,7 @@
 <Boxel::Thread
   @autoscroll={{this.autoscroll}}
   class="workflow-thread workflow-thread-animated"
-  tabindex="0"
+  tabindex="-1"
```
Aierie, line 12: @burieberry do you think this is a good change, not allowing the root to be refocused?

packages/boxel/addon/components/boxel/thread/index.hbs
```diff
@@ -4,6 +4,7 @@
   </div>
 
   <div class="boxel-thread__content-wrapper">
+    {{!-- Note that tabindex=0 is applied to ensure that this scroll context can be reached and operated by keyboard --}}
```
burieberry, line 7: I suppose a more pleasant looking way to do this is similar to how a thread on discord is navigated. There, each message is highlighted when using a keyboard, instead of the whole scrollable area. This makes is possible to focus on the elements inside of the thread message accessible without having to tab through the entire thread.

burieberry: So if there's time & priority for improvements, that's what I would recommend. Otherwise, your PR looks good to me.

backspace (approved):
I’m not deeply familiar with everything being considered here but I tried it locally and the explanation makes sense to me!

packages/boxel/addon/components/boxel/thread/index.hbs
Aierie: I think we can't do this right now because the messages in here are not meant to be interacted with and adding more tabbable items just makes it harder for non-pointer users. For the most part messages are just static text.In the future where there are richer interactions similar to discord (reply to message etc) this might be good.

PR #2771, opened 2022-03-09
Basic auth UI

Aierie:
Basic auth UI + tests to ensure that correct UI is shown in response to various states of services, and events trigger the correct callbacks. Styling improvements to come in a separate PR.

The auth steps in order are `WALLET_CONNECT` -> `HUB_AUTH` -> `DONE`, with a `LOADING` state to prevent weird transitions at first load, and a fallback state for any unknown errors.

This is to provide an auth flow that brings the user through connecting to layer 2 and authenticating with hub, in a way that is usable on mobile. The component (and services it depends on) may be moved or copied to SSR web depending on the auth method we settle on.

### DIFF
--- a/packages/web-client/app/components/auth/index.ts
+++ b/packages/web-client/app/components/auth/index.ts
@@ -3,7 +3,7 @@
 import Component from '@glimmer/component';
 import Layer2Network from '@cardstack/web-client/services/layer2-network';
 import HubAuthentication from '@cardstack/web-client/services/hub-authentication';
-// import UA from '@cardstack/web-client/services/ua';
+import UA from '@cardstack/web-client/services/ua';
 import { inject as service } from '@ember/service';
 import { action } from '@ember/object';
 import { tracked } from '@glimmer/tracking';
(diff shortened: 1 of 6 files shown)


packages/web-client/app/components/auth/index.ts
```diff
@@ -0,0 +1,92 @@
...
+import HubAuthentication from '@cardstack/web-client/services/hub-authentication';
+import UA from '@cardstack/web-client/services/ua';
+import { inject as service } from '@ember/service';
+import { action } from '@ember/object';
+import { tracked } from '@glimmer/tracking';
+import { race, rawTimeout, task, TaskGenerator } from 'ember-concurrency';
+import { taskFor } from 'ember-concurrency-ts';
+import config from '@cardstack/web-client/config/environment';
+
+const AUTH_STEPS = {
```
jurgenwerk, line 12: The other day I came across this article https://swizec.com/blog/stop-shouting-shouting/ and it struck a chord with me 😀 I know we use this pattern across the app but wondering what you think about simply using a type for this

jurgenwerk (approved):
Great, looks like a solid start for the mobile auth.

packages/web-client/app/components/auth/index.ts
Aierie: i think it's ok, but we have no types in templates.... so I'm shouting to get the point across to hbs

Aierie: If templates had types I would be happy to use types

PR #2772, opened 2022-03-09
Simple business profile page 

jurgenwerk:
Ticket: [CS-3170](https://linear.app/cardstack/issue/CS-3170/simple-business-profile-page)

Decided to not include the button on desktop because it's not possible to connect with a button there. 

TODO / to discuss:

- [x] Should top left icon have the `@model.textColor` color? Otherwise it won't be seen clearly on bright backgrounds
- [x] Tests 

Mobile:
image

Desktop:
image


backspace:
agreed re the icon

Wrapping is probably necessary for the domain:

image

also not sure what can be done re short viewports:

image

jurgenwerk:
@backspace thanks for finding these visual bugs. I think I solved the cases you pointed out:

image

Aierie:
I think this looks good. Possible improvements in https://github.com/cardstack/cardstack/pull/2777, feel free to use/discard

Aierie (approved):

kdowlin:
@jurgenwerk We should use a black card pay logo or white card pay logo based on the background color using the same logic we do for the text color.

kdowlin:
Screen Shot 2022-03-10 at 2 10 11 PM
I would expect the logo with this background color to be black and not white.

jurgenwerk:
@kdowlin yep, it's like that in the current working version of this PR. I think those screenshots are before this rule was applied.

PR #2774, opened 2022-03-09
Prep for load all fields when computeds need more than what is included in card model

habdelra:
This forces us to be much more precise about a missing field vs an empty field since we are throwing now when we encounter a missing field. the throwing will be absorbed in a subsequent PR when we use that to signal that more data needs to be loaded. This also adds a skipped test for when we are ready to implement the logic to absorb the missing field exceptions and load the data

UPDATE
This has also been updated with logic that moves knowledge about the shape of the card data into the schema instance so that we can prevent that knowledge from leaking outside of the card's schema instance. This PR reduces the need for the `CardComponentMeta` module--the only item left in this module is the serializer map. There is a little scaffolding that still exists in service of the serialization happening in the `CardModel` ultimately once the serializers can be emitted into the schema instance via the compiler, the `CardComponentMeta` module can be removed as well as a bunch of scaffolding in the `CardModel` as well as the entire `CardModelForHub` class (and the `CardModelForBrowser class becomes much simpler).


packages/hub/services/card-service.ts
```diff
@@ -107,7 +109,11 @@ export class CardService implements CardServiceInterface {
   async create(raw: RawCard<Unsaved>): Promise<Card> {
     let compiler = this.builder.compileCardFromRaw(raw);
     let compiledCard = await compiler.compile();
-    let rawCard = await this.realmManager.create(raw);
+    let allFields = fieldsAsList(compiledCard.fields);
+    let { id, realm, adoptsFrom } = raw;
+    let rawCard = await this.realmManager.create(
+      merge({ id, realm, adoptsFrom, data: makeEmptyCardData(allFields) } as RawCard, raw)
```
habdelra, line 115: Until we internalize the serializer into the card model, we'll still need to provide the shape of the card's data here since this bypasses the card model when creating a card

ef4:
Next steps: 
 - as separate PRs, unskip the two tests we had to skip here
 - backfill test coverage for the new compiler output
 - stop round-tripping the serializerMap through CardModel

packages/core/src/card-model.ts
```diff
@@ -249,46 +253,28 @@ export default abstract class CardModel implements CardModelInterface {
...
-
-  private async schemaClass() {
-    if (this._schemaClass) {
-      return this._schemaClass;
+    let result = keySensitiveGet(this.rawData, fieldPath);
+    if ('missing' in result) {
+      throw new Error(`TODO: ${result.missing}`);
+    } else {
+      // TODO it would be wonderful if the schema instance knew how to deserialize its own fields
+      return serializeField(this.schemaModule.serializerMap, fieldPath, result.value, 'deserialize');
```
habdelra, line 276: This is the last instance of the `CardModel` directly handling serialization. Perhaps we could push this into the schema instance? this is a little different in that it is deserializing a specific field rather than all the fields for the format. Maybe that means pushing this whole method into the schema instance?

PR #2775, opened 2022-03-10
Cs 3264 calculate one time drop

tintinthong:
The purpose of this pr is to add the rule for computation of the one-time-drop

```
total_reward=10,000,000 CARD
start_snapshot_block=17265698 (genesis block of cardpay)
end_snapshot_block=25165824
```

Screenshot 2022-03-10 at 9 24 22 PM


tintinthong:
Ok have to work on protocol fee payment... during registration of merchant

packages/cardpay-reward-programs/cardpay_reward_programs/rules/retro_airdrop.py
```diff
@@ -0,0 +1,92 @@
+import pandas as pd
+from cardpay_reward_programs.config import default_payment_list
+from cardpay_reward_programs.rule import Rule
+
+
+class RetroAirdrop(Rule):
+    """
+    reward_per_transaction = (total_n_payments in start_block< x <= end_block/ total_n_payments from  start_snapshot_block < x <= end_snapshot_block)
+    reward_per_payee (amount) = (total_n_payments of payee in start_block < x <= end_block) * reward_per_transaction
```
IanCal, line 9: I don't think this is correct - the start & end block are irrelevant here, it's the snapshot ones we care about for the calculation.

Start & end blocks we talked about being the range in which the *rule calculation can be triggered* but here we're doing just one - the start / end for the *snapshot* blocks are about what data range we care about in the calculation.

packages/cardpay-reward-programs/cardpay_reward_programs/rules/retro_airdrop.py
```diff
@@ -0,0 +1,92 @@
...
+        reward_per_transaction = self.get_reward_per_transaction()
+        if reward_per_transaction == 0:
+            return default_payment_list
+        if df.empty:
+            return default_payment_list.copy()
+        new_df = df.copy()
+        new_df["rewardProgramID"] = reward_program_id
+        new_df["paymentCycle"] = payment_cycle
+        new_df["validFrom"] = self.end_block
+        new_df["validTo"] = self.end_block + self.duration
```
IanCal, line 61: I think we probably want to set these explicitly

IanCal:
We'll need to remove test accounts from this, and set their reward to 100.

If a dataframe is in scope you can use it as a table in duckdb, so here test_accounts is a dataframe

```
In [34]: con.execute(f"""SELECT prepaid_card_owner,
    ...:        count(*)  AS transactions,
    ...:        count(*) * (6000000.0 - 100*{len(test_accounts)}) / ( sum(count(*)) OVER () ) AS reward
    ...: FROM   parquet_scan({local_files}) 
    ...: WHERE  _block_number < 20986720 and prepaid_card_owner not in (select prepaid_card_owner from test_accounts)
    ...: GROUP  BY prepaid_card_owner
    ...: ORDER  BY count(*) DESC """).fetchdf()

```

I'd not worry about trying to adjust the general rule class for this, connecting to a db is a small addition, and rules can have custom logic in them. The external interface can remain the same (calling run) and under that use whatever is best.

IanCal:
Code moved to https://github.com/cardstack/cardstack/pull/2792

PR #2776, opened 2022-03-10
Business accounts -> payment profiles

Aierie:
The main change is replacing business / business account  with profile / payment profile. Some specific requested changes can be seen in CS-2634, CS-3315

Where this doesn't make as much sense I've opted to omit "business". eg. "claim business revenue" -> "claim revenue". Some places have copy adjustments that fit secondary requested changes, eg. validation message change from "The Business ID can only contain..." -> "Unique ID can only contain..." since the context of use uses "Unique ID".


packages/web-client/app/components/card-pay/create-merchant-workflow/merchant-customization/index.hbs
```diff
@@ -28,7 +28,7 @@
           {{or this.merchantBgColor "#0069F9"}}
         </div>
       </Boxel::Field>
-      <Boxel::Field @label="Manager">
+      <Boxel::Field @label="Owned By Wallet">
```
Aierie, line 31: Not sure about "Owned By Wallet". Wonder if "Owned By" makes sense?

packages/ssr-web/app/components/common/payment-link/index.hbs
```diff
@@ -23,7 +23,7 @@
       href={{@deepLinkPaymentURL}}
       data-test-payment-link-deep-link
     >
-      {{or @cta "Pay Business"}}
+      {{or @cta "Pay with Card Wallet"}}
```
Aierie, line 26: Make the default similar to the payment CTA in https://github.com/cardstack/cardstack/pull/2772

packages/web-client/app/components/card-pay/create-merchant-workflow/prepaid-card-choice/index.hbs
```diff
@@ -1,23 +1,23 @@
 <ActionCardContainer
   class="prepaid-card-choice"
-  @header="Business Account"
+  @header="Payment profile"
```
backspace, line 3: This is `Payment Profile` elsewhere… does it matter? 🤔

packages/web-client/app/components/card-space/create-space-workflow/index.ts
```diff
@@ -293,7 +293,7 @@ class CreateSpaceWorkflow extends Workflow {
       errors.push(FAILURE_REASONS.RESTORATION_UNAUTHENTICATED);
     }
 
-    // TODO: check if there are any business accounts available for Card Space which haven't been used yet
+    // TODO: check if there are any payment profiles available for Card Space which haven't been used yet
```
backspace, line 296: Should we just delete this file already??? So we don’t keep updating it for no reason [CS-3326](https://linear.app/cardstack/issue/CS-3326/delete-create-card-space-workflow)

backspace (approved):
🎉

packages/web-client/app/components/card-pay/create-merchant-workflow/prepaid-card-choice/index.hbs
Aierie: It might matter if we change the formatting of text in ActionCardContainer. Will change

PR #2777, opened 2022-03-10
Profile page suggestions

Aierie:
- Some suggestions to make this page a bit better in a responsive setting. 
- accessibility of a link
- currentColor over css var, since it seems more robust

### DIFF
--- a/packages/ssr-web/app/components/card-space/user-page/index.css
+++ b/packages/ssr-web/app/components/card-space/user-page/index.css
@@ -32,7 +32,11 @@
 }
 
 .card-space-user-page__header svg {
-  fill: var(--card-space-text-color);
+  /*
+    inherit the text color so that the logo does not have contrast issues.
+    background and text color on this page are dynamic.
+  */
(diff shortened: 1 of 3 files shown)


Aierie:
We should try out what happens when there is a somewhat long merchant name with no spaces.

jurgenwerk (approved):
Great, thanks!!

packages/ssr-web/app/components/card-space/user-page/index.css
```diff
@@ -52,6 +52,7 @@
   font-size: var(--boxel-font-size-xl);
   font-weight: 600;
   margin: var(--boxel-sp-xxl) auto;
+  word-break: break-all;
```
Aierie, line 55: @jurgenwerk what do you think about `overflow-wrap` vs `word-break`? I was wondering yesterday about splitting at non-space areas and whether there was anything we could do about it. However I am not very familiar with how this affects non-Latin-alphabet languages.

PR #2778, opened 2022-03-10
ssr: Change listener for production Waypoint

backspace:
Maybe I did something wrong but I did need to change this. This branch is currently
deployed, so it just needs to be updated in `main`.


backspace:
I’m just going to merge…? It’s already deployed.

PR #2779, opened 2022-03-10
ssr: Add profile meta tags

backspace:
This adds meta tags for link previews on profile pages.

image

I originally had more tags in `error.hbs` but they don’t show as link previews so it didn’t seem worth it to me.


packages/ssr-web/app/components/card-space/user-page/index.ts
```diff
@@ -27,13 +28,22 @@ export default class CardSpaceUserPage extends Component<CardSpaceUserPageArgs>
...
   defaultAddressFetchingErrorMsg =
     'We ran into an issue while generating the payment request link. Please reload the page and try again. If the issue persists, please contact support.';
 
   get canDeepLink() {
     return this.UAService.isIOS() || this.UAService.isAndroid();
   }
 
+  get meta() {
+    return {
+      description: 'Cardstack',
```
Aierie, line 41: Would something like `Visit ${name}'s profile page on Card Space` work?

packages/ssr-web/app/components/card-space/user-page/index.ts
```diff
@@ -27,13 +28,22 @@ export default class CardSpaceUserPage extends Component<CardSpaceUserPageArgs>
...
+      title: `${this.args.model.name} Profile`,
```
Aierie, line 42: would `${this.args.model.name}'s Profile` sound nicer?

packages/ssr-web/app/components/card-space/user-page/index.ts
backspace: done, thanks

PR #2781, opened 2022-03-10
ssr: Change profile contents to be vertically centred

backspace:
This is an improvement, IMO:

image

I made the judgment call to add colours to the Mirage response so the Percy snapshot would have the roundrect wrapper, suggestions welcome if the colour scheme is too garish 😆

image


Aierie:
Overall an improvement! Long merchant names (within 50 chars) overlap with the Card Pay logo on smaller + shorter screens. Will adjust the layout to accommodate these.

PR #2782, opened 2022-03-11
ssr: Remove health check log noise

backspace:
This is currently experimental, only pushed for deployment as I can’t figure out how to run it locally via Docker 🤔

### DIFF
--- a/packages/ssr-web/deployment/fastboot-server.js
+++ b/packages/ssr-web/deployment/fastboot-server.js
@@ -3,6 +3,9 @@ const FastBootAppServer = require('fastboot-app-server');
 const fetch = require('node-fetch');
 const Sentry = require('@sentry/node');
 
+const winston = require('winston'),
+  expressWinston = require('express-winston');
+
 Sentry.init({
   dsn: process.env.SSR_WEB_SERVER_SENTRY_DSN,
   environment: process.env.SSR_WEB_ENVIRONMENT,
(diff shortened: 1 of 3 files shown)


backspace:
Fastboot header logging shows the ELB health checker user agent:

```
headers: {
host: [ '10.91.1.150:4000' ],
connection: [ 'close' ],
'user-agent': [ 'ELB-HealthChecker/2.0' ],
'accept-encoding': [ 'gzip, compressed' ]
}
}
```

But it seems like it’s still failing:

```
            "TargetHealth": {
                "State": "unhealthy",
                "Reason": "Target.ResponseCodeMismatch",
                "Description": "Health checks failed with these codes: [404]"
            }
```

It’s unclear to me why…

another approach would be to change the health check path, maybe that’s wiser? That would be a parallel infrastructure change. To be continued

backspace:
re the above comment, the problem was that I wasn’t using the `Headers` object properly. I have chosen to go ahead with the `/` check and special handling for the ELB health checker user agent.

jurgenwerk (commented):
This adds deployment-specific logic to the app which adds app behaviour that depends on where the app is deployed. What about returning 200 when visiting `/`, even if it's just a blank page for now? I imagine `/` will have some content later on anyway?

Aierie:
Would using the fastboot-server for this work? https://github.com/cardstack/cardstack/pull/2790

backspace:
@Aierie’s approach is way better 🤓

PR #2783, opened 2022-03-11
Handling compiler read errors during indexing card from notification

burieberry:
CS-3246


packages/hub/services/search-index.ts
```diff
@@ -240,6 +243,17 @@ class IndexerRun implements IndexerHandle {
     } while (rows.length > 0);
   }
 
+  async read(card: RawCard): Promise<RawCard | void> {
```
ef4, line 246: Instead of `read`, let's call this `reindex` and give it a signature like:

```js
async reindex(card: CardId, action: 'save' | 'delete'): Promise<void>;
```

Then the whole switch statement from `indexCardFromNotification` would move inside `reindex`.

packages/hub/services/search-index.ts
```diff
@@ -240,6 +243,17 @@ class IndexerRun implements IndexerHandle {
...
+    try {
+      let rawCard = await this.realmManager.read(card);
+      return rawCard;
+    } catch (err: any) {
+      let compiler = await this.builder.compileCardFromRaw(card);
```
ef4, line 251: This isn't necessary. When we have an error during read, there are no external dependencies that would cause us to need to try to index the card again. Only changes to the card itself should cause a retry. So you can pass an empty Set to `saveErrorState` below.

burieberry:
> This is close, we left some comments addressing your questions and suggesting a small refactor.

Thank you for the suggestions. The test now passes and the server doesn't blow up anymore. However, the error message still says "unhandled error" when I reproduce the error causing behavior locally:

error

PR #2784, opened 2022-03-12
Reward program refactor

IanCal:
I'll put comments inline but broadly I wanted to:

* Get to a point where I can generate multiple images
* Fix what I think are some confusions around start/end blocks
* Handle the df / paymentlist difference, without changing the logic / moving it all into the sql

One of the key things is *run* is the main part that returns a payment list. As there's logic in the df -> payment list conversion, there were issues when multiple cycles were concatenated.


packages/cardpay-reward-programs/cardpay_reward_programs/main.py
```diff
@@ -13,34 +14,19 @@
...
     boto3_session=Session(
-        aws_access_key_id=os.getenv("AWS_ACCESS_KEY_ID"),
-        aws_secret_access_key=os.getenv("AWS_SECRET_ACCESS_KEY"),
     ),
 )
 cached_client.set_as_default_client()
 
 
-def select_rule(name, core_parameters, user_defined_parameters):
-    rule_constructor = globals()[name]
```
IanCal, line 26: Instantiating a class based on allowing user input to select the code to run feels risky.

packages/cardpay-reward-programs/cardpay_reward_programs/main.py
```diff
@@ -13,34 +14,19 @@
...
-    instance = rule_constructor(core_parameters, user_defined_parameters)
-    return instance
-
-
-def to_rule(parameters):
-    core_parameters, user_defined_parameters = get_parameters(parameters)
-    name = to_camel_case(core_parameters["docker_image"])
```
IanCal, line 33: We can't use the docker image here as 

1. It'll have a different format (start acct_id.useast1.ecr/etc) rather than just "MinSpend"
2. The code here shouldn't know what docker image it's in, it's just run.

packages/cardpay-reward-programs/cardpay_reward_programs/main.py
```diff
@@ -13,34 +14,19 @@
...
-    return select_rule(name, core_parameters, user_defined_parameters)
-
-
 def run_reward_program(
-    reward_program_id: str,
```
IanCal, line 38: Positional arguments here will let us define a launch process that's easier to manage across images. The rule name being first means I can set the entrypoint on the docker images I think so that we can build one per image.

packages/cardpay-reward-programs/cardpay_reward_programs/main.py
```diff
@@ -49,11 +35,11 @@ def run_reward_program(
     """
     with open(AnyPath(parameters_file), "r") as stream:
         parameters = json.load(stream)
-    rule = to_rule(parameters)
-    start_block = rule.start_block
-    end_block = rule.end_block
-    results_df = rule.run(start_block, end_block)
-    payment_list = rule.df_to_payment_list(results_df, payment_cycle, reward_program_id)
+    
+    for subclass in Rule.__subclasses__():
```
IanCal, line 39: A safer approach than globals, but achieves the same sort of thing. Looking for a class based on name, but it's limited to Rules.

packages/cardpay-reward-programs/cardpay_reward_programs/main.py
```diff
@@ -49,11 +35,11 @@ def run_reward_program(
     """
     with open(AnyPath(parameters_file), "r") as stream:
         parameters = json.load(stream)
-    rule = to_rule(parameters)
-    start_block = rule.start_block
-    end_block = rule.end_block
-    results_df = rule.run(start_block, end_block)
```
IanCal, line 55: The start block and end block here would be the valid range of the rule, not the range for one payment cycle - and how to pick the range should be an internal decision in the rule.

packages/cardpay-reward-programs/cardpay_reward_programs/rule.py
```diff
@@ -30,8 +29,8 @@ def set_user_defined_parameters(
     def sql(self, table_query):
         raise NotImplementedError
 
-    def _get_table_query(self, table_name, min_partition: int, max_partition: int):
-        config_location = self.subgraph_config_location[table_name]
+    def _get_table_query(self, config_name, table_name, min_partition: int, max_partition: int):
```
IanCal, line 32: The table name and which subgraph config part to extract are not the same.

packages/cardpay-reward-programs/cardpay_reward_programs/rules/min_other_merchants_paid.py
```diff
@@ -39,25 +39,21 @@ def df_to_payment_list(
         new_df = df.copy().groupby("payee").agg({"merchant": "nunique"}).reset_index()
         new_df["rewardProgramID"] = reward_program_id
         new_df["paymentCycle"] = payment_cycle
-        new_df["validFrom"] = self.end_block
```
IanCal, line 42: The end block here wouldn't be right - a cashback program that is calculated each month for a year should be claimable before the end of the whole year.

packages/cardpay-reward-programs/cardpay_reward_programs/rules/min_spend.py
```diff
@@ -24,10 +24,10 @@ def sql(self, table_query):
         sum(spend_amount_uint64) as total_spent
 
         from {table_query}
-        where block_number_uint64 > ?::integer and block_number_uint64 <= ?::integer 
+        where block_number_uint64 > $1::integer and block_number_uint64 <= $2::integer 
         
         group by prepaid_card_owner
-        having(total_spent) >= ?::integer
+        having(total_spent) >= $3::integer
```
IanCal, line 30: Found that duckdb supports numbered parameters - not as nice as named ones but they are not implemented. I think this will be a little easier to keep going.

packages/cardpay-reward-programs/streamlit/views/utils.py
```diff
@@ -40,3 +46,23 @@ def download_csv(payment_list):
         file_name="payment_list.csv",
         mime="text/csv",
     )
+
+
+def view_multiple(program):
```
IanCal, line 51: These were actually pretty commonly equal across the different programs, and adding more functionality should be broadly applicable.

packages/cardpay-reward-programs/streamlit/views/utils.py
```diff
@@ -40,3 +46,23 @@ def download_csv(payment_list):
...
+    start_block, end_block = slider_partition(type="two_end")
+    progress = st.progress(0.0)
+    payment_lists = []
+    for i in range(start_block, end_block, program.payment_cycle_length):
+        progress.progress((i - start_block) / (end_block - start_block))
+        tail = min(end_block, i + program.payment_cycle_length)
+        payment_list = program.run(tail, "0x0")
+        if not payment_list.empty:
+            payment_lists.append(payment_list)
+    combined_payment_list = pd.concat(payment_lists)
```
IanCal, line 61: The concat happening here still works for everything, just shifted out of the specific rule class.

packages/cardpay-reward-programs/tests/cardpay-staging-partitioned-graph-data/data/staging_rewards/0.0.1/latest.yaml
```diff
@@ -1,3 +1,5 @@
 subgraph: habdelra/cardpay-sokol
 subgraph_deployment: QmR4zkVGxHYVKi2TJoN8Xxni3RNgQ2Vs2Tw4booFgLUWdt
 updated: 2022-01-31 22:43:31.534100
+earliest_block: 23592960
+latest_block: 24859648
```
IanCal, line 5: To support the new approach for finding files.

packages/cardpay-reward-programs/tests/rules/test_min_other_merchants_paid.py
```diff
@@ -1,5 +1,7 @@
 import hashlib
 import itertools
+from tracemalloc import start
+from turtle import st
```
IanCal, line 4: VSCode being helpful - these need to be removed.

packages/cardpay-reward-programs/tests/rules/test_min_other_merchants_paid.py
```diff
@@ -50,60 +46,48 @@ def rule(request):
...
-        h = hashlib.sha256(pd.util.hash_pandas_object(df, index=True).values).hexdigest()
+        rule.payment_cycle_length = end_block - start_block
+        payment_list = rule.run(end_block, "0x0")
         computed_summary = rule.get_summary(payment_list)
         assert computed_summary["total_reward"][0] == summary["total_reward"]
         assert computed_summary["unique_payee"][0] == summary["unique_payee"]
-        assert h == df_hash
 
 
-multiple_df_hashes = [
```
IanCal, line 73: These hashes were making the test very brittle, and quite hard to interpret failures - I'm not sure how best to update these either.

packages/cardpay-reward-programs/tests/rules/test_min_other_merchants_paid.py
```diff
@@ -50,60 +46,48 @@ def rule(request):
...
-    "e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855",
-    "f45c509908a1eba5eab02d30112b3d5153a9939df3d0e58cb4a11fadb11272c8",
-    "50bb631ce5832059f9d55de834af8aef712970adbcbe64696d261771967975c8",
-]
+
 multiple_summaries = [
     {"total_reward": 0, "unique_payee": 0},
     {"total_reward": 10, "unique_payee": 1},
-    {"total_reward": 20, "unique_payee": 2},
+    {"total_reward": 30, "unique_payee": 2},
```
IanCal, line 70: I want to verify these, but I am fairly sure these were incorrect before - because the aggregation of multiple payment cycles was then having further logic applied it was combining things across payment cycles.

packages/cardpay-reward-programs/tests/rules/test_min_spend.py
```diff
@@ -1,3 +1,4 @@
+from audioop import mul
```
IanCal, line 1: VSCode being useful again. Should be removed

packages/cardpay-reward-programs/cardpay_reward_programs/rules/min_other_merchants_paid.py
tintinthong: I **should** have put `payment_cycle`.

packages/cardpay-reward-programs/cardpay_reward_programs/main.py
tintinthong: I **should** have put this as `payment_cycle`

packages/cardpay-reward-programs/cardpay_reward_programs/rules/min_other_merchants_paid.py
tintinthong: Strongly Agreed. I **should** have put `payment_cycle` here

packages/cardpay-reward-programs/cardpay_reward_programs/main.py
tintinthong: Strongly agree!

packages/cardpay-reward-programs/cardpay_reward_programs/main.py
tintinthong: ok Agree. The `docker_image` is not the correct key to dictate the `rule_name`. 

Note: I excluded the `rule_name` from the `parameters.json`. Shall it be re-inserted into `run`. I don't have strong opinion about this.

packages/cardpay-reward-programs/cardpay_reward_programs/main.py
```diff
@@ -49,11 +35,11 @@ def run_reward_program(
...
-    payment_list = rule.df_to_payment_list(results_df, payment_cycle, reward_program_id)
+    
+    for subclass in Rule.__subclasses__():
+        if subclass.__name__ == rule_name:
+            rule = subclass(parameters["core"], parameters["user_defined"])
+    payment_list = rule.run(parameters["run"]["payment_cycle"], parameters["run"]["reward_program_id"])
```
tintinthong, line 42: I'm still unsure:
- if the `run` key is actually available after `run_reward_program` or before `run_reward_program`. I have always thought `run` key is added only after `run_reward_program` as a record to s3.

I'll leave this to you since it's highly dependent on how the scheduler works.

packages/cardpay-reward-programs/tests/rules/test_min_other_merchants_paid.py
tintinthong: Yea strongly agree. It was working initially then it was hell once I started changing the code.

packages/cardpay-reward-programs/cardpay_reward_programs/main.py
tintinthong: We fork here. 

`run(start_block, end_block)` vs `run(payment_cycle, reward_program_id)`

I always thought placing `payment_cycle` and letting the rule infer the range based off the `payment_cycle` will cause us to need to update the rule code when `payment_cycle` are conflicting across rules.

packages/cardpay-reward-programs/cardpay_reward_programs/rule.py
tintinthong: What do you mean by "subgraph_config_part"?

packages/cardpay-reward-programs/tests/rules/test_min_other_merchants_paid.py
tintinthong: The previous test was right, if you consider `aggregate` as a way to combine the rule of two or more ranges. 

for example, 

`aggregate` of: 
- `start_block_1 < x <= start_block_2`
- `start_block_2< x <= start_block_3`

will yield 
`aggregate` of 
-`start_block_1` < x <= `start_block_2`

Perhaps unnecessarily complex. Let's forget about this logic.

packages/cardpay-reward-programs/tests/rules/test_min_spend.py
```diff
@@ -50,58 +40,48 @@ def rule(request):
...
         assert computed_summary["unique_payee"][0] == summary["unique_payee"]
 
 
-multiple_df_hashes = [
-    "016f5967731732f8879315395291cd9c95ad83e5c0364b11e41b0efa3980ace4",
-    "ab33d395cbf0ebb1d5baa21ae2d3b0af5093ddef6e5c0d93f42a5a7ca1ecb08d",
-    "c3f4c2d3c1978f2d1770e2d1c564951ff6ff028d2d68b2b442d77cf41351edf6",
-]
+
 multiple_summaries = [
```
tintinthong, line 61: It follows here. That the reason these test values changed is because of the way `aggregate` works. 

My way: Combine dfs in consideration of the `min_spend` criteria
Your way: Combine payment list

Again lets remove this. Sorry if it was not clearly communicated.

tintinthong (approved):
I made some comments. I think the only major comment is around "using `payment_cycle`" as an argument.

packages/cardpay-reward-programs/cardpay_reward_programs/rule.py
IanCal: Yeah that sentence didn't make much sense

So if a rule did something using the rewards extracted data and the payments, it might have:

```
"subgraph_config_locations": {
    "prepaid_card_payment": "s3://...",
    "rewards": "s3://..."
}
```

Then the "config_name" (we could come up with a better name) here would be "prepaid_card_payment" or "rewards". Prepaid card payment example we have has just one table in it, but others (like the rewards one) have multiple.

packages/cardpay-reward-programs/cardpay_reward_programs/main.py
IanCal: I think this is before - when we get to any of the code in this package running, it's for a specific known payment cycle and for one reward program.

packages/cardpay-reward-programs/cardpay_reward_programs/main.py
IanCal: Yeah I think this is the main difference - I'm not sure what the conflict would be there, let's talk that through on Monday.

IanCal:
> I made some comments. I think the only major comment is around "using `payment_cycle`" as an argument. Thanks for making these wide-reaching changes.

No problem, it took me a couple of iterations to try out some things I was thinking about and thought it'd be better done like this than with comments on the other PR. Let's talk through the payment cycle stuff and make sure we're both talking about the same thing there.

packages/cardpay-reward-programs/cardpay_reward_programs/rule.py
tintinthong: Oh ok. Got it.

PR #2785, opened 2022-03-14
Web client/remove card space

Aierie:
Card Space has been moved to ssr-web and some stuff is still in flux so it's unknown if any of the stuff here will be used. If we need this stuff we can refer to it in the git history.

Also removed the Card Space feature flag + older Card Pay ones.


packages/web-client/app/routes/card-pay/payments.ts
```diff
@@ -23,7 +22,7 @@ const MERCHANTS_PANEL = {
         'Confirm transactions with Touch ID or Face ID in the Card Wallet mobile app',
       ],
       cta: 'Create Payment Profile',
-      isCtaDisabled: !config.features.createMerchant,
+      isCtaDisabled: false,
```
backspace, line 25: Maybe just remove? But maybe also who cares 😆

backspace (approved):
🎉

packages/web-client/app/routes/card-pay/payments.ts
Aierie: I was wondering about removing that too, but decided against it since this pattern of introducing things and then disabling them seems to be a thing.

PR #2786, opened 2022-03-14
withdrawal: take first compatible safe if no depot

Aierie:
Pick the first compatible safe if depot is not present.


jurgenwerk (approved):
Thanks, nice catch!

PR #2789, opened 2022-03-15
Fix/mobile card space layout

Aierie:
- Uncapping aspect ratio on short, small screens prevents this:

The card space for merchie1 on a short small screen shows the content reaching vertical space of the Card Pay logo/link.

-  Correction - `overflow-wrap: break-word;` is the correct value, not `break-all`

- Prevent resizing on large screens having a similar problem to the overlap problem on small screens by enforcing a min-height.


jurgenwerk (approved):
I tested this and it works as advertised!

PR #2790, opened 2022-03-15
alternative approach to health check handling

Aierie:
Haven't set the condition yet - but would this approach work?

### DIFF
--- a/packages/ssr-web/deployment/fastboot-server.js
+++ b/packages/ssr-web/deployment/fastboot-server.js
@@ -3,6 +3,21 @@ const FastBootAppServer = require('fastboot-app-server');
 const fetch = require('node-fetch');
 const Sentry = require('@sentry/node');
 
+function healthCheckMiddleware(req, res, next) {
+  let isHealthCheck = req.get('host').includes('merchie1');
+  if (isHealthCheck) {
+    console.log(
+      'intercepting request because it is a health check, host:',
+      req.get('host'),
+      ', user-agent:',
+      res.get('user-agent')
+    );
+    res.status(200).send('fastboot server is up and running');
+  } else {
+    next();
+  }
+}
+
 Sentry.init({
   dsn: process.env.SSR_WEB_SERVER_SENTRY_DSN,
   environment: process.env.SSR_WEB_ENVIRONMENT,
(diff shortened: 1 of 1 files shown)


packages/ssr-web/deployment/fastboot-server.js
```diff
@@ -4,10 +4,10 @@ const fetch = require('node-fetch');
...
 function healthCheckMiddleware(req, res, next) {
-  let isHealthCheck = req.get('host').includes('merchie1');
+  let isHealthCheck = req.get('user-agent').includes('ELB-HealthChecker');
   if (isHealthCheck) {
     console.log(
-      'intercepting request because it is a health check, host:',
+      'intercepting request because it is a health check, user-agent:',
       req.get('host'),
       ', user-agent:',
       res.get('user-agent')
```
Aierie, line 13: oops i mistakenly did res.get('user-agent') instead of req

backspace: ah yes… that explains what I’m seeing. Though I also changed the wrong thing and things are mislabeled now 🙃

but it seems like it _is_ persisting through and producing the allowlist warning? 🤔 

```
2022-03-15T09:06:23-05:00: intercepting request because it is a health check, user-agent: 10.91.1.132:4000 , user-agent: undefined
2022-03-15T09:06:39-05:00: Error while processing route: index The host header did not match a hostWhitelist entry. Host header: 10.91.1.248:4000 Error: The host header did not match a hostWhitelist entry. Host header: 10.91.1.248:4000
```

or could it be that there are still old tasks running as the switch-over happens via Waypoint?  hmmmmmmmmm

backspace: I don’t know if the Waypoint theory was correct but with the latest deployment there are no more VPC ID problems!

```
2022-03-15T09:19:56-05:00: intercepting request because it is a health check, host: 10.91.2.152:4000 , user-agent: ELB-HealthChecker/2.0
2022-03-15T09:20:02-05:00: not health check, proceeding. host: wallet-staging.stack.cards, user-agent: Checkly/1.0 (https://www.checklyhq.com)
2022-03-15T09:20:05-05:00: 2022-03-15T14:20:05.550Z 200 OK /pay/sokol/0x3a31B39bECf64aF7713CDF63aE5C760166E2CF21
2022-03-15T09:20:11-05:00: intercepting request because it is a health check, host: 10.91.1.65:4000 , user-agent: ELB-HealthChecker/2.0
```

jurgenwerk (approved):
Love it!

PR #2792, opened 2022-03-16
Add retro airdrop functionality

IanCal:
Migrated from https://github.com/cardstack/cardstack/pull/2775

I'm just excluding the test accounts here, considering having a flat drop for the test accounts before the live one, can discuss.

Noticed that the amounts here are floats, I want to investigate the precision here and what we need to do to check that we're working at the right scale.


packages/cardpay-reward-programs/cardpay_reward_programs/payment_tree.py
```diff
@@ -44,9 +44,8 @@ def encode_payment(payment: Payment) -> bytes:
         tokenType: int = 1
         payee: ChecksumAddress = payment["payee"]
         token: ChecksumAddress = payment["token"]
-        amount: int = payment["amount"]
+        amount: int = int(payment["amount"])
```
tintinthong, line 47: 👍

packages/cardpay-reward-programs/cardpay_reward_programs/rules/retro_airdrop.py
```diff
@@ -0,0 +1,71 @@
...
+        reward_per_transaction = self.total_reward / total_n_payments
+        if total_n_payments == 0:
+            return 0
+        else:
+            return reward_per_transaction
+
+    def df_to_payment_list(
+        self, df, payment_cycle, reward_program_id
+    ):
+        mask = df['payee'].str.lower().isin(self.excluded_accounts)
```
tintinthong, line 52: 👍

tintinthong (approved):
Flat drop logic looks good. Alot of the streamlit flows, github action, are broken with the intro of project.toml. But I guess that can come after.

PR #2794, opened 2022-03-17
Fix pool balance

tintinthong:
Pool balance was not being displayed when triggered from cli. The origin of the issue is that was checking balance based of `rewardTokensAvailable` which is an API-specific to tally proofs (since we have not rewarded tokens, there are no reward tokens available). 

**Solution**

I have coded `get_reward_tokens` that are available in sdk. It should change depending on the network. I have restricted only to card tokens since they are the only reward tokens. 

Screenshot 2022-03-17 at 11 40 29 AM

### DIFF
--- a/packages/cardpay-sdk/sdk/reward-pool/base.ts
+++ b/packages/cardpay-sdk/sdk/reward-pool/base.ts
@@ -63,10 +63,6 @@ export default class RewardPool {
 
   constructor(private layer2Web3: Web3) {}
 
-  async getCurrentPaymentCycle(): Promise<number> {
-    return 2;
-  }
-
   async getBalance(address: string, rewardProgramId?: string, tokenAddress?: string): Promise<BN> {
     const unclaimedProofs = await this.getProofs(address, rewardProgramId, tokenAddress, false);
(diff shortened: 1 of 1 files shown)


packages/cardpay-sdk/sdk/reward-pool/base.ts
```diff
@@ -63,11 +63,7 @@ export default class RewardPool {
 
   constructor(private layer2Web3: Web3) {}
 
-  async getCurrentPaymentCycle(): Promise<number> {
```
tintinthong, line 66: this was hanging

packages/cardpay-sdk/sdk/reward-pool/base.ts
```diff
@@ -63,11 +63,7 @@ export default class RewardPool {
...
-    return 2;
-  }
-
-  async getBalance(address: string, rewardProgramId?: string, tokenAddress?: string): Promise<BN> {
```
tintinthong, line 70: make tokenAddress a compulsory field, so that summing up logic will make sense. If there are two reward tokens, the summing up logic would be wrong

PR #2800, opened 2022-03-17
cardbot: Add $5 SKU image and remove § from $10

backspace:


lukemelia:
Hmm, did you try this out? Is the increased size going to be a problem?

Aierie (approved):

PR #2804, opened 2022-03-18
Ssr web auth

Aierie:
Move auth component and necessary services from web-client to ssr-web


backspace:
How can I exercise this? 🤔

Aierie:
@backspace https://github.com/cardstack/cardstack/pull/2804/commits/fee0d489593ae8f38ca4ea02b80cac07886b09f3 has a temporary demo route. For videos of how this looks, see CS-3407

packages/ssr-web/app/utils/wallet-info.ts
```diff
@@ -0,0 +1,43 @@
+class WalletAccount {
+  constructor(
+    readonly address: string,
+    readonly name: string,
+    readonly imageBase64: string | undefined = undefined
+  ) {}
```
backspace, line 6: What does it mean to have arguments but not assign them? 😯

backspace (approved):
I don’t love the prospect of ensuring we update the services copied from `web-client` but here we are with semi-overlapping Ember applications 😖

Thanks for adding the temporary route, I deployed to staging and it worked well. I didn’t do a deep dive on this and there are definitely some parts I don’t understand, hence the constructor question, but from a high level it looks 👍🏻

packages/ssr-web/app/utils/wallet-info.ts
Aierie: Typescript magic for assigning properties from arguments, replacing
```js
class Something {
  readonly property: number;
  constructor(property: number) { this.property = number }
}
```

https://www.typescriptlang.org/docs/handbook/2/classes.html#parameter-properties

Aierie:
> I don’t love the prospect of ensuring we update the services copied from `web-client` but here we are with semi-overlapping Ember applications 😖

it's not great. but while working on this i realised that the way that the auth is used is kind of different. i feel like for the most part we should not be updating the services here that much, unless we are doing refactors - which seem like they won't be happening that soon.

one potential update which both might benefit from is the WalletConnect deep link target (if on mobile, which wallet should open when you perform a signing request?). however for card pay there are potentially 2 WalletConnect instances (layer 1 and layer 2) so we might need a different approach than modifying a common localStorage property, + Card Pay is not mobile-ready yet.

packages/ssr-web/app/utils/wallet-info.ts
backspace: oh nice 🎉

PR #2808, opened 2022-03-18
use browserslist-recommended browsers for ssr-web

Aierie:
not you, ie 11 (it seems to actually break things if ie 11 is included btw).

Stuff works on my machine ™️.  linters (particularly stylelint) to follow in separate PR.


Aierie:
When we're clearer about requirements/if we see a need to we can remove support as necessary.

backspace (approved):

PR #2809, opened 2022-03-18
fix qr code poking out of allocated area

Aierie:
## Before
Bug: QR Code in the mobile payment request page extends outside of its allocated width

## After
QR code in the mobile payment request page now stays within its allocated width


backspace (approved):
I was able to confirm this locally by making the window narrow and increasing zoom, nice fix 🎉

PR #2810, opened 2022-03-18
publish: Update key to break cache

backspace:
This is an attempt to solve a [mysterious Boxel failure](https://github.com/cardstack/cardstack/runs/5604977632?check_suite_focus=true) when publishing.


backspace:
wrong approach!

PR #2811, opened 2022-03-18
boxel: Add explicit import of qunit-dom

backspace:
This is an attempt to address this mysterious Boxel failure
when publishing:
https://github.com/cardstack/cardstack/runs/5605597962?check_suite_focus=true#step:6:66


danibonilha (approved):
LGTM

PR #2815, opened 2022-03-21
cardbot: Change Card Wallet download copy

backspace:
The links are wrapped in angle brackets so they don’t show previews.

image


lukemelia (approved):
👍  LGTM!

PR #2816, opened 2022-03-22
add fine print to app name

Aierie:
Design notes that we should refer to the wallet as "Card Wallet by Cardstack"

Screenshot of the app name used on the merchant payment request page. App name now reads 'Card Wallet by Cardstack', where before it read 'Card Wallet'

### DIFF
--- a/packages/ssr-web/app/components/card-pay/merchant-payment-request-card/index.css
+++ b/packages/ssr-web/app/components/card-pay/merchant-payment-request-card/index.css
@@ -37,14 +37,14 @@
 .merchant-payment-request-card__card-wallet-text {
   display: flex;
   flex-direction: column;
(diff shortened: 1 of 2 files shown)


packages/ssr-web/app/components/card-pay/merchant-payment-request-card/index.css
```diff
@@ -34,6 +34,19 @@
   flex-shrink: 0;
 }
 
+.merchant-payment-request-card__card-wallet-text {
+  display: flex;
+  flex-direction: column;
+  font-weight: bold;
```
jurgenwerk, line 40: Use a weight number to be consistent with other css?

jurgenwerk (approved):

PR #2819, opened 2022-03-22
Handle unconsumed field imports

burieberry:


packages/hub/node-tests/@core/babel-plugin-card-schema-transform-test.ts
```diff
@@ -143,6 +143,53 @@ if (process.env.COMPILER) {
...
+              import { contains } from "@cardstack/types";
+              import string from "https://cardstack.com/base/string";
+              import * as foo from "../bar";
+              import { bar3 } from "../bar";
+              import bar2, { bar1, bar4 } from "../bar";
+              import { bar5 as bar } from "../bar";
+
+              export default class Test {
+                @contains(string) name;
+              }
```
burieberry, line 186: This PR works in terms of removing the unconsumed imports. 

However, if we want to enable something like this:
```
import bar2, { bar1, bar4 } from "../bar";

export default class Test {
    @contains(string) bar2;
}
```
Then the logic for the handling of imports would need to change. For example, for the above example to work, we would need to remove this part from the schema transform code:
```
if (path.node.specifiers.length !== 1 || !t.isImportDefaultSpecifier(path.node.specifiers[0])) {
            throw error(path, `expecting a default import for card`);
}
```
If all we care about for this PR to remove unconsumed imports, then it's ready, otherwise there's more work to do. I'm not sure if we want to allow all import types, so leaving this PR as is.

habdelra: sure--we require cards be default exports. however, the unused import might not actually be a card--it could just be a helper-or maybe it's stuff from lodash. in which case that card logic should not come into play

burieberry: > sure--we require cards be default exports. however, the unused import might not actually be a card--it could just be a helper-or maybe it's stuff from lodash. in which case that card logic should not come into play

Updated the test examples. Does it now cover those cases?

packages/hub/node-tests/@core/babel-plugin-card-schema-transform-test.ts
```diff
@@ -143,6 +143,54 @@ if (process.env.COMPILER) {
...
+            import { bar3 } from "../bar";
+            import bar2, { bar1, foo as foo1 } from "../bar";
+
+            export default class Test {
+              @contains(string) name;
+            }
+          `,
+        },
+      });
+      expect(card.compiled.url).to.eq(`${realm}test`);
```
habdelra, line 191: we should also assert that in the schema source we have pruned the unconsumed imports

burieberry: Done

habdelra (approved):

PR #2821, opened 2022-03-23
Adding isValid for proofs 

tintinthong:
Adding `Proof` flag `isValid`. A proof `isValid` when `validFrom<= currentBlock < validTo`. [See here](https://github.com/cardstack/card-pay-protocol/blob/58c3d0a6e99a1f9fa1895d701318a420c55e58b6/contracts/RewardPool.sol#L118)

- All balance functions `rewardTokenBalances` and `getBalance` will compute balance only for proofs which are `isValid`
- `getProofs` will return proofs with the flag
- cli adds a `isValidOnly` flag that allows you to filter proofs via the `claimable-proofs` command


IanCal:
I can confirm this works in the CLI to list my claims and marks two as invalid (as expected).

PR #2827, opened 2022-03-24
component transphone phase part 2

mattmcmanus:
- Prep for component analyze options for more information
- Track component rawHBS and whether there is a scope
- remove rpcArchiveNode from SDK--this is unsued and just creates confusion (#2823)
- Handle unconsumed field imports (#2819)
- Move inlineHBS consideration out of component transform


packages/core/src/babel-plugin-card-template.ts
```diff
@@ -306,3 +307,84 @@ function updateScope(options: NodePath<t.ObjectExpression>, names: Set<string>,
...
+      usedFields.add(`${prefix}.${fieldPath}`);
+    } else {
+      usedFields.add(fieldPath);
+    }
+  }
+}
+
+function canInlineHBS(
+  hasModifiedScope: boolean,
+  usageMeta: TemplateUsageMeta,
```
mattmcmanus, line 365: Because of the "default" format of fields, we "need" to use the usageMeta here instead of simply usedFields. This will require us, when we move this logic out of this plugin and into the main part of the compiler to expose the UsageMeta interface. If we want to drop any reference to the field list we could simply use the usedFields list, which would allow us to not have to expose it, but would complicate our ability to specific a format when we invoke a component for a field. Can we update the used fields list to include the format info?

ef4: I don't think the used field list should gain the format. We don't need to know the answer to canInlineHBS until the transform stage, and by then we should already be able to replace "default" with the real format.

ef4:
Looks good, merged so we can do the next stage as another PR.

PR #2829, opened 2022-03-24
Fix schema module loading for base card and its descendants

burieberry:
CS-3461


packages/core/src/card-model.ts
```diff
@@ -227,20 +227,23 @@ export default class CardModel {
...
-            this.state.type === 'loaded' ? 'card ' + this.url : 'new card of type ' + this.parentCardURL
-          }`
-        );
-        newError.additionalErrors = [err, ...(err.additionalErrors || [])];
-        throw newError;
-      }
-      if (this.recomputePromise !== recomputePromise) {
-        return;
+    let fields = this.schemaModule.default.loadedFields(newSchemaInstance);
+    if (fields.length) {
```
habdelra, line 231: i'm unsure why we need this guard. if `fields.length` is 0, then wont the for loop just do nothing?

burieberry: this must be left over from earlier, I'll remove it

packages/hub/node-tests/@core/babel-plugin-card-schema-transform-test.ts
```diff
@@ -333,7 +333,7 @@ if (process.env.COMPILER) {
       let source = getFileCache().getModule(compiled.schemaModule.global, 'browser');
       expect(source).to.containsSource(`
         static loadedFields(schemaInstance) {
-          return [...schemaInstance.loadedFields];
+          return allFields.length === 0 ? [] : [...schemaInstance.loadedFields];
```
habdelra, line 336: what is `schemaInstance.loadedFields` in this case? ideally it should be an empty array.

packages/core/src/compiler.ts
```diff
@@ -513,9 +513,7 @@ export class Compiler<Identity extends Saved | Unsaved = Saved> {
   }
 
   private getSchemaModuleRef(parentCard: CompiledCard | undefined, schemaPath: string | undefined): ModuleRef {
-    if (isBaseCard(this.cardSource)) {
-      return { global: `${BASE_CARD_URL}/schema` };
```
burieberry, line 517: I'm not sure how this worked before... maybe an artifact now?

packages/core/src/babel-plugin-card-schema-transform.ts
```diff
@@ -740,7 +740,7 @@ function addLoadedFieldsMethod(path: NodePath<t.Class>, state: State, babel: typ
       t.identifier('loadedFields'),
       [t.identifier('schemaInstance')],
       t.blockStatement([
-        babel.template(`return [...schemaInstance.%%loadedFields%%];`)({
+        babel.template(`return allFields.length === 0 ? [] : [...schemaInstance.%%loadedFields%%];`)({
```
habdelra, line 743: let's fix this in the constructor so that `loadedFields` is always correct. I suspect this is why you added a lot of downstream logic that didn't trust the value of `schemaInstance.loadedFields`

packages/core/src/compiler.ts
habdelra: if we don't need this, then definitely get rid of it, and also we can remove the base-card logic in the file-cache's module resolver too: https://github.com/cardstack/cardstack/blob/main/packages/hub/services/file-cache.ts#L142-L149

packages/core/src/babel-plugin-card-schema-transform.ts
habdelra: i think that is the right thing to do.

packages/core/src/babel-plugin-card-schema-transform.ts
```diff
@@ -115,7 +114,7 @@ export default function main(babel: typeof Babel) {
           // you can't upgrade a primitive card to a composite card--you are
           // either a primitive card or a composite card. so if we adopt from a
           // card that is primitive, then we ourselves must be primitive as well.
-          if (type === 'composite' && state.opts.meta.fields && keys(state.opts.meta.fields).length > 0) {
+          if (type === 'composite' && state.opts.meta.fields) {
```
habdelra, line 117: i think you did the correct thing here

packages/core/src/babel-plugin-card-schema-transform.ts
```diff
@@ -134,8 +133,9 @@ export default function main(babel: typeof Babel) {
                     ? [babel.template.ast(`super(rawData, makeComplete, isDeserialized);`) as t.Statement]
                     : []),
                   ...(babel.template(`
-                    this.%%loadedFields%% = makeComplete ? allFields : %%flattenData%%(rawData).map(([fieldName]) => fieldName);
                     this.%%isComplete%% = makeComplete;
+                    if (!rawData) { return; }
+                    this.%%loadedFields%% = makeComplete ? allFields : %%flattenData%%(rawData).map(([fieldName]) => fieldName);
```
habdelra, line 138: this looks good too

packages/core/src/babel-plugin-card-schema-transform.ts
habdelra: it should be ok to adjust the tests that thing the constructor should be empty to no longer being so. The primary job of those tests is to make it easier to see what the code that the compiler emits. since we changed the code that the compiler emits, it makes sense those tests that were expecting primitive cards to have no constructor will have to change now.

PR #2832, opened 2022-03-25
add a responsive degraded service banner to the pay page

Aierie:
On the smaller screen, the entire banner is a link and we only display the incident's name, similar to the mobile app. It's placed below the hamburger because having the hamburger's X and the hamburger button in the same place seems good - having the degraded service banner appear on top will push the hamburger down meaning that it would be lower than the X. On the large screen, we're conforming to the DApp's way of displaying a degraded service banner.

Yet to get design feedback. This aims to solve CS-3431 for the pay page.

small screen pay error page with a notice about degraded service below the top navbar.

large screen pay error page with a notice about degraded service above the top navbar, at the top of the screen.

### DIFF
--- a/packages/ssr-web/app/components/common/degraded-service-banner/index.css
+++ b/packages/ssr-web/app/components/common/degraded-service-banner/index.css
@@ -14,6 +14,7 @@
   max-width: var(--degraded-service-banner-content-max-width, 80ch);
   margin: auto;
   display: flex;
+  align-items: center;
   width: 100%;
 }
 
@@ -23,6 +24,7 @@
   display: flex;
   align-items: center;
   width: var(--degraded-service-banner-icon-area-width);
+  height: var(--degraded-service-banner-icon-area-width);
   margin-right: var(--boxel-sp-xxs);
 }
 
--- a/packages/ssr-web/app/components/common/degraded-service-banner/index.css
+++ b/packages/ssr-web/app/components/common/degraded-service-banner/index.css
@@ -4,20 +4,17 @@
   --icon-background-color: var(--degraded-service-banner-text-color);
   --icon-color: var(--degraded-service-banner-background-color);
 
-  width: 100%;
   min-height: 2rem;
-  display: flex;
-  justify-content: center;
-  align-items: center;
-  padding: var(--boxel-sp-xxs);
   font: 600 var(--boxel-font-sm);
   background: var(--degraded-service-banner-background-color);
   color: var(--degraded-service-banner-text-color);
 }
 
 .degraded-service-banner__container {
-  max-width: calc(var(--degraded-service-banner-icon-area-width) + 80ch);
+  max-width: var(--degraded-service-banner-content-max-width, 80ch);
+  margin: auto;
   display: flex;
+  width: 100%;
 }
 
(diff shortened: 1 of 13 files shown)


Aierie:
@oidoug wondering if you have some guidance for making the small screen version match the wallet?

packages/ssr-web/app/components/common/degraded-service-banner/index.css
```diff
@@ -4,20 +4,17 @@
   --icon-background-color: var(--degraded-service-banner-text-color);
   --icon-color: var(--degraded-service-banner-background-color);
 
-  width: 100%;
   min-height: 2rem;
-  display: flex;
-  justify-content: center;
-  align-items: center;
```
backspace, line 11: The removal of this makes it look a bit too high for me; here’s me twiddling this on and off:

screencast 2022-03-25 10-17-35

Even with it on it looks maybe 1px too high for me but it’s less irksome at least 😆

backspace: oh ACTUALLY I was setting this on `…_container`

backspace (changes requested):
This is a great addition, thanks. I have a sensitivity to vertical centring though 😆

packages/ssr-web/tests/acceptance/pay-test.ts
```diff
@@ -719,4 +719,55 @@ module('Acceptance | pay', function (hooks) {
...
+          }
+        );
+      });
+
+      await visit(`/pay/sok`);
+      await waitFor(MERCHANT_MISSING_MESSAGE);
+
+      assert.dom('[data-test-degraded-service-banner]').isVisible({ count: 1 });
+
+      await percySnapshot(assert);
```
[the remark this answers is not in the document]
Aierie: ya actually it doesn't matter if we only have one.

packages/ssr-web/app/components/common/degraded-service-banner/index.css
Aierie: Ya vertical centering issues become more obvious with long text, thanks

backspace (approved):
🎉

packages/ssr-web/app/components/common/degraded-service-banner/small/index.css
```diff
@@ -0,0 +1,4 @@
+.degraded-service-banner--small {
+  border-radius: 8px;
```
oidoug, line 2: We're using 10 as radius in mobile.

oidoug (commented):
Sorry I couldn't respond in a timely manner. Maybe my feedback is still valuable.

We've used three colors for these banners:
```
info: {
    iconColor: 'appleBlue',
    backgroundColor: '#F7F7F7',
    textColor: 'black',
  },
  warning: {
    iconColor: 'black',
    backgroundColor: '#F9D849',
    textColor: 'black',
  },
  error: {
    iconColor: 'black',
    backgroundColor: '#FF0000',
    textColor: 'white',
  },
```

Also, the "small" radius in mobile is always 10pts.

Screenshot_20220325-150838_Card Wallet

PR #2753, opened 2022-03-02
web: Add PR deployments

backspace:
This adds PR preview deployments for web-client, which can be useful to exercise changes or share potential bug fixes with users. They show in the “checks” list:

image

The deployment hostnames look like:

* `branch-name.web-client-preview.stack.cards` for staging
* `branch-name.web-client-preview.cardstack.com` for production

These domains are routed to S3-backed Cloudfront, which has a `viewer-request` function that rewrites requests:

* `https://branch-name.web-client-preview.stack.cards/card-pay/payments` → S3 bucket `/branch-name/index.html`
* `https://branch-name.web-client-preview.stack.cards/images/icons/card-pay-logo.svg` → S3 bucket `/branch-name/images/icons/card-pay-logo.svg`

See the related Terraform updates for more on this.

There are `s3-preview-staging` and `s3-preview-production` deployment targets that are configured similarly to their unprefixed equivalents, but deployed to the preview S3 buckets.

PR #2834, opened 2022-03-25
ssr: Add PR previews

backspace:
This builds on #2753.


packages/ssr-web/app/services/app-context.ts
```diff
@@ -46,8 +64,12 @@ export default class AppContext extends Service implements AppContextService {
...
   get cardSpaceId() {
     if (this.isCardSpace) {
-      let id = this.host.replace(this.hostSuffixPattern, '') ?? '';
-      return id;
+      if (this.queryIsCardSpace) {
+        return cardSpaceSlugQueryRegex.exec(window.location.search)![1];
+      } else {
+        let id = this.host.replace(this.hostSuffixPattern, '') ?? '';
+        return id;
+      }
```
backspace, line 72: Before I prepare this for review, thoughts on testability for these, @Aierie? I’m reluctant to add tests for this narrow use case but also this hackery is pretty uggers

Aierie: i think this is ok, since this is what the service was created for, and we have a test that can be extended to cover more tricky cases.

Aierie (approved):

packages/ssr-web/config/deploy.js
```diff
@@ -0,0 +1,76 @@
...
+    pipeline: {
+      activateOnDeploy: true,
+      disabled: {},
+      alias: {
+        s3: {
+          as: ['s3Pages', 's3WellKnown', 's3Assets'],
+        },
+        compress: {
+          as: ['compressOwnAssets', 'compressPrepaidCardPatterns'],
+        },
```
Aierie, line 25: This should also be removed

packages/ssr-web/package.json
```diff
@@ -86,6 +86,13 @@
     "ember-cli-babel": "^7.26.3",
     "ember-cli-dependency-checker": "^3.2.0",
     "ember-cli-dependency-lint": "^2.0.1",
+    "ember-cli-deploy": "^1.0.2",
+    "ember-cli-deploy-build": "^2.0.0",
+    "ember-cli-deploy-cloudfront": "^3.1.0",
+    "ember-cli-deploy-compress": "^0.3.2",
+    "ember-cli-deploy-display-revisions": "^2.1.2",
+    "ember-cli-deploy-revision-data": "^1.0.0",
```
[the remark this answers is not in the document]
backspace: we don’t! Removed, thanks. And the other comments too 🎉

PR #2835, opened 2022-03-25
Support having own card as a field

habdelra:


packages/core/src/utils/fields.ts
```diff
@@ -1,22 +1,73 @@
...
 
-  if (paths.length > 1) {
+  if (paths.length > 1 && field.card !== 'self') {
     return getFieldForPath(field.card.fields, tail.join('.'));
   }
 
   return field;
 }
 
+export function buildUsedFieldsListFromUsageMeta(
```
habdelra, line 22: it looks like this moved but an upstream merge moved this back here. I'll fix that

packages/core/src/utils/fields.ts
```diff
@@ -1,22 +1,73 @@
...
+        fieldPath,
+        fields,
+        fieldFormat === 'default' ? defaultFieldFormat : fieldFormat
+      );
+    }
+  }
+
+  return [...usedFields];
+}
+function buildUsedFieldListFromComponents(
```
habdelra, line 50: dito here--need to fix the upstream merge

packages/core/src/compiler.ts
```diff
@@ -91,9 +101,10 @@ export class Compiler<Identity extends Saved | Unsaved = Saved> {
       modules: outputModules,
       deps: [...this.dependencies],
     };
+    return this.resolveFieldsWithPlaceholders(compiledCard);
```
habdelra, line 104: @ef4 i don't think I can move this function to being one that acts only on the fields, as the nature of replacing the placeholders is that I need a handle on the compiled card that is being assembled since it is the thing that is replacing the placeholders

PR #2837, opened 2022-03-28
Retro airdrop, adds test accounts

IanCal:
The test accounts get a reward specified by an additional parameter, this does not reduce the total paid to the intended rewardees and will work even if the test accounts have no payments. Test accounts are excluded from the initial calculation as before (renamed from excluded accounts to test_accounts)

I made it a bit stricter to require checksummed addresses, which I think is reasonable.


tintinthong (approved):
LGTM

IanCal:
Thanks, will try and deal with the build issue (seems to be some setuptools fun) and will then merge.

IanCal:
Was just an issue with paths & requiring site packages to be used as well for the CI run

PR #2838, opened 2022-03-28
Fix no SDP default currency in generateMerchantPaymentUrl

topwebtek7:
Fixed `generateMerchantPaymentUrl` generates payment url with SPD currency when currency parameter is not provided which forces users to pay with SPD.

### DIFF
--- a/packages/cardpay-sdk/sdk/utils/merchant.ts
+++ b/packages/cardpay-sdk/sdk/utils/merchant.ts
@@ -31,12 +31,13 @@ export const generateMerchantPaymentUrl = ({
   merchantSafeID,
   amount,
   network,
-  currency = 'SPD',
+  currency,
 }: MerchantPaymentURLParams) => {
   const handleAmount = amount ? `amount=${amount}&` : '';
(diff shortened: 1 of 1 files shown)


packages/cardpay-sdk/sdk/utils/merchant.ts
```diff
@@ -33,11 +33,10 @@ export const generateMerchantPaymentUrl = ({
   network,
   currency,
 }: MerchantPaymentURLParams) => {
-  const handleAmount = amount ? `amount=${amount}&` : '';
-  const handleCurrency = currency ? `${handleAmount ? '&' : ''}currency=${currency}` : '';
+  const handleAmountAndCurrency = currency ? `${amount ? `amount=${amount}&` : ''}currency=${currency}` : '';
   const https = isUniversalDomain(domain) ? 'https://' : '';
 
-  return `${https}${domain}/pay/${network}/${merchantSafeID}?${handleAmount}${handleCurrency}`;
+  return `${https}${domain}/pay/${network}/${merchantSafeID}?${handleAmountAndCurrency}`;
```
Aierie, line 39: This '?' probably isn't necessary if we don't append params, but I think this is a minor detail.

PR #2852, opened 2022-03-31
Prevent bug where hub authentication service never initializes

Aierie:
Don't wait for a tracked property using `waitForProperty`. Instead we use an `initialized` event to let the hub auth service know that layer 2 network is initialized and it can check its auth.


backspace (approved):
Thanks for tracking this down and fixing it. I have sadness about there being identical code in the two packages but… learning to let go 😆

lukemelia:
@Aierie why doesn't `waitForProperty` work as expected?

backspace:
I think it’s related to [this](https://github.com/machty/ember-concurrency/issues/292)… maybe `this.set` would be a way around it instead of `this.isInitializing = true` too 🤔

Aierie:
Yes, I read the issue @backspace shared, and logged the sequence of events in initialisation of the services. waitForProperty only yields if layer2Network is already initialised, which means it cannot actually wait and respond to setting of the property , I think.

PR #2856, opened 2022-04-04
remove s3-index plugin to fix failed deployments

Aierie:
ember-cli-deploy is picking up this installed plugin and trying to use it for the staging deploy (and I guess it will for the prod deployment too?)


backspace:
eep thanks for catching this 😳

PR #2861, opened 2022-04-04
Wait for nonce advance

habdelra:
This addresses concerns in #CS-3565 around the nonce apparently not advancing in between a series of safe transactions made in the dapp in quick succession. We have introduced an additional wait for the next on-chain safe nonce so that we don't clobber the previous transaction's safe nonce. This does not eliminate the issue conceptually, but gives us a tool to help alleviate it in practice. We think that this could be well paired with an update to the relay service such that it could tell you if there is already an in-flight txn for the desired safe, so the client would know to wait for the safe txn in flight to conclude before making a new signing request. We should also consider making the client aware of these types of errors such that it can perform a retry/resign task to alleviate this scenario.


lukemelia (approved):
Checks out. 🩺

PR #2866, opened 2022-04-06
Web client 404 page

Aierie:
Design spec shows a header and a complete image shown (approx 1350px). I've changed this so that this page cannot scroll, and removed the noninteractive header so that the illustration is less hidden.

## Without header (current)
Multiple viewports' views of a 404 page. A header reads 404, with smaller text that reads "The page you are looking for does not exist". Unlike the image above, there is no noninteractive header that reads "Card Pay"

## With header (previous/matching design spec)
Multiple viewports' views of a 404 page. A header reads 404, with smaller text that reads "The page you are looking for does not exist". There is a noninteractive header that reads "Card Pay".


packages/web-client/tests/acceptance/not-found-test.ts
```diff
@@ -0,0 +1,19 @@
...
+
+    assert.equal(currentURL(), '/blah');
+
+    assert
+      .dom('[data-test-not-found-page]')
+      .containsText('The page you are looking for does not exist');
+
+    assert.dom('[data-test-back-home-link]').exists().hasAttribute('href', '/');
+  });
+});
```
backspace, line 19: Percy perhaps? 🤓

backspace (approved):
I tried this out locally, looks great. Assets are broken for me on the preview deployments 😞

packages/web-client/tests/acceptance/not-found-test.ts
Aierie: Ya I wondered about that but was unsure how much value it brings, esp since our usage is so high

PR #2872, opened 2022-04-07
Fix: scientific numbers string being 

tintinthong:
BN cant read scientific number string. 

Fix using [localeString](https://stackoverflow.com/questions/1685680/how-to-avoid-scientific-notation-for-large-numbers-in-javascript)


danibonilha (approved):
Tested on the app and it looks good

PR #2873, opened 2022-04-07
ssr: Fix link preview image on Twitter

backspace:
image

### DIFF
--- a/packages/ssr-web/app/services/app-context.ts
+++ b/packages/ssr-web/app/services/app-context.ts
@@ -83,6 +83,14 @@ export default class AppContext extends Service implements AppContextService {
         : window.location.search
     );
   }
+
+  getAbsolutePath(relativePath: string) {
+    if (this.fastboot.isFastBoot) {
+      return `${this.fastboot.request.protocol}://${this.host}${relativePath}`;
+    } else {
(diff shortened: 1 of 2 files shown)


packages/ssr-web/app/services/app-context.ts
```diff
@@ -83,6 +83,10 @@ export default class AppContext extends Service implements AppContextService {
         : window.location.search
     );
   }
+
+  getAbsolutePath(relativePath: string) {
+    return `https://${config.universalLinkDomain}${relativePath}`;
```
Aierie, line 88: What about this bothers you @backspace ?

Nit: Would go with throwing if the relative path doesn't start with / just in case folks edit or add on to this, but don't think it's a big deal.

Aierie (approved):

lukemelia (approved):

packages/ssr-web/app/services/app-context.ts
backspace: It’s just sad to me to use a hardcoded domain vs one derived from the request, but it’s not a big deal. I’ll add the leading `/`, thanks.

Ab110055:
Good project

PR #2874, opened 2022-04-08
Add build rule

tintinthong:
Manual dispatch action builds rules in staging and production

Im not totally sure but the workflow dispatch "run" should display when merged into main, now we can use the commands below to test

```
gh workflow list
gh workflow run <workflow-id> --ref cs-2957-build-rule-image -f rule=FlatPayment -f environment=production
```


IanCal:
I used this to build a staging image and got an identical MD5 sum for the resulting file as if I executed a locally built version (flat payments) :)

PR #2878, opened 2022-04-11
use leaf as source of truth

tintinthong:


tintinthong:
so far there have been no problems with this. Will resolve in a differenet way as in this pr so dont need to decode leaf on sdk necessaarily

PR #2879, opened 2022-04-11
Edit screen component

Aierie:
An edit screen component to house the editing forms for a card.xyz page. Changes between modal / full page view based on media query (max-width: 480px).


packages/ssr-web/app/components/edit-screen/wrapper/index.hbs
```diff
@@ -0,0 +1,41 @@
+<div
+  class={{cn "edit-screen-wrapper" (if this.mq (concat "edit-screen-wrapper--" (if this.showPage "page" "modal")))}}
+  ...attributes
+>
+  {{!--
+    The modal has different styling and also different event listeners.
+    We don't want to support Escape to exit editing when presentation is a page.
+    --}}
+  {{#if (not this.mq)}}
```
backspace, line 9: Is there a name for this that would make it more clear what’s happening? I gather this is shown until the media query has run…? 🤔

backspace (approved):
This seems great, we can always refine/add if we find things missing or broken!

packages/ssr-web/app/components/edit-screen/wrapper/index.hbs
Aierie: Yeah, it's shown before the media query has run. I'll rename mq to something more descriptive.

PR #2880, opened 2022-04-13
Make `amount` as optional param for withdraw rewards

tintinthong:
~This pr adds a check that the any calls to `withdrawGasEstimate` to ensure that the safe has enough balance for a withdraw. This helps us narrow down sentry errors. If we see this error, it is an indication that how we are moving `amount` as a param in the `withdrawGasEstimate` is wrong. If we don't see this error, issues with the withdraw is probably due to the deviations of gas estimation~

This PR makes sdk withdraw whole amount from the reward safe. This removes the burden of `amount-gasCost` in the cardwallet. Now, the `withdraw` function handles `amount-gasCost` within it.

### DIFF
--- a/packages/cardpay-sdk/sdk/reward-manager/base.ts
+++ b/packages/cardpay-sdk/sdk/reward-manager/base.ts
@@ -574,7 +574,7 @@ The owner of reward safe ${safeAddress} is ${rewardSafeOwner}, but the signer is
     ).methods.withdraw(rewardManagerAddress, tokenAddress, to, weiAmount);
     let token = new this.layer2Web3.eth.Contract(ERC20ABI as AbiItem[], tokenAddress);
     let safeBalance = new BN(await token.methods.balanceOf(rewardSafeAddress).call());
-    if (safeBalance.lt(new BN(amount))) {
+    if (safeBalance.lt(weiAmount)) {
       throw new Error(
         `Reward safe does not have enough balance to withdraw. The rewardSafe safe ${rewardSafeAddress} balance for token ${tokenAddress} is ${fromWei(
           safeBalance
--- a/packages/cardpay-sdk/sdk/reward-manager/base.ts
+++ b/packages/cardpay-sdk/sdk/reward-manager/base.ts
@@ -572,6 +572,15 @@ The owner of reward safe ${safeAddress} is ${rewardSafeOwner}, but the signer is
     let withdraw = await (
       await this.getRewardSafeDelegate()
     ).methods.withdraw(rewardManagerAddress, tokenAddress, to, weiAmount);
+    let token = new this.layer2Web3.eth.Contract(ERC20ABI as AbiItem[], tokenAddress);
+    let safeBalance = new BN(await token.methods.balanceOf(rewardSafeAddress).call());
+    if (safeBalance.lt(new BN(amount))) {
+      throw new Error(
+        `Reward safe does not have enough balance to withdraw. The rewardSafe safe ${rewardSafeAddress} balance for token ${tokenAddress} is ${fromWei(
+          safeBalance
+        )}, amount being claimed is ${fromWei(weiAmount)}`
+      );
+    }
     let withdrawPayload = withdraw.encodeABI();
     let estimate = await gasEstimate(
       this.layer2Web3,
--- a/packages/cardpay-sdk/sdk/reward-manager/base.ts
+++ b/packages/cardpay-sdk/sdk/reward-manager/base.ts
@@ -433,7 +433,7 @@ export default class RewardManager {
     safeAddress: string,
     to: string,
     tokenAddress: string,
-    amount: string,
+    amount?: string,
     txnOptions?: TransactionOptions,
     contractOptions?: ContractOptions
   ): Promise<SuccessfulTransactionReceipt>;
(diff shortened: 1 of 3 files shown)


packages/cardpay-sdk/sdk/reward-manager/base.ts
```diff
@@ -480,25 +477,65 @@ export default class RewardManager {
...
+      estimate = await gasEstimate(
+        this.layer2Web3,
+        safeAddress,
+        rewardSafeDelegateAddress,
+        '0',
+        withdrawPayload,
+        Operation.DELEGATECALL,
+        tokenAddress
+      );
+      let gasCost = new BN(estimate.safeTxGas).add(new BN(estimate.baseGas)).mul(new BN(estimate.gasPrice));
```
habdelra, line 503: did we want to add a safety buffer for volatile gas tokens here?

habdelra: oops, i meant this comment for line 526

packages/cardpay-sdk/sdk/reward-manager/base.ts
```diff
@@ -480,25 +477,65 @@ export default class RewardManager {
...
+          `Reward safe does not have enough to pay for gas when withdrawing rewards. The reward safe ${safeAddress} balance for token ${tokenAddress} is ${fromWei(
+            safeBalance
+          )}, amount being withdrawn is ${fromWei(amount)}, the gas cost is ${fromWei(gasCost)}`
+        );
+      }
+    } else {
+      //when amount is NOT given, we use safeBalance - gasCost as the withdraw amount
+      //Note: gasCost is estimated with safeBalance not the actual withdraw amount
+      weiAmount = safeBalance;
+      let withdraw = await rewardSafeDelegate.methods.withdraw(rewardManagerAddress, tokenAddress, to, weiAmount);
```
tintinthong, line 515: @habdelra Note this part where the `gasCost` is a function of a `gasEstimation` that uses the **safeBalance** safeBalance. I am not sure whether this will affect us. Typically, before we did `gasEstimation` that uses the **safeBalance - gasCost`. Is it a point to do the gas estimation again before executing the transaction?

packages/cardpay-sdk/sdk/reward-manager/base.ts
tintinthong: Hmmmm sure I guess we can include the buffer. Tho there is this subjective question of what the buffer should be especially with different tokens. I was under the impression that this fix would already solve the issue without the buffer.

packages/cardpay-sdk/sdk/reward-manager/base.ts
habdelra: well the thing is that the relay server uses the on chain oracle to determine how much native coin the gas token is worth. in the time between when you signed the txn and then executed the txn, the value of the gas token may have updated in teh oracle making the amount being deducted for gas insufficient. for stable coin this would almost never happen, but for volatile coins like CARD this is a possibility.

habdelra: (this is actually the reason that we introduced the rate lock idea for how prepaid cards work--this same vulnerability exists when trying to empty out a safe and still have enough left over to pay for gas when using a gas token that can fluctuate dramatically in value)

packages/cardpay-sdk/sdk/reward-manager/base.ts
habdelra: the PR is more around improving DX so that callers don't have to think about gas estimation. but the underlying problem is still there.

tintinthong: By putting all this logic in the sdk, we don't make the amount withdrawn dependent on a gasEstimate that might be lurking minutes before the transaction is executed. I think it DOES reduce the possiblity of the gas estimate deviation. 

We can add a buffer to reduce the possiblity even more.

habdelra: i mean the wallet was probably doing the gas estimation immediately before calling the SDK to do the withdraw, so it's more like you shave ms off the of delta between the estimation and the execution. but i think you shaved less than 1% off of the time that the price could have moved.

tintinthong: That is to my point the question below.

- Before, we were doing two separate gas estimates. One in the wallet. One in the sdk. 
- Now, we are doing a single gas estimate. There is no delta in gas estimations at the moment. But we have to accept that the inital gas estimate is of a DIFFERENT withdraw amount than the "actual" (accounting for gas cost)

habdelra: ah i see what you are trying to say--having a buffer actually effects the parameters to the safe tx, which would mean that it has to be re-estimated, because the withdraw amount is different now (smaller after the buffer). which in turn adds more delay because of the extra round trip to the relay server. damned if you do and damned if you don't. We can forego the buffer for now and see how this plays out without that.

habdelra: let's go with this PR and we'll followup as necessary if we continue to see issues around this.

tintinthong: I guess the buffer. For example adding gasCost= expectedGasCost+ buffer will reduce the likelihood of issues relayer erroring out.

danibonilha:
I'm gonna merge this to publish a new release, and see how it goes 🚀

PR #2882, opened 2022-04-13
Fix iOS safari's bottom bar obscuring content at the bottom of the screen

Aierie:
CS-3638

Tried out on Safari, iOS Safari, Chrome, Chrome with dvh enabled.


packages/ssr-web/app/index.html
```diff
@@ -13,6 +13,7 @@
     <link integrity="" rel="stylesheet" href="{{rootURL}}assets/app-components.css" />
     <link href="{{rootURL}}images/icon-apple-256x256.png" rel="apple-touch-icon">
     <link href="{{rootURL}}images/icon-favicon-32x32.png" rel="shortcut icon" type="image/x-icon">
+    <script src="{{rootURL}}scripts/fix-vh.js"></script>
```
Aierie, line 16: Putting this here so that the variable is set before anything renders.

Aierie:
Preview deployments won't show this fix because the script in public/ is not recognised as an asset request by the cloudfront function

packages/ssr-web/public/scripts/fix-vh.js
```diff
@@ -0,0 +1,25 @@
+(() => {
+  function supportsDvh() {
+    try {
+      return CSS.supports('(height: 1dvh)');
```
Aierie, line 4: dvh is Safari's upcoming fix for the vh problem. It's a viewport height that takes into account whether the url /bottom bar is showing.

PR #2883, opened 2022-04-13
Card spaces cleanup

Aierie:
CS-3636

Remove various now-defunct properties and related code from hub. I've tried this locally by creating a payment profile and then checking that I can access its card space.


packages/hub/db/migrations/20220413090421591_card-space-unused-data-cleanup.ts
```diff
@@ -0,0 +1,24 @@
...
+    'profile_button_text',
+    'profile_category',
+    'profile_name',
+    'profile_cover_image_url',
+  ]);
+}
+
+export async function down(_pgm: MigrationBuilder): Promise<void> {
+  console.log("not defining down migration for card_spaces' unused fields");
+}
```
backspace, line 24: Is this okay to do? 😳

backspace (approved):

burieberry (approved):

packages/hub/db/migrations/20220413090421591_card-space-unused-data-cleanup.ts
Aierie: Hmm... tbh I feel like it doesn't matter too much since these fields were never used, but I can see how it could be annoying for someone doing debugging, thinking about it a bit more. It isn't too much trouble, so I'll add the down migration.

Aierie:
I'm merging this. Will fix the css selector lint issue in a separate PR.

PR #2889, opened 2022-04-14
Don't loop forever if workflow run filtering is not working properly

lukemelia:


backspace (approved):
It’s worth trying! Thanks

PR #2893, opened 2022-04-15
Fix gas estimation issue

tintinthong:
- had to include fix to estimate gas AGAIN after subtracting gas. This applies only to max withdrawals.
- had to include a buffer because the estimate of the relayer has a tendency to have a higher base gas. Typically, 56528(relayer or estimate) vs 56516(preEstimate). The buffer ensures that when we want a max withdrawal, we overestimate the baseGas so the relayer doesn't error out. The only concern here is really if your buffer is TOO HIGH, then a user might still be able to see balance within his safe even after full withdrawal -- I think 30 is reasonable for CARD. 

- [x] check relayeer doesn't have any checks for `amount+gas < safeBalance`. 
   - [x] if so need to add buffer somewhere
- [x] check that failed addresses in sentry pass gas estimation checks
- [x] do hypothesis check that the previous bug that was fixed before was failing because of this. base gas should defere
- [x] test normal withdraw
- [x] test normal safe transfer
- [x] test full withdraw
- [x] test full safe transfer

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -267,6 +267,7 @@
         "0xD669AC924eb6812C48afA984B5efFd776d035001", // source safe address
         "0xFeDc0c803390bbdA5C4C296776f4b574eC4F30D1", // DAI.CPXD
         "0x00F0FBDEEa1cDEc029Ba6025ca726Fdcf43E9025", // destination addrress
+        "--amount",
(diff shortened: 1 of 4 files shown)


habdelra (approved):
makes sense

danibonilha:
Merging to publish

PR #2894, opened 2022-04-18
email-card-drop-requests get

Aierie:
Add a GET endpoint to the hub to check if an EOA's card drop has been claimed

```
fetch("https://hub.cardstack.com/api/email-card-drop-requests?eoa=address", { 
  method: "GET",
  headers: {
   "Accept": "application/vnd.api+json"
  }
})
```

should return 

```
{
  "data": {
    "type": "email-card-drop-request-claim-status",
    "id": "address-2022-04-20T11:52:28.908Z",
    "attributes": {
      "owner-address": "address",
      "claimed": false, // true if there is a transaction hash in the db. false otherwise, or if there is no record of the request
      "timestamp": "2022-04-20T11:52:28.908Z"
    }
  }
}
```

### DIFF
--- a/packages/hub/routes/email-card-drop-requests.ts
+++ b/packages/hub/routes/email-card-drop-requests.ts
@@ -1,0 +1,53 @@
+import Koa from 'koa';
+import autoBind from 'auto-bind';
+import { query } from '../queries';
+import { inject } from '@cardstack/di';
+
+export interface EmailCardDropRequest {
+  id: string;
+  ownerAddress: string;
+  emailHash: string;
+  verificationCode: string;
+  claimedAt?: Date;
+  requestedAt: Date;
+  transactionHash?: string;
+}
+
+export default class EmailCardDropRequestsRoute {
+  emailCardDropRequestQueries = query('email-card-drop-requests', { as: 'emailCardDropRequestQueries' });
+  emailCardDropRequestSerializer = inject('email-card-drop-request-serializer', {
+    as: 'emailCardDropRequestSerializer',
+  });
+
+  constructor() {
+    autoBind(this);
+  }
+
+  async get(ctx: Koa.Context) {
+    let timestamp = new Date();
+    let ownerAddress = ctx.request.query['eoa'] as string;
+
+    let previousRequests = await this.emailCardDropRequestQueries.query({
+      ownerAddress,
+    });
+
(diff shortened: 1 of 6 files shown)


Aierie:
Thoughts on using a `eoa` query parameter in the url, but a `ownerAddress` property in the db + hub code?

packages/hub/services/serializers/email-card-drop-request-serializer.ts
```diff
@@ -0,0 +1,29 @@
...
+      data: {
+        type: 'email-card-drop-request-claim-status',
+        id: `${model.ownerAddress}-${model.timestamp.toISOString()}`,
+        attributes: {
+          'owner-address': model.ownerAddress,
+          claimed: model.claimed,
+          timestamp: model.timestamp.toISOString(),
+        },
+      },
+    } as JSONAPIDocument;
```
danibonilha, line 21: Does it need to be casted if it's already defined on the return of the function ?

packages/hub/routes/email-card-drop-requests.ts
```diff
@@ -0,0 +1,54 @@
...
+    autoBind(this);
+  }
+
+  async get(ctx: Koa.Context) {
+    let timestamp = new Date(this.clock.now());
+    let ownerAddress = ctx.request.query['eoa'] as string;
+
+    let previousRequests = await this.emailCardDropRequestQueries.query({
+      ownerAddress,
+    });
```
danibonilha, line 33: could it be a problem  if `eoa` is undefined ?

packages/hub/services/serializers/email-card-drop-request-serializer.ts
Aierie: No, I don't think it does - and local build works without too, so I'll remove this. Thanks!

backspace (approved):
This looks good to me, with the aforenoted need to have handling for absent `eoa` 🎉

packages/hub/routes/email-card-drop-requests.ts
Aierie: Response body will now be:

```
{
      errors: [
        {
          code: '400',
          title: 'Missing required parameter: eoa',
          detail: 'Please provide an ethereum address via the eoa query parameter',
        },
      ],
}
```

for missing EOA

PR #2895, opened 2022-04-18
Add /api/config endpoint to return RPC node urls

lukemelia:


packages/hub/node-tests/routes/config-test.ts
```diff
@@ -0,0 +1,29 @@
...
+        data: {
+          type: 'config',
+          attributes: {
+            web3: {
+              layer1Network: 'kovan',
+              layer1RpcNodeHttpsUrl: 'https://infuratest.test/abc123/',
+              layer1RpcNodeWssUrl: 'wss://infuratest.test/ws/abc123/',
+              layer2Network: 'sokol',
+              layer2RpcNodeHttpsUrl: 'https://humorme.test/abc123/',
+              layer2RpcNodeWssUrl: 'wss://humorme.test/abc123/',
```
habdelra, line 22: something to keep in mind: in the not to distant future we'll need to support multiple layer 2 networks (avalanche, polygon, etc). probably don't need to change this now, but something to just keep in the back of our heads

habdelra (changes requested):
Make sure to use the WSS url in the web socket service and not the HTTPS url

packages/hub/node-tests/routes/config-test.ts
lukemelia: Yeah, this made my head spin too much to solve now since we have a lot of assumptions about layer1/layer2right now.

lukemelia:
I tested web-client and cardpay-cli and hub locally. Not sure how to test ssr-web so maybe you can help with this @backspace?

habdelra:
using wss was key for not having wallet connect use up massive amounts of connections

habdelra (approved):

packages/hub/routes/config.ts
```diff
@@ -0,0 +1,34 @@
...
+import config from 'config';
+
+export const DEGRADED_THRESHOLD = 10;
+
+export default class ConfigRoute {
+  constructor() {
+    autoBind(this);
+  }
+
+  async get(ctx: Koa.Context) {
```
Aierie, line 13: This method doesn't appear to be async

packages/hub/routes/config.ts
```diff
@@ -0,0 +1,34 @@
+import Koa from 'koa';
+import autoBind from 'auto-bind';
+import { JSONAPIDocument } from '../utils/jsonapi-document';
+import config from 'config';
+
+export const DEGRADED_THRESHOLD = 10;
+
+export default class ConfigRoute {
+  constructor() {
+    autoBind(this);
```
Aierie, line 10: Would it be good to have a check at initialization that the config for essential urls exist, and if not, report to Sentry? Or maybe a Checkly check?

Aierie (approved):

Aierie:
Have tried hub auth on ssr-web. That seems to be the only thing we use the rpc node for at the moment.

Aierie:
Perhaps it's also good to add a bit to the hub README about the new environment variables required to run web-client and ssr-web?

packages/web-client/app/utils/web3-strategies/layer-1-connection-manager.ts
```diff
@@ -439,20 +439,18 @@ class WalletConnectConnectionStrategy extends ConnectionStrategy {
         qrcodeModal: WalletConnectQRCodeModal,
       };
     }
+    let hubConfigApi = new HubConfig(config.hubURL);
```
habdelra, line 442: for the near term solution aren't we going to have 2 endpoints--one that the web client uses that has a referrer URL limitation, and one that everyone else uses that does not? if so, I think we'll need to get this endpoint from a different place (maybe the environments.js files?)

backspace (approved):
I tried this locally, success. I think we should add some robustness as part of the followup, specifically what the web-client and ssr-web behaviour should be in the edge case where they’re unable to fetch config from Hub, but for now I think this is good!

packages/hub/routes/config.ts
lukemelia: Possible future enhancement

PR #2899, opened 2022-04-20
fix fastboot URLSearchParams not being defined

Aierie:
tried this locally w `ember s` and also with the node server in the deployment package, this fixes the error we saw:

```
Error occurred:

- While rendering:
  -top-level
    application
      card-space/user-page


There was an error running your app in fastboot. More info about the error: 
 ReferenceError: URLSearchParams is not defined
```


backspace (approved):
Linting would be nice, I did a cursory search and didn’t find anything, but maybe we could delete all known globals somehow so they all have to be explicitly safelisted

Aierie:
I do also wonder if fastboot will benefit from the Typescript changes coming to ember!

PR #2902, opened 2022-04-21
hub: Add POST /api/email-card-drop-requests

backspace:
This is just a start.

### DIFF
--- a/packages/hub/routes/email-card-drop-requests.ts
+++ b/packages/hub/routes/email-card-drop-requests.ts
@@ -2,6 +2,7 @@ import Koa from 'koa';
 import autoBind from 'auto-bind';
 import { query } from '../queries';
 import { inject } from '@cardstack/di';
+import shortUuid from 'short-uuid';
 
 export interface EmailCardDropRequest {
   id: string;
@@ -14,6 +15,8 @@ export interface EmailCardDropRequest {
 }
 
 export default class EmailCardDropRequestsRoute {
+  databaseManager = inject('database-manager', { as: 'databaseManager' });
+
   emailCardDropRequestQueries = query('email-card-drop-requests', { as: 'emailCardDropRequestQueries' });
   emailCardDropRequestSerializer = inject('email-card-drop-request-serializer', {
     as: 'emailCardDropRequestSerializer',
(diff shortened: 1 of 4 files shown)


packages/hub/routes/email-card-drop-requests.ts
```diff
@@ -60,6 +68,65 @@ export default class EmailCardDropRequestsRoute {
...
+
+    let hash = crypto.createHash('sha256');
+    hash.update(email);
+    let emailHash = hash.digest('hex');
+
+    let verificationCode = (crypto.randomInt(1000000) + '').padStart(6, '0');
+
+    if (isEmail(email)) {
+      const emailCardDropRequest: EmailCardDropRequest = {
+        id: shortUuid.uuid(),
```
Aierie, line 87: Why `shortUuuid.uuid()` over `crypto.randomUUID()`?

packages/hub/services/serializers/email-card-drop-request-serializer.ts
```diff
@@ -7,6 +8,18 @@ interface EmailCardDropRequestClaimStatus {
...
+  serialize(model: EmailCardDropRequest): JSONAPIDocument {
+    return {
+      data: {
+        type: 'email-card-drop-requests',
+        id: model.id,
+        attributes: {
+          'owner-address': model.ownerAddress,
+        },
+      },
+    } as JSONAPIDocument;
```
Aierie, line 20: Dani helped me realise this casting is unnecessary if we have the return type specified and the data structure fits a `JSONAPIDocument`

packages/hub/routes/email-card-drop-requests.ts
```diff
@@ -60,6 +68,65 @@ export default class EmailCardDropRequestsRoute {
...
+        ownerAddress: ctx.state.userAddress,
```
Aierie, line 88: I don't think we should use `ctx.state.userAddress`. CS-3669 describes the payload that will be sent to this endpoint:

> Hub auth token, email address, and EOA address

Aierie: actually, thinking about this more - it feels better to use the address from the auth token to avoid a situation where we end up with a card drop request where the address from the request payload doesn't match the address in the auth token.

packages/hub/routes/email-card-drop-requests.ts
backspace: copypaste? 😆 but sure I can replace that

backspace: oh also it’s [only in Node 16](https://developer.mozilla.org/en-US/docs/Web/API/Crypto/randomUUID#browser_compatibility)

packages/hub/services/serializers/email-card-drop-request-serializer.ts
backspace: hah I even read that and still copypasted 🙃 removed now, thanks.

packages/hub/routes/email-card-drop-requests.ts
Aierie: It's in node 14 too, the note is misleading

backspace: hmm strange, autocomplete didn’t show it existing. Oh well!!

PR #2904, opened 2022-04-21
hub: Add duplication rejections for drop requests

backspace:
This builds on #2902 to add rejection of `POST /api/email-card-drop-requests` when either:
* the authenticated EOA has been used to claim
* the hashed normalised email has been used to claim

Included is a `NOT_NULL` symbol to allow the existing `buildConditions` query builder to support querying for `IS NOT NULL`.

### DIFF
--- a/packages/hub/routes/email-card-drop-requests.ts
+++ b/packages/hub/routes/email-card-drop-requests.ts
@@ -122,38 +122,7 @@ export default class EmailCardDropRequestsRoute {
       return;
     }
 
-    let verificationCode = (crypto.randomInt(1000000) + '').padStart(6, '0');
-
-    if (isEmail(email)) {
-      const emailCardDropRequest: EmailCardDropRequest = {
-        id: shortUuid.uuid(),
-        ownerAddress: ctx.state.userAddress,
-        emailHash,
-        verificationCode,
-        requestedAt: new Date(this.clock.now()),
-      };
-
-      let db = await this.databaseManager.getClient();
-
-      await this.databaseManager.performTransaction(db, async () => {
-        await this.emailCardDropRequestQueries.insert(emailCardDropRequest);
-      });
-
-      await this.workerClient.addJob('send-email-card-drop-verification', {
-        id: emailCardDropRequest.id,
-        email,
-      });
-
-      await this.workerClient.addJob('subscribe-email', {
-        id: emailCardDropRequest.id,
-        email,
-      });
-
-      let serialized = this.emailCardDropRequestSerializer.serialize(emailCardDropRequest);
-
-      ctx.status = 201;
-      ctx.body = serialized;
-    } else {
+    if (!isEmail(email)) {
       ctx.status = 422;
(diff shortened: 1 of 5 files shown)


packages/hub/routes/email-card-drop-requests.ts
```diff
@@ -74,44 +75,54 @@ export default class EmailCardDropRequestsRoute {
       return;
     }
 
+    ctx.type = 'application/vnd.api+json';
+
+    let claimedWithUserAddress = await this.emailCardDropRequestQueries.query({
+      ownerAddress: ctx.state.userAddress,
+      claimedAt: NOT_NULL,
```
Aierie, line 82: Why introduce a new utility, over using js to do the filtering?

`(await this.emailCardDropRequestQueries.query({ ownerAddress: ctx.state.userAddress, })).filter(r => r.claimedAt)`

packages/hub/routes/email-card-drop-requests.ts
```diff
@@ -74,44 +75,54 @@ export default class EmailCardDropRequestsRoute {
...
+
     let email = ctx.request.body.data.attributes.email;
 
     let hash = crypto.createHash('sha256');
     hash.update(email);
     let emailHash = hash.digest('hex');
 
-    let verificationCode = (crypto.randomInt(1000000) + '').padStart(6, '0');
+    let claimedWithUserEmail = await this.emailCardDropRequestQueries.query({
+      emailHash,
```
Aierie, line 106: This isn't reliable deduplication because of the `johnsmith+hello@gmail.com` thing.

Aierie: Maybe we need to change our hashing strategy.

Aierie: Or simply reject email addresses with +, to make things easier.

backspace: WELL the `validator` package actually [includes](https://github.com/validatorjs/validator.js#sanitizers) a `normalizeEmail` to strip those out, thoughts???

packages/hub/routes/email-card-drop-requests.ts
backspace: hmm well that does seem like it would work, I just think it’s pretty common to want to ask a database for `IS NOT NULL` and it’s sad to me that this query builder doesn’t permit it. Do you think it’s TOO ugly? Is there some better approach, maybe a `Symbol`?

packages/hub/routes/email-card-drop-requests.ts
Aierie: Seems like we could try that out!

packages/hub/routes/email-card-drop-requests.ts
Aierie: i think it's ok, was mainly asking because you describe this solution as hackish so i wondered if you would rather avoid it.

PR #2905, opened 2022-04-22
Add basic support for grafting on subgraphs

IanCal:
I want to test the scripts on this more but here's an outline.

Grafting looks like a reasonable but perhaps not default way of deploying.

The package.json is getting quite unwieldy, the steps can probably be reduced down but can certainly be made more flexible as the combination is what's exploding. Still, this is a more straightforward and less intrusive change.

I've added getting the latest deploy and block number from the nodes directly, as it needs to go into the auto-generated file. While there are notes in the readme, this means a deployment process with grafting should just look like this

```
$ yarn codegen-graft-xdai-green
$ yarn deploy-xdai-green
```


habdelra:
This looks amazing! I have made a shrine in your honor

ian

habdelra (approved):

IanCal:
> This looks amazing! I have made a shrine in your honor
> 
> ian

Now I have an answer for "what motivates you at work" in interviews :)

PR #2906, opened 2022-04-22
Allows us to detect uninitialzed safes and skip over them

habdelra:
Fixes #CS-3755
I opted to use getThreshold() instead of detecting for the `SetupSafe` event since the `SetupSafe` event was introduced in 1.3 and we have 1.2 safes that we need to be able to index. Using this approach I was able to detect the bad safe at block 21735473 in gnosis chain and skip over it on my test subgraph


alex-cardstack (approved):
LGTM

habdelra:
This has been deployed to the non-live production green subgraph. a full reindex has commenced. (so now both green and blue graph nodes are compatible)

PR #2907, opened 2022-04-22
hub: Add GET /email-card-drop/verify

backspace:
This adds the endpoint that accepts validation codes sent by email. It tries to set `claimed_at` on the relevant card drop request and triggers a `DropCard` task if it succeeds.


packages/hub/queries/email-card-drop-requests.ts
```diff
@@ -51,6 +54,28 @@ export default class EmailCardDropRequestsQueries {
...
+    ownerAddress: string;
+    verificationCode: string;
+  }): Promise<EmailCardDropRequest> {
+    let db = await this.databaseManager.getClient();
+
+    let { rows } = await db.query(
+      `
+        UPDATE email_card_drop_requests
+        SET claimed_at = $1
+        WHERE owner_address = $2 AND verification_code = $3 AND claimed_at IS NULL
```
Aierie, line 71: What do we want to do about transaction hash?

Aierie, line 71: What do we want to do about transaction hash?

packages/hub/services/email-card-drop-router.ts
```diff
@@ -0,0 +1,31 @@
+import Router from '@koa/router';
+import Koa from 'koa';
+import { route } from '@cardstack/hub/routes';
+
+export default class EmailCardDropRouter {
```
Aierie, line 5: Why do we need this as a router on its own?

Aierie, line 5: Why do we need this as a router on its own?

packages/hub/routes/email-card-drop-verify.ts
```diff
@@ -0,0 +1,67 @@
...
+      ctx.status = 400;
+      ctx.body = 'eoa is required';
+      return;
+    }
+
+    let ownerAddress = ctx.request.query.eoa as string;
+    let verificationCode = (ctx.request.query['verification-code'] as string) || '';
+
+    // Optimistically mark the request as claimed to prevent stampede attack
+    let updatedRequest = await this.emailCardDropRequestQueries.claim({ ownerAddress, verificationCode });
```
Aierie, line 29: Do we set the `claimed_at` to `null` if the `drop-card` job fails?

packages/hub/node-tests/routes/email-card-drop-verify-test.ts
```diff
@@ -0,0 +1,108 @@
...
+
+    expect(jobIdentifiers).to.be.empty;
+    expect(jobPayloads).to.be.empty;
+  });
+
+  it('rejects an unknown verification', async function () {
+    let response = await request().get(`/email-card-drop/verify?eoa=${claimedEoa.ownerAddress}&verification-code=wha`);
+
+    expect(response.status).to.equal(400);
+    expect(response.text).to.equal('Code is invalid');
```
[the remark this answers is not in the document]
backspace: This is an endpoint that people will visit with a browser vs with a client that’s using the API, so plain text response is the MVP, maybe HTML with design in the future.

packages/hub/queries/email-card-drop-requests.ts
backspace: I think the `drop-card` task will set that afterward, this is sort of “optimistic” locking to prevent multiple claims.

packages/hub/services/email-card-drop-router.ts
backspace: It’s not under `/api` 😞

packages/hub/routes/email-card-drop-verify.ts
backspace: yes, I’ll make note of that in the `drop-card` issue, thanks.

packages/hub/node-tests/routes/email-card-drop-verify-test.ts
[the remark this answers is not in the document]
Aierie: Yeah looks like there isn't. I wonder why? A legitimate user who has an error should probably get a proper page that points them to support.

PR #2908, opened 2022-04-25
add execution role for secrets read access

Aierie:
Without this we default to the default ECS task execution role ([AmazonECSTaskExecutionRolePolicy](https://us-east-1.console.aws.amazon.com/iam/home#/policies/arn%3Aaws%3Aiam%3A%3Aaws%3Apolicy%2Fservice-role%2FAmazonECSTaskExecutionRolePolicy)). It has no access to AWS Secrets Manager.


jurgenwerk (approved):
Thanks! Haven't tested but let's merge and see if a new deployment will fix it.

PR #2912, opened 2022-04-26
hub: Add handling for resending verification code

backspace:
If `POST /api/email-card-drop-requests` happens with an EOA and hashed email that already exists but `claimed_at` is absent, this updates the verification code and sends another email.

### DIFF
--- a/packages/hub/routes/email-card-drop-requests.ts
+++ b/packages/hub/routes/email-card-drop-requests.ts
@@ -140,7 +140,32 @@ export default class EmailCardDropRequestsRoute {
       return;
     }
 
-    let verificationCode = (crypto.randomInt(1000000) + '').padStart(6, '0');
+    let unclaimedButExisting = await this.emailCardDropRequestQueries.query({
+      emailHash: normalizedEmailHash,
+      ownerAddress: ctx.state.userAddress,
+    });
+
+    if (unclaimedButExisting.length > 0) {
+      let existingRequest = unclaimedButExisting[0];
+      let updatedEmailCardDropRequest = await this.emailCardDropRequestQueries.updateVerificationCode(
+        existingRequest.id,
+        generateVerificationCode()
+      );
+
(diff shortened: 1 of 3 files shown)


packages/hub/routes/email-card-drop-requests.ts
```diff
@@ -140,7 +140,32 @@ export default class EmailCardDropRequestsRoute {
...
+    let unclaimedButExisting = await this.emailCardDropRequestQueries.query({
+      emailHash: normalizedEmailHash,
+      ownerAddress: ctx.state.userAddress,
+    });
+
+    if (unclaimedButExisting.length > 0) {
+      let existingRequest = unclaimedButExisting[0];
+      let updatedEmailCardDropRequest = await this.emailCardDropRequestQueries.updateVerificationCode(
+        existingRequest.id,
+        generateVerificationCode()
```
Aierie, line 152: Why do we need to update the verification code?

packages/hub/routes/email-card-drop-requests.ts
```diff
@@ -140,7 +140,32 @@ export default class EmailCardDropRequestsRoute {
...
+      );
+
+      await this.workerClient.addJob('send-email-card-drop-verification', {
+        id: updatedEmailCardDropRequest.id,
```
Aierie, line 156: Also, not related to this PR but wondering if this mutability of the verification code is why we don't simply pass the EOA + verification code into this job?

Aierie (approved):
This ran nicely for me locally.

packages/hub/routes/email-card-drop-requests.ts
backspace: hmm… I guess I’m just following the convention, I’m not sure why the convention exists though 🤔

packages/hub/routes/email-card-drop-requests.ts
backspace: I debated including it, do you think it’s preferable?

PR #2913, opened 2022-04-26
Use @cardstack/wc-provider from the CLI to specify a rpcWss URL

lukemelia:
The lack of this started making wallet connect operations hang for some reason


habdelra (approved):

habdelra:
Let’s also do an npm publish and a new CLI publish too.

PR #2915, opened 2022-04-27
Add extraction of all safe ownerships with a high frequency

IanCal:
This is to support #CS-3781 recurring drop. The small partition size will result in 100k files at the end of the year but is required to support faster processing - we can change this at any point in the future though.

### DIFF
--- a/packages/cardpay-subgraph-extraction/config/safe_owner.yaml
+++ b/packages/cardpay-subgraph-extraction/config/safe_owner.yaml
@@ -1,0 +1,32 @@
+base: &base
+  name: safe_owner
+  tables:
+    safe_owner:
+      column_mappings:
+        created_at:
+          created_at_uint64:
+            default: 0
+            max_value: 18446744073709551615
+            type: uint64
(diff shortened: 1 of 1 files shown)


tintinthong (commented):
LGTM (Did not index locally).

tintinthong (approved):
LGTM (Did not index locally).

packages/cardpay-subgraph-extraction/config/safe_owner.yaml
```diff
@@ -0,0 +1,32 @@
+base: &base
+  name: safe_owner
+  tables:
+    safe_owner:
+      column_mappings:
+        created_at:
```
tintinthong, line 6: Cool. Had to check this `created_at` field was necessary since most `ownership_changed_at` was upon creation. Guess we wud want to be able to distinguish the changes after creation.

IanCal: Yeah technically we could leave it out of the mappings but easy to add and leave it in.

PR #2916, opened 2022-04-27
hub: Add entropy to code and email to verify

backspace:
This changes the email card drop verification code from a 6-digit string to a 10-character random `url-safe` string (from `crypto-random-string`). It also changes the email hash to be salted and required as a parameter in the `GET /email-card-drop/verify` call.

### DIFF
--- a/packages/hub/config/default.cjs
+++ b/packages/hub/config/default.cjs
@@ -34,6 +34,7 @@ module.exports = {
     messageVerificationDelayMs: 1000 * 15,
   },
   authSecret: null,
+  emailHashSecret: null,
   sentry: {
     dsn: null,
     enabled: false,
--- a/packages/hub/config/default.cjs
+++ b/packages/hub/config/default.cjs
@@ -34,7 +34,7 @@ module.exports = {
     messageVerificationDelayMs: 1000 * 15,
   },
   authSecret: null,
-  emailHashSecret: null,
+  emailHashSalt: null,
   sentry: {
     dsn: null,
     enabled: false,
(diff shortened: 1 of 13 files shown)


packages/hub/config/default.cjs
```diff
@@ -34,6 +34,7 @@ module.exports = {
     messageVerificationDelayMs: 1000 * 15,
   },
   authSecret: null,
+  emailHashSecret: null,
```
Aierie, line 37: This should be `emailHashSalt`

packages/hub/node-tests/routes/email-card-drop-verify-test.ts
```diff
@@ -96,13 +98,39 @@ describe('GET /email-card-drop/verify', function () {
     expect(jobPayloads).to.be.empty;
   });
 
+  it('rejects an unknown email-hash', async function () {
+    let response = await request().get(
+      `/email-card-drop/verify?eoa=${unclaimedEoa.ownerAddress}&verification-code=${unclaimedEoa.verificationCode}&email-hash=wha`
+    );
+
+    expect(response.status).to.equal(400);
+    expect(response.text).to.equal('Email is invalid');
```
Aierie, line 107: If we are not redirecting, then our error messages might be confusing to a user. We could go with something like a generic "Card drop link is invalid".

backspace: It’s easy to change these, I’ll leave it for future possible improvement

PR #2918, opened 2022-04-27
Add CI check to verify that waypoint.hcl-specified secrets are accessible via policy

lukemelia:


scripts/verify-waypoint-secrets.ts
```diff
@@ -0,0 +1,81 @@
...
+    result[appSlug] = {
+      executionRoleName: deployNode.use[0]['aws-ecs'][0]['execution_role_name'],
+      secretArns: secretsNode ? Object.values(secretsNode) : [],
+    };
+  }
+  return result;
+}
+
+function execute(command, options) {
+  // console.log(`executing: ${command}`);
```
IanCal, line 65: Commented out code

IanCal:
I've tested this out and it works well locally - perhaps a ticket for the backlog but at some point would be good to use APIs for this rather than triggering an aws cli command & using grep/etc. Those tools are different on mac & linux in subtle ways that can be surprising.

I think the approach otherwise makes a lot of sense. It'll hit an issue if we start wildcarding in the role permissions, but we don't do that currently and that would probably be an unsafe thing to do, so I don't see that as a big risk.

LGTM :)

IanCal (approved):

PR #2922, opened 2022-04-28
web: Add Card Drop routes

backspace:
This adds routes for Hub to redirect to from card drop email verification links. [Success](https://webemail-card-drop-cs-3671.web-client-preview.cardstack.com/card-drop/success):

image

[Already claimed](https://webemail-card-drop-cs-3671.web-client-preview.cardstack.com/card-drop/already-claimed):

image

There is a `FIXME` link for deep linking into Card Wallet, to be addressed when the link is known. There’s also a CSS question about whether a prepaid card can be scaled uniformly and dynamically vs at the hardcoded 75% I have.


packages/web-client/app/components/card-drop-page/card/index.hbs
```diff
@@ -0,0 +1,34 @@
...
+
+    <h3 class='card-drop-page-card__heading'>
+      {{yield to='heading'}}
+    </h3>
+
+    <p class='card-drop-page-card__explanation'>
+      {{yield to='explanation'}}
+    </p>
+
+    {{#if this.ua.isMobile}}
```
Aierie, line 13: `isMobile` doesn't really reflect the criteria we use, which is that the platform must support installation of the CardWallet app. There are a lot more OSes besides Android and iOS that do not.

If we want a convenience method to refer to `isAndroid` and `isIOS`, a more appropriately named one is `canInstallWallet`.

packages/web-client/app/components/common/footer/index.hbs
```diff
@@ -12,10 +12,12 @@
       </a>
     </div>
 
-    <Boxel::Button @kind="secondary-dark" @as="anchor" href="https://discord.gg/cardstack">
-      {{svg-jar "discord-logo-color" class="cardstack-footer__discord-icon"}}
-      Join us on Discord
-    </Boxel::Button>
+    {{#unless @hideDiscord}}
```
Aierie, line 15: Why hide Discord?

packages/web-client/app/components/common/footer/index.hbs
```diff
@@ -26,23 +28,25 @@
...
-        <span class="boxel-sr-only">Cardstack Monorepo</span>
-        {{svg-jar "github-logo" class="cardstack-footer__social-icon"}}
-      </a>
-      <a href="https://medium.com/cardstack" target="_blank" rel="noopener noreferrer" >
-        <span class="boxel-sr-only">Cardstack Medium</span>
-        {{svg-jar "medium-logo" class="cardstack-footer__social-icon"}}
-      </a>
-    </div>
+    {{#unless @hideSocials}}
+      <div class="cardstack-footer__socials">
```
Aierie, line 32: Why hide socials?

packages/web-client/app/components/card-drop-page/card/index.css
```diff
@@ -0,0 +1,33 @@
...
+  text-align: center;
+}
+
+.card-drop-page-card__heading {
+  font-weight: 600;
+}
+
+@media only screen and (max-width: 480px) {
+  .card-drop-page-card {
+    padding: var(--boxel-sp);
```
Aierie, line 26: This can be scaled without a breakpoint by using `clamp` or the equivalent `min` + `max` combination.

packages/web-client/app/components/card-drop-page/card/index.css
```diff
@@ -0,0 +1,33 @@
...
+  }
+
+  /* TODO is there a way to scale this dynamically and preserve aspect ratio? */
+  .card-drop-page-card .prepaid-card {
+    transform: scale(0.75);
```
Aierie, line 31: I think using a breakpoint for this is correct, unless we properly handle legibility at narrower screens. If we understand the limits of legibility properly, there is the `scale-down-to-width` modifier.

Aierie (commented):
`CardDropPage` is a copy of `UniversalLinkPage`, which is ok, but web client lacks visual regression for smaller viewports and we should have it for this case. 

We should also check with mobile team about their thoughts on the URL used and how they envision the flow going, though this probably shouldn't block merging. 

I would also defer the footer changes until we understand whether design omitted these intentionally and why.

packages/web-client/tests/acceptance/card-drop-test.ts
```diff
@@ -0,0 +1,14 @@
+import { module, test } from 'qunit';
+import { visit } from '@ember/test-helpers';
+import { setupApplicationTest } from 'ember-qunit';
+import percySnapshot from '@percy/ember';
+
+module('Acceptance | card-drop', function (hooks) {
+  setupApplicationTest(hooks);
+
+  test('it shows a success message', async function (assert) {
+    assert.expect(0);
```
Aierie, line 10: Assertion(s) that might be relevant include the button text and link.

backspace: I don’t tend to write assertions for static strings, it feels pointless to me… unless I’m misunderstanding?

PR #2926, opened 2022-05-02
Claim all revenue (make amount optional)

tintinthong:
This pr makes amount optional. So whole balances are claimed without specifying amount.


tintinthong:
Need to update docs readme

habdelra (approved):

PR #2929, opened 2022-05-03
hub: Add rate-limiting for card drops

backspace:
This inserts rate-limiting at the beginning of handling for `POST /api/email-card-drop-requests`.

It immediately 503s if the rate limit has been triggered, as stored in a `email_card_drop_state` table. If it has not, it checks whether the number of records with `claimed_at` in the rate limit period (`HUB_EMAIL_CARD_DROP_RATE_LIMIT_PERIOD_MINUTES`) exceeds the limit (`HUB_EMAIL_CARD_DROP_RATE_LIMIT_COUNT`) and triggers the rate limit and sends a `critical` error to Sentry if so.

Otherwise the request proceeds as previously implemented.

### DIFF
--- a/packages/hub/config/custom-environment-variables.json
+++ b/packages/hub/config/custom-environment-variables.json
@@ -53,7 +53,14 @@
     "provisionerSecret": "PROVISIONER_SECRET"
   },
   "cardDrop": {
-    "sku": "CARD_DROP_SKU"
+    "sku": "CARD_DROP_SKU",
+    "email": {
+      "rateLimit": {
+        "count": "HUB_EMAIL_CARD_DROP_RATE_LIMIT_COUNT",
+        "periodMinutes": "HUB_EMAIL_CARD_DROP_RATE_LIMIT_PERIOD_MINUTES",
+        "fixme": "FIXME_ABOVE_NEEDED_IN_WAYPOINT_ETC"
+      }
+    }
   },
   "web3storage": {
     "token": "WEB3_STORAGE_TOKEN"
--- a/packages/hub/config/custom-environment-variables.json
+++ b/packages/hub/config/custom-environment-variables.json
@@ -56,9 +56,8 @@
     "sku": "CARD_DROP_SKU",
     "email": {
       "rateLimit": {
-        "count": "HUB_EMAIL_CARD_DROP_RATE_LIMIT_COUNT",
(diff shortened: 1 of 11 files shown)


packages/hub/config/custom-environment-variables.json
```diff
@@ -53,7 +53,13 @@
     "provisionerSecret": "PROVISIONER_SECRET"
   },
   "cardDrop": {
-    "sku": "CARD_DROP_SKU"
+    "sku": "CARD_DROP_SKU",
+    "email": {
+      "rateLimit": {
+        "count": {"__name": "HUB_EMAIL_CARD_DROP_RATE_LIMIT_COUNT", "__format": "number"},
```
lukemelia, line 59: TIL this little syntax for typing config!

lukemelia (changes requested):
This looks good. I think we should add some inline documentation about how to reset the rate-limited flag, and add some external documentation about what to do if the rate limiting is triggered. Should we set up a sentry rule to trigger a PagerDuty incident if this is fired? https://docs.sentry.io/product/integrations/notification-incidents/pagerduty/

backspace:
There‘s no official Sentry Terraform provider, sadly, so I set up a Sentry-PagerDuty integration and alert, I was able to produce a [PagerDuty incident](https://cardstack.pagerduty.com/incidents/Q3BZBUL7AKITYY) by manually executing the same `Sentry.captureException` code from this PR. Unfortunately, the incident name showed up as `null.<anonymous>`; currently looking into how to specify more.

lukemelia (approved):

packages/hub/config/custom-environment-variables.json
backspace: ya… Javascript’s incredible laxness around typing makes this actually unnecessary but it just feels better to me to know these will come out as numbers without having to scatter `parseInt` everywhere they’re accessed 🤓

Aierie:
@backspace this doesn't currently interact with StatusPage (seems like this is by design?). Is the person on call meant to create a StatusPage incident?

backspace:
> @backspace this doesn't currently interact with StatusPage (seems like this is by design?). Is the person on call meant to create a StatusPage incident?

unsure… also this doesn’t tell Card Wallet anything, I think probably it shouldn’t suggest that people try to get a prepaid card when there are none available.

PR #2931, opened 2022-05-04
hub: Add listener for prepaid card provisioning

backspace:
This adds a push notification triggered by `PrepaidCardMarketV2`’s `PrepaidCardProvisioned` contract event:
> You were issued a new prepaid card!


packages/hub/tasks/notify-prepaid-card-drop.ts
```diff
@@ -0,0 +1,55 @@
...
+    };
+
+    let pushClientIdsForNotification = await this.notificationPreferenceService.getEligiblePushClientIds(
+      ownerAddress,
+      'customer_payment'
+    );
+
+    for (const pushClientId of pushClientIdsForNotification) {
+      let notification: PushNotificationData = {
+        sendBy: this.clock.now() + PREPAID_CARD_DROP_EXPIRY_TIME,
```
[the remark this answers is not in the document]
Aierie: Actually not very clear if this notification is a one time thing. I do think the notification for the first card a user gets shouldn't expire though.

packages/hub/tasks/notify-prepaid-card-drop.ts
```diff
@@ -0,0 +1,55 @@
...
+        pushClientId,
+        notificationBody,
+        notificationData,
+        notificationType: 'prepaid_card_drop',
+      };
+
+      await this.workerClient.addJob('send-notifications', notification, {
+        jobKey: notification.notificationId,
+        jobKeyMode: 'preserve_run_at',
+        maxAttempts: 8, // 8th attempt is estimated to run at 28 mins. https://github.com/graphile/worker#exponential-backoff
```
Aierie, line 51: This also affects expiry.

packages/hub/tasks/notify-prepaid-card-drop.ts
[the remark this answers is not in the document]
Aierie: Checked with @jurgenwerk about this, we probably will only see this event for card drop.

packages/hub/services/contract-subscription-event-handler.ts
```diff
@@ -21,6 +21,12 @@ export const CONTRACT_EVENTS = [
     eventName: 'MerchantClaim',
     taskName: 'notify-merchant-claim',
   },
+  {
+    abiName: 'prepaid-card-market-v2',
```
Aierie, line 25: abi is `prepaid-card-market-v-2` I think, but `v2` is used elsewhere.

packages/hub/tasks/notify-prepaid-card-drop.ts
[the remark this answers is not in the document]
backspace: hmm yes, for a one-time thing I agree that expiry shouldn’t exist, I’ll remove it, thanks.

packages/hub/tasks/notify-prepaid-card-drop.ts
```diff
@@ -0,0 +1,55 @@
...
+
+    let notificationData = {
+      notificationType: 'prepaid_card_drop',
+      ownerAddress,
+      network: web3Config.layer2Network,
+    };
+
+    let pushClientIdsForNotification = await this.notificationPreferenceService.getEligiblePushClientIds(
+      ownerAddress,
+      'customer_payment'
```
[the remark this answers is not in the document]
backspace: mocks 😳 thanks

packages/hub/services/contract-subscription-event-handler.ts
backspace: ugh fixed thanks

.github/workflows/pr-hub.yml
```diff
@@ -2,7 +2,6 @@ name: CI [hub]
 
 on:
   pull_request:
-    branches: [main]
```
backspace, line 5: This is temporary, will be inherited from #2984.

backspace:
I requested another review because I think there have been enough changes:
* moving the `notification_types` insertion from seeds into a migration
* adding `contractStartVersion` option to skip tests when an older contract doesn’t include an ABI
* adding generic tests for all event handlers

The latter I’m somewhat ambivalent about… is there really any point in exercising each of the entries in `CONTRACT_EVENTS` considering they’re all handled in the exact same way? 🤔

Aierie:
> is there really any point in exercising each of the entries in CONTRACT_EVENTS considering they’re all handled in the exact same way? 🤔

I'm not sure. I think there is, because special cases happen.

packages/hub/node-tests/services/contracts-test.ts
```diff
@@ -3,42 +3,48 @@ import { CONTRACT_EVENTS } from '../../services/contract-subscription-event-hand
...
-            
-  ${e}`
-          );
-        }
-        expect(contract).to.be.an.instanceOf(web3.eth.Contract);
-        expect(contract).to.exist;
-      });
+      if (
+        contractUsageInfo.contractStartVersion &&
+        semverGte(protocolVersion, contractUsageInfo.contractStartVersion)
```
Aierie, line 14: To include contracts that we didn't define a start version for, this needs to be:
```js
!contractUsageInfo.contractStartVersion ||
semverGte(protocolVersion, contractUsageInfo.contractStartVersion)
```

packages/hub/node-tests/services/contracts-test.ts
```diff
@@ -3,42 +3,48 @@ import { CONTRACT_EVENTS } from '../../services/contract-subscription-event-hand
...
+
+This may be due to a missing ABI definition. Check to see whether SDK codegen is working properly.
+
+${e}`
+            );
+          }
+          expect(contract).to.be.an.instanceOf(web3.eth.Contract);
+          expect(contract).to.exist;
+        });
+      }
```
Aierie, line 47: I'm not sure about this, but it's possible to append a skip clause so that skipped items have visibility:

```js
 else {
    it.skip(`Skipping ${contractUsageInfo.abiName} contract version ${protocolVersion} because earlier than start version ${contractUsageInfo.contractStartVersion}`, function () {});
}
```

The skipped items will be listed as pending tests.

packages/hub/node-tests/services/contracts-test.ts
backspace: ugh indeed thanks

packages/hub/node-tests/services/contracts-test.ts
backspace: I’m going to skip (😆) this for now but we can revisit if it seems like it would have helped avoid a problem 🤓

Aierie: It would have helped avoid the comment above! I think

packages/hub/tasks/notify-prepaid-card-drop.ts
```diff
@@ -0,0 +1,50 @@
...
+
+const web3Config = config.get('web3') as { layer2Network: 'sokol' | 'xdai' };
+
+export default class NotifyPrepaidCardDrop {
+  notificationPreferenceService: NotificationPreferenceService = inject('notification-preference-service', {
+    as: 'notificationPreferenceService',
+  });
+  workerClient: WorkerClient = inject('worker-client', { as: 'workerClient' });
+
+  async perform({ transactionHash, ownerAddress }: { transactionHash: string; ownerAddress: string }) {
```
backspace, line 16: This function signature is tragically broken! I don’t know why I thought I could do this… the event listener only passes `transactionHash` as a bare string.

But [here](https://github.com/cardstack/cardstack/commit/a9aa73e61a2efb9dcff23cc22c7091d9f6bd10bf) I am experimenting with passing the entire event. I think that will help because otherwise we don’t have access to the owner address needed to send the notification, and it does come with the event.

Here’s an event preserved for future reference:

```
{
  "address": "0xb52726CbDA92b5340F32A0e837e8678E68Ff9da8",
  "blockHash": "0x53f3b77bac7cdcb7f952ab9f705f405700d1edc769044343f6e8abe3c735db5f",
  "blockNumber": 26785687,
  "logIndex": 0,
  "removed": false,
  "transactionHash": "0xbaf65d22b29c635243a2e5fa748eb195577c0a7bd342081dc5907c488323e25c",
  "transactionIndex": 23,
  "transactionLogIndex": "0x0",
  "id": "log_e3b8e103",
  "returnValues": {
    "0": "0x323B2318F35c6b31113342830204335Dac715AA8",
    "1": "0x905b1faa1a56a0fd8447a469f8aa558665d550e56d60d106cd1096c526df4bdd",
    "owner": "0x323B2318F35c6b31113342830204335Dac715AA8",
    "sku": "0x905b1faa1a56a0fd8447a469f8aa558665d550e56d60d106cd1096c526df4bdd"
  },
  "event": "PrepaidCardProvisioned",
  "signature": "0x7b8bf0fde69780a9f1eef765d847919a85bff2d761a03e306d0ce4f884c0f7c3",
  "raw": {
    "data": "0x000000000000000000000000323b2318f35c6b31113342830204335dac715aa8905b1faa1a56a0fd8447a469f8aa558665d550e56d60d106cd1096c526df4bdd",
    "topics": [
      "0x7b8bf0fde69780a9f1eef765d847919a85bff2d761a03e306d0ce4f884c0f7c3"
    ]
  }
}
```

Aierie: Oops, I missed that too. It is a bit sad that Typescript doesn't catch these at the moment.

PR #2932, opened 2022-05-04
Automatic posting of PagerDuty incident status changes to Discord

lukemelia:
Note: PagerDuty has a signature header on webhooks, which can be validated using a secret that is shown to one time upon creation of the webhook subscription. However, we are creating the webhook via terraform, which doesn't allow for the capture of that secret. So, in order to prevent negative impact from spoofed requests to this callback URL, we fetch the incident via PagerDuty's API after receiving a request.


backspace (approved):
I didn’t exercise this locally but I did see it successfully post in the channel 🎉

PR #2940, opened 2022-05-06
hub: Change Card Space updates to PATCH

backspace:
Since updates will now be per-field, it doesn’t make sense to use `PUT`.


packages/hub/routes/card-spaces.ts
```diff
@@ -148,9 +148,19 @@ export default class CardSpacesRoute {
...
+    if (attributes['profile-description']) {
+      cardSpace.profileDescription = this.sanitizeText(ctx.request.body.data.attributes['profile-description']);
+    }
+
+    if (attributes['profile-image-url']) {
+      cardSpace.profileImageUrl = this.sanitizeText(ctx.request.body.data.attributes['profile-image-url']);
+    }
+
+    if (attributes['links']) {
+      cardSpace.links = ctx.request.body.data.attributes['links'];
```
Aierie, line 162: Links is effectively a PUT for now, but I don't know if that is a problem we need to handle right now.

Aierie (approved):
Works for me locally!

packages/hub/routes/card-spaces.ts
backspace: ya… I think for now sadly we just need the client to send everything in `links` 🥴

PR #2945, opened 2022-05-09
The environment does not need to be set on koa context

lukemelia:
The setting was based on NODE_ENV, which was unset, which was resulting in the exchange-rates perceiving a development environment and not enforcing its origin restriction

### DIFF
--- a/packages/hub/main.ts
+++ b/packages/hub/main.ts
@@ -7,7 +7,7 @@ import './load-dotenv';
 import config from 'config';
 
 import Koa from 'koa';
-import { environment, httpLogging, errorMiddleware } from './middleware';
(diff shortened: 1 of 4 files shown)


danibonilha (approved):
We just need to remove the `  .use(environment)` reference on hub/ main.ts and it should be good to go

packages/hub/main.ts
```diff
@@ -266,7 +265,7 @@ export class HubServer {
 
   async listen(port = 3000) {
     let instance = this.app.listen(port);
-    serverLog.info(`\n👂 Hub listening on %s\n`, port);
+    serverLog.info(`\n👂 Hubbie listening on %s\n`, port);
```
danibonilha, line 268: leftover

PR #2946, opened 2022-05-09
hub: Update drop-card task to call relay server

backspace:
This relies on an unmerged relay server PR but it’s probably fine to merge because this task won’t be produced yet.

One uncertainty: is `config.cardDrop.sku` the correct value or will it be `sku2` or somesuch?


packages/hub/node-tests/tasks/drop-card-test.ts
```diff
@@ -0,0 +1,128 @@
...
+
+    expect(testkit.reports()[0].error?.message).to.equal('provisioning should error');
+    expect(testkit.reports()[0].tags).to.deep.equal({
+      action: 'drop-card',
+    });
+
+    let requests = await emailCardDropRequestQueries.query({ id: unclaimedEoa.id });
+    let request = requests[0];
+
+    expect(request.transactionHash).to.be.null;
```
Aierie, line 126: After adding the change for `claimedAt`, I think we should have a test case that asserts that `claimedAt` is set to `null` when it was set to a `Date` at the beginning of the test. Maybe all the tests cases here should have `claimedAt` set already?

Aierie (commented):
I think this looks ok. Though I am a bit unsure about this code being in a job, because we might be prematurely redirecting people to a success screen in the route handler that calls this, and don't have a way to follow up with them to tell them it failed.

backspace:
I agree that `drop-card` shouldn’t be a task, it should be synchronous when handling the verification link. I’ll make that change now.

re auto-clearing `claimed_at`, I thought about it and decided it should be a manual process at the beginning so we can handle instances of it individually, I’m concerned about automating it and losing oversight.

backspace:
I’ve removed `drop-card` and moved its functionality into the `…/verify` route handler, ready for rereview, thanks for the prompting 😀

Aierie (approved):

PR #2947, opened 2022-05-10
Prepare sdk for the new contract version v0.9.0

jurgenwerk:
Ticket: CS-3845

This PR bumps up the version to 0.9.0 so that the SDK will know how to talk to the upgraded contracts once we perform the protocol deploy. 

It also includes addressing a breaking change in `getCreateCardPayload` (adds two empty addresses to maintain backwards compatibility).

### DIFF
--- a/packages/cardpay-sdk/contracts/abi/v0.9.0/chainlink-feed-adapter.ts
+++ b/packages/cardpay-sdk/contracts/abi/v0.9.0/chainlink-feed-adapter.ts
@@ -44,7 +44,6 @@ export default [
     type: 'event',
   },
   {
-    constant: true,
     inputs: [],
     name: 'canSnapToUSD',
     outputs: [
@@ -54,12 +53,10 @@ export default [
         type: 'bool',
       },
     ],
-    payable: false,
     stateMutability: 'view',
     type: 'function',
   },
   {
-    constant: true,
     inputs: [],
     name: 'cardpayVersion',
     outputs: [
@@ -69,12 +66,10 @@ export default [
         type: 'string',
       },
     ],
-    payable: false,
     stateMutability: 'view',
     type: 'function',
   },
   {
-    constant: true,
     inputs: [],
     name: 'daiPrice',
     outputs: [
@@ -89,12 +84,10 @@ export default [
         type: 'uint256',
       },
     ],
-    payable: false,
     stateMutability: 'view',
     type: 'function',
   },
   {
-    constant: true,
     inputs: [],
     name: 'daiUsdFeed',
     outputs: [
@@ -104,12 +97,10 @@ export default [
         type: 'address',
       },
     ],
-    payable: false,
     stateMutability: 'view',
     type: 'function',
   },
   {
-    constant: true,
     inputs: [],
     name: 'decimals',
     outputs: [
@@ -119,12 +110,10 @@ export default [
         type: 'uint8',
       },
     ],
-    payable: false,
     stateMutability: 'view',
     type: 'function',
   },
   {
-    constant: true,
     inputs: [],
     name: 'description',
     outputs: [
@@ -134,12 +123,10 @@ export default [
         type: 'string',
       },
     ],
-    payable: false,
     stateMutability: 'view',
     type: 'function',
   },
   {
-    constant: true,
     inputs: [],
     name: 'ethPrice',
     outputs: [
@@ -154,12 +141,10 @@ export default [
         type: 'uint256',
       },
     ],
-    payable: false,
     stateMutability: 'view',
     type: 'function',
   },
   {
-    constant: true,
     inputs: [],
     name: 'ethUsdFeed',
     outputs: [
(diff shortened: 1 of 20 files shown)


jurgenwerk:
1. I only updated PrepaidCardManager and PrepaidCardMarket ABIs for 0.10.0 since I think that these two are the only ones that changed. Should I update others as well? 
2. I copied the ABIs from the newly built card-pay-protocol artifacts (from the v2 contract branch), is that the correct thing to do? Not sure why, but the new ABIs removed `payable` and `constant` [from a couple of places](https://github.com/cardstack/cardstack/pull/2947/commits/62b1b32c858d960794a3fc09c1c89f0f3417db42#diff-1c0620829a8c6383e493899720f9c406d20f4c6170d2611ef249f7f29f9b9d3bL271)
3. I still need to add SDK & CLI commands for setting the SKU and ask price – should we ship this PR first and include it in the releases, then upgrade the contracts, or first wait until these commands are added?

packages/cardpay-sdk/contracts/addresses.ts
```diff
@@ -5,18 +5,18 @@ import { networkName } from '../sdk/utils/general-utils';
...
+import SplitPrepaidCardHandlerABI from './abi/v0.10.0/split-prepaid-card-handler';
+import SpendABI from './abi/v0.10.0/spend';
+import MerchantManagerABI from './abi/v0.10.0/merchant-manager';
+import DeprecatedMerchantManagerABI_0_6_7 from './abi/v0.10.0/deprecated-merchant-manager-0_6_7';
+import RegisterRewardProgramHandlerABI from './abi/v0.10.0/register-reward-program-handler';
+import RegisterRewardeeHandlerABI from './abi/v0.10.0/register-rewardee-handler';
+import SupplierManagerABI from './abi/v0.10.0/supplier-manager';
 
-export const protocolVersions = ['v0.9.0', 'v0.8.7'];
+export const protocolVersions = ['v0.10.0', 'v0.9.0', 'v0.8.7'];
```
habdelra, line 19: it should be ok to drop v0.8.7. all the consumers of the SDK are up-to-date now and using teh current version of the protocol which is 0.9.0. there is no need to be backwards compatible to 0.8.7 (so this means dropping all the 0.8.7 ABI's as well, and anything that refers to 0.8.7)

packages/cardpay-sdk/sdk/prepaid-card/v0.10.0.ts
```diff
@@ -0,0 +1,49 @@
+import Base from './base';
+import BN from 'bn.js';
+import { getAddress } from '../../contracts/addresses';
+import { ZERO_ADDRESS } from '../constants';
+import ERC677ABI from '../../contracts/abi/erc-677';
+import { AbiItem } from 'web3-utils';
+import Web3 from 'web3';
+import { Signer } from 'ethers';
+
+export default class PrepaidCard extends Base {
```
habdelra, line 10: generally i think it's better to do this in the other way--make the most recent version the default (so update base with this logic), and then make 0.9.7 the exceptional case. this way it make it easier to continue to evolve the SDK then the old behavior is the exceptional thing.

jurgenwerk: Now that I'm thinking about this, does it even make sense to make an exception for 0.9.0 if we move the new behaviour to the base class?

If we change `getCreateCardPayload` in the *base* class, and extend both 0.9.0 and 0.10.0 from the base without modifications, and deploy the SDK (while not deploying the protocol yet):
- The SDK will send 2 additional params in the data field (2 zero addresses), but the 0.9.0 `PrepaidCardManager` [will just ignore it](https://github.com/cardstack/card-pay-protocol/blob/main/contracts/PrepaidCardManager.sol#L232) because it only grabs the first 5 params in the `data` field - even though `data` has 2 more fields available (2 zero addresses), and the contract will continue executing as it should.

And when we deploy the protocol:
- The SDK behaviour will be the same but this time the 0.10.0 `PrepaidCardManager` will be able to recognize 2 additional params and choose the right execution path (there will be 2 paths in 0.10.0)

The claim where I say that `abi.decode` will just ignore the two additional params is just a hunch which I still need to verify. It's maybe also a bit odd to send params and let the handler not care about it. On the other hand this situation will be temporary and allows us to not make SDK exceptions for 0.9.0. What do you think?

/cc @alex-cardstack

habdelra: But when we try to encode with missing params it will encode to a very different value than not even having the params.

habdelra: Specifically we need to tell the ABI encoder to use a different set of types.

habdelra: You can use this online ABI encoder to see the difference https://adibas03.github.io/online-ethereum-abi-encoder-decoder/#/

jurgenwerk: Ok thanks. I thought if we just append types then it wouldn't make a difference for the rest of the encoded params.

jurgenwerk:
We discovered 0.9.0 is not actually in use. The new version will be 0.9.0 and not 0.10.0. Will make changes to address this.

jurgenwerk:
@alex-cardstack @habdelra I rewrote this PR based on our findings (currently deployed is 0.8.8, next release will be 0.9.0).

alex-cardstack (approved):

habdelra (approved):

PR #2948, opened 2022-05-10
Mailchimp

Aierie:
Posting to `/api/email-card-drop-requests` will now subscribe the provided email to our newsletter in Mailchimp.

To try this out locally, get the dev api key and assign it to `MAILCHIMP_API_KEY`:

``` 
# use staging AWS credentials
aws secretsmanager get-secret-value --secret-id dev_MAILCHIMP_API_KEY  
```

TODO:
- [x] Add `MAILCHIMP_API_KEY` variable to waypoint config for prod and staging hub + worker. There are already secrets in staging and prod AWS Secrets Manager for this

### DIFF
--- a/packages/hub/config/staging.json
+++ b/packages/hub/config/staging.json
@@ -7,6 +7,10 @@
   "cardDrop": {
     "verificationUrl": "https://hub-staging.stack.cards/email-card-drop/verify"
   },
+  "mailchimp": {
+    "serverPrefix": "us-8",
+    "newsletterListId": "c2b7a36f2d"
+  },
   "wyre": {
     "callbackUrl": "https://hub-staging.stack.cards/callbacks/wyre"
   },
--- a/packages/hub/config/staging.json
+++ b/packages/hub/config/staging.json
@@ -8,7 +8,7 @@
     "verificationUrl": "https://hub-staging.stack.cards/email-card-drop/verify"
   },
   "mailchimp": {
-    "serverPrefix": "us-8",
+    "serverPrefix": "us8",
     "newsletterListId": "c2b7a36f2d"
   },
   "wyre": {
--- a/packages/hub/node-tests/tasks/subscribe-email-test.ts
+++ b/packages/hub/node-tests/tasks/subscribe-email-test.ts
@@ -1,0 +1,93 @@
+import { Job } from 'graphile-worker';
+import { registry, setupHub } from '../helpers/server';
+import { expect } from 'chai';
+import { makeJobHelpers } from 'graphile-worker/dist/helpers';
+import waitFor from '../utils/wait-for';
+import * as Sentry from '@sentry/node';
+import sentryTestkit from 'sentry-testkit';
+import SubscribeEmail from '../../tasks/subscribe-email';
+
+const { testkit, sentryTransport } = sentryTestkit();
+
+// https://github.com/graphile/worker/blob/e3176eab42ada8f4f3718192bada776c22946583/__tests__/helpers.ts#L135
+function makeMockJob(taskIdentifier: string): Job {
+  const createdAt = new Date(Date.now() - 12345678);
+  return {
+    id: String(Math.floor(Math.random() * 4294967296)),
+    queue_name: null,
+    task_identifier: taskIdentifier,
+    payload: {},
+    priority: 0,
+    run_at: new Date(Date.now() - Math.random() * 2000),
+    attempts: 0,
+    max_attempts: 25,
+    last_error: null,
+    created_at: createdAt,
+    updated_at: createdAt,
+    locked_at: null,
+    locked_by: null,
+    revision: 0,
+    key: null,
+    flags: null,
+  };
+}
+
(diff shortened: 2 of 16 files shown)


packages/hub/node-tests/tasks/subscribe-email-test.ts
```diff
@@ -0,0 +1,93 @@
...
+  it('will throw the error and report to Sentry if an error happens while trying to subscribe', async function () {
+    StubMailchimp.shouldThrow = true;
+
+    await expect(subject.perform({ email: 'anyone@test.test' }, helpers)).to.be.rejectedWith(
+      'intentional error during test'
+    );
+
+    await waitFor(() => testkit.reports().length > 0);
+
+    let sentryReport = testkit.reports()[0];
```
Aierie, line 87: there are now even more of these and they should be extracted.

packages/hub/services/mailchimp.ts
```diff
@@ -0,0 +1,57 @@
...
+      }
+    }
+  }
+
+  /**
+   * If email is already subscribed, do nothing.
+   * If email has previously been subscribed, resend confirmation email.
+   * If email has never been subscribed, send confirmation email.
+   */
+  async subscribe(email: string) {
```
Aierie, line 40: Will add a comment on how to remove yourself and/or check the state of the list/audience.

packages/hub/config/staging.json
```diff
@@ -7,6 +7,10 @@
   "cardDrop": {
     "verificationUrl": "https://hub-staging.stack.cards/email-card-drop/verify"
   },
+  "mailchimp": {
+    "serverPrefix": "us8",
+    "newsletterListId": "c2b7a36f2d"
```
Aierie, line 12: This is the developer "newsletter" id. Shared between staging and dev.

packages/hub/routes/email-card-drop-requests.ts
```diff
@@ -222,10 +222,21 @@ export default class EmailCardDropRequestsRoute {
...
+      // for the most part, Mailchimp errors will not be retriable
+      // because of that we limit maxAttempts to 1
+      // this prevents us from retrying where we cannot do anything
+      // and also leaves dead jobs in the db that we can check later
+      // this should be supplemented by the Mailchimp errors being sent
+      // to Sentry and having appropriate alerts, especially for rate limiting
+      // could consider permanently failing the job from within conditionally, instead. probably as a utility function?
+      // https://github.com/graphile/worker/blob/e3176eab42ada8f4f3718192bada776c22946583/sql/000004.sql#L122-L126
+      {
+        maxAttempts: 1,
```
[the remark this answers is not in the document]
Aierie: Basically we want to permanently fail whenever we get one of the errors here:

https://mailchimp.com/developer/marketing/docs/errors/#error-glossary

Since it seems unlikely that we will be able to send properly then. These error codes should also trigger Sentry alerts.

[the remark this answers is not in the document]
lukemelia: If you want retry sometimes but not always, marking as permanently failed inside the job seems good. This appears to be the relevant JS function: https://github.com/graphile/worker/blob/e3176eab42ada8f4f3718192bada776c22946583/src/workerUtils.ts#L40

packages/hub/services/mailchimp.ts
```diff
@@ -0,0 +1,57 @@
+// the types from Definitely Typed are not good enough.
+// Mailchimp docs seem more reliable though the examples have the wrong node package name
+// https://mailchimp.com/developer/marketing/api
+// @ts-ignore
+import client from '@mailchimp/mailchimp_marketing';
+import config from 'config';
+
+export default class Mailchimp {
+  client = client;
+  newsletterListId = config.get('mailchimp.newsletterListId') as string;
```
lukemelia, line 10: This will fail if mailchimp.newsletterListId is not defined. That is good for staging/production, but do we really care on development?

lukemelia (commented):
Looking good 👍

packages/hub/routes/email-card-drop-requests.ts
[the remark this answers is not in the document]
Aierie: I tried using this JS function but it doesn't appear to update the job's state properly. I'm not sure if there is something preventing the permanent failure of jobs using the JS SDK.

packages/hub/services/mailchimp.ts
Aierie: I don't think having the API key missing and as such Mailchimp failing matters as much during development (there is a default `newsletterListId`), but feel hesitant to add a separate flow that can fail silently. Is there a good way to do this?

lukemelia: We should fail on/near boot if we don't have an API Key defined and we're in staging or prod. Probably can do this in the constructor of the service.

Aierie: This is done in the latest commit. Any missing essential config will cause an Error during initialization of the service in non-development environments.

PR #2949, opened 2022-05-10
[Refactor]: Remove spd from native currencies

danibonilha:
As we moved away from using spend as currency on the wallet, we can remove this currency info, because it's a bit misleading to have spend as a nativeCurrency while the api doesn't really return it, if we were to still use it, we could make the api return SPD too, but since we don't need it anymore, removing seems the best bet to avoid the need of filtering out SPD on the mobile app.


lukemelia (approved):
Seems good to me, once green.

backspace (approved):

PR #2951, opened 2022-05-11
web: Add route to display card drop error

backspace:
This is the new route referenced in #2946.

### DIFF
--- a/packages/web-client/app/templates/card-drop/error.hbs
+++ b/packages/web-client/app/templates/card-drop/error.hbs
@@ -11,7 +11,9 @@
     <p data-test-error-message>
       {{this.message}}
     </p>
+  </:explanation>
 
+  <:action>
     <Boxel::Button
       @kind='secondary'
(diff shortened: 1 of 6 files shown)


packages/web-client/app/templates/card-drop/error.hbs
```diff
@@ -0,0 +1,23 @@
...
+      {{this.message}}
+    </p>
+
+    <Boxel::Button
+      @kind='secondary'
+      @as='anchor'
+      @href={{config 'urls.mailToSupportUrl'}}
+    >
+      Contact Cardstack Support
+    </Boxel::Button>
```
Aierie, line 21: This appears together with the 'Visit Cardstack.com' button. Should we only show one CTA?

backspace: ya makes sense, I’ve added `action` named block to be able to override that default, thanks.

PR #2953, opened 2022-05-13
explicitly specify ember-test-selectors as dep

Aierie:
This seems to be required to strip out the data-test- selectors


jurgenwerk (approved):
Thanks!

PR #2955, opened 2022-05-13
Add archived reward programs

tintinthong:
Indexer tries to index reward programs that are 
- created on chain
- have root/parquet submissions in s3

We noticed that a lot of reward programs on staging didn't have parquet submissions

This pr excludes reward programs that do not have root/parquet submissions in s3

This is not to say that a reward program that has been created on chain but does **NOT** have root/parquet submissions in s3 will break the indexer.


packages/cardpay-reward-api/cardpay_reward_api/main.py
```diff
@@ -21,25 +22,29 @@
...
 SUBGRAPH_URL = os.environ.get("SUBGRAPH_URL")
 REWARDS_BUCKET = os.environ.get("REWARDS_BUCKET")
+ENVIRONMENT = os.environ.get("ENVIRONMENT")
 
 app = FastAPI()
 
 
 @app.on_event("startup")
-@repeat_every(seconds=5)
+@repeat_every(seconds=5, raise_exceptions=True)
```
tintinthong, line 38: @IanCal looks like without this exceptions on indexing were not logged.

IanCal (approved):
Code looks good, not tested locally.

PR #2957, opened 2022-05-16
Fix/ember power select selector

Aierie:
ember-power-select 4.1.7 moves splattributes to the trigger, which breaks our styling. This is unfortunately not marked as a breaking change in their changelog. I went for a simple fix, which is to move our class to a containing div. This looks ok for the components in web client (ed34e7a94298535425d8d46d2c33169a4501e327).

Some dropdowns in Boxel demos are affected by this. @burieberry are these still used, and/or maintained? (Boxel::CardPicker, Boxel::FieldRenderer, CsCardSelect, Queue)


backspace (approved):
Thanks for figuring this out. I see the proper borders again.

PR #2959, opened 2022-05-16
Root submitter - survive missing parquet files

IanCal:
The root submitter shouldn't fall over if the required parquet file is not there yet, just log and move on.


packages/reward-root-submitter/reward_root_submitter/utils.py
```diff
@@ -3,19 +3,48 @@
...
+    # we just need the utility functions attached
+    web3 = Web3()
+    for result_file in _get_potential_reward_output_locations(root):
+        reward_program_id = safe_regex_group_search(
+            r"rewardProgramID=([^/]*)", str(result_file), 1
+        )
+        payment_cycle = safe_regex_group_search(
+            r"paymentCycle=(\d*)", str(result_file), 1
+        )
+        if (
```
tintinthong, line 41: swt

PR #2967, opened 2022-05-18
Expiry of card drop requests

Aierie:
CS-3915

### Requirements
`POST /api/email-card-drop-requests` - make sure that we don't send an expired link to users since jobs may be delayed
`GET /api/email-card-drop-requests` - N/A
`GET /email-card-drop/verify` - don't allow expired links to claim a card, only allow latest request for a given EOA

### Implementation
1. Defines expiry as `requested_at` being greater than 60 mins before time of a db query, returns this as a derived `isExpired` property wherever we perform a db query returning values
2. Catches expiry when verifying and when sending verification link
3. A `POST /api/email-card-drop-requests` for an EOA with an existing request will add a new row to the DB
4. When verifying, we only look at the latest request

<details>
<summary> PREVIOUS IMPLEMENTATION</summary>
Simplifies the cases we have to handle by applying a unique constraint to `owner_address`. This means that we have a single row in the db for each EOA that requests a card drop, and any repeated `POST /api/email-card-drop-requests` for that EOA simply update that row
<br>
<br>
One consequence of this implementation is it becomes harder to find older information for the EOA since each new `POST` will overwrite the previous request.
</details>

### DIFF
--- a/packages/hub/queries/email-card-drop-requests.ts
+++ b/packages/hub/queries/email-card-drop-requests.ts
@@ -11,6 +11,12 @@ interface EmailCardDropRequestsQueriesFilter {
   verificationCode?: string;
 }
 
+const IS_NOT_OLD = `requested_at > now() - interval '60 minutes'`;
+const IS_OLD = `requested_at <= now() - interval '60 minutes'`;
+const IS_EXPIRED = `${IS_OLD} AND claimed_at IS NULL`;
+const IS_CLAIMABLE_REQUEST = `${IS_NOT_OLD} AND claimed_at IS NULL`;
+const RETURN_VALUE = `*, (${IS_EXPIRED}) as is_expired`;
+
 export default class EmailCardDropRequestsQueries {
   clock = inject('clock');
   databaseManager: DatabaseManager = inject('database-manager', { as: 'databaseManager' });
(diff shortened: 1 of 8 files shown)


lukemelia:
If I'm following this properly... my 2¢ is that I don't like the change since it results in our data being potentially lossy. I'd rather be able to look at the requests table and feel confident that I can see all the requests made. This would be helpful in troubleshooting and making sense of the system.

backspace:
> I'd rather be able to look at the requests table and feel confident that I can see all the requests made.

It was my suggestion to overwrite the row because it seemed easier to me vs having to filter through multiple rows per EOA to determine which ones counted as “reservations”, but really that’s accomplishable in SQL, and I do see the appeal of having the entire history in the table vs having to comb through logs for it, thanks.

packages/hub/queries/email-card-drop-requests.ts
```diff
@@ -11,15 +11,21 @@ interface EmailCardDropRequestsQueriesFilter {
   verificationCode?: string;
 }
 
+const IS_NOT_OLD = `requested_at > now() - interval '60 minutes'`;
+const IS_OLD = `requested_at <= now() - interval '60 minutes'`;
```
backspace, line 15: I think this would be nice to config-ify so it can be used elsewhere without duplication elsewhere, like in the text of the verification email. `HUB_EMAIL_CARD_DROP_RATE_LIMIT_PERIOD_MINUTES` is an admittedly clunky name but it’s a pattern I’ve seen to reduce ambiguity. I don’t think this needs to be an environment variable though.

backspace: In that vein I added `expiryMinutes` [here](https://github.com/cardstack/cardstack/pull/2971/files#diff-b38fc6a0079f557fa66aa5471a73510a2020d1423c74f28cc9f3f318b64ae823R87) 🤓

Aierie:
Updated this PR so that we retain a record of any requests that come in. We may need a cleanup job for this at some point?

backspace (approved):

backspace:
re cleanup, `reservations` doesn’t have it… probably the table could get quite large before it matters. But yes I always fret about these things hehe

PR #2969, opened 2022-05-19
hide payment profile creation CTAs if profile found

Aierie:


Aierie:
Yet to add explanations on why profile creation is disabled to the dashboard.

packages/web-client/app/templates/card-pay/payments.hbs
```diff
@@ -51,19 +55,21 @@
...
-            @disabled={{section.isCtaDisabled}}
-            data-test-workflow-button={{section.workflow}}
-          >
-            <span>{{section.cta}}</span>
-            {{svg-jar "plus"
-              class="card-pay-dashboard__payments-section-header-button-icon"
-              role="presentation"
-            }}
-          </Boxel::Button>
+          {{#unless this.merchantSafes.length}}
```
backspace, line 58: maybe this reaches the threshold where it would make sense to extract? Ideally it would be near-global to the “account” somehow but even just a getter is fine probs

backspace (approved):

packages/web-client/app/templates/card-pay/payments.hbs
Aierie: `safes.hasMerchant` might work

Aierie:
Added copy to explain why profile creation button is removed

Text. Heading reads: Payment Profiles, subtitle reads: You already have a profile for this account, so you can't create another.

PR #2970, opened 2022-05-19
ci: wait targetgroup healthy before release

pcjun97:
https://github.com/cardstack/cardstack/pull/2965 removes the `fix-listener` script which does a few things, one of it was waiting for targetgroup to become healthy before updating the weights in the listener.

Waypoint update the weights of the targetgroups in the listener in the `release` step. I've modified the `fix-listener` script into `wait-targetgroup-healthy`, which is added to a `after` hook in the `deploy` step.

[CS-3857](https://linear.app/cardstack/issue/CS-3857/migrate-services-away-from-the-usage-of-listener-arn)


waypoint.hcl
```diff
@@ -341,6 +346,11 @@ app "ssr-web" {
       when    = "after"
       command = ["node", "./scripts/purge-target-groups.mjs", "ssr-web"]
     }
+
+    hook {
+      when    = "after"
```
lukemelia, line 351: did you want this to be "before"?

lukemelia (changes requested):

waypoint.hcl
pcjun97: I tried using a `before` hook in the `release` block, but waypoint complains when the block does not contain the `use` stanza: https://github.com/cardstack/cardstack/runs/6509127447?check_suite_focus=true

I tried `use = ""` as well, but waypoint then ignores any hooks that we defined in the `release` block: https://github.com/cardstack/cardstack/runs/6521301141?check_suite_focus=true

Using the `after` hook in the `deploy` block does allow the hooks to work properly: https://github.com/cardstack/cardstack/runs/6511150292?check_suite_focus=true

PR #2971, opened 2022-05-19
hub: Add expiry notice to verification email

backspace:


packages/hub/tasks/send-email-card-drop-verification.ts
```diff
@@ -3,6 +3,7 @@ import config from 'config';
 import { query } from '@cardstack/hub/queries';
 import { inject } from '@cardstack/di';
 import * as Sentry from '@sentry/node';
+import { formatDuration } from 'date-fns'
```
Aierie, line 6: Do we need an external library? I tried out `formatDuration`. It doesn't provide anything beyond `minute` -> `minutes`.

backspace: it was already present but really I’m just missing Rails-y niceties that would turn 90 into `1 hour and 30 minutes` or whatever… but we can think about that in the future if the expiry window changes.

backspace:
I rebased this onto #2967 to update that to use `expiryMinutes`, rereview? 🤓

Aierie (approved):
Looks good, thanks!

PR #2974, opened 2022-05-20
hub: Add reservation/quantity check for drop requests

backspace:
This adds a 503 response from `POST /api/email-card-drop-requests` if SDK’s `PrepaidCardMarketV2.getQuantity` ≤ the number of “active” reservations, as in new records in the table from the past 60 minutes that haven’t been claimed. It also 503s if the contract is paused.

There’s a Sentry alert to `web-team` if `getQuantity` is < 25 (in `config`).

### DIFF
--- a/packages/hub/config/staging.json
+++ b/packages/hub/config/staging.json
@@ -6,7 +6,6 @@
   },
   "cardDrop": {
     "sku": "0x18367eee0b96d5f6e46e172ee78585d22f464884e27e7e843031c03a05af16d2",
-    "FIXME": "remove the above?",
     "verificationUrl": "https://hub-staging.stack.cards/email-card-drop/verify"
   },
   "mailchimp": {
--- a/packages/hub/config/staging.json
+++ b/packages/hub/config/staging.json
@@ -5,7 +5,7 @@
     "allowedChannels": "898941895525425172"
   },
   "cardDrop": {
-    "sku": "0x18367eee0b96d5f6e46e172ee78585d22f464884e27e7e843031c03a05af16d2",
+    "sku": "0x905b1faa1a56a0fd8447a469f8aa558665d550e56d60d106cd1096c526df4bdd",
     "verificationUrl": "https://hub-staging.stack.cards/email-card-drop/verify"
   },
   "mailchimp": {
--- a/packages/hub/config/staging.json
+++ b/packages/hub/config/staging.json
@@ -5,6 +5,8 @@
     "allowedChannels": "898941895525425172"
   },
   "cardDrop": {
+    "sku": "0x18367eee0b96d5f6e46e172ee78585d22f464884e27e7e843031c03a05af16d2",
+    "FIXME": "remove the above?",
     "verificationUrl": "https://hub-staging.stack.cards/email-card-drop/verify"
   },
   "mailchimp": {
--- a/packages/hub/node-tests/routes/email-card-drop-request-test.ts
+++ b/packages/hub/node-tests/routes/email-card-drop-request-test.ts
@@ -180,7 +180,7 @@ describe('POST /api/email-card-drop-requests', function () {
     registry(this).register('clock', FrozenClock);
 
     mockPrepaidCardMarketContractPaused = false;
-    mockPrepaidCardQuantity = 100;
+    mockPrepaidCardQuantity = 50;
   });
 
   let { request, getContainer } = setupHub(this);
@@ -191,6 +191,22 @@ describe('POST /api/email-card-drop-requests', function () {
   });
 
   it('persists an email card drop request and triggers jobs', async function () {
+    let emailCardDropRequestsQueries = await getContainer().lookup('email-card-drop-requests', { type: 'query' });
+    let insertionTimeInMs = fakeTime - 50 * 60 * 1000;
+
+    // Create no-longer-active reservations
+
+    for (let i = 0; i < mockPrepaidCardQuantity + 1; i++) {
+      await emailCardDropRequestsQueries.insert({
+        ownerAddress: '0xother',
+        emailHash: 'other-email-hash',
+        verificationCode: 'x',
+        id: shortUUID.uuid(),
+        requestedAt: new Date(insertionTimeInMs),
+        claimedAt: new Date(),
+      });
+    }
+
     let email = 'valid@example.com';
 
     const payload = {
@@ -215,7 +231,6 @@ describe('POST /api/email-card-drop-requests', function () {
         resourceId = res.body.data.id;
       });
 
-    let emailCardDropRequestsQueries = await getContainer().lookup('email-card-drop-requests', { type: 'query' });
     let emailCardDropRequest = (await emailCardDropRequestsQueries.query({ ownerAddress: stubUserAddress }))[0];
 
     expect(emailCardDropRequest.ownerAddress).to.equal(stubUserAddress);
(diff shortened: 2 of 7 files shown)


packages/hub/queries/email-card-drop-requests.ts
```diff
@@ -62,6 +62,21 @@ export default class EmailCardDropRequestsQueries {
...
+  // TODO does this cover everything? 🤔
+  async activeReservations() {
+    let db = await this.databaseManager.getClient();
+    const query = `
+      SELECT COUNT(*)
+      FROM  email_card_drop_requests AS t1
+      WHERE
+        requested_at=(SELECT MAX(requested_at) FROM email_card_drop_requests WHERE t1.owner_address=email_card_drop_requests.owner_address)
+        AND claimed_at IS NULL
+    `;
```
Aierie, line 74:
```
SELECT DISTINCT p1.owner_address FROM email_card_drop_requests p1
LEFT JOIN (SELECT DISTINCT owner_address FROM email_card_drop_requests WHERE claimed_at IS NOT NULL) as p2
ON p1.owner_address = p2.owner_address
WHERE p2.owner_address IS NULL AND p1.requested_at > (now() - interval '60 minutes')
```

This selects unique owner addresses for recent requests that do not have a claim by using a LEFT JOIN to exclude those that do have a claim.

https://stackoverflow.com/a/48646883/9809532

backspace: thanks, there’s definitely some brokenness in what I had, I’m adjusting the test cases to properly reveal that. But this part seems wrong: `WHERE p2.owner_address IS NULL`… right?

Aierie: Without that, we just get all addresses

Aierie: Wasn't clear earlier - `WHERE p2.owner_address IS NULL` is necessary to exclude the addresses that have claimed. It was not intuitive to me either, but it works!

Aierie: Oh I didn't spot the new change. The new query seems correct! And ya easier to understand

packages/hub/config/staging.json
```diff
@@ -5,6 +5,8 @@
     "allowedChannels": "898941895525425172"
   },
   "cardDrop": {
+    "sku": "0x18367eee0b96d5f6e46e172ee78585d22f464884e27e7e843031c03a05af16d2",
+    "FIXME": "remove the above?",
```
backspace, line 9: I’m unsure whether the SKU will differ between testnet and mainnet but will follow up on this once I know.

packages/hub/routes/email-card-drop-requests.ts
```diff
@@ -85,12 +96,46 @@ export default class EmailCardDropRequestsRoute {
...
+        {
+          tags: {
+            action: 'drop-card',
+            alert: 'web-team',
+          },
+        }
+      );
+
+      lastNotifiedOnQuantity = Date.now();
+    }
```
Aierie, line 129: If we're going to be sending the notification by using Sentry alerts, an alternative is: 

1. Set up a Sentry alert that is specifically for having a low quantity of prepaid cards that emails web-team, and people responsible for adding funds (right now everyone who can do this should have access to Sentry as a member)
2. Set an [action interval](https://docs.sentry.io/product/alerts/create-alerts/issue-alert-config/#action-interval-rate-limit) of 30 min to limit the notification-sending there.

This means less state to manage in the hub, and it's easier to change notification interval. I'm not clear if there are any drawbacks to this... are there? I guess having more Sentry issues is one.

packages/hub/routes/email-card-drop-requests.ts
```diff
@@ -85,12 +96,46 @@ export default class EmailCardDropRequestsRoute {
...
 
+    let prepaidCardMarketV2 = await this.cardpay.getSDK('PrepaidCardMarketV2', this.web3.getInstance());
+
+    if (await prepaidCardMarketV2.isPaused()) {
+      return respondWith503(ctx, 'The prepaid card market contract is paused');
+    }
+
+    let quantityAvailable = await prepaidCardMarketV2.getQuantity(cardDropSku);
+
+    let quantityIsBelowNotificationThreshold = quantityAvailable < notifyWhenQuantityBelow;
```
Aierie, line 107: If we only look at `quantityAvailable` without counting active reservations, we could get into a situation where an alert is not triggered despite running out of prepaid cards. 

The situation where we don't get notified is sudden rushes in requesting card drops (We get `>= notifyWhenQuantityBelow` requests, before any of them start claiming).

I think this is not a big deal right now, but probably good to note?

packages/hub/routes/email-card-drop-requests.ts
```diff
@@ -85,12 +96,46 @@ export default class EmailCardDropRequestsRoute {
...
+        }
+      );
+
+      lastNotifiedOnQuantity = Date.now();
+    }
+
+    let activeReservations = await this.emailCardDropRequestQueries.activeReservations();
+
+    if (quantityAvailable <= activeReservations) {
+      return respondWith503(ctx, 'There are no prepaid cards available');
```
Aierie, line 134: This should probably also trigger a Sentry alert, since this is the "out of prepaid card" situation we're trying to prevent/respond to. It is possible to be a false positive, but I think we would want to know anyway since this is affecting user experience.

packages/hub/node-tests/routes/email-card-drop-request-test.ts
```diff
@@ -170,6 +191,21 @@ describe('POST /api/email-card-drop-requests', function () {
   });
 
   it('persists an email card drop request and triggers jobs', async function () {
+    let emailCardDropRequestsQueries = await getContainer().lookup('email-card-drop-requests', { type: 'query' });
+    let insertionTimeBeyondExpiry = Date.now() - emailVerificationLinkExpiryMinutes * 2 * 60 * 1000;
```
Aierie, line 195: Should this be relative to `fakeTime`?

Aierie (approved):
Looks good, suggested some changes for the Sentry alerting.

packages/hub/routes/email-card-drop-requests.ts
```diff
@@ -85,12 +96,46 @@ export default class EmailCardDropRequestsRoute {
...
+        ),
+        {
+          tags: {
+            action: 'drop-card',
+            alert: 'web-team',
+          },
+        }
+      );
+
+      lastNotifiedOnQuantity = Date.now();
```
Aierie, line 128: For consistency with the rest of the things, should this also use `this.clock.now()`?

packages/hub/node-tests/routes/email-card-drop-request-test.ts
backspace: I was using that before but because the SQL query uses a non-fake `now()` it didn’t work… [mocking `now()` exists apparently](https://gist.github.com/thehesiod/d0314d599c721216f075375c667e2d9a) but seems pretty dramatique 😳

packages/hub/routes/email-card-drop-requests.ts
backspace: ya I wish we could manage Sentry alerts in Terraform, I don’t like the untracked-ness of it, but I agree with letting it manage the threshold, thanks.

packages/hub/routes/email-card-drop-requests.ts
backspace: I changed the threshold calculation to consider reservations, thanks.

packages/hub/routes/email-card-drop-requests.ts
backspace: Now that I‘ve included reservations in the threshold calculation, I think this would be redundant…??

PR #2976, opened 2022-05-24
Cs 3936 support empty payment lists

IanCal:
When we don't have any payments to make, we still need something written on chain. I've chosen a marker of all Fs as a clear distinction, I'd normally use a 0 distinguished from NULL but the on chain structure is essentially already filled with 0s.

This should be written only when

* There is a results file
* The results file has no rows

The tests should cover  this distinction.

I've also added a small check that the results file not only exists but is actually a file - it was throwing an error on some badly written data before and crashing.


tintinthong (approved):
LGTM!

PR #2979, opened 2022-05-25
Add SDK+CLI functions for PrepaidMarketV2 contract

jurgenwerk:
Ticket: CS-3840

Adds SDK & CLI commands for interacting with the PrepaidCardMarketV2 contract, which is currently deployed to Sokol. 
After all commands listed below are implemented and tested on Sokol, we can deploy to mainnet. 

- [x] addTokens
- [x] addSKU
- [x] setAsk
- [x] getQuantity
- [x] provisionPrepaidCard


jurgenwerk:
Currently the only SDK function missing is `provisionPrepaidCard` - I'm first trying to get this working by calling the contract's method directly from the console against Sokol deploy but hitting an error where the contract is [sending tokens to the PrepaidCardManager](https://github.com/cardstack/card-pay-protocol/blob/main/contracts/PrepaidCardMarketV2.sol#L193). So once I get this resolved I can add the provisioning function to the SDK.

packages/cardpay-cli/prepaid-card-market-v-2/add-tokens.ts
```diff
@@ -0,0 +1,47 @@
...
+  builder(yargs: Argv) {
+    return yargs
+      .positional('issuerSafe', {
+        type: 'string',
+        description: 'The safe address used to fund the transfer to the contract',
+      })
+      .positional('issuer', {
+        type: 'string',
+        description: 'The address of the issuer',
+      })
```
Aierie, line 18: I was a bit confused by the `issuer` positional, I learnt through trial and error that this is an EOA. Does it have to be the EOA that owns the `issuerSafe`, and do the funds come from it or from the `issuerSafe`?

habdelra: Yes, the issuer is an EOA, and yes it has to be the same EOA that owns the issuer safe. The funds come from the issuer safe though. The idea is that the issuer can find their inventory from multiple safes but still have a stable SKU

jurgenwerk:
I added the command for provisioning, but unable to provision. I'm currently trying to get this running using the local relayer and see if I can get a better error message out. 

```
Provisioning a prepaid card from the SKU 0x905b1faa1a56a0fd8447a469f8aa558665d550e56d60d106cd1096c526df4bdd to the EOA 0xeBCC5516d44FFf5E9aBa2AcaeB65BbB49bC3EBe1 in the staging environment...
Transaction hash: https://blockscout.com/poa/sokol/tx/0x9f129773c76255e9be6e00b6f4bb85382692ed68f344b0575a6382a6c78556f6/token-transfers
  waiting for txn consistency for txn=0x9f129773c76255e9be6e00b6f4bb85382692ed68f344b0575a6382a6c78556f6

Error: Transaction with hash "0x9f129773c76255e9be6e00b6f4bb85382692ed68f344b0575a6382a6c78556f6" was reverted
```

packages/cardpay-sdk/sdk/prepaid-card-market-v2/base.ts
```diff
@@ -0,0 +1,393 @@
+import Web3 from 'web3';
```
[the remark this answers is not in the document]
jurgenwerk: Yep! Fixed in https://github.com/cardstack/cardstack/pull/2979/commits/85a162b4d486bad1f5a72e90d9cb2812e98d1a24

Aierie (approved):
We've managed to provision a prepaid card on staging using the work in this branch!

packages/cardpay-sdk/sdk/prepaid-card-market-v-2/base.ts
```diff
@@ -0,0 +1,446 @@
...
+
+  private async getAddTokensPayload(issuerAddress: string, tokenAddress: string, amount: BN): Promise<string> {
+    let token = new this.layer2Web3.eth.Contract(ERC677ABI as AbiItem[], tokenAddress);
+    let prepaidCardMarketV2Address = await getAddress('prepaidCardMarketV2', this.layer2Web3);
+    let data = this.layer2Web3.eth.abi.encodeParameters(['address'], [issuerAddress]);
+    return token.methods.transferAndCall(prepaidCardMarketV2Address, amount, data).encodeABI();
+  }
+
+  async addTokens(txnHash: string): Promise<SuccessfulTransactionReceipt>;
+  async addTokens(
```
habdelra, line 45: Hos is this different than the normal Safe,sendTokens API? https://github.com/cardstack/cardstack/blob/main/packages/cardpay-sdk/sdk/safes.ts#L313 it seems like duplicate code.

jurgenwerk: It's different where we add the issuer's address to the payload: https://github.com/cardstack/cardstack/pull/2979/files/85a162b4d486bad1f5a72e90d9cb2812e98d1a24#diff-3e7b707c9d90125a88f86bb8d8c85c1f4514dbb328c263bf955d7124876ba443R40

Do you think it's fine to have similar code in this place but with this difference, or should we refactor sendTokens to support adding params to the `data`?

PR #2982, opened 2022-05-25
Freestyle:Usage cleanup

lukemelia:
This commit fixes two incorrect usages:

1) Standalone cases of FreestyleAnnotation -- moved to <:description> block of Freestyle::Usage
2) Case of Freestyle::Usage with an @description property -- use an @name property instead

This results in all content being associated with a component so filtering to a particular component works more successfully.


Aierie (approved):
Thanks. The filtering improvements are when clicking on the button next to the header right?

Component header for Action Chin, has text 'Action Chin' and a button next to it

PR #2984, opened 2022-05-27
hub: Update workflow to test all PRs

backspace:
This means tests will run on temporarily-sub PRs such as #2931.


Aierie (approved):
I think this works, because hub doesn't really use any paid services (like Percy) for testing.

PR #2985, opened 2022-05-30
Fix bug when using multiple proofs

tintinthong:
Get [error](https://sentry.io/share/issue/5f591c76ffbe44a29194efa2959f8050/) whenever attempting to claim their reward. This error is only seen when doing airdrop to multiple addresses.

Main reason is that 

new tally tally dont used 0x prefix in proofArray

Screenshot 2022-05-30 at 4 26 08 PM

old tally use 0x prefix proofArray (changed back to this)

Screenshot 2022-05-30 at 4 26 42 PM

- this error occurs only when rewarding multiple multiple addresses otherwise the proofArray would be empty
- this change requires no sdk updates

### DIFF
--- a/packages/cardpay-reward-api/cardpay_reward_api/indexer.py
+++ b/packages/cardpay-reward-api/cardpay_reward_api/indexer.py
@@ -78,7 +78,7 @@ class Indexer:
                 proofArray=payment["proof"],
                 rewardProgramId=payment["rewardProgramID"],
                 amount=amount,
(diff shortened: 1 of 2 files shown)


IanCal:
LGTM with the code change, I like the validators. I know those ensure the output format but adding a test so we don't regress in the future.

IanCal (approved):
If we can add a test easily about the output, that would be good but otherwise LGTM.

tintinthong:
Will add test later on

PR #2993, opened 2022-06-02
Add test for fast api 

tintinthong:
Adding test for fastapi. The test are divided into two parts
- the api (mostly validating fields)
- the indexer


packages/cardpay-reward-api/tests/test_reward_api.py
```diff
@@ -0,0 +1,74 @@
...
+    We need to use monkeypatch when patching a fixture
+    But the issue is that, we will always receive a ScopeMismatch error
+    when wanting the fixture to run once in a specific scope (in this case a module)
+    https://stackoverflow.com/questions/53963822/python-monkeypatch-setattr-with-pytest-fixture-at-module-scope
+    """
+    from _pytest.monkeypatch import MonkeyPatch
+
+    mpatch = MonkeyPatch()
+    yield mpatch
+    mpatch.undo()
```
tintinthong, line 35: One of the issuees with monkeypatch is that I could not get it to work with scoped fixtures very well. So, I had to patch the monkeypatch st it is scoped. That way I can instantiate the fixture for a single file and not run it every test

packages/cardpay-reward-api/cardpay_reward_api/models.py
```diff
@@ -28,6 +28,6 @@ class Root(Base):
     id = Column(Integer, primary_key=True, index=True)
     rewardProgramId = Column(String)
     rootHash = Column(String, unique=True)
-    paymentCycle = Column(Integer)
+    paymentCycle = Column(Integer, unique=True)
     blockNumber = Column(Integer)
-    timestamp = Column(Time)
+    timestamp = Column(DateTime)
```
tintinthong, line 33: Some mistakes here in the model. Will need to clear out dbs during deploy

packages/cardpay-reward-api/cardpay_reward_api/indexer.py
```diff
@@ -62,8 +62,8 @@ def add_root_and_proofs(self, db: Session, root, payment_list):
         root = models.Root(
             rootHash=root["id"],
             rewardProgramId=root["rewardProgram"]["id"],
-            paymentCycle=root["paymentCycle"],
-            blockNumber=root["blockNumber"],
+            paymentCycle=int(root["paymentCycle"]),
+            blockNumber=int(root["blockNumber"]),
```
tintinthong, line 66: Also some mistakes here also but non-critical

packages/cardpay-reward-api/cardpay_reward_api/indexer.py
```diff
@@ -77,7 +77,7 @@ def add_root_and_proofs(self, db: Session, root, payment_list):
                 payee=payment["payee"],
                 proofArray=payment["proof"],
                 rewardProgramId=payment["rewardProgramID"],
-                amount=amount,
+                amount=str(amount),
```
tintinthong, line 80: This makes our setup compatible with sqlite. Sqlite has a limitation of big numbers so must specify string. I don't have an issue with this. 

I think with this pr. The db can be replaced easily with an sqlite file

tintinthong: @IanCal I know u were investigating this earlier. Is there anything changes u made to make the orm work for postgres and sqlite that I shud know about?

tintinthong:
I've made some refactors of the fast-api-test. https://github.com/cardstack/cardstack/pull/2998.

packages/cardpay-reward-api/cardpay_reward_api/indexer.py
IanCal: @tintinthong There was something around dates, I think a datetime rather than time was one but you've already addressed that. String conversion was one that had to happen explicitly (as here) - wasn't much that needed changing.

packages/cardpay-reward-api/cardpay_reward_api/models.py
```diff
@@ -28,6 +28,6 @@ class Root(Base):
     id = Column(Integer, primary_key=True, index=True)
     rewardProgramId = Column(String)
     rootHash = Column(String, unique=True)
-    paymentCycle = Column(Integer)
+    paymentCycle = Column(Integer, unique=True)
```
IanCal, line 31: This one shouldn't be unique right?

packages/cardpay-reward-api/cardpay_reward_api/models.py
IanCal: Worth just moving to sqlite?

packages/cardpay-reward-api/cardpay_reward_api/models.py
tintinthong: Good spot!

packages/cardpay-reward-api/cardpay_reward_api/models.py
tintinthong: I dont see why not. I looked a bit. it shud be faster and ecs filesystem shud support. 

But, there is some procedures that I need to figure out (like how to tunnel). did u do work on that previously?

IanCal:
@tintinthong can we split out the web3 stuff into the other PR? I think this one looks pretty good but am not sure the web3 stuff belongs in this package.

tintinthong:
> @tintinthong can we split out the web3 stuff into the other PR? I think this one looks pretty good but am not sure the web3 stuff belongs in this package.

right how did that slip in.

tintinthong:
> @tintinthong can we split out the web3 stuff into the other PR? I think this one looks pretty good but am not sure the web3 stuff belongs in this package.

Fixed it. Can this go in @IanCal

IanCal (approved):
LGTM, not run locally.

PR #2994, opened 2022-06-02
Better withdrawal waiting UX

Aierie:
Lower timeout duration since Gnosis chain only needs 5s per new block. Identify the timeout error when it happens, and ask users to try refreshing their browser as a first course of action.


backspace (approved):
I hope we can fix the failure to detect success eventually but this is definitely an improvement 🎉

PR #2996, opened 2022-06-02
wc: Add subscription override to remove parameters

backspace:
This is a patch to `eth-block-tracker`’s `SubscribeBlockTracker` to fix bugs relating to `web-client` missing blockchain events, which has resulted in symptoms like withdrawal never completing and needing to be refreshed and the depot balances not updating after a deposit and needing to be refreshed. The patch exists simply to remove [this line](https://github.com/MetaMask/eth-block-tracker/blob/e7c075a80e46c18ecbd44d007f2257ba31079c8b/src/SubscribeBlockTracker.ts#L46). Extensive explanation:

We checked `getBlockHeight` and found that the number was not changing on production but was on staging. @Aierie suggested looking at the websocket logs and we found this error:

Wallet · Card Pay 2022-06-02 14-18-25

[This comment](https://github.com/ethereum/go-ethereum/issues/21588#issuecomment-708992169) seemed relevant, despite being for `go-ethereum`:

> The extra object parameter is not supported

We tried `eth_subscribe` to `newHeads` with and without the empty parameters object:

```
quickNode = new WebSocket(QUICKNODE_URL);
quickNode.onmessage = m => console.log(m.data);

quickNode.send('{"method":"eth_subscribe","params":["newHeads",{}],"id":1,"jsonrpc":"2.0"}')
> {"jsonrpc":"2.0","error":{"code":-32602,"message":"Couldn't parse parameters: newHeads","data":"\"Expected no parameters.\""},"id":1}

quickNode.send('{"method":"eth_subscribe","params":["newHeads"],"id":1,"jsonrpc":"2.0"}')
> {"jsonrpc":"2.0","result":"0xcfa1d5e8020c5a7f","id":1}
```

The Sokol POA RPC node accepts both:

```
sokol = new WebSocket('wss://sokol.poa.network/wss');
sokol.onmessage = m => console.log(m.data);

sokol.send('{"method":"eth_subscribe","params":["newHeads",{}],"id":1,"jsonrpc":"2.0"}')
> {"jsonrpc":"2.0","result":"0x0d2bad2d11bb49c79c17e72275d0a3e2","id":1}

sokol.send('{"method":"eth_subscribe","params":["newHeads"],"id":1,"jsonrpc":"2.0"}')
> {"jsonrpc":"2.0","result":"0x0223b203bae7435eb7ddef6fa0cefd2c","id":1}
```

With this patch in place, the aforementioned withdrawal and deposit bugs were fixed!

Ultimately we hope to get this change into the upstream library. It’s maintained by MetaMask, which is backed by Infura, which appears to be indifferent to the absence of the `{}` parameter. But for now, this will address some bugs that have puzzled us for a while.

### DIFF
--- a/packages/wc-provider/block-tracker.ts
+++ b/packages/wc-provider/block-tracker.ts
@@ -12,6 +12,23 @@ export default class PatchedSubscribeBlockTracker extends SubscribeBlockTracker 
     //@ts-ignore need to expose the provider for updating
     return this._provider;
   }
+
+  // Some RPC nodes don’t support subscribing to newHeads with {} parameters, so this removes that
+  protected async _start(): Promise<void> {
+    //@ts-ignore access private this._subscriptionId
+    if (this._subscriptionId === undefined || this._subscriptionId === null) {
+      try {
+        //@ts-ignore access private this._call
+        const blockNumber = (await this._call('eth_blockNumber')) as string;
+        //@ts-ignore access private this._subscriptionId and this._call
+        this._subscriptionId = (await this._call('eth_subscribe', 'newHeads')) as string;
+        this.provider.on('data', this._handleSubData.bind(this));
+        this._newPotentialLatest(blockNumber);
+      } catch (e) {
+        this.emit('error', e);
+      }
+    }
+  }
 }
 
 type Constructor = new (...args: any[]) => {};


packages/wc-provider/block-tracker.ts
```diff
@@ -12,6 +12,23 @@ export default class PatchedSubscribeBlockTracker extends SubscribeBlockTracker
     //@ts-ignore need to expose the provider for updating
     return this._provider;
   }
+
+  // Some RPC nodes don’t support subscribing to newHeads with {} parameters, so this removes that
+  protected async _start(): Promise<void> {
+    //@ts-ignore access private this._subscriptionId
+    if (this._subscriptionId === undefined || this._subscriptionId === null) {
```
habdelra, line 19: a more succinct way to express this is:
`this._subscriptionId == null`, its the one time that `==` makes sense to use.

habdelra (approved):

PR #3002, opened 2022-06-07
Add reward pool balance endpoint

tintinthong:
The point of this is to expose an endpoint `/reward-pool-balance/{rewardProgramId}/{token}` so that checkly can hit the endpoint and send notification to alarms-internal channel on discord

This is paired with this pr https://github.com/cardstack/infra/pull/171


pcjun97 (approved):

IanCal (approved):
LGTM, if there's a sensible place in the future we can move this to (e.g. hub?) we should look at extracting it out.

We may want an API that can query an arbitrary contract for view functions for other checks in the future, but this works right now,

PR #3009, opened 2022-06-08
ssr: Update exchange-fetch for new Hub interface

backspace:
This updates payment request conversion to use the new
Hub CryptoCompare caching proxy in #2992.


Aierie:
This needs to be deployed first, right? before the hub PR is merged.

backspace:
yes… maybe I’ll actually merge it into the Hub PR just so they’re bound

PR #3010, opened 2022-06-10
ssr: Update to Ember 3.28

backspace:
This probably has a long way to go but I want to see if a fresh build reveals anything.


backspace:
The reason for [`qunit/require-expect`](https://github.com/cardstack/cardstack/runs/6834018408?check_suite_focus=true#step:6:99) errors is that the `percySnapshot(assert)` calls trigger the [rule](https://github.com/platinumazure/eslint-plugin-qunit/blob/master/docs/rules/require-expect.md). `assert` is only passed in for an [internal `generateName`](https://github.com/percy/percy-ember/blob/6e263861a6f58eb3cf44cccfbb0abcd5e95b2ad7/addon-test-support/%40percy/ember/index.js#L16) that constructs a snapshot name agnostically in both Mocha and QUnit contexts but the lint rule is concerned:

> The rationale here is that by wrapping `assert` statements in conditional blocks or callbacks, developers are at risk of creating tests that incorrectly pass by skipping assertions.

There’s a plan to [change the default](https://github.com/platinumazure/eslint-plugin-qunit/issues/222) for this rule from the above `except-simple`

I appreciate the spirit of this rule and think that it could be helpful in cases like [this](https://github.com/cardstack/cardstack/pull/3010/commits/14362124ce7afcd192011686e2d8bd2cb26dd09e#diff-6bf2c86d24b7518874acca763adba6fd3dd5b5fe31b9362cc4a45587d8ff8af2R18). I generally prefer to stay as close to the Ember CLI blueprints as possible and not add/remove default rules whenever possible. I’m surprised I haven’t found any mentions of this regarding `@percy/ember`. This will surely happen with `web-client` as well.

Any thoughts @Aierie?

Aierie:
> The reason for [`qunit/require-expect`](https://github.com/cardstack/cardstack/runs/6834018408?check_suite_focus=true#step:6:99) errors is that the `percySnapshot(assert)` calls trigger the [rule](https://github.com/platinumazure/eslint-plugin-qunit/blob/master/docs/rules/require-expect.md). `assert` is only passed in for an [internal `generateName`](https://github.com/percy/percy-ember/blob/6e263861a6f58eb3cf44cccfbb0abcd5e95b2ad7/addon-test-support/%40percy/ember/index.js#L16) that constructs a snapshot name agnostically in both Mocha and QUnit contexts but the lint rule is concerned:
> 
> > The rationale here is that by wrapping `assert` statements in conditional blocks or callbacks, developers are at risk of creating tests that incorrectly pass by skipping assertions.
> 
> There’s a plan to [change the default](https://github.com/platinumazure/eslint-plugin-qunit/issues/222) for this rule from the above `except-simple`
> 
> I appreciate the spirit of this rule and think that it could be helpful in cases like [this](https://github.com/cardstack/cardstack/pull/3010/commits/14362124ce7afcd192011686e2d8bd2cb26dd09e#diff-6bf2c86d24b7518874acca763adba6fd3dd5b5fe31b9362cc4a45587d8ff8af2R18). I generally prefer to stay as close to the Ember CLI blueprints as possible and not add/remove default rules whenever possible. I’m surprised I haven’t found any mentions of this regarding `@percy/ember`. This will surely happen with `web-client` as well.
> 
> Any thoughts @Aierie?

I think the rule is useful for some specific cases, though in Percy's case it seems more of a hindrance. But why introduce `eslint-plugin-qunit` now, instead of in a separate PR with less going on?

backspace:
> why introduce `eslint-plugin-qunit` now, instead of in a separate PR with less going on?

it’s just in the [Ember CLI 3.27 blueprint](https://github.com/ember-cli/ember-cli/releases/tag/v3.27.0) 🥴

ef4:
> it’s just in the [Ember CLI 3.27 blueprint](https://github.com/ember-cli/ember-cli/releases/tag/v3.27.0) 🥴

When a default eslint config change like this introduces significant code changes, I think a good option is to take the new changes but immediately disable the problematic rules, and then address the churn in a separate PR that re-enables the rule.

In the case of the qunit rules, some of them are pretty opinionated and it's reasonable to decide to disable whole rules that we don't care about and not plan to re-enable them.

lukemelia:
Maybe obsoleted by #3068?

backspace:
> Maybe obsoleted by #3068?

indeed, I’m curious what was different for you, I’ve done a lot of Ember updates over the years but this one stumped me twice!

PR #3011, opened 2022-06-10
hub: Add job_tickets migration

backspace:
This is temporarily set to merge into #2992 as that contains an unmerged migration.

The issue only describes the table, is there anything more to it, @jurgenwerk?


jurgenwerk (changes requested):
Since we're going to query by job type, owner address, and state, I think it would be a good idea to make an index for that.

backspace:
> Since we're going to query by job type, owner address, and state, I think it would be a good idea to make an index for that.

👍🏻 added now, thanks 😀

PR #3016, opened 2022-06-13
hub: Add CreateProfile task

backspace:
These simple setup commits are making me wish for:
* extraction for `StubWorkerClient` and its assertions
* automatic discovery of tasks

### DIFF
--- a/packages/hub/tasks/create-profile.ts
+++ b/packages/hub/tasks/create-profile.ts
@@ -1,7 +1,9 @@
-import { Helpers } from 'graphile-worker';
+import { inject } from '@cardstack/di';
 
 export default class CreateProfile {
-  async perform(_payload: any, helpers: Helpers) {
-    helpers.logger.info(`Placeholder for CreateProfile`);
+  workerClient = inject('worker-client', { as: 'workerClient' });
+
+  async perform(_payload: any) {
+    this.workerClient.addJob('persist-off-chain-merchant-info', { 'merchant-safe-id': 'fixme' });
   }
 }
--- a/packages/hub/tasks/create-profile.ts
+++ b/packages/hub/tasks/create-profile.ts
@@ -1,0 +1,7 @@
+import { Helpers } from 'graphile-worker';
+
+export default class CreateProfile {
+  async perform(_payload: any, helpers: Helpers) {
+    helpers.logger.info(`Placeholder for CreateProfile`);
+  }
+}
--- a/packages/hub/main.ts
+++ b/packages/hub/main.ts
@@ -62,6 +62,7 @@ import ReservedWords from './services/reserved-words';
 import CardpaySDKService from './services/cardpay-sdk';
 import WorkerClient from './services/worker-client';
 import { Clock } from './services/clock';
+import CreateProfile from './tasks/create-profile';
 import Web3HttpService from './services/web3-http';
 import Web3SocketService from './services/web3-socket';
 import ExchangeRatesService from './services/exchange-rates';
@@ -120,6 +121,7 @@ export function createRegistry(): Registry {
   registry.register('cardpay', CardpaySDKService);
   registry.register('clock', Clock);
   registry.register('contracts', Contracts);
+  registry.register('create-profile', CreateProfile);
   registry.register('custodial-wallet-route', CustodialWalletRoute);
   registry.register('database-manager', DatabaseManager);
   registry.register('development-config', DevelopmentConfig);
--- a/packages/hub/node-tests/tasks/create-profile-test.ts
+++ b/packages/hub/node-tests/tasks/create-profile-test.ts
@@ -1,0 +1,41 @@
+import { Job, TaskSpec } from 'graphile-worker';
+import { registry, setupHub } from '../helpers/server';
+import { expect } from 'chai';
+import CreateProfile from '../../tasks/create-profile';
+
+let addedJobIdentifiers: string[] = [];
+let addedJobPayloads: string[] = [];
+
+class StubWorkerClient {
+  async addJob(identifier: string, payload?: any, _spec?: TaskSpec): Promise<Job> {
+    addedJobIdentifiers.push(identifier);
+    addedJobPayloads.push(payload);
+    return Promise.resolve({} as Job);
+  }
+}
+
+describe('CreateProfileTask', function () {
+  let subject: CreateProfile;
+
+  this.beforeEach(function () {
+    registry(this).register('worker-client', StubWorkerClient);
+  });
+
+  let { getContainer } = setupHub(this);
+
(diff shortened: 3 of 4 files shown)


packages/hub/services/relay.ts
```diff
@@ -82,6 +82,33 @@ export default class RelayService {
     let body = await response.json();
     return body.txHash;
   }
+
+  async registerProfile(userAddress: string, did: string): Promise<string> {
+    let relayUrl = getConstantByNetwork('relayServiceURL', web3config.layer2Network);
+    if (!provisionerSecret) {
```
jurgenwerk, line 88: Should we use the "provisioner secret", which is used for prepaid card provisioning, for registering merchants as well? Or would it be better to add another secret to the relayer,  the "merchant registrar secret"? This would be clearer to identify which secret is for what action. But this doesn't really scale well in case we add even more actions to the relayer... Or are we okay with "provisioner secret" being a good umbrella name for the trusted actions we run thru the relayer? Also pinging @habdelra for this one

packages/hub/services/relay.ts
```diff
@@ -82,6 +82,33 @@ export default class RelayService {
...
+      throw new Error(`Could not register profile because relay.provisionerSecret config is not set.`);
+    }
+    let response = await fetch(`${relayUrl}/v1/profiles/register?eoa=${userAddress}&did=${did}`, {
```
jurgenwerk, line 91: Current URL that I have in my local relayer feature branch is `v1/merchant/register`. Params are `owner` and `info_did`.

packages/hub/tasks/create-profile.ts
```diff
@@ -0,0 +1,36 @@
...
+    'merchant-info-id': merchantInfoId,
+  }: {
+    'job-ticket-id': string;
+    'merchant-info-id': string;
+  }) {
+    try {
+      let merchantInfo = (await this.merchantInfoQueries.fetch({ id: merchantInfoId }))[0];
+      let did = encodeDID({ type: 'MerchantInfo', uniqueId: merchantInfoId });
+
+      let merchantSafeAddress = await this.relay.registerProfile(merchantInfo.ownerAddress, did);
```
jurgenwerk, line 26: I think the relayer will return the transaction id and we need to wait here for the transaction to be finished or failed, and update the job ticket accordingly.

packages/hub/services/relay.ts
backspace: I can’t think of a reason to have separate secrets unless it feels safer to isolate a leak to one contract. But this is definitely wild speculation for me hehe

packages/hub/tasks/create-profile.ts
backspace: ah okay, does that mean subgraph support will be needed?

packages/hub/services/relay.ts
backspace: Updated now, thanks.

packages/hub/services/relay.ts
habdelra: i agree. we can segment them in the future, but I think the funds at risk right now is low enough to not worry about having separate secrets

packages/hub/tasks/create-profile.ts
habdelra: good point, the relayer is not indexing V2 market contract txns. currently the SDK using the subgraph to look for txn hashes so that there is consistency between what is on chain and what is in the subgraph. not using the subgraph would force clients of this to have to use web3 to monitor txn completion instead of the SDK

packages/hub/services/relay.ts
jurgenwerk: Thanks, noted!

packages/hub/tasks/create-profile.ts
jurgenwerk: @habdelra we're not dealing with the `PrepaidCardMarketV2` contract here, just the `MerchantManager` (we use the `registerMerchant` function). It looks like `MerchantCreation` event is indexed in the subgraph so I think we can use the SDK for waiting for the transaction to finish.

backspace: It looks like `waitForTransactionConsistency` [isn’t exported](https://github.com/cardstack/cardstack/blob/69dd3bad2341981cdcf9afb12445940ef49f29ff/packages/cardpay-sdk/index.ts#L38), is there a higher-level function I should be using? Or should I add an export for it?

I tried that locally but I think I’d need to publish the SDK for it to work…?

backspace: okay, I was finally able to use that function locally, I tried it with [this prepaid card issuance transaction](https://blockscout.com/xdai/mainnet/tx/0xf47180533b1eada183ec94b29db7dc2707041b88607a1fca4ac21d27ca23fd82/internal-transactions):

```
result = await waitForTransactionConsistency(
  window['@cardstack/WebClient'].__container__.lookup('service:layer2-network').strategy.web3,
  '0xf47180533b1eada183ec94b29db7dc2707041b88607a1fca4ac21d27ca23fd82')
```

I didn’t see the address of the prepaid card (`0x362b921cFe6a72731219AE9366784DF742cE06f9`) in the result though

What would it look like to extract the merchant safe address from the result of this function?

lukemelia (approved):

packages/hub/tasks/create-profile.ts
```diff
@@ -0,0 +1,77 @@
...
+      let merchantInfo = (await this.merchantInfoQueries.fetch({ id: merchantInfoId }))[0];
+      let did = encodeDID({ type: 'MerchantInfo', uniqueId: merchantInfoId });
+
+      let profileRegistrationTxHash = await this.relay.registerProfile(merchantInfo.ownerAddress, did);
+
+      await this.cardpay.waitForTransactionConsistency(this.web3.getInstance(), profileRegistrationTxHash);
+
+      let merchantCreationsSubgraphResult = await this.cardpay.gqlQuery(network, merchantCreationsQuery, {
+        txn: profileRegistrationTxHash,
+      });
```
lukemelia, line 55: Is there a risk of this graph query failing and causing the job ticket to fail but, the profile was actually already created? In that situation will retrying the job ticket succeed?

backspace: I’m unsure. The issue specifies that this job should not automatically be retried. But the next issue I’m going to work on lets Card Wallet retry a job from its ticket id… but that would indeed create a second merchant safe, undesirable 🤔

habdelra: Yeah, if a txn hash exists, the system needs to assume that the safe also exists, despite the inability to resolve the merchant safe id if the graph node is unavailable.

PR #3017, opened 2022-06-14
Add script dependency to python subpackages

tintinthong:
It looks like the dependency `js-hcl-parser` in the `wait-service-stable` script is unavailable to python subpackages. This pr adds `js-hcl-parser` within the github action so the script can run. I didn't want to install the entire monorepo since it seemed unnecessary for now.


pcjun97 (approved):
LGTM!

PR #3021, opened 2022-06-15
ci: Use hcl2-parser

pcjun97:
HCL parser
-
Replace `js-hcl-parser` with `hcl2-parser` due to the fact that `js-hcl-parser` is unable to parse the new hcl2 syntax (eg: [gitrefpretty](https://tip.waypointproject.io/docs/waypoint-hcl/functions/all#gitrefpretty))

Non Javascript services hook and action dependencies
-

All deploy workflows are updated to install `hcl2-parser` and `@actions/core` as part of their dependencies. Some non javascript services (such as python ones) does not require `yarn` to fetch all the dependencies for the entire monorepo, thus a separate `yarn add <dependency> -W` makes more sense.

### DIFF
--- a/.github/actions/deploy-cardie/action.yml
+++ b/.github/actions/deploy-cardie/action.yml
@@ -21,7 +21,6 @@ runs:
         curl -fsSL https://apt.releases.hashicorp.com/gpg | sudo apt-key add -
         sudo apt-add-repository "deb [arch=amd64] https://apt.releases.hashicorp.com $(lsb_release -cs) main"
         sudo apt-get install awscli waypoint=${{ inputs.waypoint_version }}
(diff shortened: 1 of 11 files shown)


lukemelia (commented):
Should we consider adding hcl2-parser and @actions/core to the devDependencies of the root package.json? It seems annoying to have to include this in every deployment action.

lukemelia:
> Should we consider adding hcl2-parser and @actions/core to the devDependencies of the root package.json? It seems annoying to have to include this in every deployment action.

I belatedly see the note about this in the PR description. I still think we should consider only requiring a specific install in the python package cases then.

pcjun97:
Added `hcl2-parser` to dependencies and removed it from js services

PR #3026, opened 2022-06-16
Add script to check sync of indexing

tintinthong:
Due to sentry errors, I was wary on whether the indexing occured correctly. I made a script to reassure me of any diffs between s3 and the db.

Screenshot 2022-06-16 at 4 13 04 PM

### DIFF
--- a/packages/cardpay-reward-indexer/README.md
+++ b/packages/cardpay-reward-indexer/README.md
@@ -1,0 +1,6 @@
+
+## Scripts 
+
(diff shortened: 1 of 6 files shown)


tintinthong:
@IanCal this sync is confirmed on both prod and staging

tintinthong:
@IanCal can we move get this in before I split stuff up?

lukemelia (approved):

PR #3028, opened 2022-06-16
hub: Fix card drop availability calculation

backspace:
A combination of discrepancy in mock response types and test setup coincidences were masking a bug in the determination of whether there are sufficient funds for a user to initiate the card drop process. To exercise #3025 in staging, I checked `getQuantity` (48) and added 100 `email_card_drop_requests` records to simulate the number of reserved cards exceeding available funds and found that it was still showing `available: true`:

image

With this change, `available: false` was the result in the same circumstances.

This updates SQL queries to not use `NOW()`, instead passing in `this.clock.now()`, which gets around the [difficulty in mocking `NOW()`](https://github.com/cardstack/cardstack/pull/2974#discussion_r882664945). So a few tests now use `fakeTime`.

### DIFF
--- a/packages/hub/node-tests/routes/email-card-drop-request-test.ts
+++ b/packages/hub/node-tests/routes/email-card-drop-request-test.ts
@@ -49,7 +49,8 @@ class StubCardpaySDK {
     switch (sdk) {
       case 'PrepaidCardMarketV2':
         return Promise.resolve({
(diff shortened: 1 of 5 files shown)


Aierie (approved):
The date change makes sense for another reason too, I think... since the value of `requested_at` is set by us, and we aren't handling time zones, there's potential for a time zone difference between the SQL db and hub to mess things up?

PR #3038, opened 2022-06-22
Update reward endpoint

tintinthong:


tintinthong:
@danibonilha just flagging here that I have changed the endpoint. Both endpoints will work still but a new sdk release will point to the new one.

pcjun97 (approved):

PR #3042, opened 2022-06-22
Rename internals xdai->gnosis and allow --network=gnosis in cardpay CLI

lukemelia:


packages/cardpay-sdk/sdk/constants.ts
```diff
@@ -89,7 +89,8 @@ const constants: {
   sokol: SOKOL,
   kovan: KOVAN,
   mainnet: MAINNET,
-  xdai: XDAI,
+  gnosis: GNOSIS,
+  xdai: GNOSIS,
 });
 
 export const networks: { [networkId: number]: string } = Object.freeze({
```
backspace, line 96: a few lines down from here is one other place that needs updating; if you visit [the `web-client` productionpreview deployment] there‘s this error in the console:

```
Error: chainId is required to set custom session storage for parallel connections in WalletConnect
```

backspace (changes requested):
This mostly looks good to me apart from the web breakage but also this breaks existing payment links, you can see that on [this preview deployment](https://cs-4120-cardpay-cli-should-accept-gnosis-as-a.ssr-web-preview.cardstack.com/pay/xdai/0xf9c0E2B59824f33656CC5A94423FcF62892dad60?amount=0.5&currency=USD). It works if you change `xdai` in the URL to `gnosis`, but I always prefer to keep old links working, so maybe this means forwarding in `ssr-web`? Or maybe the `networkIds` change would fix this too, I’m unsure.

packages/cardpay-sdk/sdk/constants.ts
lukemelia: Actually, I think ssr-web is good but web-client still has issues.

lukemelia: @backspace Now this should be all good.

backspace (approved):
I’m glad the unraveling ended eventually!

jurgenwerk (approved):
👏

PR #3045, opened 2022-06-23
hub: Add retry endpoint for job tickets

backspace:
`POST /api/job-tickets/:id/retry` queues a duplicate of the retried job and returns the id and attributes of the new job ticket for that job.

The `CreateProfile` job is intended to be queued with spec `maxAttempts: 1`. I added the `job_tickets.spec` column to preserve the spec a job was started with so it can be used again when retrying.

### DIFF
--- a/packages/hub/node-tests/routes/job-tickets-test.ts
+++ b/packages/hub/node-tests/routes/job-tickets-test.ts
@@ -1,5 +1,6 @@
 import { registry, setupHub } from '../helpers/server';
 import shortUUID from 'short-uuid';
+import JobTicketsQueries from '../../queries/job-tickets';
 
 class StubAuthenticationUtils {
   validateAuthToken(encryptedAuthToken: string) {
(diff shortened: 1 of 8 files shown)


packages/hub/routes/job-tickets.ts
```diff
@@ -3,13 +3,16 @@ import autoBind from 'auto-bind';
 import { query } from '@cardstack/hub/queries';
 import { ensureLoggedIn } from './utils/auth';
 import { inject } from '@cardstack/di';
+import shortUUID from 'short-uuid';
+import { JOB_TICKETS_STATE_DEFAULT } from '../db/migrations/20220610203119883_create-job-tickets';
```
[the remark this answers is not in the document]
Aierie: Alternatively, we could also remove the SQL default value, put this constant in the db query file, and have the db query file be responsible for adding it as a default where not specified.

packages/hub/node-tests/routes/job-tickets-test.ts
```diff
@@ -101,3 +103,146 @@ describe('GET /api/job-tickets/:id', function () {
...
+          state: 'pending',
+        });
+      })
+      .expect('Content-Type', 'application/vnd.api+json');
+
+    expect(getJobIdentifiers()).to.deep.equal(['a-job']);
+    expect(getJobPayloads()).to.deep.equal([{ 'a-payload': 'yes' }]);
+    expect(getJobSpecs()).to.deep.equal([{ 'a-spec': 'yes' }]);
+
+    let allTickets = await jobTicketsQueries.findAll();
```
Aierie, line 161: If we assign the return value of `await request()...` to a variable, we can retrieve the id via `res.body.data.id` and use `jobTicketQueries.find(id)` so we don't have to introduce `findAll` as a test-only method.

packages/hub/routes/job-tickets.ts
```diff
@@ -69,6 +73,81 @@ export default class JobTicketsRoute {
...
+
+      return;
+    }
+
+    if (jobTicketToRetry.state !== 'failed') {
+      ctx.body = {
+        errors: [
+          {
+            status: '422',
+            title: 'Job ticket state is not failed',
```
jurgenwerk, line 123: A nitpick - would prefer `Can only retry a job with failed state (current state: ${jobTicketToRetry.state})` for a little more clarity

jurgenwerk (approved):
Looking good! 

Something unrelated to this PR that occurred to me - would be beneficial to have some monitoring in place that would tell us if the job tickets are stuck in pending state for too long. Probably this would include making a runner that periodically queries for this, or maybe adding it to the status endpoint and leverage Checkly for checking. We could include so many other sanity checks, for example checking if push notifications are working (have any been sent recently), etc.

packages/hub/routes/job-tickets.ts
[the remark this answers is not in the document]
backspace: I originally had  `state: null` when creating the new job, allowing the database to set the default, since the Javascript doesn’t really need to know the string, but it caused Typescript problems. In 4bbf50 I have adjusted the model type to allow that, thanks.

packages/hub/node-tests/routes/job-tickets-test.ts
backspace: 👍🏻 done in a757345

packages/hub/routes/job-tickets.ts
backspace: done in 3959414, thanks

PR #3056, opened 2022-06-30
Task interface registry

Aierie:
Typescript enforcement of task names and types. There's room to make a couple of tasks' payloads stricter.

We might want to modify the type interface of `helpers` in a task definition.... if we can find a way to do it nicely.

### DIFF
--- a/packages/hub/routes/job-tickets.ts
+++ b/packages/hub/routes/job-tickets.ts
@@ -4,10 +4,11 @@ import { query } from '@cardstack/hub/queries';
 import { ensureLoggedIn } from './utils/auth';
 import { inject } from '@cardstack/di';
 import shortUUID from 'short-uuid';
+import { KnownTasks } from '@cardstack/hub/tasks';
 
 export interface JobTicket {
   id: string;
-  jobType: string;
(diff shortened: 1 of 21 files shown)


packages/hub/routes/job-tickets.ts
```diff
@@ -4,10 +4,11 @@ import { query } from '@cardstack/hub/queries';
 import { ensureLoggedIn } from './utils/auth';
 import { inject } from '@cardstack/di';
 import shortUUID from 'short-uuid';
+import { KnownTasks } from '@cardstack/hub/tasks';
 
 export interface JobTicket {
   id: string;
-  jobType: string;
+  jobType: keyof KnownTasks; // TODO: how do we know that this is true?
```
backspace, line 11: 🤔

backspace (approved):
nice, thanks. Does this open up the possibility of auto-discovery of tasks? `main.ts` and `worker.ts` are sad-making every time I add a new one 😞

Aierie:
Auto-discovery is a separate thing that was always possible, and shouldn't be too much effort.

PR #3059, opened 2022-07-04
hub: Add IAP verification

backspace:
This replaces the stubbed implementation of the `InAppPurchases` service. In the case of Google Play purchase validation, it does not include the required authentication code, which looks to be quite involved and not worth adding until it can be truly exercised, pending completion of necessary setup tasks. There’s a new issue to ensure followup on this.

It introduces `msw` for HTTP mocking, but I had to use an older version because of [import problems](https://github.com/mswjs/msw/issues/1267).


packages/hub/config/test.cjs
```diff
@@ -16,6 +16,15 @@ module.exports = {
     },
     verificationUrl: 'https://card-drop-email.test/email-card-drop/verify',
   },
+  iap: {
+    apple: {
+      verificationUrl: 'https://test.itunes.apple.com/verifyReceipt',
+    },
+    google: {
+      verificationUrlBase:
+        'https://test.googleapis.com/androidpublisher/v3/applications/com.cardstack.cardpay/purchases/products/0001/tokens',
```
Aierie, line 25: If these are not real urls, would we be able to use `.test` domains instead?

packages/hub/services/in-app-purchases.ts
```diff
@@ -1,8 +1,56 @@
...
+
+interface InAppPurchaseValidationResult {
+  valid: boolean;
+  response: any;
+}
+
 export default class InAppPurchases {
-  async validate(provider: string, receipt: any) {
-    console.error('In-app purchase validation is not implemented, arguments are', provider, receipt);
+  async validate(provider: string, receipt: any): Promise<InAppPurchaseValidationResult> {
```
Aierie, line 10:
```suggestion
  async validate(provider: 'google' | 'apple', receipt: string): Promise<InAppPurchaseValidationResult> {
```

If `string` is the correct type for both receipts, then we could do this, is it really `string`?

packages/hub/services/in-app-purchases.ts
```diff
@@ -1,8 +1,56 @@
+import config from 'config';
+import fetch from 'node-fetch';
```
Aierie, line 2: Do we need to import `fetch` here? We assign `global.fetch` in `main.ts`

packages/hub/node-tests/services/in-app-purchases-test.ts
```diff
@@ -0,0 +1,201 @@
...
+
+      mockServer.listen({ onUnhandledRequest: 'error' });
+    });
+
+    this.afterEach(function () {
+      mockServer.close();
+    });
+
+    it('passes along a successful validation', async function () {
+      let subject = await getContainer().lookup('in-app-purchases');
```
Aierie, line 30: `subject` could move into a `beforeEach`

packages/hub/services/in-app-purchases.ts
```diff
@@ -1,8 +1,56 @@
...
+      method: 'POST',
+      body: JSON.stringify(requestBody),
+    });
+
+    let json = await response.json();
+
+    if (json.status === 0) {
+      return { valid: true, response: json };
+    } else {
+      return { valid: false, response: json };
```
Aierie, line 37: There are some cases where we want to report to Sentry listed here:

https://developer.apple.com/documentation/appstorereceipts/status

Most things besides 21003 seem like they should be reported to Sentry?

packages/hub/services/in-app-purchases.ts
```diff
@@ -1,8 +1,56 @@
...
+      method: 'GET',
+    });
+
+    let json = await response.json();
 
-    return { valid: true, response: {} };
+    if (response.ok && json.resource.purchaseState === 0) {
+      return { valid: true, response: json };
+    } else {
+      return { valid: false, response: json };
```
Aierie, line 52: Are there cases for Google where we want to report to Sentry too?

Aierie (commented):
This is neat! How can I try stuff out for this PR?

packages/hub/services/in-app-purchases.ts
backspace: it’s true, this is something I didn’t appreciate initially, thanks

packages/hub/services/in-app-purchases.ts
backspace: news to me, thanks!

PR #3061, opened 2022-07-05
fix task instantiation in tests

Aierie:
we should use `instantiate`, as we do in `worker.ts`. `instantiate` does not keep shared state between different tasks, while `lookup` does since tasks share one instance. this means we have no need of registering in `main.ts` either


lukemelia (approved):
Good find 👍

PR #3062, opened 2022-07-05
Rename xdai -> gnosis in subgraph scripts

IanCal:
This extends the work done elsewhere changing references from xdai to gnosis.

Not included here is moving the whole subgraph name to gnosis, which will require more co-ordination.


IanCal:
This is the mininum change required to get things building again. Two further steps could be

* Renaming the package commands
* Renaming the subgraph itself (a much bigger task)

lukemelia (approved):
👍

PR #3066, opened 2022-07-07
Adopt VSCode Multi-Root & Github Action Linting (Python)

tintinthong:
This PR adds 
1. the use of a VS code [multi-root workspace](https://code.visualstudio.com/docs/editor/multi-root-workspaces) for python developer workflow
2. pr-linting github actions that runs flake8, black and isort in succession to ensure code quality. This is made to work with the above multi-root-workspace.

In order to open the workspace, Go to File > Open Workspace From File > Select `python-cardstack.code-workspace`. There are 2 files added to root of monorepo `pyproject.toml` (which is only for formatting settings) and `python-cardstack.code-workspace` (which is for vscode use). 

Few notable changes: 
- **Get VS Code to work well with PEP582**. Since we follow a `node-modules`-type approach (the equivalent being `__pypackages__`) with our python subpackages, we need VSCode to understand these paths. Most dev dependencies remain local to the subpackage, for example, `pytest` or `black` but are pointed to at this root workspace. 
- **Pytest Discovery of all subpackages**. Developers previously have to open VSCode workspace in each subpackage individually to get pytest discovery to work within the console. Now, we can literally open the workspace that displays all test in python subpackages defined in the workspace file.  
- **Python 3.9 and 3.10 support**. This particular setup should support python interpreter choice of 3.9 or 3.10. For a seamless VSCode workflow, you should select the python interpreter inside your user space settings.

Screenshot 2022-07-20 at 10 49 48

Screenshot 2022-07-20 at 10 54 39

Downsides: 
- **Exclusion of all other files in the monorepo**. Opening multi-root workspace file seems to exclude all other files that are not specifed as a `"folder"`. 
- **Still need to manually pdm install**. This is purely a DX change, not an opiniated monorepo change. There is no `pdm install` that occurs at the root that will make ALL dependencies available. One still needs to manually go into each subpackage and `pdm install` or `pdm install -d`. **I believe this is better handled when building python-utils subpackage**


cardstack.code-workspace
```diff
@@ -0,0 +1,43 @@
...
+    }
+  ],
+  "settings": {
+    "[python]": {
+      "editor.formatOnSave": true,
+      "editor.codeActionsOnSave": {
+        "source.organizeImports": true
+      }
+    },
+    "python.defaultInterpreterPath": "/opt/homebrew/Cellar/pdm/1.15.3/libexec/bin/python3.10",
```
tintinthong, line 23: I thnk theres no way around this. Must just specify the interpreter path given by `pdm use`

tintinthong: Resolved. Just have to specify python path in user space settings. Include in readme to use an interpreter of python3.10.

tintinthong:
Just a note: Black formatting only ever refers to a single `pyproject.toml` file (for settings).

So
- if we execute `black <root>` it will use the `pyproject.toml` file in the `<root>`. 
- if we execute `black <path>/<child>` it will use the `pyproject.toml` file in `<path>`

This is as expected

Vscode workspace conforms to this so if you save a file in `<path>/<child>`, it will follow `pyproject.toml` in `<path>`

So what that means is that. We should **NOT** have subpackage specific black settings.

tintinthong:
Big reformat occured [0295083](https://github.com/cardstack/cardstack/pull/3066/commits/02950834d4af420d5e3483eb3e7f6fd663235dc8)

IanCal:
I'll make sure I see the same stuff locally but looking over the changes they make sense to me, thanks for this work!

IanCal (approved):
LGTM if we add flake8 as a dev dependency in each package, otherwise I get errors every time I open a file. Otherwise works great :)

PR #3068, opened 2022-07-07
hub: Change card drop email to use template

backspace:
I chose to add a new `sendTemplate` function even though `send` will now be not used because maybe it’ll be useful again in the future?

I’ve tested this via `yarn console`:

```
Hub > await (await container.lookup('email')).sendTemplate({to: 'buck.doyle@cardstack.com', from: 'no-reply@stack.cards', title: 'a title'}, 'card-drop-email', JSON.stringify({verificationUrl: 'http://example.com'}))
{
  '$metadata': {
    httpStatusCode: 200,
    requestId: 'ac747700-ffd0-4f37-952f-bb47bb5d77fb',
    extendedRequestId: undefined,
    cfId: undefined,
    attempts: 1,
    totalRetryDelay: 0
  },
  MessageId: '01000181d9eec303-1cc84ace-9e13-4dec-a223-9e609f782b4b-000000'
}
```

Verify your identity to receive a prepaid card - buck doyle@cardstack com - Cardstack Mail 2022-07-07 14-05-48

I tried deploying to staging but the banner isn’t showing for me 😞

### DIFF
--- a/packages/hub/services/email.ts
+++ b/packages/hub/services/email.ts
@@ -40,7 +40,7 @@ export default class Email {
     );
   }
 
-  async sendTemplate(email: { to: string; from: string; templateName: string; templateData: any }) {
+  async sendTemplate(email: { to: string; from: string; templateName: string; templateData: Record<string, any> }) {
     return this.client.send(
       new SendEmailCommand({
         Destination: {


packages/hub/services/email.ts
```diff
@@ -39,6 +39,25 @@ export default class Email {
       })
     );
   }
+
+  async sendTemplate(email: { to: string; from: string; templateName: string; templateData: any }) {
```
Aierie, line 43: We could type `templateData` as `Record<string, any>`, I think.

Aierie (approved):

PR #3069, opened 2022-07-08
ci: Fix security groups ids

pcjun97:
Fix the wrong security groups in `waypoint.hcl` that points to security groups created in the wrong VPC. Updated the security groups with the newly created ones from the fix in https://github.com/cardstack/infra/pull/193


IanCal (approved):
LGTM - can confirm the staging extractor is running fine, thanks!

PR #3071, opened 2022-07-08
wip removing autoinjectable services

Aierie:


packages/hub/node-tests/services/merchant-info-test.ts
```diff
@@ -28,9 +28,9 @@ describe('MerchantInfoService', function () {
       }
     }
 
-    registry(this).register('merchant-info', PatchedMerchantInfoService);
+    registry(this).register('merchant-info', PatchedMerchantInfoService, { type: 'service' });
 
-    let subject = await getContainer().lookup('merchant-info');
+    let subject = await getContainer().lookup('merchant-info', { type: 'service' });
```
backspace, line 33: there are build errors here and in many tests that I’m not sure how to resolve… was it wrong to change these from `@cardstack/di` to `@cardstack/hub/services`? 🤔

Aierie: Ya, we need it to be `@cardstack/di`

PR #3072, opened 2022-07-11
Don't rely on the issuer in CreatePrepaidCard to be correct

jurgenwerk:


jurgenwerk:
This is a short term solution for populating the correct issuer and owner of the prepaid cards in the subgraph. Once we deploy https://github.com/cardstack/card-pay-protocol/pull/237 we should implement versioning where after the protocol deploy we start getting the needed data from the event params, and use eth calls only for historical events. This is so that we will be able to improve mapping performance and take advantage of the new subgraph streaming features. 

I'm yet to test this new mapping - still dealing with some issues while running subgraph in forked mode.

jurgenwerk:
Closing in favor of https://github.com/cardstack/cardstack/pull/3080 which includes an upgrade.

PR #3073, opened 2022-07-11
hub: Add Prisma

backspace:
This is a start at using Prisma to reduce boilerplate SQL. It updates code that touches these tables to use the generated Prisma client:
* `prepaid_color_chemes`
* `prepaid_card_customizations`
* `prepaid_card_patterns`
* `push_notification_registrations`

We can proceed with migrating other tables piecemeal. There’s sometimes entanglement because of how `@chax-at/transactional-prisma-testing` sets up a transaction in tests, but most of the time tables can be updated individually.

Unlike the spike in #2672, this does not add the [mapping](https://www.prisma.io/docs/concepts/components/prisma-schema/names-in-underlying-database#mapping-collectiontable-and-fieldcolumn-names) necessary to have Prisma client use camel case instead of snake case. That would require constant upkeep as every `prisma db pull` would clobber what was already there, it didn’t seem worth it.

There are a few unresolved `FIXME`s which I have left comments for.


packages/hub/services/serializers/prepaid-card-pattern-serializer.ts
```diff
@@ -30,20 +25,16 @@ export default class PrepaidCardPatternSerializer {
...
     }
-    let row = queryResult.rows[0];
-    return {
-      id: row['id'],
-      patternUrl: row['pattern_url'],
-      description: row['description'],
-    };
+
+    // FIXME why? If it’s already known to not be null
+    return pattern as unknown as prepaid_card_patterns;
```
backspace, line 37:
```
Type 'prepaid_card_patterns | null' is not assignable to type 'prepaid_card_patterns'.
  Type 'null' is not assignable to type 'prepaid_card_patterns'.
```

Why isn’t this addressed by the return from `if (!pattern)`?

packages/hub/services/prisma-manager.ts
```diff
@@ -0,0 +1,97 @@
...
+    }
+
+    return this.client;
+  }
+
+  async teardown() {
+    this.prismaTestingHelper?.rollbackCurrentTransaction();
+    // FIXME I hoped the below would address this warning but no:
+    // warn(prisma-client) There are already 10 instances of Prisma Client actively running.
+    return this.client?.$disconnect();
```
backspace, line 89: This is discussed [here](https://www.prisma.io/docs/guides/database/troubleshooting-orm/help-articles/nextjs-prisma-client-dev-practices#solution). Should we have some `global` hack for testing? I think it only happens then because the service is constantly being recreated.

lukemelia: I really would like to avoid the global thing. However, it seems that the warning is emitted based on module-scoped data that cannot be reset and actually would be the source of a memory leak itself the way we are using PrismaClient: https://github.com/prisma/prisma/blob/67dabe14b33748d64d4c08a16fa1871836e4cedd/packages/engine-core/src/library/LibraryEngine.ts#L96-L109 🙀

packages/hub/services/prisma-manager.ts
```diff
@@ -0,0 +1,101 @@
...
+  }
+
+  private addConvenienceFunctions(client: PrismaClient) {
```
lukemelia, line 60: Nice name. 💯

packages/hub/services/serializers/push-notification-registration-serializer.ts
```diff
@@ -1,16 +1,16 @@
-import { PushNotificationRegistration } from '../../routes/push_notification_registrations';
 import { JSONAPIDocument } from '../../utils/jsonapi-document';
+import { push_notification_registrations } from '@prisma/client';
 
 export default class PushNotificationRegistrationSerializer {
-  serialize(model: PushNotificationRegistration): JSONAPIDocument {
+  serialize(model: push_notification_registrations): JSONAPIDocument {
```
lukemelia, line 5: It is confusing that a singular model is typed as a plural thing. Not sure if that is just a fact of life we have to deal with using Prisma or if there is an alternative.

lukemelia (approved):

packages/hub/README.md
```diff
@@ -133,7 +133,12 @@ yarn db:migrate create <migration-name>`
 
 Documentation on how to create migration scripts is available at https://salsita.github.io/node-pg-migrate/#/migrations
 
-After you have completed running your new DB migration script create a pg_dump of the DB in the `config/structure.sql` file using:
+After you have completed running your new DB migration script, update Prisma’s schema and client:
+
+    yarn run prisma db pull
+    yarn run prisma generate
+
+And create a pg_dump of the DB in the `config/structure.sql` file using:
```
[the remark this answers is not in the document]
backspace: 👍🏻 that would also be a good place to insert a schema-transforming script if that works out (like [Pal](https://paljs.com/cli/schema#convert-to-camel-case), though that appears to only address fields, not table names). Captured in 4253.

packages/hub/services/prisma-manager.ts
```diff
@@ -0,0 +1,97 @@
+import { Prisma, PrismaClient } from '@prisma/client';
+import config from 'config';
+import { PrismaTestingHelper } from '@chax-at/transactional-prisma-testing';
+
+type push_notification_registrations_getter = Prisma.push_notification_registrationsDelegate<any>;
+// FIXME this is <GlobalRejectSettings> but how to import?
```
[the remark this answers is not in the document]
backspace: My belief: it doesn’t matter.

packages/hub/services/prisma-manager.ts
backspace: I’m sure we can find a way to accomplish this without excessive hacks. It does seem like we should be able to construct the client once in tests and go with it forever. Noted in 4254.

packages/hub/services/serializers/prepaid-card-pattern-serializer.ts
backspace: Extracted to 4255

packages/hub/services/serializers/push-notification-registration-serializer.ts
backspace: yes, this reflects the larger problem of Prisma’s opinion of database naming conventions not matching our conventions. I’ve made [4256](https://linear.app/cardstack/issue/CS-4256/look-at-automated-prisma-schema-transformation-to-convert-between) to track addressing this. It’s what I’ll focus on next because it will affect all future conversions.

PR #3074, opened 2022-07-12
Cs 4159 create did resolver

tintinthong:
The main motivation for this was that previously each reward program will store a rule blob that was a [hex encoded json blob](https://github.com/cardstack/card-pay-protocol/blob/fa93c409be1d65869f4360a5119ee13806978a5f/contracts/RewardManager.sol#L45). But as we increased the size of that json blob by encoding more rules, we hit into a [size limitation on data signed](https://discord.com/channels/584043165066199050/877196997344522270/990908115576848494) by a trezor. 

This PR change enables us to write the hex encoding as a did, so that offchain services can get the did and resolve to document in s3 offchain storage. In a more general sense, did functionality is now available for our 2nd love -- python. 

**Where to start?**
- Here is an [example](https://github.com/cardstack/cardstack/blob/ab021f1942a82ea76d83bb51e418232c6644eff2/packages/python-did-resolver/scripts/resolve.py) on how to use the resolver in python `pdm run resolver`
- Run test `pdm run pytest tests` in the subpackage

**Notable changes**

- **Fork of did-resolver**. In order to complete this, we had to create a complete fork of the [did resolver library](https://github.com/decentralized-identity/did-resolver) but in python. The [PR](https://github.com/cardstack/did-resolver/pull/1) is available here. 
- **Property based test**. We are using property based testing using python-hypothesis, therefore, the test are sampled from a set of defined cases.
- **Compatibality with existing JS subpackage**. This repo is meant to mirror [this](https://github.com/cardstack/cardstack/tree/main/packages/did-resolver). To ensure compatibility with our current implementation, we run test from within python to call a ts-node process that imports functions from javascript. 
- **Wrong short code specification for CardSpace**. CardSpace did had a "cs" short code which is wrong, it is meant to be "c" only. The reason this was not flagged because CardSpace did's were never used to resolve to documents, if it did we would hit _Error: Invalid DID identifier: unknown type "c"_. **I need to make a migration of s3 documents corresponding cardspace**
- **Update CLI add-rule**. Typically, if we want to add a rule to a reward program we will use the cli and a trezor. We have made updates to the CLI so hex encoded reward rule blob -> converted to json and written to s3 -> did created -> hex encoded did written to reward manager
- **Usage of uuidv5**. Since we commit to the use of shortened uuid, I have decided to use uudv5 to hash json form of our rule to obtain a unique id. A constant uuid4 namespace is specified. There is a tricky thing about using `shortuuid` in python because it is incompatible to it's javascript counter part due to sorting of alphabets; I eventually patched the library [here](https://github.com/cardstack/cardstack/blob/ab021f1942a82ea76d83bb51e418232c6644eff2/packages/python-did-resolver/python_did_resolver/main.py#L13-L19). 

Note: I intentionally didn't use ipfs cid into the uid of the cardstack method identifier, if we want to generate it we'd need to use params configured of the ipfs node (which is variable depending on the node that generates the hash) and [ipfs-only-hash](https://github.com/alanshaw/ipfs-only-hash) seemed incompatible with the FLICKRBASE58 shortcode. So right now its only s3.


packages/did-resolver/index.ts
```diff
@@ -103,18 +109,30 @@ function isFlickrBase58(candidate: string) {
...
+}
+
 export function parseIdentifier(identifier: string): CardstackIdentifier {
   let data = identifier.substring(0, identifier.length - HASH_LENGTH);
   let checksum = identifier.substring(identifier.length - HASH_LENGTH, identifier.length);
   if (hashFunc(data) !== checksum) {
     throw new Error('Invalid DID identifier: checksum failed');
   }
-  let version = versionFromChar(data[0]);
-  let type = SHORT_TYPE_TO_DID_TYPE[data[1]];
```
tintinthong, line 113: @lukemelia is this a strict restriction on the cardstack method identifier. My property test failed because there was ["cs" short type](https://github.com/cardstack/cardstack/blob/3590a2666d4c7f3f3663cf36fdb81392cb050b1a/packages/did-resolver/index.ts#L17), which is two characters long. 

Initially, I thought this was a mistake but then I realise you specify within the README that this short-type is strictly one character

tintinthong: cc @jurgenwerk

tintinthong: I am going with the direction of using single character to represent our did type (That means migrating  all s3 docs which have did wrong)

For future reference. Test on python fail

```
Error: Invalid DID identifier: unknown type "c"
```

IanCal:
> Wrong short code specification for CardSpace. CardSpace did had a "cs" short code which is wrong, it is meant to be "c" only. The reason this was not flagged because CardSpace did's were never used to resolve to documents, if it did we would hit Error: Invalid DID identifier: unknown type "c". I need to make a migration of s3 documents corresponding cardspace

Do we need to migrate the s3 docs? The short code should return the longer name which is then used for the folder, but the file names use the contents after the first two characters IIUC.

We'd need to replace all the ids right? As they currently are invalid DIDs?

tintinthong:
> > Wrong short code specification for CardSpace. CardSpace did had a "cs" short code which is wrong, it is meant to be "c" only. The reason this was not flagged because CardSpace did's were never used to resolve to documents, if it did we would hit Error: Invalid DID identifier: unknown type "c". I need to make a migration of s3 documents corresponding cardspace
> 
> Do we need to migrate the s3 docs? The short code should return the longer name which is then used for the folder, but the file names use the contents after the first two characters IIUC.
> 
> We'd need to replace all the ids right? As they currently are invalid DIDs?

Hmmm. I think maybe im not very clear on what migration meant. Migration meant that I needed to change a "did" key inside all s3 documents (in s3 I guess this means rewriting the documents).

Here is an example

Right now 

Look at https://storage.cardstack.com/card-space/12LLwAK2fsp3pHv1ZCddTP.json 
- 12LLwAK2fsp3pHv1ZCddTP is the uid and the folder name
- The did **INISIDE** of the document says did:cardstack:1cs12LLwAK2fsp3pHv1ZCddTP852cec1573ddd86c. This does not seem to be in use anywhere tho

The did should actually be did:cardstack:1c12LLwAK2fsp3pHv1ZCddTP416d12bda3f43f00. Because this uid is randomly generated we can continue to use it (it is just a uuid which is shortened). All we have to do is change the did **INISIDE** of the document because the did corresponds to the uid given, the checksum (852cec1573ddd86c) and the shortcode (cs) have changed.

jurgenwerk:
@tintinthong @IanCal I think we aren't saving cardspaces to S3 anymore. I think we don't wanna bother with the migration of s3 documents.

IanCal:
When we say the dids aren't being used, they aren't stored on chain or in any databases? I can see that if we were ever trying to resolve them we'd hit an error.

>  All we have to do is change the did INISIDE of the document because the did corresponds to the uid given; the checksum (852cec1573ddd86c) and the shortcode (cs) have changed.

Ah I see why there's a need to change inside the files now.

Given that the existing ones didn't work I'm happy to say it's outside the scope of this change.

I'd say more broadly we make sure we don't have invalid dids. Either migrating them, or if they're in no external data then possibly just removing the files. Not a blocker for this PR though.

tintinthong:
> When we say the dids aren't being used, they aren't stored on chain or in any databases?

From what I understand the did's of card space are not stored but are encoded into the document in s3. I have just moved the ticket of migration (cs-4302) into next sprint as their not-related to this PR and worth discussing.

jurgenwerk:
@tintinthong from what I can see they aren't even written to S3. They were in the past but we are not doing that anymore. So I'm leaning towards @IanCal's suggestion to maybe just delete those.

tintinthong:
> @tintinthong from what I can see they aren't even written to S3. They were in the past but we are not doing that anymore. So I'm leaning towards @IanCal's suggestion to maybe just delete those.

Hi @jurgenwerk can u clarify "delete those" to mean 
1. delete entire /cardspace folder of s3 documents 
2. just the did key in each document

jurgenwerk:
Entire card space folder. Would be good to get @lukemelia's opinion too.

IanCal (approved):
I think this looks good reviewing the code, is there anything  you want me to run locally and test?

tintinthong:
> I think this looks good reviewing the code, is there anything you want me to run locally and test?

nope all should be good.

PR #3076, opened 2022-07-13
improve safes sentry logging

Aierie:
Improve logging of application state so that we know more about the problem described in CS-3429


packages/web-client/app/resources/safes.ts
```diff
@@ -195,8 +196,27 @@ export class Safes extends Resource<Args> {
...
+        selectedSafe = individualUpdate.safe;
       } else {
-        selectedSafe = graphDataSafe ?? individualUpdate.safe;
+        let e = new Error(
+          `Safe with address "${address}" was not found in graph data or individual update data`
+        );
+
+        Sentry.withScope((scope) => {
+          // we want to know if there are actually any addresses available
+          scope.setExtras({
```
Aierie, line 210: This is deprecated but still works, and allows a bigger payload (16kb) than the alternative (8kb)

backspace (approved):

PR #3077, opened 2022-07-13
hub: Add text-color as required merchant attribute

backspace:
This is a required field in the database. There are no `POST /api/merchant-infos` tests exercising this but I added it there anyway since it would cause the same failure.

Without this, a `POST /api/profile-purchases` without `text-color` for the merchant produced a 500, which was only intelligible by looking at Sentry or Hub logs. I think we should consider catching 500s and attaching Postgres errors to the response so the problem is more immediately obvious and doesn’t require looking elsewhere. That way when @oidoug encountered this, it would have been clear that adding `text-color` would allow the exercise to continue.

But is there a danger of exposing information that shouldn’t be known? Maybe it could be staging-only 🤔 thoughts @Aierie?

### DIFF
--- a/packages/hub/node-tests/routes/profile-purchases-test.ts
+++ b/packages/hub/node-tests/routes/profile-purchases-test.ts
@@ -416,6 +416,11 @@ describe('POST /api/profile-purchases', function () {
             title: 'Missing required attribute: color',
             detail: 'Required field color was not provided',
           },
+          {
+            status: '422',
+            title: 'Missing required attribute: text-color',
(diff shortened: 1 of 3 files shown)


backspace:
> Numeric error codes are a possible solution if we are concerned about exposing internal details.

How would that work in this case? I’m thinking we could attach the entire Postgres error in the `meta` of the response, so it’s difficult to know in advance what the errors actually are. It’s probably not going to leak anything but it’s a bit scary to me 😯

PR #3078, opened 2022-07-13
hub: Fix merchant creation query parameter name

backspace:
This was revealed by the failure of `create-profile`:

```
Error: Could not register profile card v2 for customer 0x323B2318F35c6b31113342830204335Dac715AA8, did did:cardstack:1m7w3wcQJvzCqjfrTxPvZiJi3bca2c583012aa35, received 400 from relay server: {\"infoDid\":[\"This field is required.\"]}
```

The relay server expects a `POST` body like this:

```
{"owner":"0xexample","infoDid":"did:etc"}
```

I updated this to use HTTP mocks to better assert on that format and expected responses.


backspace:
I deployed to staging to actually try this and found that it’s still broken, it needs to be in the `POST` body, not query parameters 😬

lukemelia (approved):

PR #3081, opened 2022-07-14
Fix an eslint issue in web-client

lukemelia:


backspace:
Is this an expected diff? 😯 
Cardstack - Percy | Visual testing as a service 2022-07-14 09-42-45

lukemelia:
> Is this an expected diff? 😯

I had something similar in the previous PR that I ended up approving. Maybe it was a symptom of what was fixed in this PR, and this percy diff reflects it.

PR #3082, opened 2022-07-14
hub: Change Prisma client naming conventions

backspace:
This adds a new `db transform-prisma-schema` command that runs an adapted version of [this script](https://github.com/prisma/prisma/discussions/2530#discussioncomment-17535) to change names to camel case so we don’t have a chaos of snake case mixed in. It updates Prisma Client calls to use the converted names.

A good way to exercise this (from `packages/hub`):

```
git checkout 76d66f1 prisma/schema.prisma
node dist/hub.js db transform-prisma-schema
yarn run prisma format
```

This runs the transformation on the stock schema in `main`. After these commands, there should be no Git diff.

Another thing to try is to simulate a migration and schema update:

```
psql hub_development -c "ALTER TABLE job_tickets ADD COLUMN test_test_test jsonb;"
HUB_DATABASE_URL=postgres://postgres:postgres@localhost:5432/hub_development yarn run prisma db pull
node dist/hub.js db transform-prisma-schema
yarn run prisma format
```

This should show the schema with the new field and nothing else.


packages/hub/README.md
```diff
@@ -136,7 +136,10 @@ Documentation on how to create migration scripts is available at https://salsita
 After you have completed running your new DB migration script, update Prisma’s schema and client:
 
     yarn run prisma db pull
+    node dist/hub.js db transform-prisma-schema
+    yarn run prisma prisma format
     yarn run prisma generate
+    yarn add @prisma/client
```
backspace, line 142: 4253 exists for wrapping this all up 😬

packages/hub/README.md
```diff
@@ -136,7 +136,10 @@ Documentation on how to create migration scripts is available at https://salsita
 After you have completed running your new DB migration script, update Prisma’s schema and client:
 
     yarn run prisma db pull
+    node dist/hub.js db transform-prisma-schema
+    yarn run prisma prisma format
     yarn run prisma generate
```
Aierie, line 141: Since skipping any one of these will result in invalid code or types, it feels like they should all be bundled into one command

packages/hub/cli/db/transform-prisma-schema.ts
```diff
@@ -0,0 +1,161 @@
...
+    // Changes fields listed in @@index or @@unique to camel case
+    const indexUniqueFieldsMatch = fixedLine.match(/@@\w+\(\[([\w,\s]+)\]/);
+    if (indexUniqueFieldsMatch) {
+      const fields = indexUniqueFieldsMatch[1];
+      fixedLine = fixedLine.replace(fields, fixFieldsArrayString(fields));
+    }
+
+    fixedText.push(fixedLine);
+  }
+
```
Aierie, line 157: We can use `getDMMF` from `@prisma/internals` to assert that the data model matches what we expect.

Aierie:
Would https://paljs.com/cli/schema/ work for us?

packages/hub/README.md
backspace: ya will do that in 4253 🤓

PR #3084, opened 2022-07-14
Use prisma for job_tickets queries

lukemelia:


packages/hub/routes/job-tickets.ts
```diff
@@ -131,18 +119,20 @@ export default class JobTicketsRoute {
...
-      id: shortUUID.uuid(),
-      jobType: jobTicketToRetry.jobType,
-      ownerAddress: jobTicketToRetry.ownerAddress,
-      payload: jobTicketToRetry.payload,
-      result: null,
-      spec: jobTicketToRetry.spec,
-    };
-
-    newJobTicket = await this.jobTicketsQueries.insert(newJobTicket);
+    let taskName = jobTicketToRetry.job_type as unknown as keyof KnownTasks;
```
lukemelia, line 122: Is there a way to type this more strongly?

backspace:
I think it’s worth holding off on these conversions until #3082 is done, since it changes the names of the models and fields 😯

packages/hub/routes/job-tickets.ts
backspace: I can’t think of a good way, it would have to be guaranteed that when we insert a ticket the type is a known task type 🤔

backspace (approved):
The camelCase change is merged so this needs updating, but otherwise looks good, thanks!

packages/hub/routes/job-tickets.ts
backspace: runtime-only though, right? Better than nothing!

Aierie: it's a runtime check, and also removes the necessity of doing type assertions (`as Type`):

https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates

packages/hub/node-tests/routes/profile-purchases-test.ts
```diff
@@ -143,7 +145,7 @@ describe('POST /api/profile-purchases', function () {
     let cardSpaceRecord = (await cardSpacesQueries.query({ merchantId }))[0];
     expect(cardSpaceRecord).to.exist;
 
-    let jobTicketRecord = await jobTicketsQueries.find({ id: jobTicketId! });
+    let jobTicketRecord = await prisma.jobTicket.findUnique({ where: { id: jobTicketId! } });
```
backspace, line 148: I’m surprised to find that `prisma.jobTicket` and `prisma.JobTicket` both work here 🤔 maybe we should make only one of them work, but if we do, the build will fail so it shouldn’t hard to update everything as needed

backspace (approved):

PR #3087, opened 2022-07-15
hub: Fix Prisma in deployed environments

backspace:
This updates the Webpack and Docker builds to include the files needed to use Prisma.

With this deployed, Prisma works as expected:

```
curl \
 -s -S -H "Accept: application/vnd.api+json" \
 -H "Accept-Encoding: gzip" \
 -H "Content-Type: application/vnd.api+json" \
 "https://hub-staging.stack.cards/api/prepaid-card-patterns"
{"data":[{"id":"3b8b436b-a99e-4346-ab20-3ba5f9963321","type":"prepaid-card-patterns","attributes":{"pattern- …
```

### DIFF
--- a/packages/hub/Dockerfile
+++ b/packages/hub/Dockerfile
@@ -9,11 +9,10 @@ COPY dist/sql node_modules/graphile-worker/sql
 COPY dist/services packages/hub/services
 
 # Copy Prisma necessities
-COPY dist/@prisma/client node_modules/.prisma/client
 COPY dist/@prisma/client node_modules/@prisma/client
 COPY prisma/schema.prisma node_modules/.prisma/client/schema.prisma
 COPY prisma/schema.prisma node_modules/@prisma/client/schema.prisma
-COPY dist/.prisma/client/ node_modules/@prisma/client/
+COPY dist/.prisma/client/ node_modules/.prisma/client/
 
 # Replicate enough of the node_modules structure to support node-pg-migrate
 COPY dist/node-pg-migrate node_modules/node-pg-migrate
--- a/packages/hub/Dockerfile
+++ b/packages/hub/Dockerfile
@@ -8,6 +8,8 @@ COPY dist/sql node_modules/graphile-worker/sql
 # This includes the discord bot image assets that are hosted by the hub
 COPY dist/services packages/hub/services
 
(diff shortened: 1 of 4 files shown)


backspace:
Maybe we need something like [this](https://gist.github.com/malteneuss/a7fafae22ea81e778654f72c16fe58d3#file-dockerfile-L16). But I don’t know if people build locally and that would cause problems

packages/hub/Dockerfile
```diff
@@ -8,6 +8,13 @@ COPY dist/sql node_modules/graphile-worker/sql
 # This includes the discord bot image assets that are hosted by the hub
 COPY dist/services packages/hub/services
 
+# Copy Prisma necessities
+COPY dist/@prisma/client node_modules/.prisma/client
```
Aierie, line 12: Is this a typo? If so we might not need the subsequent COPY instructions for `dist/.prisma/<somefile>`?

```suggestion
COPY dist/.prisma/client node_modules/.prisma/client
```

packages/hub/Dockerfile
```diff
@@ -8,6 +8,13 @@ COPY dist/sql node_modules/graphile-worker/sql
...
+COPY dist/@prisma/client node_modules/@prisma/client
+COPY prisma/schema.prisma node_modules/.prisma/client/schema.prisma
+COPY prisma/schema.prisma node_modules/@prisma/client/schema.prisma
+COPY dist/.prisma/client/*.node node_modules/@prisma/client/
```
Aierie, line 16: Is this a typo? If so we might not need the subsequent COPY instructions for `dist/.prisma/<somefile>`, or for `schema.prisma`? My understanding is `@prisma/client` requires `.prisma/client`, so as long as we copy both of them into `node_modules` we should be ok. This is the code from `@prisma/client/index.js`:

```js
module.exports = {
  ...require('.prisma/client/index'),
}
```

```suggestion
COPY dist/.prisma/client node_modules/.prisma/client
COPY dist/@prisma/client node_modules/@prisma/client
```

Aierie: Also just realised we probably don't need to `yarn add @prisma/client` as a final step, as specified in the readme.

lukemelia (approved):

PR #3090, opened 2022-07-18
hub: Update exchange-rates to use Prisma

backspace:


packages/hub/services/prisma-extensions/exchange-rate.ts
```diff
@@ -0,0 +1,33 @@
...
+}
+
+export function getExchangeRateExtension(client: PrismaClient) {
+  return {
+    async select(
+      from: string,
+      to: string[],
+      date: string,
+      exchange: string
+    ): Promise<CryptoCompareConversionBlock | null> {
```
lukemelia, line 17: I think this would be more readable with a named params style argument:

```ts
    let cachedValue = await prisma.exchangeRate.select({
        from: 'BTC',
        to: ['USD'],
        date: '2022-06-01',
        exchange: DEFAULT_CRYPTOCOMPARE_EXCHANGE
    });
```

lukemelia (approved):

PR #3092, opened 2022-07-19
hub: Fix Prisma warning about excess instances in tests

backspace:
This updates the Prisma service to construct `PrismaClient` when the service is loaded instead of instantiated. This fixes a warning discussed [here](https://github.com/prisma/prisma/discussions/4399) and explained [here](https://www.prisma.io/docs/concepts/components/prisma-client/working-with-prismaclient/instantiate-prisma-client#use-a-single-shared-instance-of-prismaclient) in the documentation. The previous implementation would construct a new Prisma client for each test, which would quickly exceed the database connexion limit.

[Here](https://github.com/cardstack/cardstack/runs/7413985693?check_suite_focus=true#step:9:70)’s an example of the warning in `main`. One cannot link to an absence, but in [this job log](https://github.com/cardstack/cardstack/runs/7436119671?check_suite_focus=true) you cannot find that string.


packages/hub/services/prisma-manager.ts
```diff
@@ -7,51 +7,58 @@ import {
...
+let singletonClient = new PrismaClient({
+  datasources: { db: { url: dbConfig.url } },
+  log: dbConfig.prismaLog,
+});
+
+// In tests, do not add extensions until the in-transaction proxy client exists
+
+if (!dbConfig.useTransactionalRollbacks) {
+  addCardstackPrismaExtensions(singletonClient);
+}
```
Aierie, line 31: Should this be in the `ready` hook for the service instead?

backspace: sure, done in 56ee76c, thanks

PR #3094, opened 2022-07-20
await prisma promise in prepaid card pattern serializer

Aierie:
lint rule to follow later


backspace (approved):
Thanks for figuring this out! 😵‍💫

PR #3106, opened 2022-07-22
move deprecation workflow config to specified configPath

Aierie:
If in the incorrect location, our deprecation workflow config will not be read by the addon.


lukemelia:
Whoops, thanks!

lukemelia (approved):

PR #3107, opened 2022-07-22
hub: Change card drop state to use Prisma

backspace:
This is otherwise unremarkable, but it introduces a slight change to the extension pattern, as I noticed while creating this extension (copied from `LatestEventBlock`) that Typescript was unable to enforce that the extension class actually implemented the declared interface. So I had `rateLimited: string` in the function implementation despite the interface saying it would be `rateLimited: boolean`. I think it’s preferable to have it more explicit so any discrepancies result in a build failure:


lukemelia (approved):
I like this. I was bit by a similar mistake. Can we apply this to the other extensions in this PR?

PR #3108, opened 2022-07-25
Fix the mapping to use the correct owner and issuer 

jurgenwerk:
This is a stripped down version of https://github.com/cardstack/cardstack/pull/3080 - it includes the mapping fix only, without the upgrade. The mentioned PR is blocked until https://github.com/graphprotocol/graph-node/issues/3747 is done. We don't know when the graph node will get the new release that we need, so let's proceed with the adjusted mapping only because it's a blocker for email card drop feature.

TODO:

- [x] Deploy to staging blue node, rebuild the graph, test, switch active deployment green to blue
- [x] Deploy to production blue node, rebuild the graph, test, switch active deployment green to blue

### DIFF
--- a/packages/cardpay-subgraph/src/mappings/prepaid-card.ts
+++ b/packages/cardpay-subgraph/src/mappings/prepaid-card.ts
@@ -28,7 +28,13 @@ import { log } from '@graphprotocol/graph-ts';
 
 export function handleCreatePrepaidCard(event: CreatePrepaidCard): void {
   let prepaidCard = toChecksumAddress(event.params.card);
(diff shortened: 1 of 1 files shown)


jurgenwerk:
Rebuilt the graph on sokol, and did the switch (green to blue). All looks fine, the data in wallet seems to load ok. 

I have also deployed this to the inactive node (blue) to production. I'm waiting for it to sync (1-3 hours left) and then we can switch. The accompanying infra PR for switching the live graph instance will follow.

habdelra (approved):

jurgenwerk:
Tested this on staging and production, works ok.

PR #3109, opened 2022-07-25
ember: Fix routing.transition-methods deprecations

backspace:
This is mostly a mechanical replacement of `transitionTo`/`transitionToRoute` with `this.router.transitionTo`. It also applies the same fix to the Boxel “dummy” application, which wasn’t uncovered through tests.

There’s a new test helper in `web-client` called `visitWithQueryFix` to bypass failures described in [this issue](https://github.com/emberjs/ember.js/issues/17118). A local example of this shows in [this job](https://github.com/cardstack/cardstack/runs/7508755707?check_suite_focus=true#step:8:100). The workaround is to `visit('/')` before some `visit` calls with query parameters. The failure does not happen in _all_ `visit` calls with query parameters for unknown reasons. This is a test-only issue; you can confirm this by [directly visiting a workflow](https://emberroutingtransition-methods-cs-4282.web-client-preview.stack.cards/card-pay/wallet?flow=issue-prepaid-card) on a preview deployment.


packages/web-client/app/routes/card-pay/base.ts
```diff
@@ -23,7 +28,7 @@ export default class CardPayTabBaseRoute extends Route {
         return;
 
       transition.abort();
-      this.transitionTo(this.routeName, {
+      this.router.transitionTo(this.routeName, {
```
backspace, line 31: This is doing nothing…!

lukemelia (approved):

backspace:
I believe the remaining failures are an example of [this](https://github.com/emberjs/ember.js/issues/17118). Will look into the fixes suggested there later.

backspace:
@lukemelia see the edited initial comment for updates on a workaround for test failures.

lukemelia (approved):

PR #3111, opened 2022-07-25
Eliminate ensure-safe-component.string deprecation from boxel

lukemelia:


packages/boxel/tests/dummy/config/deprecation-workflow.js
```diff
@@ -8,7 +8,7 @@ self.deprecationWorkflow.config = {
     { handler: 'silence', matchId: 'ember-modifier.function-based-options' },
     { handler: 'silence', matchId: 'ember-modifier.no-args-property' },
     { handler: 'silence', matchId: 'ember-modifier.no-element-property' },
-    { handler: 'silence', matchId: 'ensure-safe-component.string' },
```
backspace, line 11: I was removing the lines in my PRs, should I be using `throw` instead? 😯

backspace (approved):

packages/boxel/tests/dummy/config/deprecation-workflow.js
lukemelia: I think so. It will prevent us from backsliding.

PR #3112, opened 2022-07-26
Ssr web sentry fix

Aierie:
SSR web's Sentry has not been working for a while. This attempts to fix the client side issues - it includes [this change](https://github.com/mansona/ember-get-config/commit/4b6b103683008e8f27f9fc9c04ba14ae41a77004) in ember-get-config which fixes a bug where we overwrite Sentry's `enabled` configuration as `false`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -91,7 +91,8 @@
   },
   "resolutions": {
     "ember-a11y-testing": "4.0.3",
-    "axe-core": "4.1.4"
+    "axe-core": "4.1.4",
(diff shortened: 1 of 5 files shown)


backspace (approved):
We tried deploying this to staging and it successfully reported an error. Strangely, `/client-boom` seems to not work anymore, but visiting a known non-existent Card Space (`jortleby.pouty.pizza` for instance) does show an error in Sentry.

PR #3114, opened 2022-07-27
hub: Move drop requests and more to Prisma

backspace:
This changes `email_card_drop_requests` code to use Prisma. Interval calculations being used in the now-removed `queries` file were mysteriously incorrect (time zones?) so I switched them to use purely-numeric comparison using `epoch`.

This also removes the now-empty `queries` directory and the `WHERE` clause utilities formerly used there. There are a some `db.query` calls remaining in the codebase in parts of it I’m unfamiliar with (and `RemoveOldSentNotifications`, which had similar timezone problems), but this should be fine indefinitely.


lukemelia (approved):
Thanks for getting the stragglers as part of this! 👍

PR #3118, opened 2022-07-27
hub: Change Sentry.init to only be called once

backspace:
This fixes a memory leak in tests caused by Sentry.init
being called for every test. There’s no reason to call
it more than once in one process.

See [here](https://github.com/cardstack/cardstack/runs/7547445989?check_suite_focus=true#step:9:808) for a pre-fix example. It doesn’t exist in [this job](https://github.com/cardstack/cardstack/runs/7549271955?check_suite_focus=true).

### DIFF
--- a/packages/hub/node-tests/helpers/sentry.ts
+++ b/packages/hub/node-tests/helpers/sentry.ts
@@ -6,15 +6,22 @@ import { Suite } from 'mocha';
 const { testkit, sentryTransport } = sentryTestkit();
 const DUMMY_DSN = 'https://acacaeaccacacacabcaacdacdacadaca@sentry.io/000001';
 
+let sentryNeedsInit = true;
+
 export function setupSentry(context: Suite) {
-  context.beforeAll(function () {
-    Sentry.init({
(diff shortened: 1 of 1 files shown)


Aierie (approved):
🙏 🧹 

There seems to be some [way to use Sentry to check if Sentry is already initialized](https://github.com/getsentry/sentry-javascript/pull/3860/files#diff-0942a26140e4282a046cc2e4bde7cb0ea0170e49b0a52a3f6f3873e4b10367ceR45), but this relies on `@sentry/core`, which they've labeled as internal.

backspace:
> There seems to be some [way to use Sentry to check if Sentry is already initialized](https://github.com/getsentry/sentry-javascript/pull/3860/files#diff-0942a26140e4282a046cc2e4bde7cb0ea0170e49b0a52a3f6f3873e4b10367ceR45), but this relies on `@sentry/core`, which they've labeled as internal... though it's reexported in `@sentry/node`, hmm.

heh Next seems to operate similarly, there’s specific documentation for it with the parallel Prisma problem

PR #3120, opened 2022-07-28
hub: Replace card_spaces/merchant_infos with profiles

backspace:
This is mostly a mechanical process of combining the creation of separate models into a single profile. Since the surviving fields on `card_spaces` were all optional and it had a foreign key for merchants, it’s as if it merged into `merchant_infos`.

This preserves the existing endpoints at `/api/card-spaces` and `/api/merchant-infos` so we don’t have to update all clients simultaneously. Then we can add `/api/profiles` and eventually remove those once everything has migrated. I did delete `POST /api/card-spaces` because it’s no longer applicable.

### DIFF
--- a/packages/hub/config/structure.sql
+++ b/packages/hub/config/structure.sql
@@ -823,22 +823,6 @@ CREATE TABLE public.card_drop_recipients (
 ALTER TABLE public.card_drop_recipients OWNER TO postgres;
 
 --
--- Name: card_spaces; Type: TABLE; Schema: public; Owner: postgres
---
-
-CREATE TABLE public.card_spaces (
-    id uuid NOT NULL,
-    profile_image_url text,
-    profile_description text,
-    created_at timestamp without time zone DEFAULT CURRENT_TIMESTAMP NOT NULL,
-    links json[] DEFAULT '{}'::json[] NOT NULL,
-    merchant_id uuid NOT NULL
-);
-
-
-ALTER TABLE public.card_spaces OWNER TO postgres;
-
---
 -- Name: cards; Type: TABLE; Schema: public; Owner: postgres
 --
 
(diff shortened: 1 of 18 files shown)


packages/hub/config/structure.sql
```diff
@@ -1095,6 +1062,26 @@ CREATE TABLE public.prepaid_card_patterns (
...
 ALTER TABLE public.prepaid_card_patterns OWNER TO postgres;
 
+--
+-- Name: profiles; Type: TABLE; Schema: public; Owner: postgres
+--
+
+CREATE TABLE public.profiles (
+    id uuid NOT NULL,
+    name text NOT NULL,
+    slug text NOT NULL,
```
jurgenwerk, line 1072: We're querying by slug so would be good to add an index

packages/hub/db/migrations/20220728144935996_add-profiles.ts
```diff
@@ -0,0 +1,81 @@
...
+    INSERT INTO profiles
+    SELECT
+      m.id, m.name, m.slug, m.color, m.text_color, m.owner_address,
+      c.links, c.profile_image_url, c.profile_description,
+      m.created_at
+    FROM merchant_infos as m
+    LEFT JOIN card_spaces AS c ON m.id = c.merchant_id
+  `);
+
+  await pgm.db.query('DROP TABLE card_spaces');
```
jurgenwerk, line 43: I'm extra paranoid about dropping tables in production before verifying the migration + new code works ok 🫥

jurgenwerk (approved):
Looking good!

packages/hub/config/structure.sql
backspace: eep ya, I missed [that](https://github.com/cardstack/cardstack/pull/3120/files#diff-e03f308a75e917163fbf9fe5bc34d68936b918230eea5447cb6f580bf7af8c70L1484), thanks! Included in 1361d9d

packages/hub/db/migrations/20220728144935996_add-profiles.ts
backspace: yes, it’s scary! I ran this many times locally to refine it. I’ll be careful about preserving everything before running it in production.

PR #3121, opened 2022-07-28
hub: Add GET /api/job-tickets

backspace:
This returns all tickets for the EOA associated with the auth token.

### DIFF
--- a/packages/hub/routes/job-tickets.ts
+++ b/packages/hub/routes/job-tickets.ts
@@ -17,6 +17,19 @@ export default class JobTicketsRoute {
     autoBind(this);
   }
 
+  async list(ctx: Koa.Context) {
+    if (!ensureLoggedIn(ctx)) {
+      return;
+    }
+
+    let prisma = await this.prismaManager.getClient();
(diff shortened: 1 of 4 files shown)


danibonilha (approved):
Nice!

packages/hub/routes/job-tickets.ts
```diff
@@ -17,6 +17,19 @@ export default class JobTicketsRoute {
     autoBind(this);
   }
 
+  async list(ctx: Koa.Context) {
+    if (!ensureLoggedIn(ctx)) {
+      return;
+    }
+
+    let prisma = await this.prismaManager.getClient();
+    let jobTickets = await prisma.jobTicket.findMany({ where: { ownerAddress: ctx.state.userAddress } });
```
lukemelia, line 26: Do we need any guarantees of order here? And what happens if a user has hundreds of job tickets?

jurgenwerk (approved):

packages/hub/routes/job-tickets.ts
backspace: This is definitely the most naïve possible implementation, we can refine as needed 😀

PR #3123, opened 2022-07-29
Eliminate deprecated-run-loop-and-computed-dot-access and upgrade Sentry

lukemelia:
- nohoist for wc-provider prevents its transitive dep on older sentry version from wreaking havoc
- Also fixes CS-4280 and CS-4281
- And eliminates argument-less-helper-paren-less-invocation deprecation from web-client


packages/web-client/package.json
```diff
@@ -107,7 +107,7 @@
     "ember-concurrency-decorators": "^2.0.3",
     "ember-concurrency-ts": "^0.2.2",
     "ember-css-url": "v1.0.0-alpha.0",
-    "ember-data": "~3.25.0",
+    "ember-data": "~3.28.0",
```
backspace, line 110: Should we just remove this?

backspace (approved):

packages/web-client/package.json
lukemelia: Yes! :D

PR #3125, opened 2022-07-29
Add resolver to scheduler

tintinthong:


packages/cardpay-reward-scheduler/reward_scheduler/reward_program.py
```diff
@@ -122,9 +124,10 @@ def get_all_payment_cycles(self, rule):
         return set(range(start_block, end_block, payment_cycle_length))
 
     def get_rules(self):
-        rule_blob = self.reward_manager.caller.rule(self.reward_program_id)
-        if rule_blob:
-            rules = json.loads(rule_blob)
+        did_blob = self.reward_manager.caller.rule(self.reward_program_id)
+        if did_blob and did_blob != b"":
```
IanCal, line 128: Am ok with either approach, but either we need to deploy this and swap over the current rule blob at the same time to minimise errors - or we can try and parse as a did and fall back to json.

tintinthong: Yea. I think fallback is better now I think about it. Lets keep it backward compatible.

tintinthong:
@IanCal I made the blob decoding backward compatible

tintinthong:
> @IanCal I made the blob decoding backward compatible

I swapped to prefer did resolution over json

PR #3127, opened 2022-08-01
Reward root submitter lambda

IanCal:
This converts the reward submitter from something that scans over s3 to one that is triggered on a file being uploaded to S3 (this config exists in the infra side, pr 205, which is already deployed)

Supports:

* Manual execution
* Lambda execution
* Sentry
* GH action deploys


packages/reward-root-submitter/README.md
```diff
@@ -78,11 +72,15 @@ You need to build and then copy from the subdirectory `artifacts` all of the non
...
+
+     waypoint build -app=reward-submit-lambda
+
+Then updating the function in lambda (this also waits for the new one to be live)
 
-     waypoint up -app=reward-submit -prune-retain=0
+     aws lambda update-function-code --function-name reward_root_submitter --image-uri $(aws lambda get-function --function-name reward_root_submitter | jq -r '.Code.ImageUri') && time aws lambda wait function-updated --function-name reward_root_submitter
 
-This will remove the previous deployment, build the docker image and deploy a new service.
+These steps are combined in the github
```
lukemelia, line 83:
```suggestion
These steps are combined in the github action
```

packages/reward-root-submitter/reward_root_submitter/config.py
```diff
@@ -0,0 +1,39 @@
...
+
+
+class Config(BaseSettings):
+
+    environment: str
+    evm_full_node_url: str
+    reward_root_submitter_address: str
+    reward_root_submitter_private_key: str
+    reward_root_submitter_sentry_dsn: str
+    log_level: str = Field("WARNING")
```
tintinthong, line 26: By [CS-4182](https://linear.app/cardstack/issue/CS-4182/root-submitter-no-logs), do we want to make the defaults into "INFO" or make a logging class which is more informative

packages/reward-root-submitter/README.md
```diff
@@ -28,42 +28,36 @@ To run test, run
...
+
+The lambda can be triggered manually in the AWS UI, and the code can be run from your local machine.
 
-    docker build -t reward_root_submitter .
+Make sure you are in the correct AWS profile for the required environment and run the manual submitter endpoint with a file on S3.
 
-And run the image with
+For example
+
+    ENVIRONMENT=staging python -m reward_root_submitter.manual  s3://cardpay-staging-reward-programs/rewardProgramID=0x5E4E148baae93424B969a0Ea67FF54c315248BbA/paymentCycle=27071744/results.parquet
```
tintinthong, line 60: Can you document information on the event triggers the lambda (in the context of our services)?

packages/reward-root-submitter/reward_root_submitter/main.py
```diff
@@ -1,58 +1,97 @@
...
+        r"rewardProgramID=([^/]*)", str(reward_output_filename), 1
     )
+    if not Web3().isChecksumAddress(reward_program_id):
+        raise Exception(
+            f"{reward_output_filename} does not have a valid checksummed address for the reward program ID"
+        )
+    payment_cycle = safe_regex_group_search(
+        r"paymentCycle=(\d*)", str(reward_output_filename), 1
+    )
+    if not (payment_cycle or "").isdigit():
```
tintinthong, line 54: This particular if condition is confusing. 

```
payment_cycle = "1"
not (payment_cycle or "").isdigit()
```

yields false 

Is that meant to be?

packages/reward-root-submitter/reward_root_submitter/main.py
```diff
@@ -1,58 +1,97 @@
...
+            if HexBytes(reward_program_id) != HexBytes(first_row["rewardProgramID"]):
+                raise Exception(
+                    f"{reward_output_filename} reward program ID in path and in the file do not match"
+                )
+            if payment_cycle != str(first_row["paymentCycle"]):
+                raise Exception(
+                    f"{reward_output_filename} payment cycle in path and in the file do not match"
+                )
+        except StopIteration:
+            root = EMPTY_MARKER_HEX
```
tintinthong, line 74: Just wanna check theres no data issues here. 

- We do create EMPTY parquet files right? 
- All EMPTY parquet files are meant to have an EMPTY_MARKER_HEX written to the blockchain?

tintinthong (approved):
I left some comments. These changes are nicer and more simplifying (abides by the "do one thing really well" thinking). LGTM!

tintinthong:
It also occured to me that we can piggy back on matics and fadhlans local protocol work. As long as we make the subgraph extractor point to their local protocol, we can then have a docker up for reward services too.

packages/reward-root-submitter/reward_root_submitter/main.py
IanCal: > We do create EMPTY parquet files right?

Yes, currently when there are no payments, the file is empty. 

> All EMPTY parquet files are meant to have an EMPTY_MARKER_HEX written to the blockchain?

Yes.

IanCal:
> It also occured to me that we can piggy back on matics and fadhlans local protocol work. As long as we make the subgraph extractor point to their local protocol, we can then have a docker up for reward services too.
> 
> EDIT: Actually maybe not, we need a local running lambda and a local s3 data store.

We could probably work towards that - the lambda section is very thin so having something that watches for files written locally and calls the same code would work. Or there may already be a lambda test setup that we could use / something like minio.

packages/reward-root-submitter/reward_root_submitter/main.py
IanCal: I think it's correct but awkward. I'll have a quick think if casting to an int and raising on an exception there with a better error message would change anything.

IanCal: Changed the regex so it requires at least one digit, and it's only pulling in digits so a "is not None" is enough.

packages/reward-root-submitter/README.md
IanCal: Added a note on what triggers this service to the readme

packages/reward-root-submitter/reward_root_submitter/config.py
IanCal: Good idea, done.

PR #3128, opened 2022-08-01
Ssr web server sentry

Aierie:
Changes:
1. Node/server-side Sentry has to be imported and initialized outside the FastBoot sandbox because it relies on certain environment globals. We access the injected global `NodeSentry` in our app.
2. Use `getSentry` and `initSentry` utils to use the correct Sentry package depending on where the code is running
3. `application.ts` is responsible for detecting some errors using the `error` action because for some reason `Ember.onerror` doesn't seem to fire for route errors...? However we still add Sentry handling to `Ember.onerror` because it could get called for other errors.


lukemelia (approved):
Looks good. Not sure about the answers to your TODOs though.

Aierie:
I think the todos are both real problems and actionable, I'll fix them in separate PRs.

PR #3130, opened 2022-08-01
[CS-4168] Change wallet name to Cardstack Wallet

paulinhapenedo:
This PR replaces all the instances of the card wallet name with the new name: "Cardstack Wallet". 

I'm not sure if a single PR with these changes is enough or if I need to open a PR for each package, so guidance is welcome.


Aierie (approved):

Aierie:
Single PR seems good enough for these changes to me!

jurgenwerk (approved):
Looking good!

PR #3131, opened 2022-08-02
Add withinLimit guard to unlock sdk function

tintinthong:
During bridging process, `unlock` and `relay` steps occur 

**Currently**
If a user bridges above the token limits, he will succesfully `unlock` but fail `relay`

**Now**
If a user bridges above the token limits, he will fail `unlock` even before reaching the `relay` step


packages/cardpay-sdk/sdk/token-bridge-foreign-side.ts
```diff
@@ -85,6 +85,22 @@ export default class TokenBridgeForeignSide implements ITokenBridgeForeignSide {
...
+      throw new Error(`
+        unlock action NOT within configured limits
+
+          Minimum Amount Per Tx: ${fromWei(minPerTx)}
+          Maximum Amount Per Tx: ${fromWei(maxPerTx)}
+          Total Amount Today: ${fromWei(totalSpentPerDay)}
+          Daily Amount Limit: ${fromWei(dailyLimit)}
+
+      `);
+    }
```
Aierie, line 103: For this and for `relayTokens`, do we want to include the deposited amount in the error message?

jurgenwerk (approved):
Looking good. Do we also have to handle this error and add an explanation message in the workflow?

Aierie (approved):

packages/cardpay-sdk/sdk/token-bridge-foreign-side.ts
Aierie: The types for minPerTx, maxPerTx, dailyLimit, and totalSpentPerDay may not be correct - they don't seem to be BNs, leading to a problem here: https://github.com/cardstack/cardstack/pull/3134#partial-pull-merging . Should they be strings or BNs?

tintinthong: Thanks for the suggestions! I will make the corresponding change. 
- add deposited amount to `unlock` and `relay` 
- covert those minPerTx, maxPerTx, dailyLimit, and totalSpentPerDay  to BNs

tintinthong:
> Looking good. Do we also have to handle this error and add an explanation message in the workflow?

This prevents the error state? https://github.com/cardstack/cardstack/pull/3134

In any case, if we handle the error in the workflow, it should be handled in another PR.

packages/cardpay-sdk/sdk/token-bridge-foreign-side.ts
tintinthong: EDIT: It looks like the call() is returning a string. I will adjust the types. And ur conversion occured it other PR will stay as it is. Sorry bout that

tintinthong: @Aierie I made the BN change. And I removed the conversion u did to getDepositLimits() previously 
as stated here https://github.com/cardstack/cardstack/pull/3134#discussion_r935719355

Aierie (approved):
Thanks for the fixes! 😄

PR #3134, opened 2022-08-02
add deposit limit input validation

Aierie:
Copy change to come later, since that might be a bit more finicky / need more tweaking.

### DIFF
--- a/packages/web-client/app/utils/web3-strategies/layer1-chain.ts
+++ b/packages/web-client/app/utils/web3-strategies/layer1-chain.ts
@@ -8,6 +8,7 @@ import * as Sentry from '@sentry/browser';
 import { Emitter, SimpleEmitter, UnbindEventListener } from '../events';
 import {
   BridgeableSymbol,
+  bridgeableSymbols,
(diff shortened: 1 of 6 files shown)


packages/web-client/app/utils/web3-strategies/layer1-chain.ts
```diff
@@ -363,8 +363,8 @@ export default abstract class Layer1ChainWeb3Strategy
           return [
             tokenSymbol,
             {
-              min,
-              max,
+              min: new BN(min),
+              max: new BN(max),
```
Aierie, line 367: Doing this makes sure we get BNs where we expect them.

backspace (approved):
I was able to exercise this on the staging preview deployment by hacking the `cardBalance`: 
Deposit   Withdrawal · Card Pay 2022-08-02 10-59-28

I think the conversion shouldn’t be 0 but that’s a preexisting problem.

backspace:
I’m not sure what’s happening with Percy, maybe `main` needs to be merged in? Or approval was missed somewhere?

PR #3136, opened 2022-08-03
Fix blur redundant validation

Aierie:


backspace (approved):
Thanks for looking into this! 🎉

PR #3145, opened 2022-08-04
Upgrade apps to Ember 4.6

lukemelia:


ef4:
The cardhost build failure was caused by ember-simple-set-helper. It has a style of AST transform that is no longer supported in Ember 4. That addon is deprecated in favor of ember-set-helper. But it looks like it wasn't even being used.

lukemelia:
I held back ember-template-lint from 4.x upgrade due to https://github.com/ember-template-lint/ember-template-lint/issues/2479

lukemelia:
Almost there... still have lint execution issues and violations to resolve...

packages/boxel/.github/workflows/ci.yml
```diff
@@ -0,0 +1,79 @@
...
+      - uses: actions/checkout@v3
+      - name: Install Node
+        uses: actions/setup-node@v3
+        with:
+          node-version: 14.x
+          cache: npm
+      - name: Install Dependencies
+        run: npm ci
+      - name: Run Tests
+        run: ./node_modules/.bin/ember try:one ${{ matrix.try-scenario }}
```
backspace, line 79: Should this (and all its equivalents in other packages) be deleted? It seems confusing to have in the repository since it will never be used; I guess every `ember update` will produce a merge conflict without it but that still seems preferable to me.

backspace (approved):
🎉

packages/boxel/.github/workflows/ci.yml
lukemelia: Yes. Removed.

packages/web-client/tests/acceptance/issue-prepaid-card-persistence-test.ts
```diff
@@ -75,14 +75,15 @@ module('Acceptance | issue prepaid card persistence', function (hooks) {
       'service:workflow-persistence'
     );
 
+    // eslint-disable-next-line ember/no-array-prototype-extensions
```
ef4, line 78: This made me investigate what the heck this rule is doing that would cause it to misfire here, and the whole rule is extremely dubious. We can turn it off globally.

ef4 (approved):

packages/web-client/tests/acceptance/issue-prepaid-card-persistence-test.ts
ef4: https://github.com/ember-cli/eslint-plugin-ember/pull/1562

packages/web-client/tests/unit/services/layer2-network-test.ts
```diff
@@ -13,9 +13,9 @@ module('Unit | Service | Layer2Network', function (hooks) {
   test('it clears the auth token when onDisconnect is called', function (assert) {
     let hubAuthentication = this.owner.lookup('service:hub-authentication');
     hubAuthentication.authToken = 'something';
-    assert.equal(hubAuthentication.authToken, 'something');
+    assert.strictEqual(hubAuthentication.authToken, 'something');
     let layer2Network = this.owner.lookup('service:layer2-network');
     layer2Network.onDisconnect();
-    assert.equal(hubAuthentication.authToken, null);
+    assert.strictEqual(hubAuthentication.authToken, undefined);
```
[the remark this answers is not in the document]
lukemelia: Good catch. Fixed.

PR #3146, opened 2022-08-05
hub: Add /api/profiles

backspace:
This adds a set of conventional endpoints at `/api/profiles` that operate on the new `profiles` table. They mirror the combined endpoints of `/api/card-spaces` and `/api/merchant-infos`. Once we have updated clients that use those to use `profiles`, we can remove them.

This also removes `PATCH /api/card-spaces/:id`, which is unused; a superset of its functionality is now present at `PATCH /api/profiles/:id`.

The `ProfileSerializer` accepts an optional type so can return the shapes expected that `CardSpaceSerializer` and `MerchantInfoSerializer` used to return. When the old JSON documents on S3 and endpoints are no longer consumed by any clients we can simplify this serialiser.


packages/hub/node-tests/routes/profiles-test.ts
```diff
@@ -0,0 +1,669 @@
...
+    it('persists profile', async function () {
+      const payload = {
+        data: {
+          type: 'profiles',
+          attributes: {
+            name: 'Satoshi Nakamoto',
+            slug: 'satoshi',
+            color: 'ff0000',
+            'text-color': 'ffffff',
+            'owner-address': '0x00000000000',
```
Aierie, line 224: Pre-existing problem - `ownerAddress` is derived from the auth token, so we should probably remove it from payloads used for profile creation tests?

packages/hub/node-tests/routes/profiles-test.ts
```diff
@@ -0,0 +1,669 @@
...
+          data: {
+            type: 'profiles',
+            id: 'the-id',
+            attributes: {
+              name: 'Satoshi Nakamoto',
+              slug: 'satoshi',
+              did: 'the-did',
+              color: 'ff0000',
+              'text-color': 'ffffff',
+              'owner-address': '0x2f58630CA445Ab1a6DE2Bb9892AA2e1d60876C13',
```
Aierie, line 255: Pre-existing problem - We could use `stubUserAddress` here to be clearer about how we know this address.

packages/hub/node-tests/routes/profiles-test.ts
```diff
@@ -0,0 +1,669 @@
...
+              links: [],
+              'profile-description': 'description',
+              'profile-image-url': 'http://example.com/image.jpg',
```
Aierie, line 258: I wonder if we should prevent storage of these details until we are clearer about what kind of validation/constraints we want to apply?

packages/hub/node-tests/routes/profiles-test.ts
```diff
@@ -0,0 +1,669 @@
...
+let stubUserAddress = '0x2f58630CA445Ab1a6DE2Bb9892AA2e1d60876C13';
+function handleValidateAuthToken(encryptedString: string) {
+  expect(encryptedString).to.equal('abc123--def456--ghi789');
+  return stubUserAddress;
+}
+
+describe('/api/profiles', function () {
+  describe('GET /api/profiles/validate-slug/:slug', function () {
+    this.beforeEach(function () {
+      registry(this).register('authentication-utils', StubAuthenticationUtils);
```
Aierie, line 33: Pre-existing problem - we don't check for worker jobs to update S3, which is the more important source of truth for web client + wallet most of the time.

packages/hub/routes/profiles.ts
```diff
@@ -0,0 +1,253 @@
...
+    });
+
+    let serialized = this.profileSerializer.serialize(profile);
+
+    ctx.status = 201;
+    ctx.body = serialized;
+    ctx.type = 'application/vnd.api+json';
+  }
+
+  async patch(ctx: Koa.Context) {
```
Aierie, line 109: Should this also have an action to update S3?

packages/hub/routes/profiles.ts
```diff
@@ -0,0 +1,253 @@
...
+    }
+
+    ctx.status = 200;
+    ctx.body = this.profileSerializer.serialize(profile);
+    ctx.type = 'application/vnd.api+json';
+  }
+
+  sanitizeText(value: string | unknown): any {
+    if (typeof value === 'string') {
+      return value.trim().replace(/ +/g, ' ');
```
Aierie, line 242: More pre-existing problems - Is this method doing what it claims to do? It seems like it's just normalizing spaces at the moment. If the goal is to normalize spaces and we don't care about newlines, we should use `/\s+/g` instead. 

Not sure if we need to fix this elsewhere too

packages/hub/routes/profiles.ts
```diff
@@ -0,0 +1,253 @@
...
+          slugAvailable: false,
+          detail: 'This ID is not allowed',
+        };
+      } else {
+        let prisma = await this.prismaManager.getClient();
+        let profile = await prisma.profile.findFirst({ where: { slug } });
+        return {
+          slugAvailable: profile ? false : true,
+          detail: profile ? 'This ID is already taken. Please choose another one' : 'ID is available',
+        };
```
Aierie, line 217: Nit-ish (not suggestion because formatting)

```js
if (profile) {
  return { slugAvailable: false, detail: 'This ID is already taken. Please choose another one' }
} else {
  return { slugAvailable: true, detail: 'ID is available' }
}
```

packages/hub/node-tests/routes/profiles-test.ts
Aierie: We could also consider changing the way that we handle addresses because it's possible to end up with a "stale" auth token/ an auth token for another address - so the auth token's address is checked against the supplied `owner-address` and we return a 401 if they are not the same.

packages/hub/routes/profiles.ts
Aierie: Maybe in this case it should not be a job to update, but rather the hub calling and waiting for the S3 update.

Aierie:
Just tried this locally - ssr-web profile viewing and web-client profile creation work fine

packages/hub/routes/card-spaces.ts
```diff
@@ -34,84 +25,12 @@ export default class CardSpacesRoute {
       return;
     }
 
-    let serialized = await this.cardSpaceSerializer.serialize(profile);
```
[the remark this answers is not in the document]
backspace: yes… why does it warn about it but not produce a linting error 🤔

packages/hub/node-tests/services/card-space-validator-test.ts
```diff
@@ -1,11 +1,11 @@
 import { Profile } from '@prisma/client';
 import { setupHub } from '../helpers/server';
 
-describe('CardSpaceValidator', function () {
+describe('ProfileValidator', function () {
```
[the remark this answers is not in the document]
backspace: yes, fixed, thanks

packages/hub/services/validators/profile.ts
```diff
@@ -24,17 +19,21 @@ function isValidUrl(url: string): boolean {
   }
 }
 
-export default class CardSpaceValidator {
+export default class ProfileValidator {
   prismaManager = inject('prisma-manager', { as: 'prismaManager' });
 
   async validate(cardSpace: Partial<Profile>): Promise<CardSpaceErrors> {
```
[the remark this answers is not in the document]
backspace: it’s true. I’m not even convinced this external validator is necessary. But I want to resist too much refactoring hehe

packages/hub/node-tests/routes/profiles-test.ts
backspace: 👍🏻 definite improvement, thanks

packages/hub/node-tests/routes/profiles-test.ts
backspace: Maybe! Probably? We can make another issue for that if so. These card space properties are near-orphaned 😖

packages/hub/routes/profiles.ts
backspace: hmm why not use a job? We expect a delay after `POST`, why not `PATCH`? But agreed that this should trigger an update, added now, thanks.

packages/hub/routes/profiles.ts
backspace: updated, thanks

packages/hub/routes/profiles.ts
backspace: I don’t know the intent but I’m reluctant to dig into it really, my goal was to just preserve existing behaviour from the two merged endpoints, but I do think we should look at what we really want here

packages/hub/routes/profiles.ts
Aierie: Because for `PATCH`, users can refresh after the action seemingly completes, and get back a profile that does not include the changes, possibly leading to confusion about whether changes actually were submitted, and if so, when they will be reflected.

packages/hub/routes/profiles.ts
Aierie: Can we rename this so that we don't forget that it is not really sanitising text atm?

packages/hub/routes/profiles.ts
backspace: let’s revisit when we see `PATCH` in action and refine as needed 🤓

PR #3147, opened 2022-08-08
hub: Add type to job serialisations

backspace:


oidoug (approved):
Thanks for the quick response. LGTM.

PR #3148, opened 2022-08-10
Add tables: scheduled_payments, scheduled_payment_attempts

jurgenwerk:
Tickets: CS-4371, CS-4372

### DIFF
--- a/packages/hub/db/migrations/20220810123016866_create-scheduled-payments.ts
+++ b/packages/hub/db/migrations/20220810123016866_create-scheduled-payments.ts
@@ -1,0 +1,37 @@
+import { MigrationBuilder } from 'node-pg-migrate';
+
+const TABLE = 'scheduled_payments';
+
+export async function up(pgm: MigrationBuilder): Promise<void> {
+  pgm.createTable(TABLE, {
+    id: { type: 'uuid', primaryKey: true },
+    sender_safe_address: { type: 'string', notNull: true },
+    module_address: { type: 'string', notNull: true },
+    token_address: { type: 'string', notNull: true },
+    amount: { type: 'bigint', notNull: true },
+    payee_address: { type: 'string', notNull: true },
+    execution_gas_estimation: { type: 'bigint', notNull: true },
+    max_gas_price: { type: 'bigint', notNull: true },
+    fee_fixed_usd: { type: 'integer', notNull: true },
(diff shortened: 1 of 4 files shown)


lukemelia:
Let's add the new fee fields before me merge this.

packages/hub/db/migrations/20220810123029047_create-scheduled-payment-attempts.ts
```diff
@@ -0,0 +1,23 @@
+import { MigrationBuilder } from 'node-pg-migrate';
+
+const TABLE = 'scheduled_payment_attempts';
+const STATUS_ENUM = `${TABLE}_status_enum`;
+
+export async function up(pgm: MigrationBuilder): Promise<void> {
+  pgm.createType(STATUS_ENUM, ['in_progress', 'finished', 'failed']);
```
jurgenwerk, line 7: Note to self: actually use the enum 😅

packages/hub/db/migrations/20220810123016866_create-scheduled-payments.ts
```diff
@@ -0,0 +1,37 @@
...
+    execution_gas_estimation: { type: 'bigint', notNull: true },
+    max_gas_price: { type: 'bigint', notNull: true },
+    fee_fixed_usd: { type: 'integer', notNull: true },
+    fee_percentage: { type: 'integer', notNull: true },
+    nonce: { type: 'bigint', notNull: true },
+    pay_at: { type: 'timestamp' },
+    recurring_day_of_month: { type: 'integer' },
+    recurring_until: { type: 'timestamp' },
+    valid_for_days: { type: 'integer', default: 3 },
+    sp_hash: { type: 'string', notNull: true },
```
jurgenwerk, line 22: Note to self: add uniqueness constraint

backspace (commented):
The migrations look fine, is this meant to include the changes `structure.sql` file that come from actually running them? And it would make sense to then include the [Prisma changes](https://github.com/cardstack/cardstack/blob/main/packages/hub/README.md#database-migrations)

jurgenwerk:
@backspace thanks - I added the new tables to the structure and the Prisma models as well. There are currently ongoing discussions whether we actually need the `nonce` (and if it would be good to constrain uniqueness by sender, recipient, amount, token, pay date (or recurring day in month)) so let's wait until we clarify that.

PR #3149, opened 2022-08-10
ssr/web: Update profiles wording and endpoints

backspace:
This updates two Ember applications to use the new `/api/profiles` endpoints. It also includes vast string replacements:
* `merchant` becomes `profile`
* `payment profile` becomes `profile`
* `Card Space` becomes `card.xyz`

Since `MerchantSafe` will continue to exist, not every instance of that string is gone.

There’s also a seemingly-superfluous change to a Mirage route handler, I don’t understand why but it was not returning 404 as expected with a missing profile.


Aierie (approved):
I think key functionality is working - I can get through the workflows fine, and there are no overt visual bugs. That said, this could have been easier to review if we had split things up into several small(er), focused PRs.

I hope we don't need to do this too often in the future, but what do you think of this suggestion for splitting up such renames?
1. Any logic changes that change behaviour of non-local-dev environments (in this case using the renamed endpoints) 
2. Copy changes, so that Percy diffs clearly indicate that ONLY the copy has changed. Reviewer can focus on making sure that all copy reflects required changes, and that assumptions about content dimensions are still valid.
3. CSS class and variable renaming - this potentially has global implications so it seems good to have it separate and focused - there should be no Percy diffs/complaints here. Reviewer can focus on making sure nothing looks broken and checking on the more global implications of class + variable changes (clashing/setting where we don't expect a variable to be set or we rely on the default).
4. File + Ember template + JS import renames (+ JS variable renames? unsure if should be separate) - This seems necessary to group together. There should also be no Percy diffs here. Reviewer can focus on making sure that essential flows work correctly and all assets are reachable.
5. Test renames (`module(moduleName); test(testName)`) - since test name changes will result in Percy having new snapshots that do not have an old version to compare against, having this isolated seems good.

backspace:
yes, that seems like a good approach, I’m sorry for the general chaos that this ended up being, I should have been more strategic.

PR #3151, opened 2022-08-11
hub: Add GET /api/profiles/:slug

backspace:
This is extracted from #3149, as it in the same deployment
would result in Card Space downtime as ssr-web will expect
this endpoint to exist.


Aierie (approved):
This worked for me w ssr-web!

PR #3154, opened 2022-08-15
don't run fixPrismaFile as part of initial file evaluation

Aierie:
CS-4396


packages/hub/cli/db/transform-prisma-schema.ts
```diff
@@ -47,6 +47,10 @@ function fixFieldsArrayString(fields: string) {
 }
 
 async function fixPrismaFile() {
+  if (!fs.existsSync(PRISMA_FILE_PATH)) {
+    throw new Error('Attempting to apply Prisma file transformation but file does not exist');
```
backspace, line 51:
```suggestion
    throw new Error('Skipping Prisma schema transformation because file does not exist');
```

Maybe this is more clear that this is not a problem?

backspace (approved):
Thanks for looking into this!

PR #3155, opened 2022-08-15
Step over reward claims 

tintinthong:
Claim of proofs fail if `rewardAmount < gasFees`. This will mean that if a user has 3 valid proofs and 1 proof with a `rewardAmount < gasFees`, the claim may fail somewhere in the middle. This PR circumvents the possibility this by computing gasFees for each claim and filtering the proofs based upon that. **This is backward compatible (does not affect any existing) with wallet implementation of reward sdk.**

- Additionally, previously wallet loops through all available proofs  to `claim`. Now, `claimAllGasEstimates` and `claimAll` function in the sdk to simplify the wallet greatly

EDIT: Will update readme after review 
EDIT: `rewardTokenBalances` will not filter for this. Will handle this in another PR based on some product discussion
EDIT: As per discussion in protocol team, step over reward claims should be handled using a multi-call batching the reward claim function (which requires no smart contract mutation). Nevertheless, this pr does give a "multi-call" function which ignores small claim amounts, which we will fall back to if the multi-call fails.


packages/cardpay-sdk/sdk/reward-pool/base.ts
```diff
@@ -62,7 +63,10 @@ export interface RewardTokenBalance {
   balance: BN;
 }
 
-export type WithSymbol<T extends Proof | RewardTokenBalance> = T & {
+export interface HasTokenAddress {
+  tokenAddress: string;
+}
+export type WithSymbol<T extends HasTokenAddress> = T & {
```
tintinthong, line 69: Made `WithSymbol` type more generic to accept types that have `tokenAddress` attribute

packages/cardpay-sdk/sdk/reward-pool/base.ts
```diff
@@ -93,35 +97,12 @@ export default class RewardPool {
...
-    let json = await response.json();
-    return json['tokenAddresses'];
-  }
-
   async getProofs(
     address: string,
     rewardProgramId?: string,
     tokenAddress?: string,
     knownClaimed?: boolean,
+    safeAddress?: string,
```
tintinthong, line 105: There was some discussion of using `override` type for function and extend the `Proof` type. I intentionally found that using `optional` parameters much easier and straightforward. `type` based switch case don't seem too straightforward and I still need to use it in other functions -- this solution is very easy to grok despite needing to check existence of parameter `safeAddress` in multiple places.

packages/cardpay-sdk/sdk/reward-pool/base.ts
```diff
@@ -71,15 +75,6 @@ export default class RewardPool {
...
 
-  async getBalance(address: string, tokenAddress: string, rewardProgramId?: string): Promise<BN> {
-    const unclaimedValidProofs = (await this.getProofs(address, rewardProgramId, tokenAddress, false)).filter(
-      (o) => o.isValid
-    );
-    return unclaimedValidProofs.reduce((total, { amount }) => {
-      return total.add(amount);
-    }, new BN('0'));
-  }
-
```
tintinthong, line 82: Remove `getBalance` it was confusing. Case covered by `rewardTokenBalances`

packages/cardpay-sdk/sdk/reward-pool/base.ts
```diff
@@ -241,19 +228,6 @@ export default class RewardPool {
...
-    rewardProgramId?: string
-  ): Promise<RewardTokenBalance> {
-    let balance = await this.getBalance(address, tokenAddress, rewardProgramId);
-    return {
-      rewardProgramId,
-      tokenAddress,
-      balance,
-    };
-  }
-
```
tintinthong, line 256: Remove `rewardTokenBalance` it was confusing. Case covered by `rewardTokenBalances`

packages/cardpay-sdk/sdk/reward-pool/base.ts
tintinthong: EDIT: Ended up introducing new interface `ClaimableProof` and using override. It was useful after ian told me to use `proofs` as the only input to `claimAll` and `claimAllGasEstimates`

packages/cardpay-cli/rewards/claimable-proofs.ts
```diff
@@ -43,7 +43,12 @@ export default {
     };
     let { web3 } = await getEthereumClients(network, getConnectionType(args));
     let rewardPool = await getSDK('RewardPool', web3);
-    const proofs = await rewardPool.getProofs(address, rewardProgramId, tokenAddress, false, safeAddress);
+    let proofs;
+    if (safeAddress) {
+      proofs = await rewardPool.getProofs(address, safeAddress, rewardProgramId, tokenAddress, false);
+    } else {
+      proofs = await rewardPool.getProofs(address, undefined, rewardProgramId, tokenAddress, false);
+    }
```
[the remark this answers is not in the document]
tintinthong: Sweet!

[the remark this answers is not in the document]
tintinthong: I think this doesn't work. My first and second overload of the `getProofs` function get in the way of the flexible type implementation. 

```
[{
	"resource": "/Users/tintinthong/Github/cardstack/packages/cardpay-cli/rewards/claimable-proofs.ts",
	"owner": "typescript",
	"code": "2769",
	"severity": 8,
	"message": "No overload matches this call.\n  Overload 1 of 2, '(address: string, safeAddress: string, rewardProgramId?: string | undefined, tokenAddress?: string | undefined, knownClaimed?: boolean | undefined, offset?: number | undefined, limit?: number | undefined): Promise<...>', gave the following error.\n    Argument of type 'string | undefined' is not assignable to parameter of type 'string'.\n      Type 'undefined' is not assignable to type 'string'.\n  Overload 2 of 2, '(address: string, safeAddress: undefined, rewardProgramId?: string | undefined, tokenAddress?: string | undefined, knownClaimed?: boolean | undefined, offset?: number | undefined, limit?: number | undefined): Promise<...>', gave the following error.\n    Argument of type 'string | undefined' is not assignable to parameter of type 'undefined'.\n      Type 'string' is not assignable to type 'undefined'.",
	"source": "ts",
	"startLineNumber": 46,
	"startColumn": 56,
	"endLineNumber": 46,
	"endColumn": 80,
	"relatedInformation": [
		{
			"startLineNumber": 115,
			"startColumn": 9,
			"endLineNumber": 115,
			"endColumn": 18,
			"message": "The call would have succeeded against this implementation, but implementation signatures of overloads are not externally visible.",
			"resource": "/Users/tintinthong/Github/cardstack/packages/cardpay-sdk/sdk/reward-pool/base.ts"
		}
	]
}]
```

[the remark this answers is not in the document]
oidoug: Ah, I see, worth the shot :)

packages/cardpay-sdk/sdk/reward-pool/base.ts
```diff
@@ -561,6 +588,32 @@ The reward program ${rewardProgramId} has balance equals ${fromWei(
     };
   }
 
+  async claimAllGasEstimate(unclaimedValidProofsWithSingleToken: WithSymbol<Proof>[]): Promise<GasEstimate> {
```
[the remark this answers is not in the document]
tintinthong: Oh Yes. Leftover from previous. Thanks

IanCal (approved):
I think this makes sense but I don't have a good way of testing it out. Not sure where a test would fit best, or if this just spawns a manual testing ticket.

tintinthong:
> I think this makes sense but I don't have a good way of testing it out. Not sure where a test would fit best, or if this just spawns a manual testing ticket.

This PR is only test manually. It is not in action in wallet yet, but I will be sure to do manual testing in the wallet. https://linear.app/cardstack/issue/CS-4429/add-claimall-and-claimallgasestimate-to-wallet

tintinthong:
Also this has been tested on CLI

PR #3156, opened 2022-08-15
fix for slow web-client build

Aierie:
Because of Watchman problems, I run `ember s --watcher=polling` to serve web client - this generally takes really long. The slowdown  only occurs with `ember s`, `ember build` finishes fairly quickly. With the fix in this PR, initial build for `ember s --watcher=polling` finishes in under 10 mins .


ef4:
My best guess as to what's going on here is that the performance comparison is being confounded by persistent caches. The package.json is part of the cache key, so arbitrary changes to it could cause significant performance swings.

In general, poor performance without watchman is not unexpected, and rather than invest effort into optimizing that case I think it's fine for us to require watchman (which @Aierie says he already got working). I'm trying hard to get us past the point where we can drop all of this legacy broccoli watching machinery entirely and get a much faster build.

PR #3159, opened 2022-08-16
ci: Improve caching and support parallel builds

pcjun97:
Managed to bring down hub's deployment time from ~30+ mins to ~13mins.

Changes made:
- cache `node_modules` instead of yarn's cache folder: This brings caching operation (restorations and upload) up to about 2 mins, but brings yarn initialisation's time from few minutes down to less then 5 seconds when there are no dependency changes.
- For apps that has multiple waypoint deployments that based on the same build (eg: hub, hub-worker, etc), the build step is performed once, then uploaded to GitHub's artifact. Waypoint deployment stage is then performed as parallel jobs.

Sample workflow: https://github.com/cardstack/cardstack/actions/runs/2867369913


backspace (approved):
I’m looking forward to faster deployments, thank you 🤩

PR #3013, opened 2022-06-13
Staking rewards

paulin-alcoser:
Two main changes - 

Addition of a dummy rule testing out running against two different tables

Staking V1 (flat interest)

PR #3162, opened 2022-08-16
Staking

IanCal:
Replaces #3013, easier to manage the branches after divergence.

Implements staking and adds an example 'dummy' rule. Rules to be cleaned up for demonstration/implementation in a future PR.

Staking is rewarding based on balances held in CARD.cpxd in depots, which must take into account changing balances through the time it is held.

Due to uint256s being enormous and being handled very poorly, token amounts are scaled down 1e9 in the subgraph export and so rewards need to be scaled back up at the end.


IanCal:
Note - this has already gone through some rounds of review. I'll leave some inline comments too.

packages/cardpay-reward-programs/cardpay_reward_programs/rule.py
```diff
@@ -34,7 +34,7 @@ def set_user_defined_parameters(
         raise NotImplementedError
 
     @abstractmethod
-    def sql(self, table_query):
+    def sql(self, table_query, aux_table_query=None):
```
IanCal, line 37: this was extended to support requiring more than one table. There's a refactor coming that removes this requirement entirely using duckdbs "register" functionality which will also clean up the tests significantly but I want to do that alongside the shift to using the parquet metdata files.

packages/cardpay-reward-programs/cardpay_reward_programs/rules/staking.py
```diff
@@ -0,0 +1,107 @@
...
+        if table_query == "parquet_scan([])" or aux_table_query == "parquet_scan([])":
+            df = pd.DataFrame(columns=["payee", "rewards"])
+        else:
+            df = self.run_query(table_query, vars, aux_table_query)
+        df["rewardProgramID"] = reward_program_id
+        df["paymentCycle"] = payment_cycle
+        df["validFrom"] = payment_cycle
+        df["validTo"] = payment_cycle + self.duration
+        df["token"] = self.token
+        df["amount"] = df["rewards"] * 1_000_000_000
```
IanCal, line 105: The balances exported are scaled down by 1e9 to make them manageable but with still pretty high precision (9 decimal places) and so the rewards are scaled up at the end here.

IanCal:
The dummy rule will be refactored into some example cases, and I'll clear through the other rules we have that we don't actually use in a following change.

packages/cardpay-reward-programs/.vscode/settings.json
```diff
@@ -1,3 +1,12 @@
...
+    "python.testing.pytestArgs": [
+        "tests",
+        "-rs",
+    ],
+    "python.testing.unittestEnabled": false,
+    "python.testing.pytestEnabled": true,
+    "python.analysis.extraPaths": [
+        "./streamlit",
+        "./__pypackages__/3.9/lib"
+    ]
```
tintinthong, line 11: Cheers. Actually perhaps I should include this in all subpackages. So ppl wanting to look at single subpackage dont have to use python codeworkspace for this. Sorry if this impeded u

packages/cardpay-reward-programs/cardpay_reward_programs/rules/dummy_rule.py
```diff
@@ -0,0 +1,86 @@
+import duckdb
+import pandas as pd
+from cardpay_reward_programs.rule import Rule
+
+
+class DummyRule(Rule):
+    """
+    This rules rewards acounts with
```
tintinthong, line 8: missing description. Its not clear to me immediately what the dummy rule is

tintinthong:
I like that all `input` rules now have example `parameters.json`

packages/cardpay-reward-programs/cardpay_reward_programs/rules/weighted_usage.py
```diff
@@ -21,7 +21,7 @@ def set_user_defined_parameters(
         self.token = token
         self.duration = duration
 
-    def sql(self, table_query):
+    def sql(self, table_query, aux_table_query=None):
```
tintinthong, line 24: don't think this is needed?

packages/cardpay-reward-programs/tests/rules/test_dummy_rule.py
```diff
@@ -0,0 +1,194 @@
...
+
+    core_config = {
+        "payment_cycle_length": 32768,
+        "start_block": 0,
+        "end_block": 10000,
+        "subgraph_config_locations": {
+            "prepaid_card_payment": "s3://cardpay-staging-partitioned-graph-data/data/prepaid_card_payments/0.0.3/",
+            # "spend_accumulation": "s3://tall-data-dev/paulin/spend_accumulation/0.0.1/",
+            "safe_owner": "s3://tall-data-dev/paulin/safe_owner/0.0.1/",
+        },
```
tintinthong, line 22: I don't think this data is available in the repo. Hence, we r probably making real s3 calls rather than mocking

IanCal: yeah I'm going to pull the dummy rule work out, we can bring it back in for examples at some point but I think there's more cleanup required than I want to do for getting staking out.

packages/cardpay-reward-programs/cardpay_reward_programs/rules/weighted_usage.py
IanCal: I'll try and either add it everywhere or clear it out of here - it was added I think to match the signature of the function in the base class. With the register refactor this can be standard for all again and take no arguments at all.

packages/cardpay-reward-programs/cardpay_reward_programs/rules/staking.py
```diff
@@ -0,0 +1,107 @@
...
+        table_query = self._get_table_query(
+            "token_holder", "token_holder", self.start_block, payment_cycle
+        )
+
+        aux_table_query = self._get_table_query(
+            "safe_owner", "safe_owner", self.start_block, payment_cycle
+        )
+
+        if table_query == "parquet_scan([])" or aux_table_query == "parquet_scan([])":
+            df = pd.DataFrame(columns=["payee", "rewards"])
```
tintinthong, line 97: ["payee", "rewards"]->["payee", "amount"]

packages/cardpay-reward-programs/cardpay_reward_programs/rules/staking.py
```diff
@@ -0,0 +1,107 @@
...
+        self.token = token
+        self.duration = duration
+        self.interest_rate_monthly = interest_rate_monthly
+
+    def sql(self, table_query, aux_table_query):
+        token_holder_table = table_query
+        safe_owner_table = aux_table_query
+        return f"""
+            with filtered_safes as (
+                SELECT safe, _block_number, type, max(owner) as owner
```
tintinthong, line 23: why are we using `max(owner)`?

IanCal: So this was uncovered with the hypothesis tests, it's an edge case and I think non-exploitable in the real world given restrictions on depots but as we go forwards it might not always be the case. 

Case:

Depot changes owner twice in one block. This CTE then emits *two* rows for the safe, and *both* owners get rewarded with the entire amount. 

We could use "any" for this, but then the outcome of the query is random.

There is a remaining issue that we actually don't know who should be rewarded, to solve that we'd need to do some more careful work looking at the chain of transfers, but using max here prevents the rewards from totalling more than they should.

packages/cardpay-reward-programs/cardpay_reward_programs/rules/staking.py
IanCal: This gets swapped out on lines 105/106

packages/cardpay-reward-programs/cardpay_reward_programs/rules/dummy_rule.py
IanCal: deleted

tintinthong:
Ahh cool I was running into errors with test. Looks like u resolved it

IanCal:
> Ahh cool I was running into errors with test. Looks like u resolved it

Yeah sorry about that, I did a "quick bit of cleanup" with the default args that broke the tests, fixed now.

packages/cardpay-reward-programs/cardpay_reward_programs/rules/staking.py
```diff
@@ -0,0 +1,107 @@
...
+                )
+                group by safe
+                ),
+            all_data as (select * from original_balances union all select * from balance_changes)
+
+            select owner as payee, sum(change * ((interest_rate**compounding_rate) - 1)) as rewards
+            from all_data
+            left join filtered_safes using (safe)
+            group by payee
+        """
```
tintinthong, line 73: This is real hard to grok. So im layout the data here. 

filtered_safes

    safe  _block_number   type   owner
0  safe1              0  depot  owner1
1  safe2              0  depot  owner2
2  safe3              0  depot  owner3

filtered_balances

    safe  balance_int64  _block_number   token
0  safe1           1000              0  card-0
1  safe1           1500             10  card-0
2  safe1           1750             20  card-0
3  safe1           1500             25  card-0
4  safe2           1000              0  card-0
5  safe2           1500             10  card-0
6  safe2           1200             15  card-0
7  safe2           1700             20  card-0
8  safe3           1000              0  card-0

balance_changes

    safe  change  interest_rate  compounding_rate
0  safe1    1000           1.06          1.000000
1  safe1     500           1.06          0.666667
2  safe1     250           1.06          0.333333
3  safe1    -250           1.06          0.166667
4  safe2    1000           1.06          1.000000
5  safe2     500           1.06          0.666667
6  safe2    -300           1.06          0.500000
7  safe2     500           1.06          0.333333
8  safe3    1000           1.06          1.000000

original_balances

Empty DataFrame
Columns: [safe, change, interest_rate, compounding_rate]
Index: []


all_data

    safe  change  interest_rate  compounding_rate
0  safe1    1000           1.06          1.000000
1  safe1     500           1.06          0.666667
2  safe1     250           1.06          0.333333
3  safe1    -250           1.06          0.166667
4  safe2    1000           1.06          1.000000
5  safe2     500           1.06          0.666667
6  safe2    -300           1.06          0.500000
7  safe2     500           1.06          0.333333
8  safe3    1000           1.06          1.000000


end_reward

    payee    rewards
0  owner3  59.999943
1  owner1  82.268583
2  owner2  80.742585

tintinthong: @IanCal what is the compounding **period**??

tintinthong: If there is a link to how you guys computed interest, it would be good so I can check by it.

tintinthong: Might be a silly question. The rule has cycle length of 30 in the test. But here we see data from 0,10,15,20. Is this meant to be?

IanCal: > Might be a silly question. The rule has cycle length of 30 in the test. But here we see data from 0,10,15,20. Is this meant to be?

The users balance will change during the month, so we need to calculate based on that.

IanCal: > If there is a link to how you guys computed interest, it would be good so I can check by it.

Doing it by summing looking at the individual changes is in the RFC - other than that it's raising to a power of the number of periods. I can add code comments explaining what it's doing through the query

IanCal: > If there is a link to how you guys computed interest, it would be good so I can check by it.

Doing it by summing looking at the individual changes is in the RFC - other than that it's raising to a power of the number of periods. I can add code comments explaining what it's doing through the query.

I'd like to experiment with a tool that would let us split the query up and test the parts individually in some upcoming work.

tintinthong: > > Might be a silly question. The rule has cycle length of 30 in the test. But here we see data from 0,10,15,20. Is this meant to be?
> 
> The users balance will change during the month, so we need to calculate based on that.

right this makes sense

packages/cardpay-reward-programs/tests/rules/test_staking.py
```diff
@@ -0,0 +1,492 @@
...
+    rule = staking.Staking(core_config, user_config)
+    return rule
+
+
+def get_amount(result, payee, idx):
+    idx = result.index[result["payee"] == payee].tolist()[0]
+    return result.loc[idx, "amount"]
+
+
+def test_correct_calc_rewards(monkeypatch):
```
tintinthong, line 106: would be good to have a test that the rewards for DATA 
- block 0, balance:100
- block 30, balance: 100

- block 0, balance:100
- block 10, balance: 100
- block 20, balance: 100
- block 30, balance: 100

will give the same rewards in total

packages/cardpay-reward-programs/cardpay_reward_programs/rules/staking.py
tintinthong: > > > Might be a silly question. The rule has cycle length of 30 in the test. But here we see data from 0,10,15,20. Is this meant to be?
> > 
> > 
> > The users balance will change during the month, so we need to calculate based on that.
> 
> right this makes sense

Think this just means that the compounding period is the payment cycle length

packages/cardpay-reward-programs/tests/rules/test_staking.py
IanCal: yeah good idea, will add.

IanCal:
@tintinthong there's a refactor of the query here https://github.com/cardstack/cardstack/blob/staking_query_refactor/packages/cardpay-reward-programs/cardpay_reward_programs/rules/staking.py#L21

I think it's a bit clearer, am happy to go either way.

tintinthong:
> @tintinthong there's a refactor of the query here https://github.com/cardstack/cardstack/blob/staking_query_refactor/packages/cardpay-reward-programs/cardpay_reward_programs/rules/staking.py#L21
> 
> I think it's a bit clearer, am happy to go either way.

good enuff -- the comments are great.

tintinthong (approved):

tintinthong:
Pending add the refactor and minor tweaks by @IanCal

PR #3170, opened 2022-08-19
Update rewardtokenbalance to handle cryptodust

tintinthong:
This PR is related to this [PR in the wallet](https://github.com/cardstack/cardwallet/pull/1020)

This PR adds/edits SDK functions: 
- `claimAll`: This gets proofs internally (not like b4) and loops thru claims. It filters crypto dust. 
- `claimAllGasEstimate`: This gets proofs internally and cumulative gas estimates. It filters crypto dust. It also doesn't assume that the reward is in a particular token. 
- `rewardTokenBalancesWithoutDust`: This gets rewards tokens balance minus crypto dust. If a user has this output > 0 then he should be able to go into a claim state (click Claim and visit RewardClaimConfirmationSheet). The original `rewardTokensBalances` gets reward token balances with crypto dust, this is so the user can see his accumulating wealth in the wallet but that full balance should not be claimed. 

I struggled with two things with typescript and sdk organisation 
- The existence of [proof var](https://github.com/cardstack/cardwallet/blob/4d0165e968780786fca4827c96b2811636eea88a/cardstack/src/services/rewards-center/rewards-center-service.ts#L74) cache made me feel that I needed to support sdk functions that takes in proof array. After some discussion, I preferred to make `claimAll` and `claimAllGasEstimate` as abstracted functions, and I kept caching optional  responsibility of wallet by creating the intermediary function of `claimAllProofs`. It turns out caching was not too important. 
- I have bothered myself by introducing this [type](https://github.com/cardstack/cardstack/blob/263bb2cd209c48c908a524802547604d7fab6788/packages/cardpay-sdk/sdk/reward-pool/base.ts#L41-L44) and this [override](https://github.com/cardstack/cardstack/blob/263bb2cd209c48c908a524802547604d7fab6788/packages/cardpay-sdk/sdk/reward-pool/base.ts#L97-L115). Which made things rather complex when handling proof arrays like [here](https://github.com/cardstack/cardstack/blob/263bb2cd209c48c908a524802547604d7fab6788/packages/cardpay-cli/rewards/claimable-proofs.ts#L81-L92) and the use of this [guard check](https://github.com/cardstack/cardstack/blob/263bb2cd209c48c908a524802547604d7fab6788/packages/cardpay-sdk/sdk/reward-pool/base.ts#L818-L824) which was annoying. This made me always create new sdk methods which handled both types, but I benefited from just assuming having methods that handled a **SINGLE,** type e.g. a one method to handle `ClaimableProof` and another to handle `Proof` -- more methods better (my hypotheiss is toggling data types probs better in the wallet). Turns out the complexity did not blow up in the wallet either.

Post review TODO
- [ ] Will update CLI and SDK READMEs
- [ ] Cut a new release of the SDK


packages/cardpay-sdk/sdk/reward-pool/base.ts
```diff
@@ -833,8 +868,7 @@ but the balance is the reward pool is ${fromWei(rewardPoolBalanceForRewardProgra
   }
   private async getRewardTokens(): Promise<string[]> {
     let cardTokenAddress = await getAddress('cardCpxd', this.layer2Web3);
-    let daiTokenAddress = await getAddress('daiCpxd', this.layer2Web3);
-    return [cardTokenAddress, daiTokenAddress];
+    return [cardTokenAddress];
```
tintinthong, line 871: For some odd reason, claiming for dai was failing. Which I thought was actually working. So I just ensured I only served proofs with card token.

packages/cardpay-sdk/sdk/reward-pool/base.ts
```diff
@@ -267,6 +268,26 @@ export default class RewardPool {
...
+  async rewardTokenBalancesWithoutDust(
+    address: string,
+    safeAddress: string,
+    rewardProgramId?: string
+  ): Promise<WithSymbol<RewardTokenBalance>[]> {
+    const unclaimedValidProofs = (await this.getProofs(address, safeAddress, rewardProgramId, undefined, false))
+      .filter((o) => o.isValid)
+      .filter((o) => {
+        return o.gasEstimate.amount.lt(new BN(o.amount));
+      });
```
danibonilha, line 280: I've noticed there are lot's of isValid and  gasEstimates filtering, maybe we can a function that returns valid proofs with/without crypto dust.

packages/cardpay-sdk/sdk/reward-pool/base.ts
```diff
@@ -513,28 +534,36 @@ The reward program ${rewardProgramId} has balance equals ${fromWei(
...
+    const unclaimedValidProofsWithoutCryptoDust = unclaimedValidProofs.filter((o) => {
+      return o.gasEstimate.amount.lt(new BN(o.amount));
+    });
+    console.log(
+      `Claiming ${unclaimedValidProofsWithoutCryptoDust.length} out of ${unclaimedValidProofs.length} proofs`
+    );
+    return this.claimAllProofs(unclaimedValidProofsWithoutCryptoDust, safeAddress, txnOptions, contractOptions);
+  }
+
+  async claimAllProofs(
```
danibonilha, line 557: nit:  Maybe this could be called `claimProofs`,  to avoid confusing with claiming all available proofs, when I first read it I got confused on the different of `claimAll`  and `claimAllProofs`

tintinthong: Yea actually this makes a lot of sense. 

When using "All" it seems to mean **claim what you can**.  `claimAllProofs` doesn't make the assumption that mean to **claim what you can**, so you can try claiming a proof which has crypto dust too.

oidoug (approved):
LGTM, nothing to add.

packages/cardpay-sdk/sdk/reward-pool/base.ts
tintinthong: I do like the suggestion. I would prefer to separate them out like this rather than have a single method

```
  async getUnclaimedValidProofs(
    address: string,
    rewardProgramId?: string,
    tokenAddress?: string
  ): Promise<WithSymbol<Proof>[]> {
  
  async getUnclaimedValidProofsWithoutDust(
    address: string,
    safeAddress: string,
    rewardProgramId?: string,
    tokenAddress?: string
  ): Promise<WithSymbol<ClaimableProof>[]> {
```

packages/cardpay-sdk/sdk/reward-pool/base.ts
```diff
@@ -788,6 +829,14 @@ but the balance is the reward pool is ${fromWei(rewardPoolBalanceForRewardProgra
     }
   }
 
+  isClaimableProof(proof: WithSymbol<Proof> | WithSymbol<ClaimableProof>): proof is WithSymbol<ClaimableProof> {
+    if ((proof as WithSymbol<ClaimableProof>).gasEstimate) {
+      return true;
+    } else {
+      return false;
+    }
+  }
```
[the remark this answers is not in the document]
tintinthong: I think its a boolean returned tho

tintinthong:
Ok resolved reviewed points. Will be testing here with comments on wallet before release

PR #3171, opened 2022-08-20
Enable glint in boxel, web-client, ssr-web

ef4:
Opening a draft PR to aid collaboration on @lukemelia's boxel-glint-wip branch.

- [x] boxes tests passing
- [x] type feedback in vscode working
- [x] `yarn glint` passing and running in CI
- [x] consuming apps testing green


ef4:
https://github.com/cardstack/cardstack/pull/3171/commits/9b38abce9a7c008cb406eab6a62ea79da78c031b is enough to get Boxes building and its test suite passing, as well as getting Glint type feedback in vscode with the Glint plugin installed.

packages/boxel/addon/components/boxel/action-chin/action-status-area/index.gts
```diff
@@ -0,0 +1,23 @@
...
+  Blocks: {
+    default: [];
+  }
+}
+
+export default class ActionChin extends Component<ActionChinSignature> {
+  <template>
+    <div class="boxel-action-chin__action-status-area" ...attributes data-test-boxel-action-chin-action-status-area>
+      {{#if @icon}}
+        {{svgJar @icon class="boxel-action-chin__action-status-area-icon" width="20" height="20" role="presentation"}}
```
ef4, line 18: We have to import this helper now.

packages/boxel/index.js
```diff
@@ -81,7 +81,7 @@ module.exports = {
     // grab all js files from the addon directory, except utils
     let addonFiles = stew.find(
       this.treeGenerator(path.join(__dirname, 'addon')),
-      '**/*.+(js|ts|hbs)'
+      '**/*.+(js|ts|gts|hbs)'
```
ef4, line 84: You were getting trolled by this clever code that builds the app reexports tree automatically. It didn't know about gts file extensions.

packages/boxel/package.json
```diff
@@ -56,6 +56,7 @@
     "ember-qunit": "^5.1.5",
     "ember-set-body-class": "^1.0.2",
     "ember-svg-jar": "^2.3.3",
+    "ember-template-imports": "^3.0.1",
```
ef4, line 59: Since we're extending the build pipeline of a v1 addon, this needed to be in dependencies, not devDependencies.

packages/boxel/package.json
```diff
@@ -149,7 +154,7 @@
     "qunit-dom": "^2.0.0",
     "stylelint": "^14.3.0",
     "stylelint-config-standard": "^24.0.0",
-    "typescript": "^4.2.3",
+    "typescript": "^4.7.4",
```
ef4, line 157: This was the source of the `f.createClassStaticBlockDeclaration` declaration. I guessed babel version problem originally, but it was actually typescript version problem.

packages/boxel/tsconfig.json
```diff
@@ -14,7 +14,7 @@
     "noUnusedLocals": true,
     "noUnusedParameters": true,
     "noImplicitReturns": true,
-    "noEmitOnError": true,
+    "noEmitOnError": false,
```
ef4, line 17: This is necessary because ember-cli-typescript is trying to typecheck the app with tsc, which won't work because it doesn't understand gts files. Instead we are separately type checking the app with glint, which covers both ts and gts. This also means you see type errors in the console, although they are not really relevant.

We can fix this once https://github.com/embroider-build/embroider/pull/1236 releases and we upgrade to that embroider version. With that, you can remove ember-cli-typescript while still keeping typescript transpiration support.

packages/boxel/addon/components/boxel/action-chin/action-status-area/index.gts
```diff
@@ -0,0 +1,23 @@
+import Component from '@glimmer/component';
```
ef4, line 1: It's not really correct that glint is pushing us to convert from a template-only component to a glimmer component. Those are not the same thing.

We should be able to drop the wrapping class and just have a bare `<template></template>` here, and that is legal and runs but glint doesn't provide us a place to put component signature in that case. I filed https://github.com/typed-ember/glint/issues/387 to ask about this.

ef4: I got help from the dev-typescript discord channel and resolved this.

packages/boxel/addon/components/boxel/action-chin/action-status-area/index.gts
```diff
@@ -0,0 +1,24 @@
...
+const StatusArea: TemplateOnlyComponent<Signature> = <template>
+  <div class="boxel-action-chin__action-status-area" ...attributes data-test-boxel-action-chin-action-status-area>
+    {{#if @icon}}
+      {{svgJar @icon class="boxel-action-chin__action-status-area-icon" width="20" height="20" role="presentation"}}
+    {{/if}}
+    {{yield}}
+  </div>
+</template>
+
+export default StatusArea;
```
ef4, line 23: At the moment, if you want to put a correct signature on a template-only component you can't inline it into the `export default`. You need to assign it to a variable such that you can give the variable the type.

ef4:
I think the remaining work here is all in giving types to the existing components. There isn't really an incremental adoption story for glint -- it is now type-checking all the templates, so there is a lot of untyped stuff generating errors.

This is where my recommendation is to go all the way to `.gts` files, because if you try to make the types work in the traditional hbs files you'll end up doing a ton of:

```ts
declare module '@glint/environment-ember-loose/registry' {
  export default interface Registry {
    "Boxel::Button": ButtonComponent;
  }
}
```

which later becomes throwaway code when we switch to `.gts` and `<template>` anyway.

lukemelia:
Thanks for figuring all that out! I'm excited to keep pushing this forward.

packages/web-client/app/components/card-pay/dashboard-panel/section/index.gts
```diff
@@ -0,0 +1,52 @@
...
+    disclaimer: []
+  }
+}
+
+export default class CardPayDashboardPanelSection extends Component<Signature> {
+  <template>
+    <section class={{cn "dashboard-panel-section" dashboard-panel-section--bottom-aligned=(not @section.title) dashboard-panel-section--has-disclaimer=(has-block "disclaimer")}} ...attributes>
+      <div class="dashboard-panel-section__body">
+        {{#if @section.icon}}
+          {{svgJar @section.icon role="presentation" width="72"}}
```
Aierie, line 23: Oh, I didn't realise that `alt` does not work for inline svg. We should probably reconsider the use of `role=presentation` too (in another PR), it may not the right way to get inline svgs ignored by assistive tech.

Aierie (commented):
I've only looked at `web-client` so far.  Some things noted:
- there's a web3 version difference that is causing problems with the `TransactionReceipt` type, I wonder if we should have the SDK declare a peer dependency version and have consumers try to adhere to it
- There are now two ways of accessing components 
    - arbitrary names without `::` separators when imported
    - filename/directory-determined names with `::` separators when using as a global component in route templates
    - there's a similar situation for helpers and modifiers

I guess we have to live with having two or more names to refer to components, helpers, and modifiers for a while, until `.gts` route templates are supported?

A devx issue while reviewing was having to turn of Typescript and Javascript language features in order to not get noise around templates, I wonder if there's a good way to do this for specific files only?

packages/web-client/app/components/card-pay/header/workflow-tracker/item/index.gts
```diff
@@ -0,0 +1,206 @@
+import Component from '@glimmer/component';
+import { inject as service } from '@ember/service';
+import { action } from '@ember/object';
+
+import cn from '@cardstack/boxel/helpers/cn';
+import { on } from '@ember/modifier';
+import { svgJar } from '@cardstack/boxel/utils/svg-jar';
+import tippy from '@cardstack/web-client/modifiers/tippy';
+import BoxelButton from '@cardstack/boxel/components/boxel/button';
```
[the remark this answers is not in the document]
ef4: There are pros and cons to reexporting from one file. It makes the API nicer, but it requires spec-non-compliant optimizations if you want your build to only include components that get used. How much that matters probably depends on how likely people are to use only some parts of the component library and how big they are.

But even if we keep separate files for optimizability, we should definitely reorganize them into shorter paths, like:

```diff
-import BoxelButton from '@cardstack/boxel/components/boxel/button';
+import BoxelButton from '@cardstack/boxel/button';
```

packages/web-client/app/controllers/application.ts
```diff
@@ -6,6 +6,7 @@ import { action } from '@ember/object';
 import Layer2Network from '@cardstack/web-client/services/layer2-network';
 
 class ApplicationController extends Controller {
+  //@ts-ignore
```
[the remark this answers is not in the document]
lukemelia: Apparently nothing! I've removed it. This, incidentally, is the reason that I've learned we should prefer `@ts-expect-error` over `@ts-ignore`.

packages/wc-provider/tests/wallet-connect-provider-test.ts
```diff
@@ -265,6 +265,7 @@ describe('WalletConnectProvider', function () {
   describe('Ethers', () => {
     let web3Provider: providers.Web3Provider;
     before(async () => {
+      // @ts-expect-error it would be nice to resolve this
```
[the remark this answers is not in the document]
ef4: The difference is pretty thin and debatable, but there is some upstream guidance at https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-9.html#ts-ignore-or-ts-expect-error

packages/web-client/app/models/workflow/network-aware-message.ts
```diff
@@ -18,7 +18,7 @@ export default class NetworkAwareWorkflowMessage extends WorkflowMessage {
 
   get layer1Network() {
     let postable = this as WorkflowPostable;
-    let layer1Network = getOwner(postable.workflow).lookup(
+    let layer1Network = getOwner(postable.workflow)?.lookup(
```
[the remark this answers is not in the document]
ef4: I agree with this, non-null assertion seems better here. This is the kind of thing it's made for: we can see that it should be impossible to ever return undefined, but typescript is not smart enough to see that.

packages/web-client/tests/acceptance/deposit-withdrawal-test.ts
```diff
@@ -27,9 +28,9 @@ module('Acceptance | deposit and withdrawal', function (hooks) {
   });
 
   test('Depot and profile balances are listed when wallet is connected and update when the account changes', async function (this: Context, assert) {
-    let layer2Service: Layer2TestWeb3Strategy = this.owner.lookup(
-      'service:layer2-network'
-    ).strategy;
+    let layer2Service = (
+      this.owner.lookup('service:layer2-network') as Layer2Network
+    ).strategy as Layer2TestWeb3Strategy;
```
ef4, line 33: While it's sometimes necessary to do this kind of aggressive casting, when it's necessary I think it belongs in a single documented utility function, not spread throughout the codebase.

But before you go changing these all, I'm going to send you a PR that extends the type of `owner.lookup` so it works the same as service injection, such that our `declare module '@ember/service' { ... }` stuff will inform it.

packages/boxel/addon/components/boxel/org-switcher/index.gts
```diff
@@ -0,0 +1,42 @@
...
+import { on } from '@ember/modifier';
+import eq from 'ember-truth-helpers/helpers/eq';
+import or from 'ember-truth-helpers/helpers/or';
+import '@cardstack/boxel/styles/global.css';
+import './index.css';
+
+interface Signature {
+  Element: HTMLUListElement;
+  Args: {
+    orgs: any; // TODO: better types
```
backspace, line 14: is this for a future PR?

packages/web-client/app/components/card-pay/dashboard-panel/index.gts
```diff
@@ -0,0 +1,71 @@
...
+    <details
+      class={{cn
+        "dashboard-details"
+        dashboard-details--no-collapse=@noCollapse
+        dashboard-details--no-bottom-padding=@noBottomPadding
+        dashboard-details--has-footer=(has-block 'footer')
+      }}
+      ...attributes
+      {{! @glint-ignore }}
+      open={{or @noCollapse @open}}
```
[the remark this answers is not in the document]
ef4: To clarify (and I think @Aierie already meant this but it took me a while to parse it correctly in my head): the `open` attribute on native HTML `<dialog>` already accepts a boolean. The question is about whether the `@open` argument to this component should change from `Function` to `@boolean`.

I think this is a place where we've uncovered a preexisting bug. If anybody does pass a function for `@open` it will do the wrong thing and try to `toString()` the function into the value of the `open` attribute. 

If we don't want to change the call signature, the fix is:

```ts
const asBool = (value: any): boolean => Boolean(value);
<template>
  <details open={{or @noCollapse (asBool @open)}} />
</template>
```

Tangentially, it *should* be possible to just say `open={{or @noCollapse (Boolean @open)}}` since `Boolean` is a function and all functions can be helpers, but typescript and/or glint doesn't like that because of the way `Boolean` is also a constructor.

packages/web-client/tests/acceptance/deposit-withdrawal-test.ts
ef4: https://github.com/cardstack/cardstack/pull/3211

ef4:
I apparently can't review my own PR as "request changes" but I do think both of these are worth addressing before merging:

 - https://github.com/cardstack/cardstack/pull/3171#discussion_r966049167
 - https://github.com/cardstack/cardstack/pull/3171#discussion_r966076164

packages/web-client/app/components/card-pay/dashboard-panel/index.gts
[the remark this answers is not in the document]
Aierie: Sorry, I meant that `@open` is actually meant to be a `boolean` argument, but is incorrectly typed as `Function` here, maybe because we often use `@is<adjective>` to refer to states, rather than `@<adjective>`.

packages/boxel/addon/components/boxel/progress-icon/index.gts
```diff
@@ -0,0 +1,65 @@
...
+import cn from '@cardstack/boxel/helpers/cn';
+import not from 'ember-truth-helpers/helpers/not';
+import or from 'ember-truth-helpers/helpers/or';
+import '@cardstack/boxel/styles/global.css';
+import './index.css';
+
+interface Signature {
+  Element: HTMLDivElement;
+  Args: {
+    isCancelled?: boolean;
```
backspace, line 16: Unrelated to the PR, the spelling of this is something we tried to [standardise on](https://github.com/cardstack/cardstack/commit/f90762ea7feab2ca50893541106a53f172a44396), but thankfully Glint should make it easier to fix now!

packages/boxel/types/ember-truth-helpers/helpers/not.d.ts
```diff
@@ -0,0 +1,15 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
+import Helper from '@ember/component/helper';
+
+// NOTE: These types are somewhat imperfect.
+// For instance, the limit at 5 is arbitrary. Also, the actual helpers
+// use JavaScript truthiness along with special handling for empty arrays
+// and objects with an `isTruthy` method. In these cases, we just won't
+// narrow as much as we potentially could.
```
backspace, line 8: Was this meant to be in `or`?

backspace (approved):
I only looked at `packages/boxel` and it’s difficult to do a close reading but it seems good, a few minor questions/suggestions

packages/boxel/addon/components/boxel/action-chin/index.gts
```diff
@@ -0,0 +1,166 @@
+import Component from '@glimmer/component';
+import { equal } from 'macro-decorators';
+import '@cardstack/boxel/styles/global.css';
+import './index.css';
+import BoxelButton from '../button';
+import ActionStatusArea, { StatusAreaSignature } from './action-status-area';
+import InfoArea, { InfoAreaSignature } from './info-area';
+import cn from '@cardstack/boxel/helpers/cn';
+
+//@ts-ignore glint does not think this is consumed-but it is consumed in the template
```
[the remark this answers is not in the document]
ef4: Also I think it would be good to include a link to the underlying bug in the comment: https://github.com/typed-ember/glint/issues/374

packages/boxel/addon/components/boxel/card-container/usage.ts
```diff
@@ -1,7 +1,7 @@
 import Component from '@glimmer/component';
 import { tracked } from '@glimmer/tracking';
 
-export default class extends Component {
+export default class CardContainerUsage extends Component {
```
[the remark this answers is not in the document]
ef4: This is one of the reasons people argue against default exports. If we want a consistent convention it's much easier if we switch to named exports for all components. But we can't do that until we can get rid of all non-strict templates.

packages/web-client/app/models/workflow/network-aware-message.ts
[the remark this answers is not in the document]
lukemelia: Great. I'll introduce a `strongGetOwner` util we can use that bakes in this assertion.

packages/boxel/addon/components/boxel/modal/usage.hbs
```diff
@@ -25,7 +25,7 @@
 </Boxel::Modal>
 </:example>
 <:api as |Args|>
-  <Args.String @name="size" @description="Can be `small`, `medium`, `large`, or unspecified. Sets a predetermined max-width to inner modal." @value={{this.size}} @options={{array "small" "medium" "large" "n/a"}} @onInput={{fn (mut this.size)}} @defaultValue="65rem" />
+  <Args.String @name="size" @description="Can be `small`, `medium`, `large`, or unspecified. Sets a predetermined max-width to inner modal." @value={{this.size}} @options={{array "small" "medium" "large" undefined}} @onInput={{fn (mut this.size)}} @defaultValue="65rem" />
```
backspace, line 28: I tried this out in Freestyle and found that it doesn’t allow `undefined` to be chosen after another value has been chosen.

packages/web-client/tests/acceptance/deposit-withdrawal-test.ts
lukemelia: 👍🏼

packages/boxel/types/ember-truth-helpers/helpers/not.d.ts
lukemelia: Seems likely. I copied and pasted this from https://github.com/Gavant/glint-template-types/blob/main/types/ember-truth-helpers/not.d.ts

packages/boxel/addon/helpers/css-var.ts
```diff
@@ -1,4 +1,6 @@
+// prettier-ignore
 import { helper } from '@ember/component/helper';
+// prettier-ignore
```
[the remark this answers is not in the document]
lukemelia: Ah, this was leftover from the weird "import value" bug I mentioned that is fixed by restarting the VS Code extension host. I'll remove them.

packages/boxel/addon/components/boxel/org-switcher/index.gts
lukemelia: This is now TODONE!

packages/boxel/addon/components/boxel/dashboard/usage.ts
```diff
@@ -1,7 +1,7 @@
 import Component from '@glimmer/component';
 import { tracked } from '@glimmer/tracking';
 
-export default class BoxelDashboardComponent extends Component {
+export default class BoxelDashboardUsage extends Component {
```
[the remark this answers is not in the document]
lukemelia: Updated

packages/boxel/addon/components/boxel/action-chin/index.gts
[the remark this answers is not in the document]
lukemelia: Added the link and switched to ts-expect-error so we'll be notified when we upgrade to a version of glint with a fix.

packages/web-client/app/components/card-pay/header/workflow-tracker/item/index.gts
```diff
@@ -0,0 +1,206 @@
...
+  @action hideDeleteConfirmation() {
+    this.deleteConfirmDialogShown = false;
+  }
+
+  @action deleteWorkflow() {
+    this.workflowPersistence.clearWorkflowWithId(this.workflowId);
+    this.deleteConfirmDialogShown = false;
+  }
+
+  <template>
```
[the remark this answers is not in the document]
lukemelia: I've inquired in #topic-typescript on the Ember Discord, but there is no guidance as yet. Good thing to think about has we work with gts files.

packages/web-client/app/components/card-pay/dashboard-panel/section/index.gts
lukemelia: I'm open to changes around this in a separate PR.

packages/boxel/addon/components/boxel/card-container/usage.ts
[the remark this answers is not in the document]
lukemelia: For now, we can go with '<component name without "Boxel::" prefix>Usage'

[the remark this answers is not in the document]
lukemelia: Updated to make these all consistent

packages/web-client/app/utils/web3-strategies/layer1-chain.ts
```diff
@@ -337,11 +337,12 @@ export default abstract class Layer1ChainWeb3Strategy
   ): Promise<TransactionReceipt> {
     if (!this.web3) throw new Error('Cannot unlock tokens without web3');
     let tokenBridge = await getSDK('TokenBridgeForeignSide', this.web3);
-    return tokenBridge.unlockTokens(
+    let receipt = await tokenBridge.unlockTokens(
       new TokenContractInfo(tokenSymbol, this.networkSymbol).address,
       amountInWei.toString(),
       { onTxnHash }
     );
+    return receipt as TransactionReceipt;
```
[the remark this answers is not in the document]
lukemelia: I added a resolution to lock to web3-core 1.7.x

packages/cardpay-sdk/index.ts
```diff
@@ -24,6 +24,7 @@ export type { IHubAuth } from './sdk/hub-auth';
 export { default as HubConfig } from './sdk/hub-config';
 export type { AddressKeys } from './contracts/addresses';
 export type { TransactionOptions } from './sdk/utils/general-utils';
+export type { SuccessfulTransactionReceipt } from './sdk/utils/successful-transaction-receipt';
```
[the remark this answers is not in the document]
lukemelia: Not used. Removed.

packages/web-client/app/components/card-pay/header/workflow-tracker/item/index.gts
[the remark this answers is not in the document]
lukemelia: I'll save this for a future PR

packages/web-client/app/components/card-pay/header/workflow-tracker/item/index.gts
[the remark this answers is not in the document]
lukemelia: I inquired in topic-typescript on the Ember discord, but have not gotten any good feedback yet. Let's look for a standard as we start working with gts files

packages/cardpay-sdk/sdk/utils/successful-transaction-receipt.ts
```diff
@@ -1,5 +1,5 @@
-import type { TransactionReceipt as MaybeSuccessfulTransactionReceipt } from 'web3-core';
+import { TransactionReceipt } from 'web3-core';
 
-export interface SuccessfulTransactionReceipt extends MaybeSuccessfulTransactionReceipt {
+export interface SuccessfulTransactionReceipt extends TransactionReceipt {
```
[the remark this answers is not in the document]
lukemelia: I didn't think it was adding value.

lukemelia:
@Aierie wrote:

>there's a web3 version difference that is causing problems with the TransactionReceipt type, I wonder if we should have the SDK declare a peer dependency version and have consumers try to adhere to it

I added a resolution to the root package.json to try to address this.

>There are now two ways of accessing components
>arbitrary names without :: separators when imported
>filename/directory-determined names with :: separators when using as a global component in route templates
>there's a similar situation for helpers and modifiers
>I guess we have to live with having two or more names to refer to components, helpers, and modifiers for a while, until .gts route templates are supported?

Yep. That's the price of backward compatibility and a progressive migration process.

> A devx issue while reviewing was having to turn of Typescript and Javascript language features in order to not get many red squigglies around templates in gts files, I wonder if there's a good way to do this for specific files only?

I updated the README to reflect the VS Code extensions that I am using. This makes the experience pretty good on my machine.

packages/web-client/app/components/card-pay/dashboard-panel/index.gts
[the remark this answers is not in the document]
lukemelia: Elegant resolution to this: `@open` was not being used by anything!

packages/boxel/addon/components/boxel/modal/usage.hbs
lukemelia: Fixed this.

packages/boxel/types/ember-truth-helpers/helpers/not.d.ts
lukemelia: Fixed.

README.md
```diff
@@ -42,6 +42,13 @@ In development, we use `lerna` to manage the inter-dependencies of all the packa
  4. Launch the typescript compiler with `yarn compile --watch`
  5. Start ember-cli and the hub node server with `yarn start` and/or run tests with `yarn test`
 
+### Editors
+
+If you use Visual Studio Code and work on the boxel, web-client or ssr-web packages, we recommend the following extensions to make working with glint and gts files better:
+
+* [Glint language server integration for VS Code](https://marketplace.visualstudio.com/items?itemName=typed-ember.glint-vscode)
+* [VSCode extension for glimmer](https://marketplace.visualstudio.com/items?itemName=chiragpat.vscode-glimmer)
```
habdelra, line 50: perhaps it might be worth noting that you'll generally see TS errors duplicated if you have the built-in VS Code typescript extension running. When looking at errors you may also see false positives if the built-in VS Code typescript extension is running--so make sure to check the error message to see if it is being reported from glint or tsc--as TSC will not understand glint-isms. (sometimes it might actually be beneficial to turn off the built-in TS extension when working purely in a workspace that is configured for glint, since glint is performing a super set of TS validation)

ef4: Also, we might want to include instructions on how to turn off the built in typescript extension, because it's not obvious. To find it in the extensions list you need to search for `@builtin typescript`.

habdelra (approved):

README.md
lukemelia: Good ideas. Added.

PR #3176, opened 2022-08-23
Update reward services with new staging contracts

tintinthong:
- [x] register reward program. Wait for relayer
- [ ] update reward program
- [ ] deploy (all services). Wait for subgraph
- [x] Cleanup reward program bucket 
- [ ] Drop proofs db of old card token address


IanCal (approved):
Changes all look consistent to me, any issues should be clear if addresses are mixed up but I can't see any.

PR #3180, opened 2022-08-24
Update wyreIssuer and SKU after inventory rebuild

jurgenwerk:
After testnet redeploy we had to recreate the PrepaidCardMarket inventory. I did this using my wallet EOA as the issuer. The mobile app is filtering the inventory by `wyreIssuer` so this value needs to be updated now.


habdelra (approved):
thanks--sorry I didnt catch this in your earlier PR

PR #3181, opened 2022-08-24
Handle case when there are no merkle submissions in the parquet data

IanCal:
Running into this for the first time as we've reset staging.

Now checks to see if there are any submissions stored in the parquet data, if not we default to block 0.


tintinthong (approved):
Nice! Thanks!

PR #3186, opened 2022-08-25
Upgrade typescript

lukemelia:
This is a prerequisite for getting glint working in Boxel.


packages/firebase-functions/push/functions/src/routes/push.ts
```diff
@@ -29,7 +29,7 @@ export const push = async (req: Request): Promise<IResponse> => {
       code: 200,
       success: true,
     };
-  } catch (error) {
+  } catch (error: any) {
```
Aierie, line 32: Would it be better to start with `Error` as the default type for `catch`, so that any non-standard error properties require explicit checking?

Aierie (approved):

packages/firebase-functions/push/functions/src/routes/push.ts
lukemelia: I thought so too, but Typescript didn't agree with us: https://fettblog.eu/typescript-typing-catch-clauses/

backspace (approved):

habdelra (approved):
looks good

packages/firebase-functions/push/functions/src/routes/push.ts
habdelra: you could use an `unknown` instead which means that in your catch you need to do positive work to understand what the type of the error is.

PR #3188, opened 2022-08-25
Reward programs - When the min partition is not specified, look back across all time

IanCal:
The staking rule was using the start_block from the config as the first block to scan in data. This is wrong.

This is incorrect, the first block should be the earliest one available in the data. 0 doesn't quite work as the partition calculations won't work with that, but a "None" for the min partition I think is logical - in that case it will use the first block there's data. We need to look back as we only get data about a balance or safe ownership when it *changes*.

The calculations otherwise are correct, and the results would be correct if we were applying this retroactively, but otherwise we couldn't set staking to be calculated from and of Aug.

Included is a little cleanup on the tests.

### DIFF
--- a/packages/cardpay-reward-programs/tests/rules/test_staking.py
+++ b/packages/cardpay-reward-programs/tests/rules/test_staking.py
@@ -7,7 +7,6 @@ from hypothesis import example, given
 from hypothesis.extra.pandas import column, data_frames, range_indexes
 from pytest import MonkeyPatch
 
-
 END_BLOCK = 60
 CYCLE_LENGTH = 30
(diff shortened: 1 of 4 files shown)


packages/cardpay-reward-programs/tests/rules/test_staking.py
```diff
@@ -98,7 +98,7 @@ def run_query(self, table_query, vars, aux_table_query):
     return rule
 
 
-def get_amount(result, payee, idx):
+def get_amount(result, payee):
```
IanCal, line 101: idx was never used here, so I've removed it and fixed the calls in the rest of the tests.

tintinthong (approved):

PR #3189, opened 2022-08-26
hub: Add a service to fetch scheduled payments

jurgenwerk:
Ticket: CS-4379

The idea is that there will be a scheduled payments scheduler that will trigger every 5 minutes and check whether there are any scheduled payments due, and pass them to the scheduled payments executor. This service will return the scheduled payments that are due to be executed at the time of checking.

Since `pay_at` is used to query for both one-time (`pay_at` is set initially only) and recurring payments (`pay_at` is updated on every sucessful payment), we had some discussions whether it would make sense to rename it to `next_pay_at` (or have both `pay_at` and `next_pay_at`). I opted in for keeping the `pay_at` and use it for both types of scheduled payment so that we can keep the fetching query as simple as possible.

TODO: 

- [x] Fetch one time payments
- [x] Fetch recurring payments

### DIFF
--- a/packages/hub/node-tests/services/scheduled-payments/fetcher-test.ts
+++ b/packages/hub/node-tests/services/scheduled-payments/fetcher-test.ts
@@ -2,19 +2,27 @@ import shortUUID from 'short-uuid';
 import ScheduledPaymentsFetcherService from '../../../services/scheduled-payments/fetcher';
 import { setupHub } from '../../helpers/server';
 import crypto from 'crypto';
-import { addDays, addMinutes, subDays, subMinutes } from 'date-fns';
+import { addDays, addMinutes, addMonths, subDays, subMinutes, subMonths } from 'date-fns';
 import { ScheduledPayment, ScheduledPaymentAttemptStatusEnum } from '@prisma/client';
+import { convertDateToUTC } from '../../../utils/dates';
+import { ExtendedPrismaClient } from '../../../services/prisma-manager';
 
 describe('fetching scheduled payments that are due', function () {
   let subject: ScheduledPaymentsFetcherService;
+  let now: Date;
+  let validForDays: number;
+  let prisma: ExtendedPrismaClient;
+
   let { getPrisma, getContainer } = setupHub(this);
 
   this.beforeEach(async function () {
     subject = await getContainer().lookup('scheduled-payments-fetcher');
+    now = convertDateToUTC(new Date());
+    validForDays = subject.validForDays;
+    prisma = await getPrisma();
   });
 
-  let createScheduledPayment = async (payAt: Date) => {
-    let prisma = await getPrisma();
+  let createScheduledPayment = async (params: any) => {
     return prisma.scheduledPayment.create({
       data: {
         id: shortUUID.uuid(),
@@ -28,8 +36,13 @@ describe('fetching scheduled payments that are due', function () {
         feeFixedUsd: 1,
         feePercentage: 1,
         nonce: 'abc',
-        payAt,
+        payAt: params.payAt,
+        recurringDayOfMonth: params.recurringDayOfMonth,
+        recurringUntil: params.recurringUntil,
+        validForDays: params.validForDays,
         spHash: crypto.randomBytes(20).toString('hex'),
+        canceledAt: params.canceledAt,
+        creationBlockNumber: 'creationBlockNumber' in params ? params.creationBlockNumber : 1,
         chainId: 1,
       },
     });
@@ -41,7 +54,6 @@ describe('fetching scheduled payments that are due', function () {
     endedAt: Date | null,
     status: ScheduledPaymentAttemptStatusEnum
   ) => {
-    let prisma = await getPrisma();
     return prisma.scheduledPaymentAttempt.create({
       data: {
         id: shortUUID.uuid(),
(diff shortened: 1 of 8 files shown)


packages/hub/services/scheduled-payments/fetcher.ts
```diff
@@ -0,0 +1,61 @@
...
+
+  // This query fetches scheduled payments that are due to be executed now.
+  // Payment scheduler will periodically try to fetch these payments at regular intervals and execute them if they are due.
+  // This query mostly relies on the payAt field, which is set to the time when the payment should be executed.
+  // In case of one-time payments, payAt is set only initially, and then never changed.
+  // In case of recurring payments, payAt will update on every successful payment (to 1 month in the future).
+  // It is also important to note that validForDays tells us for how many days the payment is still valid to retry in
+  // case it is failing for some reason.
+
+  async fetchScheduledPayments(): Promise<any> {
```
backspace, line 17: if this used a more specific return type I think you wouldn’t need the `sp: ScheduledPayments` type in the tests when using `map`

packages/hub/services/scheduled-payments/fetcher.ts
```diff
@@ -0,0 +1,61 @@
+import { inject } from '@cardstack/di';
+import { startOfDay, subDays } from 'date-fns';
+import { convertDateToUTC } from '../../utils/dates';
+
+export default class ScheduledPaymentsFetcherService {
+  prismaManager = inject('prisma-manager', { as: 'prismaManager' });
+  validForDays = 10; // Keep this in sync with the value in the config contract
```
backspace, line 7: Can this be read out of the contract? Or how will it be kept in sync?

backspace (approved):

packages/hub/node-tests/services/scheduled-payments/fetcher-test.ts
```diff
@@ -0,0 +1,194 @@
...
+        senderSafeAddress: '0x123',
+        moduleAddress: '0x456',
+        tokenAddress: '0x789',
+        amount: 1,
+        payeeAddress: '0xabc',
+        executionGasEstimation: 1,
+        maxGasPrice: 1,
+        feeFixedUsd: 1,
+        feePercentage: 1,
+        nonce: 'abc',
```
habdelra, line 38: renamed to `salt`, right?

packages/hub/node-tests/services/scheduled-payments/fetcher-test.ts
```diff
@@ -0,0 +1,194 @@
...
+        payAt: params.payAt,
+        recurringDayOfMonth: params.recurringDayOfMonth,
+        recurringUntil: params.recurringUntil,
+        validForDays: params.validForDays,
+        spHash: crypto.randomBytes(20).toString('hex'),
```
habdelra, line 43: does the hub care what the spHash is? I thought this was something that was computed on-chain based on the parameters to contract method that the crank calls to make a payment--where the contract stores the spHash and the subsequent calls to make a payment must hash to the spHash that the contract stored

packages/hub/services/scheduled-payments/fetcher.ts
jurgenwerk: The idea is to read it from the contract. But I think the contract is not finalized yet - I'll make a ticket to adjust this once we get it deployed.

packages/hub/services/scheduled-payments/fetcher.ts
jurgenwerk: Good catch. Will fix it.

packages/hub/node-tests/services/scheduled-payments/fetcher-test.ts
jurgenwerk: Good catch. Will rename it.

packages/hub/node-tests/services/scheduled-payments/fetcher-test.ts
jurgenwerk: I think the hub doesn't care about it, it just has it saved so that it can use it to execute payments. The idea is that the client is the one who will form the `spHash` and send it to the hub which will save it along with all the params and attempt the contract call to schedule a payment. It will then use the `spHash` later when trying to execute the payment.

packages/hub/node-tests/services/scheduled-payments/fetcher-test.ts
jurgenwerk: Renamed

packages/hub/services/scheduled-payments/fetcher.ts
jurgenwerk: Fixed

packages/hub/node-tests/services/scheduled-payments/fetcher-test.ts
lukemelia: We may want to have the hub verify that the spHash it has is present in the module state. Otherwise, we should likely assume that a payment was either made or canceled out-of-band.

packages/hub/prisma/schema.prisma
```diff
@@ -356,8 +360,8 @@ enum WalletOrdersStatusEnum {
   @@map("wallet_orders_status_enum")
 }
 
-enum ScheduledPaymentAttemptsStatusEnum {
-  in_progress
+enum ScheduledPaymentAttemptStatusEnum {
+  inProgress @map("in_progress")
```
lukemelia, line 364: There seem to be a number of string versions of "inProgress" in this code still. Are we sure everything is lined up right with the formatting of this value?

lukemelia (approved):

packages/hub/node-tests/services/scheduled-payments/fetcher-test.ts
jurgenwerk: Good point. I would put that check in the service that will execute the scheduled payment.

packages/hub/prisma/schema.prisma
jurgenwerk: This just tells prisma that inProgress is maps to `in_progress` when executing the query. So it's expected the code is using `inProgress`. 

Proof (note that `in_progress` is used in SQL while `inProgress` is used in the code) :

```
Hub > service.fetchScheduledPayments()
Hub > {
  timestamp: 2022-09-02T09:42:11.516Z,
  query: 'SELECT "public"."scheduled_payments"."id", "public"."scheduled_payments"."sender_safe_address", "public"."scheduled_payments"."module_address", "public"."scheduled_payments"."token_address", "public"."scheduled_payments"."amount", "public"."scheduled_payments"."payee_address", "public"."scheduled_payments"."execution_gas_estimation", "public"."scheduled_payments"."max_gas_price", "public"."scheduled_payments"."fee_fixed_usd", "public"."scheduled_payments"."fee_percentage", "public"."scheduled_payments"."salt", "public"."scheduled_payments"."pay_at", "public"."scheduled_payments"."recurring_day_of_month", "public"."scheduled_payments"."recurring_until", "public"."scheduled_payments"."valid_for_days", "public"."scheduled_payments"."sp_hash", "public"."scheduled_payments"."chain_id", "public"."scheduled_payments"."creation_transaction_hash", "public"."scheduled_payments"."creation_block_number", "public"."scheduled_payments"."cancelation_transaction_hash", "public"."scheduled_payments"."cancelation_block_number", "public"."scheduled_payments"."created_at", "public"."scheduled_payments"."updated_at", "public"."scheduled_payments"."canceled_at" FROM "public"."scheduled_payments" WHERE ("public"."scheduled_payments"."canceled_at" IS NULL AND "public"."scheduled_payments"."creation_block_number" > $1 AND "public"."scheduled_payments"."pay_at" > $2 AND "public"."scheduled_payments"."pay_at" <= $3 AND (("public"."scheduled_payments"."recurring_day_of_month" IS NULL AND ("public"."scheduled_payments"."id") NOT IN (SELECT "t0"."id" FROM "public"."scheduled_payments" AS "t0" INNER JOIN "public"."scheduled_payment_attempts" AS "j0" ON ("j0"."scheduled_payment_id") = ("t0"."id") WHERE (("j0"."status" = $4 OR "j0"."status" = $5) AND "t0"."id" IS NOT NULL))) OR ("public"."scheduled_payments"."recurring_day_of_month" > $6 AND "public"."scheduled_payments"."recurring_until" >= $7 AND ("public"."scheduled_payments"."id") NOT IN (SELECT "t0"."id" FROM "public"."scheduled_payments" AS "t0" INNER JOIN "public"."scheduled_payment_attempts" AS "j0" ON ("j0"."scheduled_payment_id") = ("t0"."id") WHERE ("j0"."status" = $8 AND "t0"."id" IS NOT NULL))))) OFFSET $9',
  params: '[0,2022-08-22 22:00:00 UTC,2022-09-02 07:42:11 UTC,"succeeded","in_progress",0,2022-09-02 07:42:11 UTC,"in_progress",0]',
  duration: 5,
  target: 'quaint::connector::metrics'
}
```

PR #3190, opened 2022-08-26
hub: Add Prisma integration for Sentry

backspace:
This adds an integration that detects various Prisma errors classes and attaches Sentry tags based on their conditions.

To avoid filling Hub on Sentry with test issues I created a temporary project. You can see an example issue [here](https://sentry.io/organizations/cardstack/issues/3534571136/?project=6691637&query=is%3Aunresolved).


backspace:
Did you have a bigger vision for this, @Aierie? This seems like an addition vs doing anything about the long error text that you previously addressed.

Aierie:
Prisma has a couple more errors, I think getting any of these should also warrant an alert: 

> PrismaClientUnknownRequestError
PrismaClientRustPanicError
PrismaClientInitializationError 

There might be action needed from the on-call person for `PrismaClientRustPanicError` ([Prisma docs](https://www.prisma.io/docs/reference/api-reference/error-reference) mention that this may require a restart of the node process) and `PrismaClientInitializationError`.

I think it makes sense to be something that adds on to the previous long string fix.

backspace:
👍🏻 thanks, I’ve updated it now with the other errors classes. This doesn’t involve any on-call notifications though, am I correct that we don’t currently have that ability directly from Hub?

Aierie:
Sentry can send PagerDuty notifications now, via alerts, so we don't need to implement that bit in hub I think. That said, I found a `pagerduty-api` service in hub that can provide an API client, if we want to do that.

PR #3193, opened 2022-08-29
Enable scheduled payment module and meta guard

FadhlanR:
Ticket: CS-4511

- [x] enable scheduled payment module.
- [x] set meta guard.

### DIFF
--- a/packages/cardpay-cli/scheduled-payment/enable-module.ts
+++ b/packages/cardpay-cli/scheduled-payment/enable-module.ts
@@ -1,0 +1,43 @@
+import { Argv } from 'yargs';
+import { getConstant, getSDK } from '@cardstack/cardpay-sdk';
+import { getEthereumClients, getConnectionType, NETWORK_OPTION_ANY } from '../utils';
+import { Arguments, CommandModule } from 'yargs';
+
+export default {
+  command: 'enable-module <safeAddress> <gasTokenAddress>',
+  describe: 'Enable scheduled payment module on the safe',
+  builder(yargs: Argv) {
+    return yargs
+      .positional('safeAddress', {
+        type: 'string',
+        description: 'The address of the safe whose enables the scheduled payment module',
+      })
+      .positional('gasTokenAddress', {
+        type: 'string',
+        description: 'The token address (defaults to Kovan DAI)',
+      })
+      .option('network', NETWORK_OPTION_ANY);
+  },
+  async handler(args: Arguments) {
+    let { network, safeAddress, gasTokenAddress } = args as unknown as {
+      network: string;
+      safeAddress: string;
+      gasTokenAddress: string;
+    };
+    let { web3, signer } = await getEthereumClients(network, getConnectionType(args));
+    let scheduledPaymentModule = await getSDK('ScheduledPaymentModule', web3, signer);
+    let blockExplorer = await getConstant('blockExplorer', web3);
+
+    console.log(`Enabling scheduled payment module to the safe ...`);
+    let onTxnHash = (txnHash: string) =>
+      console.log(`Transaction hash: ${blockExplorer}/tx/${txnHash}/token-transfers`);
+    let { scheduledPaymentModuleAddress, metaGuardAddress } = await scheduledPaymentModule.enableModule(
+      safeAddress,
+      gasTokenAddress,
+      { onTxnHash }
+    );
+    console.log(`scheduled payment module deployed to: ${scheduledPaymentModuleAddress}`);
+    console.log(`meta guard deployed to: ${metaGuardAddress}`);
+    console.log('done');
+  },
+} as CommandModule;
--- a/packages/cardpay-sdk/sdk/constants.ts
+++ b/packages/cardpay-sdk/sdk/constants.ts
@@ -43,6 +43,8 @@ const KOVAN = {
   // https://docs.tokenbridge.net/kovan-sokol-amb-bridge/about-the-kovan-sokol-amb shows 1 for finalization rate
   // but making this the same as mainnet so that the dev experience matches prod
   ambFinalizationRate: '20',
+  relayServiceURL: 'http://localhost:8888/api',
+  subgraphURL: '',
 };
 const MAINNET = {
   apiBaseUrl: 'https://api.etherscan.io/api',
--- a/packages/cardpay-sdk/sdk/scheduled-payment-module.ts
+++ b/packages/cardpay-sdk/sdk/scheduled-payment-module.ts
@@ -9,11 +9,9 @@ import { TransactionOptions, waitForTransactionConsistency } from './utils/gener
 import { ContractOptions } from 'web3-eth-contract';
 import { executeTransaction, gasEstimate, getNextNonceFromEstimate, Operation } from './utils/safe-utils';
 import { Signer } from 'ethers';
-import { signSafeTxAsRSV } from './utils/signing-utils';
+import { signSafeTx } from './utils/signing-utils';
 import { BN } from 'bn.js';
 import { ERC20ABI } from '..';
-import { ZERO_ADDRESS } from './constants';
-
 export default class ScheduledPaymentModule {
   constructor(private layer2Web3: Web3, private layer2Signer?: Signer) {}
 
@@ -61,22 +59,6 @@ export default class ScheduledPaymentModule {
         onNonce(nonce);
       }
     }
-    let signatures = await signSafeTxAsRSV(
-      this.layer2Web3,
-      multiSendTransaction.to,
-      0,
-      multiSendTransaction.data,
-      multiSendTransaction.operation,
-      estimate.safeTxGas,
-      estimate.baseGas,
-      estimate.gasPrice,
-      estimate.gasToken,
-      ZERO_ADDRESS,
-      nonce,
-      from,
-      safeAddress,
-      this.layer2Signer
-    );
     let gnosisTxn = await executeTransaction(
       this.layer2Web3,
       safeAddress,
@@ -85,7 +67,17 @@ export default class ScheduledPaymentModule {
       multiSendTransaction.operation,
       estimate,
       nonce,
-      signatures
+      await signSafeTx(
+        this.layer2Web3,
+        safeAddress,
+        multiSendTransaction.to,
+        multiSendTransaction.data,
+        Operation.CALL,
+        estimate,
+        nonce,
+        from,
+        this.layer2Signer
+      )
     );
 
     if (typeof onTxnHash === 'function') {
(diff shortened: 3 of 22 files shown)


packages/cardpay-sdk/sdk/constants.ts
```diff
@@ -43,6 +43,8 @@ const KOVAN = {
   // https://docs.tokenbridge.net/kovan-sokol-amb-bridge/about-the-kovan-sokol-amb shows 1 for finalization rate
   // but making this the same as mainnet so that the dev experience matches prod
   ambFinalizationRate: '20',
+  relayServiceURL: 'http://localhost:8888/api',
+  subgraphURL: '',
```
FadhlanR, line 47: @lukemelia @pcjun97 can we have relay service and subgraph for kovan?

packages/cardpay-sdk/sdk/scheduled-payment-module.ts
```diff
@@ -0,0 +1,147 @@
...
+        )}, the the gas cost is ${fromWei(gasCost)} ${symbol}`
+      );
+    }
+    if (nonce == null) {
+      nonce = getNextNonceFromEstimate(estimate);
+      if (typeof onNonce === 'function') {
+        onNonce(nonce);
+      }
+    }
+    let signatures = await signSafeTxAsRSV(
```
habdelra, line 64: Let's use `signSafeTX` from the cardpay-sdk/sdk/utils/signing-utils.ts module--it's more concise

packages/cardpay-sdk/sdk/utils/module-utils.ts
```diff
@@ -0,0 +1,89 @@
...
+  };
+}
+
+const encodePacked = (tx: any) =>
+  pack(
+    ['uint8', 'address', 'uint256', 'uint256', 'bytes'],
+    [tx.operation || Operation.CALL, tx.to, tx.value, hexDataLength(tx.data), tx.data]
+  );
+
+const remove0x = (hexString: string) => hexString.substr(2);
```
habdelra, line 89: for these 2 functions is it necessary to use `=>`? fat arrows perform a .`bind()` to the calling context, but we are not using `this` in either of the functions. Probably we can just use normal module scoped functions here:
```js
function encodePacked() { ... }
function remove0x() { ... }
```

Also for the type of `tx`, could we use a better type than `any`? In the `encodeMultiSend()` you are using a `Transaction[]` type--is this not the type that we should be using?.

Also, you couse use an idempotent remove0x() that use the logic `hexString.replace(/^0x/, '')` this logic could be run mulitple times against the hex and always return the current value, where `hexString.substr(2)` will not.

packages/cardpay-sdk/sdk/utils/module-utils.ts
```diff
@@ -0,0 +1,89 @@
...
+  data: string;
+  operation: Operation;
+}
+
+export async function deployAndSetUpModule(
+  web3: Web3,
+  masterCopy: Contract,
+  setupArgs: {
+    types: string[];
+    values: any[];
```
habdelra, line 24: can we use a better type here?

packages/cardpay-sdk/sdk/constants.ts
habdelra: this is possible, but it's not an insignificant amount of work--likely this would need a separate project on its own to setup

packages/cardpay-sdk/sdk/scheduled-payment-module.ts
```diff
@@ -0,0 +1,147 @@
...
+    );
+
+    if (typeof onTxnHash === 'function') {
+      await onTxnHash(gnosisTxn.ethereumTx.txHash);
+    }
+
+    return {
+      scheduledPaymentModuleAddress: enableModuleTxs.expectedModuleAddress,
+      metaGuardAddress: setGuardTxs.expectedModuleAddress,
+      txnReceipt: await waitForTransactionConsistency(this.layer2Web3, gnosisTxn.ethereumTx.txHash, safeAddress, nonce),
```
habdelra, line 98: so based on our protocol discussion, it sounds like there is client communication to the server to enable a scheduled payment. Perhaps, the client sends the hub the all the parameters that we are currently sending to `executeTransaction`, and the hub actulaly executes the transaction and waits for it, during which it updates it's DB state, and then returns the the client at this point, so that the promise includes all the hub communication as well as txn mining. One argument, though, might be that it's not feasible for the web server to hold on to a client request for the amount of time that it takes to mine the transaction--which likely will be many minutes in ethereum mainnet. That is simply too long for a web server to hold on to a connection. Rather the hub could return the txn hash to the client, and the client can just use web3 to wait for the txn to be mined. @lukemelia @jurgenwerk  it would be good to get your thoughts here too.

jurgenwerk: @habdelra you're right, that's how we specified the general mechanics in the RFC - the client sends params, and waits for the hub to do "everything": executing the transaction, waiting for the transaction to be mined, and returning the result. 

I remember our talk with @alex-cardstack and @FadhlanR where we were discussing the feasibility of giving a single http request so much work and I believe we didn't anticipate these transactions to be taking more than a couple of seconds (maybe we were thinking this is going to be on gnosis chain only?) - I believe that's why we opted for simplifying this by wrapping all the work in a single request. 

In case this can take several minutes then this surely needs to be done async. I'm not sure giving the client the transaction hash and let it wait for it to resolve is the right way to go since our hub is planned to keep track of the transactions so that it can know whether they succeeded or not which then tells whether the scheduled payment has been registered on the contract successfully. One alternative option is to wrap this into a background job in the hub and let the client poll for the job status (the same we did recently for profile creation feature)

But we're talking about creating/canceling a scheduled payment right? Which is a feature separate from this PR which doesn't include any Hub communication as far as I can see

FadhlanR: @jurgenwerk do we need to store mapping data of user address, safe address, and module address in the crank? If yes, I think it's better we use the same pattern as scheduling/canceling for enabling module.

packages/cardpay-sdk/sdk/scheduled-payment-module.ts
FadhlanR: Updated

packages/cardpay-sdk/sdk/utils/module-utils.ts
FadhlanR: Updated

packages/cardpay-sdk/sdk/scheduled-payment-module.ts
jurgenwerk: Currently, the eoa, the safe and module address are saved in the scheduled payment row in the DB. When client creates a scheduled payment it sends all these addresses together with the rest of the params and all this gets saved into the scheduled payments table. So we might not need additional actions to maintain that mapping of addresses

habdelra: background job is a nice idea

FadhlanR: Based on discussions after the daily meeting, we decided to execute the transaction in the SDK, since the hub would have nothing to do with the enabling process, and I removed `await waitForConsistency` because we don't need it here.

habdelra: @FadhlanR the hub doesn't have anything to do with enabling--but it still needs to know that there are schedule payments and populate the DB with the schedule info, so there is definitely communication required between the client and the hub as part of the establishing a scheuled payment. I would think that it would be more reliable to have the server broker the on-chain txn and grab from it any relavant details that it needs.

FadhlanR: @habdelra I agree to schedule a payment there must be communication between the client and the hub, and the `executeTransaction` must be triggered by the hub. And I also agree we can't rely on the web client to wait for transaction status. But this function is to enable the scheduled payment module, not to schedule payment, So I think we can execute the transaction in SDK and then send a callback via onTxHash callback function and let client do anything with txHash such as send it to the server broker to retrieve the transaction status as you said or just return the etherscan link to the user.

alex-cardstack (approved):
This is great, thanks, I like the use of multisend for setup I hadn't thought of doing that before

jurgenwerk (approved):
LGTM!

packages/cardpay-cli/scheduled-payment/enable-module.ts
```diff
@@ -0,0 +1,43 @@
...
+  describe: 'Enable scheduled payment module on the safe',
+  builder(yargs: Argv) {
+    return yargs
+      .positional('safeAddress', {
+        type: 'string',
+        description: 'The address of the safe whose enables the scheduled payment module',
+      })
+      .positional('gasTokenAddress', {
+        type: 'string',
+        description: 'The token address (defaults to Kovan DAI)',
```
lukemelia, line 17: If anything, this should default to Goerli DAI since we have decided to use goerli as our testnet. This can wait for a separate PR.

packages/cardpay-cli/scheduled-payment/enable-module.ts
```diff
@@ -0,0 +1,43 @@
...
+    let { network, safeAddress, gasTokenAddress } = args as unknown as {
+      network: string;
+      safeAddress: string;
+      gasTokenAddress: string;
+    };
+    let { web3, signer } = await getEthereumClients(network, getConnectionType(args));
+    let scheduledPaymentModule = await getSDK('ScheduledPaymentModule', web3, signer);
+    let blockExplorer = await getConstant('blockExplorer', web3);
+
+    console.log(`Enabling scheduled payment module to the safe ...`);
```
lukemelia, line 31: Let's use `Enabling scheduled payment module on safe ${safeAddress}`

packages/cardpay-cli/scheduled-payment/enable-module.ts
```diff
@@ -0,0 +1,43 @@
...
+    let onTxnHash = (txnHash: string) =>
+      console.log(`Transaction hash: ${blockExplorer}/tx/${txnHash}/token-transfers`);
+    let { scheduledPaymentModuleAddress, metaGuardAddress } = await scheduledPaymentModule.enableModule(
+      safeAddress,
+      gasTokenAddress,
+      { onTxnHash }
+    );
+    console.log(`scheduled payment module deployed to: ${scheduledPaymentModuleAddress}`);
+    console.log(`meta guard deployed to: ${metaGuardAddress}`);
+    console.log('done');
```
lukemelia, line 41: Let's use `Done. The module and guard has now been enabled on safe ${safeAddress}`

packages/cardpay-sdk/sdk/constants.ts
lukemelia: To confirm, we have decided to use Goerli for this, so we'll need to configure that in a future PR. We will need a relay server but not a subgraph for now.

FadhlanR: Correct, we need to configure this after we have relay server for goerli

PR #3194, opened 2022-08-29
claimable proofs must only be valid 

tintinthong:
since I estimated the gas high up in data chain (during `getProofs` call), that means non valid proofs were being estimated too. This is a wrong state, if a proof is non valid it cannot be estimated (`eth_call` would give error). This adds the filter to `getProofs`


danibonilha (approved):
LGTM

PR #3195, opened 2022-08-29
ssr: Update to not report 404s to Sentry

backspace:
With this deployed to staging, a 404 like [this](https://sentry.io/organizations/cardstack/issues/3494423291/events/a776cd978cc9408eb1be591eb1c6398c/?project=6223279&query=is%3Aunresolved) is no longer logged to Sentry.

### DIFF
--- a/packages/ssr-web/app/routes/not-found.ts
+++ b/packages/ssr-web/app/routes/not-found.ts
@@ -3,6 +3,8 @@ import { inject as service } from '@ember/service';
 import ProfileService from '@cardstack/ssr-web/services/profile';
 import RouterService from '@ember/routing/router-service';
 
+export const NotFoundError = new Error('404: Not Found');
+
 export default class NotFoundRoute extends Route {
   @service declare router: RouterService;
   @service('profile') declare profile: ProfileService;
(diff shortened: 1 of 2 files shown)


packages/ssr-web/app/routes/not-found.ts
```diff
@@ -3,6 +3,8 @@ import { inject as service } from '@ember/service';
 import ProfileService from '@cardstack/ssr-web/services/profile';
 import RouterService from '@ember/routing/router-service';
 
+export const NotFoundError = new Error('404: Not Found');
```
Aierie, line 6: I would make this a subclass of `Error` in case there is some handler that attaches some extra data to the error.

Aierie (commented):
Should this also be applied to profiles?

packages/ssr-web/app/routes/not-found.ts
backspace: 👍🏻 done, thanks

backspace:
> Should this also be applied to profiles?

I didn’t originally think so since none of the errors in Sentry are from that but now I think it’s fine hehe thanks

PR #3197, opened 2022-08-31
boxel: Add vertical option for Participant

backspace:
This has been removed from the AP designs but follows a pattern of other components.


lukemelia (approved):
👍

PR #3198, opened 2022-08-31
boxel: Add multiline option for Input

backspace:
`textarea`-specific attributes don’t need special handling thanks to `…attributes`. The change is rendered [here](https://boxel.stack.cards/boxel/input-multiline-cs-4467/#/docs?s=Components&ss=%3CBoxel%3A%3AInput%3E).


lukemelia (approved):
I didn't expect it to be that simple...

PR #3199, opened 2022-08-31
boxel: Add CancelButton to default ActionChin

backspace:
Some ERP designs include a `CancelButton` in the `ActionChin` when the state is not `in-progress`, so this enables that. It can be seen [here](https://boxel.stack.cards/boxel/action-chin-cancel-4466/#/docs?s=Components&ss=%3CBoxel%3A%3AActionChin%3E) when the preview deployment is complete.


lukemelia (approved):
Looks good

PR #3200, opened 2022-09-01
textarea uses input focus color

Aierie:


backspace (approved):
Thanks for catching this!

PR #3201, opened 2022-09-01
Add command to provision a safe

FadhlanR:
Ticket: CS-4525


lukemelia (dismissed):
I think the intention is to use CREATE2 for our safe creation. /cc @alex-cardstack @habdelra

habdelra:
@lukemelia we’ll need to think thru what the salt should be if we want create2. Like what is the “vary” for making a safe. I’m pretty sure we want EOA’s to make more than a single safe in a network, but I’m not sure what the limit is beyond that.

lukemelia:
Thanks @habdelra. That makes sense. I confirmed with @christse that we should NOT use CREATE2 for this. Sorry for the distraction @FadhlanR!

packages/cardpay-sdk/sdk/utils/safe-utils.ts
```diff
@@ -75,6 +86,35 @@ export interface GnosisExecTx extends RelayTransaction {
...
+      'Content-Type': 'application/json',
+      Accept: 'application/json',
+    },
+    body: JSON.stringify({
+      owners,
+      threshold,
+      saltNonce,
+      paymentToken,
+    }),
+  };
```
habdelra, line 110: I don't think we can use this--this is _specifically_ for create2 safes. As @lukemelia mentioned chris won't like to not use create2 safes since we don't have a salt that we want to use (a "vary" that derives the conterfactual safe addresses). Rather we need to use non-create2 safes that don't have a salt as an input

habdelra: we probably need to  do this all using web3, we can just call the gnosis safe proxy factory to create a non-create2 safe, and then another txn to send our tokens into the newly created safe.

An alternative to this though could be that we make a custom contract whose job is to create a safe and then transfer tokens from the msg.sender to the safe. This would probably need to be a 2 txn though (because of erc-20 approve flow), as the first txn would need to set an allowance on this new contract, and the 2nd txn would be the txn that creates the safe and moves the tokens into the safe. since this is also 2 txns, not sure this would buy us much... if we had erc-677 tokens though, it would be just a single txn. but that is wishful thinking....

packages/cardpay-sdk/sdk/safes.ts
```diff
@@ -217,6 +226,63 @@ export async function viewSafe(network: 'gnosis' | 'xdai' | 'sokol', safeAddress
 export default class Safes implements ISafes {
   constructor(private layer2Web3: Web3, private layer2Signer?: Signer) {}
 
+  async createSafe(
+    owners: string[],
+    threshold: number,
+    tokenAddress: string,
+    saltNonce?: string,
+    txnOptions?: TransactionOptions
+  ) {
```
habdelra, line 235: generally all the SDK functions have the ability to accept a txn hash as well so that if the txn is still not confirmed the SDK function will return a promise after the txn has been mined. This allows us to resume a workflow that is in progress and waiting for txns to be mined. Here are a bunch of examples: https://github.com/cardstack/cardstack/blob/main/packages/cardpay-sdk/sdk/prepaid-card/base.ts#L108-L111, https://github.com/cardstack/cardstack/blob/main/packages/cardpay-sdk/sdk/prepaid-card/base.ts#L183-L186, 
https://github.com/cardstack/cardstack/blob/main/packages/cardpay-sdk/sdk/revenue-pool/base.ts#L114-L117,
https://github.com/cardstack/cardstack/blob/main/packages/cardpay-sdk/sdk/revenue-pool/base.ts#L207-L214,
https://github.com/cardstack/cardstack/blob/main/packages/cardpay-sdk/sdk/prepaid-card-market-v-2/base.ts#L59-L62,
https://github.com/cardstack/cardstack/blob/main/packages/cardpay-sdk/sdk/prepaid-card-market-v-2/base.ts#L308-L311

FadhlanR:
I just updated the code, I added two approaches to provide a new safe and enable modules:

In the 1st approach, we have two different commands to provision new safe and enable modules.
1. Use this command to create a new safe `yarn cardpay safe create [threshold] [tokenAddress] [owners...]`. In this command, the user can create a new safe with a non-native coin, because we create the safe with relay server help.
2. After the safe have created, use this command to enable the SP module and set meta guard on the safe, `yarn cardpay schedule-payment enable-module <safeAddress> <gasTokenAddress>`. Basically, this command wraps four transactions into one multi-send transaction, which is deploying SP module proxy, enabling SP module, deploying meta guard, and setting meta guard. The multi-send transaction will be executed as a safe transaction because treated as a safe transaction user can pay the gas with a non-native coin.

The 2nd approach wraps safe creation and module enablement transactions into one multi-send-call-only transaction. The transaction will be executed by the user so he needs to pay the gas with a native coin. This is the command for the second approach `yarn cardpay schedule-payment create-safe`.

habdelra:
@FadhlanR I don't understand why we are using the relay server to create safe--as I said before that is to for create2 safes. What are we supposed to use for salt in that scenario?

habdelra (changes requested):
We need to come to a consensus on create2 safes. this solution is predicated on using create2 safes, but i don't think we can support that because we have not identified a "vary" that can work for this.

habdelra:
The main challenge is that this could easily be attacked. If an evil entity wanted to wreak havok for us and they knew our salt, they could race ahead of us and create safes for all the salt that we plan to use especially if our salt is easily guessable--like a sequental time series. then all the safe creations that we attempt would fail because a safe already exists at that address.

habdelra:
if we want to have an ever changing vary so that the `create2` behavior mimics the original `create` behavior, perhaps a random UUID might be more successful for a salt instead of a time stamp.

habdelra:
yeah, this is not about stealing funds--this is about trolling. the malicious entity would be creating safes that they could collect no funds with, but that would hinder our operational ability.

FadhlanR:
Hmm, I see, do attacks like that happen often? because it will harm hackers without any funds to be stolen.

FadhlanR:
For a more advanced solution, I think we can modify the factory contract and include `tx.origin` in `salt`.

`salt = saltNonce + initializer + tx.origin`

habdelra:
that's an interesting idea that definitely solves using create2 for malicious reasons, but may have unintended side effects for making counterfactual safes

packages/cardpay-sdk/sdk/utils/module-utils.ts
```diff
@@ -26,7 +22,7 @@ export async function deployAndSetUpModule(web3: Web3, masterCopy: Contract, set
 
   let encodeInitParams = web3.eth.abi.encodeParameters(setupArgs.types, setupArgs.values);
   let moduleSetupData = masterCopy.methods.setUp(encodeInitParams).encodeABI();
-  let saltNonce = new Date().getTime().toString();
+  let saltNonce = String(getRandomInt(1000000));
```
FadhlanR, line 25: @habdelra I updated the saltNonce value with random integer, it is not possible to use UUID V4 as the value of saltNonce because the data type of saltNonce in GnosisSafeProxFactorty and ModuleProxyFactory is uint256.

habdelra: a uuid v4 has 128 bits. a random number between 1 and 1000000 has 20 bits. we might wanna have a much larger random number to prevent collisions, or perhaps concat the random number to the current time.

habdelra: another idea is to convert a uuid v4 string to it's code points (aka char codes) which are numbers.

habdelra: we also might wanna think about namespacing too. if we decide to make safes for some other thing that uses create2, that will prevent the safes from colliding. like a prefix of "cardstack-scheduled-payments" in front of the UUID beefore we convert it into code points.

FadhlanR: Got out of range error when using uuidv4 codepoint as saltNonce

habdelra: did you get the out of range in solidity or javascript--I think that number above can fit into uint256. in javascript you need to treat this as string concat and not integer math (like my example above shows).

habdelra: I just checked and the max uint256 looks like 115792089237316195423570985008687907853269984665640564039457584007913129639935, which looks like  4 decimal points larger than the example I posted above

FadhlanR: I got that error in javascript because the library function like `ethers.utils.solidityKeccak256` will treat uint256 as BigNumber and it causes the out-of-range error. I need to use that library function to calculate the module proxy address.

habdelra: that seems like a bug--since this is a legit uint256 number. can you try truncating the number by a single digit at a time and see what the largest number is that works? (in terms of how many digits you can use)

packages/cardpay-sdk/sdk/utils/safe-utils.ts
```diff
@@ -252,12 +298,75 @@ export async function executeSend(
...
+
+  return { expectedSafeAddress, create2SafeTx };
+}
+
+async function calculateCreateProxyWithNonceAddress(
+  gnosisSafeProxyFactory: Contract,
+  masterCopyAddress: string,
+  initializer: string,
+  saltNonce: number
```
habdelra, line 349: wont this be a problem as a number if we are using really large uint256 vlaues?

FadhlanR: right, will change it to string

packages/cardpay-sdk/sdk/utils/module-utils.ts
habdelra: ok, that actually makes sense since 78 digits could be larger than max uint256. let's just truncate to to 77 digits and I think we'll be ok.

FadhlanR: updated

packages/cardpay-sdk/sdk/utils/safe-utils.ts
FadhlanR: updated

PR #3202, opened 2022-09-01
boxel: Add TabBar

backspace:
The issue called for the interface to include `@selectedItem` and `@onClickItem` but since this is so menu-like I chose to follow the pattern of `Boxel::Menu`, which takes a list of `(menu-item)` as `@items`. Then the items have their actions specified alongside them, or can use `(link)`, which is likely the most common usage. `ember-link` supplies the `isActive` property for `(link)`, which is used for CSS styling.

Though the design didn’t specify any hover behaviour, it seemed sensible to me to make the text bold. This includes an [approach from CSS-Tricks](https://css-tricks.com/bold-on-hover-without-the-layout-shift/) that lets the bolding not cause the following items to shift; without it, it would look like this:

bold-on-hover-shift

One possible avenue for improvement is that the links are currently `href="#"`, which means they can’t be copied or opened in a new tab. The current implementation supports arbitrary actions in the items, but if they were all `(link)`, they would act as expected. Is this the way to go? 🤔

### DIFF
--- a/packages/boxel/addon/components/boxel/tab-bar/index.hbs
+++ b/packages/boxel/addon/components/boxel/tab-bar/index.hbs
@@ -1,6 +1,6 @@
 <nav role="tablist" class={{cn "boxel-tab-bar" @class}} ...attributes>
-  {{#if @items}}
-    {{#each (compact @items) as |tab|}}
+  {{#if this.linkItems}}
+    {{#each this.linkItems as |tab|}}
       <div
         role="none"
         class={{cn
@@ -9,14 +9,14 @@
         }}
       >
         {{!-- template-lint-disable require-context-role --}}
-        <a
+        <LinkTo
+          @route={{tab.action.routeName}}
+          @query={{or tab.action.queryParams (hash)}}
           role="tab"
-          href={{tab.action.url}}
           data-text={{tab.text}}
-          {{on "click" (fn this.invokeMenuItemAction tab.action)}}
         >
           {{tab.text}}
-        </a>
+        </LinkTo>
       </div>
     {{/each}}
   {{/if}}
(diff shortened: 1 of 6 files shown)


lukemelia (approved):

packages/boxel/addon/components/boxel/tab-bar/index.hbs
```diff
@@ -0,0 +1,23 @@
...
+    {{#each (compact @items) as |tab|}}
+      <div
+        role="none"
+        class={{cn
+          "boxel-tab-bar__item"
+          boxel-tab-bar__item--is-active=tab.action.isActive
+        }}
+      >
+        {{!-- template-lint-disable require-context-role --}}
+        <a
```
Aierie, line 12: Why did we choose `a` over `button` for this? If we do `event.preventDefault` and the `href` is not representative of triggered navigation,  it seems like we could use `button` instead.

Aierie (approved):
Looks good!

It's possible to improve this with keyboard interaction handlers:

https://www.w3.org/WAI/ARIA/apg/patterns/tabpanel/#:~:text=.-,Keyboard%20Interaction,-For%20the%20tab

packages/boxel/addon/components/boxel/tab-bar/index.hbs
backspace: True! I just copied this from `Menu`, I’m not sure why it uses `a`. BUT I am now convinced I want these to be true links, I always resent when I know something is link-like and developer decisions have made me unable to use it like one 😬

backspace:
I failed to visit the WAI link until now, thanks for that @Aierie, I’ll revisit with that in mind.

packages/boxel/addon/components/boxel/tab-bar/index.hbs
Aierie: We still can't open things in a new tab if we're preventing the default event and navigating with Javascript. We might need different HTML + event handling for actions and navigation?

PR #3203, opened 2022-09-03
Add setup account cli command 

tintinthong:
This is a devX change that sets up safes based on our card pay protocol. 

When a tester wants to debug a particular state of an account (particularly for UI), 

A dev will execute this command
```
cardpay testing create-account true false false true --network kovan --mnemonic <ur mnemonic>
```

And share the following information with a tester

```
  WARNING: NOT A SECURE MNEMONIC. ONLY INTENDED FOR TESTING!

      Mnemonic: tornado reduce emotion palm asthma rural solid family million coral laptop grant 
      Address: 0x79212DA5173b4e42c1e05C0eF0c98D065497D11b
      Prepaid card: No prepaid card created 
      Merchant safe: No merchant safe created
      Reward safe: No reward safe created
      Depot safe: 0x63f4B6C796E2b74e9DB49FABeB17D785ff8C248B
```

The tester than includes loads the mnemonic into their wallet account.

**Points**
- the signing of this command occurs from the account specified `-w`, `-t`, or `--mnemonic`. The signing only truly occurs when a `createDepot=true`. Getting a prepaid card occurs just from prepaid card market provisioning. Registering a merchant and reward safe will use the new mnemonic generated to sign the prepaid card action. 
- prepaid cards are provisioned by a hardcoded sku on staging. This is setup by me and has 1000 spend per prepaid card by default. This command requires provisioner rights otherwise, the command will not work. production is not set up.
- As above, this means that the dev can provide funds to this testing account without additional work of funding a new account. A lot of times, in order to recreate an cardpay account a person has to `fund a new account`,`bridge`, `create a prepaid card`; now we don't have to `fund a new account`. 
- One can decide which token to bridge and how much using `bridgeTokenAmount`(the default is 10)  and `bridgeTokenAddress` (the default is dai). 

**Extras**
- added some market v2 commands
- added guard for `add-sku`
- rearrange launch file

### DIFF
--- a/packages/cardpay-cli/testing/create-account.ts
+++ b/packages/cardpay-cli/testing/create-account.ts
@@ -59,7 +59,7 @@ export default {
     const createdMnemonic = generateMnemonic();
 
     const web3OptsL1 = getConnectionType(args);
-    const createdWeb3Opts: Web3OptsMnemonic = {
+    const createdWeb3OptsL2: Web3OptsMnemonic = {
       connectionType: 'mnemonic',
       mnemonic: createdMnemonic,
     };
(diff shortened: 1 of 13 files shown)


tintinthong:
@jurgenwerk @backspace @oidoug. Open to suggestions on how this cli command could be used

packages/cardpay-sdk/sdk/prepaid-card-market-v-2/base.ts
```diff
@@ -181,6 +181,11 @@ export default class PrepaidCardMarketV2 {
     marketAddress = marketAddress ?? (await getAddress('prepaidCardMarketV2', this.layer2Web3));
     let { nonce, onNonce, onTxnHash } = txnOptions ?? {};
     let from = contractOptions?.from ?? (await this.layer2Web3.eth.getAccounts())[0];
+    let contract = new this.layer2Web3.eth.Contract(PrepaidCardMarketV2ABI as AbiItem[], marketAddress);
+    const issuerEOA = await contract.methods.issuers(issuerSafe).call();
+    if (issuerEOA == ZERO_ADDRESS) {
+      throw new Error('Issuer does not exists on prepaid card market v2 contract. Consider adding tokens first.');
+    }
```
tintinthong, line 188: @jurgenwerk I added a guard rail here that may affect the sdk

packages/cardpay-cli/testing/create-account.ts
```diff
@@ -0,0 +1,270 @@
...
+  const prepaidCardMgr = await getSDK('PrepaidCard', web3);
+  let blockExplorer = await getConstant('blockExplorer', web3);
+
+  console.log(
+    `Provisioning a prepaid card from the SKU ${sku} to the EOA ${createdOwner} in the ${environment} environment...`
+  );
+
+  let quantity = await prepaidCardMarketV2.getQuantity(sku);
+
+  if (quantity === 0) {
```
jurgenwerk, line 152: Will this comparison work given it says here the quantity is returned as string? https://github.com/cardstack/cardstack/blob/main/packages/hub/node-tests/routes/email-card-drop-request-test.ts#L54

jurgenwerk (approved):
This will be very useful! Just one minor comment

packages/cardpay-cli/testing/create-account.ts
tintinthong: Good catch!

PR #3204, opened 2022-09-05
Update wording of enable SP module command

FadhlanR:


packages/cardpay-cli/scheduled-payment/enable-module.ts
```diff
@@ -38,6 +38,6 @@ export default {
     );
     console.log(`scheduled payment module deployed to: ${scheduledPaymentModuleAddress}`);
```
jurgenwerk, line 39: It would look better if the capitalization of the sentences in console log outputs was consistent

FadhlanR: updated

jurgenwerk (approved):

PR #3205, opened 2022-09-05
Estimate scheduled payment execution

FadhlanR:
Tickets: CS-4540, CS-4541

- [x] Estimate execution gas of scheduled one-time payment.
- [x] Estimate execution gas of scheduled recurring payment.

### DIFF
--- a/packages/cardpay-cli/scheduled-payment/estimate-execution.ts
+++ b/packages/cardpay-cli/scheduled-payment/estimate-execution.ts
@@ -96,7 +96,7 @@ export default {
       salt: string;
       payAt: number;
       gasPrice: string;
-      recurringDayOfMonth: number,
+      recurringDayOfMonth: number;
       recurringUntil: number;
     };
     let { web3, signer } = await getEthereumClients(network, getConnectionType(args));
--- a/packages/cardpay-cli/scheduled-payment/estimate-execution.ts
+++ b/packages/cardpay-cli/scheduled-payment/estimate-execution.ts
@@ -5,7 +5,7 @@ import { Arguments, CommandModule } from 'yargs';
 
 export default {
   command:
-    'estimate-execution <moduleAddress> <tokenAddress> <amount> <payeeAddress> <fixedUSDFee> <percentageFee> <maxGasPrice> <gasTokenAddress> <salt> <payAtOrRecursDayOfMonth> <gasPrice>',
+    'estimate-execution <moduleAddress> <tokenAddress> <amount> <payeeAddress> <fixedUSDFee> <percentageFee> <maxGasPrice> <gasTokenAddress> <salt> <payAt> <gasPrice>',
   describe: 'Enable scheduled payment module on the safe',
   builder(yargs: Argv) {
     return yargs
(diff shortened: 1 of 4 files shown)


packages/cardpay-sdk/sdk/scheduled-payment-module.ts
```diff
@@ -135,4 +142,101 @@ export default class ScheduledPaymentModule {
...
+    tokenAddress: string,
+    amount: string,
+    payeeAddress: string,
+    fee: Fee,
+    maxGasPrice: string,
+    gasTokenAddress: string,
+    salt: string,
+    recursDayOfMonth: number,
+    gasPrice: string,
+    until?: number
```
jurgenwerk, line 169: Is this recurringUntil?

packages/cardpay-sdk/sdk/scheduled-payment-module.ts
```diff
@@ -135,4 +142,101 @@ export default class ScheduledPaymentModule {
...
+  async estimateExecutionGas(
+    moduleAddress: string,
+    tokenAddress: string,
+    amount: string,
+    payeeAddress: string,
+    fee: Fee,
+    maxGasPrice: string,
+    gasTokenAddress: string,
+    salt: string,
+    recursDayOfMonth: number,
```
jurgenwerk, line 167: Can we name this `recurringDayOfMonth` to be consistent with the namings in from the RFC?

packages/cardpay-cli/scheduled-payment/estimate-execution.ts
```diff
@@ -0,0 +1,120 @@
...
+      })
+      .positional('gasTokenAddress', {
+        type: 'string',
+        description: 'The address of the gas token',
+      })
+      .positional('salt', {
+        type: 'string',
+        description: 'Arbitrary string to make SP unique',
+      })
+      .positional('payAtOrRecursDayOfMonth', {
```
jurgenwerk, line 52: Can we have two different args for these fields? Since they can have quite different values I feel it's better to separate them

jurgenwerk (approved):
Can we get rid of some of the params that need to be provided in the CLI command? I was thinking we could generate some default values in the SDK part. For example, will the execution gas change with different values of salt, payAt, recurringDayOfMonth or can we just use some hardcoded values for that? 

Other than that I think this looks good 👍

habdelra (approved):
I'd 2nd the comments that matic made. aside from the minor naming nits looks good.

packages/cardpay-sdk/sdk/scheduled-payment-module.ts
FadhlanR: renamed

packages/cardpay-sdk/sdk/scheduled-payment-module.ts
FadhlanR: renamed

packages/cardpay-cli/scheduled-payment/estimate-execution.ts
FadhlanR: updated

jurgenwerk:
@FadhlanR was it not possible to remove some of the CLI command params and use default values for that as suggested in the comment?

PR #3210, opened 2022-09-07
boxel: Add documentation for CSS variables

backspace:
This adds Freestyle documentation for `--boxel-*` CSS variables. Some components already had them documented ([`OrgHeader` for instance](https://boxel.stack.cards/main/#/docs?s=Components&ss=%3CBoxel%3A%3AOrgHeader%3E)), this attempts to cover the rest, though it can be difficult to know which are meant to be externally-configurable and which are internal-only; maybe in the future we could adopt a naming convention for “intimate” ones.

An example of that is `--boxel-searchbox-input-color`; since it’s [on the `input`](https://github.com/cardstack/cardstack/blob/823185ccf3348221988aff2dd0496ee4d54b26bd/packages/boxel/addon/components/boxel/searchbox/index.css#L13) child of the component which isn’t exposed to the consumer, it cannot be overridden. `--boxel-searchbox-icon-size` can however!

`Button` is more ambiguous; there are various size and colour overrides available but since they’re `@kind`- and `@size`-linked (and `:hover`/`:disabled` etc for colour), it seems to me they are not meant to be overridden. I had an elaborate setup for allowing overrides but removed it in 2dc66df.

I used the `unbound` helper despite it being forbidden by default to allow templates to derive default values from `usage.ts` but not change when a new value is entered. This means we have defaults duplicated in only two places instead of three. But the true dream would be to be able to extract the defaults from CSS!

This will need rebasing if #3171 is merged first.


packages/boxel/addon/components/boxel/modal/usage.hbs
```diff
@@ -25,7 +26,8 @@
 </Boxel::Modal>
 </:example>
 <:api as |Args|>
-  <Args.String @name="size" @description="Can be `small`, `medium`, `large`, or unspecified. Sets a predetermined max-width to inner modal." @value={{this.size}} @options={{array "small" "medium" "large" "n/a"}} @onInput={{fn (mut this.size)}} @defaultValue="65rem" />
+  <Args.String @name="size" @description="Can be `small`, `medium`, `large`, or unspecified. Sets a predetermined max-width to inner modal." @value={{this.sizeAsString}} @options={{array "small" "medium" "large" "<undefined>"}} @onInput={{this.updateSize}} @defaultValue="65rem" />
```
Aierie, line 29: Is this `defaultValue` appropriate here? `65rem`  might also be confusing when the options are strings, not CSS lengths.

packages/boxel/addon/components/boxel/modal/usage.hbs
```diff
@@ -10,6 +10,7 @@
   @isOpen={{this.isOpen}}
   aria-labelledby="boxel-modal-usage-example-id"
   style={{css-var
+    boxel-modal-max-width=this.maxWidth
     boxel-modal-offset-top=this.offsetTop
     boxel-modal-offset-left=this.offsetLeft
     boxel-modal-offset-right=this.offsetRight
```
Aierie, line 16: If we do this:

```
style={{html-safe (
      concat (css-var
        boxel-modal-offset-top=this.offsetTop
        boxel-modal-offset-left=this.offsetLeft
        boxel-modal-offset-right=this.offsetRight
      )
      "; "
      (if this.maxWidthOverride (css-var boxel-modal-max-width=this.maxWidthOverride) "")
    )
  }}
  ```
  
  We can avoid having two inputs affect the `this.maxWidthOverride` value.

**Review by Aierie — APPROVED** [review]
Should we enforce the convention of using `unbound` + first value for defaults? I don't mind it, but I don't think it's easily noticeable at first glance, especially if it is only for css variables.

**Comment by lukemelia** [comment]
@backspace do you think that Freestyle should expose a :css block supporting slightly more purpose-fit structure for this? I don't think it would be too hard to add.

**Comment by backspace** [comment]
yes, I think that would be great, @lukemelia!

File: packages/boxel/addon/components/boxel/action-chin/usage.hbs [hunk 3]
```diff
@@ -164,6 +173,56 @@
       @value={{this.stepNumber}}
       @onInput={{fn (mut this.stepNumber)}}
     />
+    {{!-- template-lint-disable no-unbound --}}
+    <Args.String
+      @name="--boxel-action-chin-padding-horizontal"
+      @description="padding CSS style"
```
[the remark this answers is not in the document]
backspace: ya… I copied an existing `Args.String` for a CSS variable before noticing that some were more descriptive than others, I agree that they should all really specify what they’re doing, not just name the property being changed, will do

packages/boxel/addon/components/boxel/modal/usage.hbs
backspace: 👍🏻 thanks, I incorporated that in the merge commit 13728f0

packages/boxel/addon/components/boxel/modal/usage.hbs
backspace: I actually think maybe this isn’t meant to be overridden…? I got that from [here](https://github.com/cardstack/cardstack/blob/05b0f342a863670d0a8b019d1770eb0d21ea9e6a/packages/boxel/addon/components/boxel/modal/index.css#L50) but it’s true that it’s superseded if a `size` is set.

backspace: oh I misunderstood. No, I don’t think this should be `65rem`.

packages/boxel/addon/components/boxel/action-chin/usage.hbs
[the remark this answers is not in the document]
backspace: I ended up removing most of them because I thought the names were clear enough alone, in a few cases I added a more specific description.

PR #3211, opened 2022-09-08
Extend ember's Owner type

ef4:
This makes it so that `owner.lookup("service:whatever")` gets the right type automatically, so long as you have already registered your service's type like we already do: https://github.com/cardstack/cardstack/blob/bd47275d064fdd34cd45e278de353c3bb2791f83/packages/ssr-web/app/services/layer2-network.ts#L101-L105

Out-of-the-box, the types only make that work for the `@service` decorator. This makes it also work with `owner.lookup`.

@lukemelia I targeted the boxel-glint-wip branch here because I think this is relevant to those changes and can be tested that context.

### DIFF
--- a/packages/web-client/app/utils/owner.ts
+++ b/packages/web-client/app/utils/owner.ts
@@ -1,0 +1,22 @@
+import { Registry } from '@ember/service';
+
+// This file augments Ember's `Owner.lookup` so that it understands services in
+// the same way that service injection does.
+
+// This reexport is here because you cannot augment the type of a default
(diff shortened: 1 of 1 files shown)


ef4:
(The CI failures are somewhat expected here, there is followup work in the other PR to take advantage of these better types.)

lukemelia:
Thanks! This seems like something that should be in Ember right?

ef4:
Probably yes. But the current Ember 4.8 betas have already moved the types into ember itself instead of using the DefinitelyTyped ones and that is going to change some things, so the fix would go there.

PR #3215, opened 2022-09-12
sdk: Add SDK & CLI command for cancel scheduled payment

FadhlanR:
Tickets: CS-4516, CS-4513


packages/cardpay-cli/scheduled-payment/cancel.ts
```diff
@@ -0,0 +1,136 @@
+import { Argv } from 'yargs';
+import { getConstant, getSDK } from '@cardstack/cardpay-sdk';
+import { getEthereumClients, getConnectionType, NETWORK_OPTION_ANY } from '../utils';
+import { Arguments, CommandModule } from 'yargs';
+
+export default {
+  command:
+    'cancel <safeAddress> <moduleAddress> <tokenAddress> <amount> <payeeAddress> <fixedUSDFee> <percentageFee> <executionGas> <maxGasPrice> <gasTokenAddress> <salt> <payAt>',
+  describe: 'Enable scheduled payment module on the safe',
```
jurgenwerk, line 9: Wrong description?

packages/cardpay-cli/scheduled-payment/cancel.ts
```diff
@@ -0,0 +1,136 @@
+import { Argv } from 'yargs';
+import { getConstant, getSDK } from '@cardstack/cardpay-sdk';
+import { getEthereumClients, getConnectionType, NETWORK_OPTION_ANY } from '../utils';
+import { Arguments, CommandModule } from 'yargs';
+
+export default {
+  command:
+    'cancel <safeAddress> <moduleAddress> <tokenAddress> <amount> <payeeAddress> <fixedUSDFee> <percentageFee> <executionGas> <maxGasPrice> <gasTokenAddress> <salt> <payAt>',
```
jurgenwerk, line 8: I think this is too many params to define a single scheduled payment. Would be simpler to just use a single param -  `spHash`

packages/cardpay-sdk/sdk/scheduled-payment-module.ts
```diff
@@ -239,4 +240,186 @@ export default class ScheduledPaymentModule {
...
+      gasTokenAddress
+    );
+    let gasCost = new BN(estimate.safeTxGas).add(new BN(estimate.baseGas)).mul(new BN(estimate.gasPrice));
+
+    let token = new this.layer2Web3.eth.Contract(ERC20ABI as AbiItem[], gasTokenAddress);
+    let symbol = await token.methods.symbol().call();
+    let balance = new BN(await token.methods.balanceOf(safeAddress).call());
+    if (balance.lt(gasCost)) {
+      throw new Error(
+        `Safe does not have enough balance to enable scheduled payment module. The gas token ${gasTokenAddress} balance of the safe ${safeAddress} is ${fromWei(
```
jurgenwerk, line 389: Error message seems wrong

packages/cardpay-cli/scheduled-payment/cancel.ts
```diff
@@ -0,0 +1,136 @@
...
+      recurringUntil
+    );
+
+    console.log(`Cancel scheduled payment with spHash: ${spHash} ...`);
+    let onTxnHash = (txnHash: string) => console.log(`Transaction hash: ${blockExplorer}/tx/${txnHash}`);
+    await scheduledPaymentModule.cancelScheduledPayment(safeAddress, moduleAddress, spHash, gasTokenAddress, {
+      onTxnHash,
+    });
+
+    console.log(`Success to cancel scheduled payment, spHash: ${spHash}`);
```
jurgenwerk, line 134: I'd say: "Scheduled payment <spHash> canceled successfuly (spHash: xxx)"

jurgenwerk (changes requested):

packages/cardpay-cli/scheduled-payment/cancel.ts
FadhlanR: updated

packages/cardpay-sdk/sdk/scheduled-payment-module.ts

packages/cardpay-cli/scheduled-payment/cancel.ts
FadhlanR: updated

packages/cardpay-cli/scheduled-payment/cancel.ts
FadhlanR: updated

packages/cardpay-cli/scheduled-payment/cancel.ts
```diff
@@ -0,0 +1,49 @@
...
+      gasTokenAddress: string;
+      spHash: string;
+    };
+    let { web3, signer } = await getEthereumClients(network, getConnectionType(args));
+    let scheduledPaymentModule = await getSDK('ScheduledPaymentModule', web3, signer);
+    let blockExplorer = await getConstant('blockExplorer', web3);
+
+    console.log(`Cancel scheduled payment with spHash: ${spHash} ...`);
+    let onTxnHash = (txnHash: string) => console.log(`Transaction hash: ${blockExplorer}/tx/${txnHash}`);
+    await scheduledPaymentModule.cancelScheduledPayment(safeAddress, moduleAddress, spHash, gasTokenAddress, {
```
jurgenwerk, line 43: What will happen if the spHash doesn't exist on the contract? Should we handle that?

habdelra: probably we can just return gracefully since the schedule doesn't exist in the first place

packages/cardpay-sdk/sdk/scheduled-payment-module.ts
```diff
@@ -239,4 +240,96 @@ export default class ScheduledPaymentModule {
...
+    safeAddressOrTxnHash: string,
+    moduleAddress?: string,
+    spHash?: string,
+    gasTokenAddress?: string,
+    txnOptions?: TransactionOptions,
+    contractOptions?: ContractOptions
+  ): Promise<SuccessfulTransactionReceipt | void> {
+    if (isTransactionHash(safeAddressOrTxnHash)) {
+      let txnHash = safeAddressOrTxnHash;
+      return await waitUntilTransactionMined(this.layer2Web3, txnHash);
```
habdelra, line 263: i just noticed that the web3 instance variable is called `layer2Web3`. i think that is not totally accurate. We used layer1Web3 and layer2Web3 in the previous card pay work because it was really important to be clear aboutwhat layer the web3 instance was. for all the scheduled payment work, there is no special need to have a separate web3 instance for each layer. I think we can just call this class property `this.web3`. (same for all the other schedule payment related SDK classes)

habdelra (approved):
just a minor wording nit. looks good

packages/cardpay-cli/scheduled-payment/cancel.ts
FadhlanR: added validation

packages/cardpay-sdk/sdk/scheduled-payment-module.ts

packages/cardpay-cli/scheduled-payment/cancel.ts
```diff
@@ -0,0 +1,49 @@
...
+import { Arguments, CommandModule } from 'yargs';
+
+export default {
+  command: 'cancel <safeAddress> <moduleAddress> <gasTokenAddress> <spHash>',
+  describe: 'Cancel a scheduled payment',
+  builder(yargs: Argv) {
+    return yargs
+      .positional('safeAddress', {
+        type: 'string',
+        description: 'The address of the safe that will fund the scheduled payment',
```
jurgenwerk, line 13: Probably the description should be that it will fund the removal of scheduled payment

jurgenwerk (approved):
Looking good, just another very minor description change suggestion

PR #3218, opened 2022-09-13
Convert Boxel::HelpBox component to .gts

lukemelia:


lukemelia:
This is an example of a PR that converts a single Boxel component to .gts

backspace (approved):

PR #3219, opened 2022-09-13
sdk: Add SDK & CLI command to execute scheduled payment

FadhlanR:
Tickets: CS-4514, CS-4517


habdelra (approved):
looks good

packages/cardpay-cli/scheduled-payment/execute.ts
```diff
@@ -0,0 +1,137 @@
...
+        type: 'string',
+        description: 'The address of the gas token',
+      })
+      .positional('salt', {
+        type: 'string',
+        description: 'Arbitrary string to make SP unique',
+      })
+      .positional('payAt', {
+        type: 'number',
+        description: 'Time to execute scheduled payments (in seconds)',
```
jurgenwerk, line 59: We can copy this description from the other PR where we described this as UTC unix time

packages/cardpay-sdk/sdk/scheduled-payment-module.ts
```diff
@@ -424,4 +425,135 @@ export default class ScheduledPaymentModule {
...
+    moduleAddressOrTxnHash: string,
+    tokenAddress: string,
+    amount: string,
+    payeeAddress: string,
+    fee: Fee,
+    executionGas: number,
+    maxGasPrice: string,
+    gasTokenAddress: string,
+    salt: string,
+    payAtOrRecurringDayOfMonth: number,
```
jurgenwerk, line 440: Would it be better to split these into two optional params? Their values are two quite different things

jurgenwerk (changes requested):
Looks ok. Minor comments.

packages/cardpay-cli/scheduled-payment/execute.ts
FadhlanR: copied

packages/cardpay-cli/scheduled-payment/execute.ts
```diff
@@ -0,0 +1,137 @@
...
+      })
+      .positional('gasPrice', {
+        type: 'string',
+        description: 'Gas price (in the smallest units of gas token)',
+      })
+      .option('recurringDayOfMonth', {
+        type: 'number',
+        description: 'Days of the month in the range of 1-28 to make recurring payments',
```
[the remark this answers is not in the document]
FadhlanR: copied

packages/cardpay-sdk/sdk/scheduled-payment-module.ts

packages/cardpay-cli/scheduled-payment/execute.ts
```diff
@@ -0,0 +1,134 @@
...
+      payeeAddress,
+      fixedUSDFee,
+      feePercentage,
+      executionGas,
+      maxGasPrice,
+      gasTokenAddress,
+      salt,
+      payAt,
+      gasPrice,
+      recurringDayOfMonth,
```
jurgenwerk, line 127: When `recurringDayOfMonth` and `recurringUntil` are provided as empty strings, they are transformed to 0 by yargs right? What are the implications of passing 0s as values for these params into `executeScheduledPayment`? Maybe better to make sure to transform them to null in case they aren't given in the cli command (empty strings)

jurgenwerk:
Shouldn't this implementation also check whether the safe balance has enough funds to cover for the gas estimate + the tokens to transfer?

habdelra:
alternatively we could just let the safe txn revert if there are not enough tokens and then handle it as a failure. (notifications and all that) Either way it's a failure. I think  letting it blow up organically is easier implementation wise, but it won't help to answer the question why it failed if that's important. Albeit, answering "why" it failed via blockchain analysis is not actually that difficult via tools like tenderly but probably requires a human to interpret the results.

packages/cardpay-cli/scheduled-payment/execute.ts
FadhlanR: updated

packages/cardpay-sdk/sdk/utils/general-utils.ts
```diff
@@ -312,6 +314,18 @@ export async function sendTransaction(web3: Web3, transaction: Transaction): Pro
   return txHash;
 }
 
+export function extractSendTransactionError(revert: Revert, abiInterface: Interface) {
+  let error;
+  try {
+    let funcError = abiInterface.parseError(revert.data);
+    error = `${funcError.name} ${funcError.args.join(',')}`;
```
jurgenwerk, line 321: I think it would be great to document all the types of errors/reasons that we know that could happen here. This way we can handle them in the hub, for example if gas limit is reached we can try again later, or if the sp is unknown or the time period is invalid, we can stop retrying.

jurgenwerk (changes requested):

packages/cardpay-sdk/sdk/scheduled-payment-module.ts
```diff
@@ -918,4 +920,137 @@ export default class ScheduledPaymentModule {
...
+    };
+    let txnHash;
+    try {
+      txnHash = await sendTransaction(this.web3, executeScheduledPaymentTx);
+      if (typeof onTxnHash === 'function') {
+        await onTxnHash(txnHash);
+      }
+      return await waitUntilTransactionMined(this.web3, txnHash);
+    } catch (e: any) {
+      // UnknownHash: payment details generate unregistered spHash
```
FadhlanR, line 1045: @jurgenwerk I document the function error here since the extractSendTransaction is a general function that can be used for non-scheduled payment module contracts.

jurgenwerk: Perfect, thank you!

jurgenwerk (approved):

PR #3220, opened 2022-09-13
sdk: Add SDK & CLI command to create SP hash

FadhlanR:


packages/cardpay-cli/scheduled-payment/create-sp-hash.ts
```diff
@@ -0,0 +1,126 @@
...
+        type: 'string',
+        description: 'The address of the gas token',
+      })
+      .positional('salt', {
+        type: 'string',
+        description: 'Arbitrary string to make SP unique',
+      })
+      .positional('payAt', {
+        type: 'number',
+        description: 'Time to execute scheduled payments (in seconds)',
```
jurgenwerk, line 58: We can more thoroughly describe it and say it's the UTC Unix time in seconds

packages/cardpay-cli/scheduled-payment/create-sp-hash.ts
```diff
@@ -0,0 +1,126 @@
...
+      })
+      .option('recurringDayOfMonth', {
+        type: 'number',
+        description: 'Days of the month in the range of 1-28 to make recurring payments',
```
jurgenwerk, line 62: Maybe better: "Day of the month on which the payment will be made recurringly (range: 1-28)"

packages/cardpay-cli/scheduled-payment/create-sp-hash.ts
```diff
@@ -0,0 +1,126 @@
...
+      })
+      .option('recurringUntil', {
+        type: 'number',
+        description: 'End date of recurring payment',
```
jurgenwerk, line 66: We could add that this is a Unix timestamp

packages/cardpay-cli/scheduled-payment/create-sp-hash.ts
```diff
@@ -0,0 +1,126 @@
+import { Argv } from 'yargs';
+import { getSDK } from '@cardstack/cardpay-sdk';
+import { getEthereumClients, getConnectionType, NETWORK_OPTION_ANY } from '../utils';
+import { Arguments, CommandModule } from 'yargs';
+
+export default {
+  command:
+    'create-sp-hash <moduleAddress> <tokenAddress> <amount> <payeeAddress> <fixedUSDFee> <percentageFee> <executionGas> <maxGasPrice> <gasTokenAddress> <salt> <payAt>',
+  describe: 'Create scheduled payment hash',
```
jurgenwerk, line 9: "Generates a scheduled payment hash which is calculated from the given arguments. The hash serves as a unique identifier used for scheduling, executing, and canceling scheduled payments"

jurgenwerk (approved):
Looking great! Just a couple of minor copy comments

packages/cardpay-cli/scheduled-payment/create-sp-hash.ts
habdelra: maybe we can use 1-31, and then we could on the server use moment.js (or your favorite date tool) to detect how many days there are in the current current month, and if the number provided by the user is larger than the current day, and the current day is the last day of the month, then we could execute the payment. And the way we could explain this is that we accept a range between 1 - 31, and that for months with less days that the specified value, we'll execute the payment on the last day of the month. (I generally have bills that I like to schedule to be paid on the last day of the month--it's probably a popular option actually).

habdelra (approved):

PR #3227, opened 2022-09-14
boxel/web: Update TabBar for web-client use

backspace:
This retrofits `Boxel::TabBar` so it can replace `web-client`’s `CardPay::HeaderNav`. It will also be of use for schedule payments:

Wallet Not Connected - Banner Expanded - MVP - Zeplin 2022-09-16 12-31-53

There’s a Percy diff for `web-client` but IMO the existing centring is broken:

Quick Payments · Card Pay 2022-09-16 12-28-03

I also converted it to use Glint in a final commit but the template changes before that are viewable [here](https://github.com/cardstack/cardstack/compare/2d4a3a85e4d55e49b7299b8994a25daf38707d2a...9aef95a562a3c06c23505b6423bbcb92f47d69aa#diff-51fd3134d9334d069bd3c27b39bd90bf3cffe3432fa51eff86f80262ae57e21d).

One major caveat: the hover-jumping I mentioned in #3202 is happening! So far I’ve been unable to find why. Would you be into looking at this together, @Aierie?


packages/boxel/addon/components/boxel/tab-bar/index.css
```diff
@@ -19,13 +52,12 @@
...
-.boxel-tab-bar__item > a {
-  display: inline-flex;
+.boxel-tab-bar__item-text {
+  display: flex;
   flex-direction: column;
-  padding: var(--boxel-sp-xs) var(--boxel-sp-lg);
 }
 
-.boxel-tab-bar__item > a::after {
+.boxel-tab-bar__item-text::after {
```
Aierie, line 60: `font-weight` for this should be `var(--boxel-tab-bar-font-weight-hover)`, since we allow configuring it.

packages/boxel/addon/components/boxel/tab-bar/index.css
```diff
@@ -42,7 +74,7 @@
   }
 }
 
-.boxel-tab-bar__item > a:hover {
+.boxel-tab-bar__item > a:hover .boxel-tab-bar__item-text {
```
Aierie, line 77: This rule will not be applied because the `a` tag is the `.boxel-tab-bar__item`. We can safely remove it and preserve existing behaviour.

packages/boxel/addon/components/boxel/tab-bar/index.gts
```diff
@@ -0,0 +1,102 @@
...
+            {{! template-lint-disable require-context-role }}
+            <LinkTo
+              id={{or tab.menuItem.id tab.link.routeName}}
+              @route={{tab.link.routeName}}
+              @query={{or tab.link.queryParams (hash)}}
+              role="tab"
+              class={{
+                cn
+                "boxel-tab-bar__item"
+                "boxel-tab-bar__item"
```
Aierie, line 76:
```suggestion
```

packages/boxel/addon/components/boxel/tab-bar/index.css
```diff
@@ -1,16 +1,49 @@
...
+  gap: var(--boxel-sp-xs);
+  border-bottom: 2px solid transparent;
+}
+
+.boxel-tab-bar__item--is-inactive {
+  opacity: 0.45;
+  cursor: text;
 }
 
 .boxel-tab-bar__item:hover {
```
Aierie, line 35:
```suggestion
.boxel-tab-bar__item:not(.boxel-tab-bar__item--is-inactive):hover {
```

Currently this is more specific than the inactive rule above, and the cursor shown is still pointer.

packages/boxel/addon/components/boxel/tab-bar/index.css
```diff
@@ -19,13 +52,12 @@
...
+.boxel-tab-bar__item-text {
+  display: flex;
   flex-direction: column;
-  padding: var(--boxel-sp-xs) var(--boxel-sp-lg);
 }
 
-.boxel-tab-bar__item > a::after {
+.boxel-tab-bar__item-text::after {
   content: attr(data-text);
   content: attr(data-text) / "";
```
Aierie, line 62: Do we need both these content properties?

packages/boxel/addon/components/boxel/tab-bar/index.gts
```diff
@@ -0,0 +1,102 @@
...
+  }
+
+  <template>
+    <div
+      role="tablist"
+      class={{cn "boxel-tab-bar" @class boxel-tab-bar--spread=@spread}}
+      style={{
+        cssVar
+        boxel-tab-bar-background-color="var(--boxel-tab-bar-background-color)"
+        boxel-tab-bar-font="var(--boxel-tab-bar-font)"
```
Aierie, line 37: I don't think we should set these variables here, this invalidates defaults for these variables because this is a circular definition.

packages/boxel/addon/components/boxel/tab-bar/index.gts
```diff
@@ -0,0 +1,102 @@
...
+      }}
+      ...attributes
+    >
+      {{#if this.linkItems}}
+        {{#each this.linkItems as |tab|}}
+          {{#if tab.menuItem.inactive}}
+            <div
+              role="tab"
+              data-text={{tab.text}}
```
Aierie, line 46: `data-text` should move down to `.boxel-tab-bar__item-text`, too? Could also omit it.

packages/boxel/addon/components/boxel/tab-bar/index.gts
```diff
@@ -0,0 +1,102 @@
...
+        cssVar
+        boxel-tab-bar-background-color="var(--boxel-tab-bar-background-color)"
+        boxel-tab-bar-font="var(--boxel-tab-bar-font)"
+      }}
+      ...attributes
+    >
+      {{#if this.linkItems}}
+        {{#each this.linkItems as |tab|}}
+          {{#if tab.menuItem.inactive}}
+            <div
```
Aierie, line 44: Using a div instead of a link is nice here!

packages/boxel/addon/components/boxel/tab-bar/index.css
backspace: it’s true, this is a relic, thanks!

packages/boxel/addon/components/boxel/tab-bar/index.css
backspace: hmm well Firefox [doesn’t support](https://developer.mozilla.org/en-US/docs/Web/CSS/content#browser_compatibility) the `/` for fallback but I tried removing `data-text` and it was fine (with the jumping of course) so I guess not…??? Thanks

packages/boxel/addon/components/boxel/tab-bar/index.gts
backspace: whew, thanks for pointing this out, in some ways I’m still confused re CSS variables

packages/boxel/addon/components/boxel/tab-bar/index.gts
backspace: re this and the below, I recently (re)learned that `LinkTo` supports `@disabled`, which I think is preferable to having a whole parallel tree, for reasons like this oversight! So I’ve updated with that.

PR #3229, opened 2022-09-14
Update aws-sdk versions

lukemelia:


backspace (approved):
I’ve deployed this to staging and ran a `persist-off-chain-profile` to confirm that it fixed the problem.

Is there some lint-esque tool that would help prevent these from getting out of sync in the future?

lukemelia:
>Is there some lint-esque tool that would help prevent these from getting out of sync in the future?

I think the preventable problem is that this code was not exercised by tests.

PR #3230, opened 2022-09-15
[CS-4573]: Create Boxel::Input::RangedNumberPicker

danibonilha:
In progress: better style dropdown positioning 
Todo: Move .hbs to .gts


packages/boxel/addon/components/boxel/input/ranged-number-picker/index.css
```diff
@@ -0,0 +1,45 @@
...
+  background-color: var(--boxel-light-100);
+}
+
+.boxel-ranged-number-picker-menu__item {
+  font-weight: 600;
+  font-size: var(--boxel-font-size-sm);
+  padding: var(--boxel-sp-xxs) var(--boxel-sp-sm);
+}
+
+.ember-power-select-placeholder,
```
backspace, line 39:
```suggestion
.boxel-ranged-number-picker-menu .ember-power-select-placeholder,
```

I think this would prevent this style from applying to other uses of Ember Power Select outside of this component 🤓

backspace (approved):
This is looking good! 🥳

I briefly looked at how to make the caret lighter despite the colour being hardcoded in the SVG and had some success with the CSS filters technique described [here](https://codepen.io/noahblon/post/coloring-svgs-in-css-background-images#css-filters-2):

Boxel Components 2022-09-16 16-21-19

It would need some tweaking but it involved using the `::after` hack to move the background in that “pseudo-element” so `filter` only applies to it and not the input’s value:

```
.boxel-ranged-number-picker-menu {
  display: inline-block;
  padding-right: 1em;
}

.boxel-ranged-number-picker-menu::after {
  content: '';
  background: url('/@cardstack/boxel/images/icons/caret-down.svg') no-repeat;
  width: 11px;
  height: 6px;
  display: inline-block;
  filter: opacity(29%);
}
```

This is somewhat unpleasant and would make for an awkward customisation interface (`filter` CSS variable?) but is maybe fine?

packages/boxel/addon/components/boxel/input/ranged-number-picker/index.ts
```diff
@@ -0,0 +1,32 @@
+import { action } from '@ember/object';
+import Component from '@glimmer/component';
+import { tracked } from '@glimmer/tracking';
+import './index.css';
+
+interface RangedBoxelInputRangedNumberPickerInputArgs {
+  min: number;
+  max: number;
+  onNumberSelected: (num: number) => void;
+}
```
lukemelia, line 10: This is an older style of component interface. We should use the newer style that Glint uses.

lukemelia (changes requested):
Looking good. One TypeScript thing to update that I left a comment on, and I agree with @backspace 's recommendation about matching the css class name to the component name.

lukemelia (approved):

PR #3231, opened 2022-09-16
hub: Add an endpoint to create a scheduled payment

jurgenwerk:
Ticket: CS-4373

It goes like this:

1. Client will generate the signature for creating the transaction to add a scheduled payment on the blockchain
2. Client will send the scheduled payment params to the endpoint, along with the signature
3. Endpoint will save the scheduled payment row in the db, and submit a blockchain transaction. The request will wait until we are able to get the transaction hash. If there are problems submitting the transaction, the request will fail, and the scheduled payment row will be discarded
4. At this point, the transaction is not mined yet. This could take many minutes to finish, so we do it async in a background worker which will update the scheduled payment row once done (it will write the `creation_block_number`). This way the client can poll the scheduled payment and when `creation_block_number` gets written by the background job, it means the scheduled payment has been added in both blockchain and the crank - the client can then show it as "added successfully", otherwise it will be "pending".

TODO:
- [ ] Test using a real chain 
- [ ] More thorough validator checks + validator tests
- [ ] Add a route test for a recurring scheduled payment (important: see if `payAt` gets set correctly)
- [ ] Prevent double save (at the beginning of the request, check if there was a very similar scheduled payment created very recently)

### DIFF
--- a/packages/cardpay-sdk/sdk/utils/signing-utils.ts
+++ b/packages/cardpay-sdk/sdk/utils/signing-utils.ts
@@ -435,3 +435,15 @@ function sortSignaturesAsBytes(
     return [contractSignature, ownerSignature];
   }
 }
+
+export function signatureToVRS(rawSignature: string) {
+  const signature = strip0x(rawSignature);
+  const v = parseInt(signature.substr(64 * 2));
+  const r = signature.substr(0, 32 * 2);
+  const s = signature.substr(32 * 2, 32 * 2);
+  return { v, r, s };
+}
+
+export function strip0x(s: string) {
+  return Web3.utils.isHexStrict(s) ? s.substr(2) : s;
+}
--- a/packages/hub/node-tests/routes/scheduled-payments-test.ts
+++ b/packages/hub/node-tests/routes/scheduled-payments-test.ts
@@ -1,0 +1,307 @@
+import { PrismaClient } from '@prisma/client';
+
+import { registry, setupHub } from '../helpers/server';
+import { setupStubWorkerClient } from '../helpers/stub-worker-client';
+
+class StubAuthenticationUtils {
+  validateAuthToken(encryptedAuthToken: string) {
+    return handleValidateAuthToken(encryptedAuthToken);
+  }
+}
+
+let stubUserAddress = '0x2f58630CA445Ab1a6DE2Bb9892AA2e1d60876C13';
+function handleValidateAuthToken(encryptedString: string) {
+  expect(encryptedString).to.equal('abc123--def456--ghi789');
+  return stubUserAddress;
+}
+
+let prisma: PrismaClient;
+let forceSchedulePaymentOnChainTimeout = false;
+let mockTxnHash = '0xc13d7905be5c989378a945487cd2a1193627ae606009e28e296d48ddaec66162';
+
+class StubCardpaySDK {
+  getSDK(sdk: string) {
+    switch (sdk) {
+      case 'ScheduledPaymentModule':
+        return Promise.resolve({
+          schedulePayment: async (
+            _safeAddress: any,
+            _moduleAddress: any,
+            _tokenAddress: any,
+            _spHash: any,
+            _signature: any,
+            { onTxnHash }: any
+          ) => {
+            if (forceSchedulePaymentOnChainTimeout) {
+              // no-op. txn hash will never be returned
+            } else {
+              await onTxnHash(mockTxnHash);
+            }
+          },
+        });
+      default:
+        throw new Error(`unsupported mock cardpay sdk: ${sdk}`);
+    }
+  }
+}
+
+describe('POST /api/scheduled-payments', async function () {
+  let { getJobIdentifiers, getJobPayloads } = setupStubWorkerClient(this);
+
+  this.beforeEach(async function () {
+    registry(this).register('authentication-utils', StubAuthenticationUtils);
+    registry(this).register('cardpay', StubCardpaySDK);
+  });
+
+  let { getPrisma, request } = setupHub(this);
+
+  this.beforeEach(async function () {
+    prisma = await getPrisma();
+  });
+
+  it('returns a 401 if the auth token is invalid', async function () {
+    await request()
+      .post('/api/scheduled-payments')
+      .send({})
+      .set('Accept', 'application/vnd.api+json')
+      .set('Content-Type', 'application/vnd.api+json')
+      .expect(401)
+      .expect({
+        errors: [
+          {
+            status: '401',
+            title: 'No valid auth token',
+          },
+        ],
+      })
+      .expect('Content-Type', 'application/vnd.api+json');
+  });
+
+  it('responds with errors when attrs are missing', async function () {
+    await request()
+      .post('/api/scheduled-payments')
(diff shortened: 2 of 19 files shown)


jurgenwerk:
This is ready for a draft review, I'm currently not planning any significant changes, only the things listed in TODO.

packages/hub/node-tests/routes/scheduled-payments-test.ts
```diff
@@ -0,0 +1,307 @@
...
+        },
+      })
+      .expect('Content-Type', 'application/vnd.api+json');
+
+    expect(getJobIdentifiers()[0]).to.equal('scheduled-payment-on-chain-creation-waiter');
+    expect(getJobPayloads()[0]).to.deep.equal({ scheduledPaymentId });
+  });
+
+  it('does not persist a scheduled payment when something goes wrong with on chain call', async function () {
+    forceSchedulePaymentOnChainTimeout = true;
```
backspace, line 273: Maybe this should be reset to `false` in a `beforeEach` in case any tests get added after this?

packages/hub/routes/scheduled-payments.ts
```diff
@@ -0,0 +1,149 @@
...
+                data: {
+                  creationTransactionHash: txnHash,
+                },
+              });
+
+              // It might take several minutes for the transaction to be mined, so we wait for it in the background. The client
+              // can poll the record to see when it's mined (creation_block_number attribute will be set).
+              await this.workerClient.addJob('scheduled-payment-on-chain-creation-waiter', {
+                scheduledPaymentId: scheduledPayment.id,
+              });
```
backspace, line 106: The profile purchase code uses a “job ticket” pattern for a slow/chain operation; the endpoint returns a newly-created ticket that the consumer can poll at `/api/job-tickets/:id` to track progress. Would it make sense to use that pattern here?

packages/hub/tasks/scheduled-payment-on-chain-creation-waiter.ts
```diff
@@ -0,0 +1,49 @@
...
+        data: {
+          creationBlockNumber: receipt.blockNumber, // To let us know that the scheduled payment has been created on chain and that the process of scheduling payment was completed
+        },
+        where: {
+          id: scheduledPayment.id,
+        },
+      });
+    }
+  }
+}
```
backspace, line 43: Is it worth having a couple of tests for this task?

habdelra:
Love your approach, especially #4 :pray:

packages/cardpay-sdk/sdk/utils/signing-utils.ts
```diff
@@ -435,3 +435,15 @@ function sortSignaturesAsBytes(
     return [contractSignature, ownerSignature];
   }
 }
+
+export function signatureToVRS(rawSignature: string) {
```
habdelra, line 439: is this essentially the same as `ethSignSignatureToRSVForSafe()` in this module?

packages/cardpay-sdk/sdk/utils/signing-utils.ts
```diff
@@ -435,3 +435,15 @@ function sortSignaturesAsBytes(
...
+  const signature = strip0x(rawSignature);
+  const v = parseInt(signature.substr(64 * 2));
+  const r = signature.substr(0, 32 * 2);
+  const s = signature.substr(32 * 2, 32 * 2);
+  return { v, r, s };
+}
+
+export function strip0x(s: string) {
+  return Web3.utils.isHexStrict(s) ? s.substr(2) : s;
```
habdelra, line 448: using `s.replace(/^0x/, '')` is better since it's idempotent--you can run it as many times as you want and it will always return the correct thing.

packages/hub/routes/scheduled-payments.ts
jurgenwerk: I thought about it. I think it's easier to just query for the scheduled payments. If `creation_block_number` is present on the record, it will mean the job finished successfully. In this case that pattern sounds a bit roundabout

packages/cardpay-sdk/sdk/utils/signing-utils.ts
jurgenwerk: It looks like it. However, there is one diference, in the recovery identifier (`v`):

`signatureToVRS` uses `parseInt(value)` while `ethSignSignatureToRSVForSafe` uses `parseInt(value, 16)`

So for example, let's have:
`signature = 0x21fbf0696d5e0aa2ef41a2b4ffb623bcaf070461d61cf7251c74161f82fec3a4370854bc0a34b3ab487c1bc021cd318c734c51ae29374f2beb0e6f2dd49b4bf41c`

`v` is the last 2 bytes and that is `1c`

`signatureToVRS`:

`parseInt("1c") `
`=> 1`

`ethSignSignatureToRSVForSafe`:

`parseInt("1c", 16)`
`=> 28`

Judging by [this article](https://medium.com/mycrypto/the-magic-of-digital-signatures-on-ethereum-98fe184dc9c7) it looks like the `v` can be either either 27 (0x1b) or 28 (0x1c). That makes me think `ethSignSignatureToRSVForSafe` is the correct implementation and we should use it instead of `signatureToVRS` (and get rid of it). 

However I'm not sure how this even worked then, given the parsing might be wrong. 🙇‍♂️ Judging by this comment https://github.com/ethereum/go-ethereum/issues/19751#issuecomment-504900739, `v` can also be 0 or 1 and this is the reason both implementations that we have are working.

packages/cardpay-sdk/sdk/utils/signing-utils.ts
jurgenwerk: Agree! Will change it

packages/hub/node-tests/routes/scheduled-payments-test.ts
jurgenwerk: Sounds good. Will change

jurgenwerk:
@habdelra I think @FadhlanR used `waitUntilTransactionMined` for the act of enabling the scheduled payment module on the safe. I believe we discussed that doesn't concern the hub and that we want to delegate the waiting to the client. @FadhlanR can you confirm?

FadhlanR:
From what I understand for now the development in the hub is only related to the crank work which is regarding scheduling the payment and triggering the execution of the scheduled payment. I don't think we plan to use the hub to pull any type of transactions, such as creating safe and enabling modules that will not be handled by the hub/crank. CMIIW

habdelra:
@jurgenwerk @FadhlanR so i'm confused them about about this function in the SDK: https://github.com/cardstack/cardstack/blob/main/packages/cardpay-sdk/sdk/scheduled-payment-module.ts#L672. This function uses `waitUntilTransactionMined()` https://github.com/cardstack/cardstack/blob/main/packages/cardpay-sdk/sdk/scheduled-payment-module.ts#L739. Shouldn't this function be responsible for calling the endpoint in this PR to create the schedule payment, and then based on the comments in #4 of this PR also poll the hub to confirm that the sceduled payment is setup?

To be clear is this PR about establishing the schedule for the payments or to execute a specific payment that is part of the schedule?

PR #3232, opened 2022-09-16
[Chore]: Override boxel js prettier to match template

danibonilha:
using VsCode, if we had format on save enabled it was actually adding singles quotes to templates. In order to correctly auto format we need to override the rule. There's this [great article ](https://dev.to/jelhan/format-glimmer-templates-with-prettier-ipa)with more info.

https://user-images.githubusercontent.com/20520102/190681693-9e8e8bdd-b29c-42d9-bcef-4f565124aa6c.mov

### DIFF
--- a/packages/boxel/.prettierrc.js
+++ b/packages/boxel/.prettierrc.js
@@ -2,4 +2,12 @@
 
 module.exports = {
   singleQuote: true,
+  overrides: [
+    {
+      files: '*.hbs',
+      options: {
+        singleQuote: false,
+      },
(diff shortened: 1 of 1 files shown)


backspace (approved):
Nice, thanks for figuring this out!

I tried it locally and with it I no longer need to avoid autoformat on save 🤩 

re whether to apply it to `web-client` and `ssr-web` I’m unsure, I see we don’t have double quotes configured on either of those. Any thoughts @Aierie?

PR #3237, opened 2022-09-20
Css adjustments for ranged number input

Aierie:


packages/boxel/addon/components/boxel/input/ranged-number-picker/index.css
```diff
@@ -18,19 +18,20 @@
...
-  max-height: var(--boxel-xs-container);
-  overflow: auto;
 }
 
 .boxel-ranged-number-picker__dropdown ul {
   list-style: none;
   padding: 0;
   text-align: center;
+  overflow: auto;
+  max-height: var(--boxel-xs-container);
```
Aierie, line 28: The `ul` is what we want the overflow on, because of the way that `ember-power-select` is handling keyboard input (try navigating with arrow keys up and down). The keyboard input handler finds this `ul` and adjusts its scroll position:

https://github.com/cibernox/ember-power-select/blob/cb4c57f94d9d8ff8868832cd3e38182eea4a2c04/addon/components/power-select.ts#L425-L443

This has bitten us multiple times before, and will happen again anywhere we have a dropdown whose contents are bigger than the allocated height. @backspace do you think it's worth making a base component so that we don't have to apply styles like this each time we create a new dropdown? We would either bake the scroll styles in or try the `scrollTo` arg to see if we can make something more robust.

backspace: yes, maybe we should be using a wrapper. And even linting for other uses of EPS???

CS-4566 is about creating `Boxel::Dropdown` but it’s meant to be based off Ember Basic Dropdown 🤔

backspace (approved):
Thanks for catching this!

danibonilha (approved):
Nice catch

PR #3239, opened 2022-09-20
Deploy scheduled payment module to goerli and mumbai

FadhlanR:


habdelra (approved):
Looks good

packages/cardpay-sdk/sdk/scheduled-payment-module.ts
```diff
@@ -419,9 +419,14 @@ export default class ScheduledPaymentModule {
       }
     } catch (e: any) {
       let _interface = new utils.Interface(['error GasEstimation(uint256 gas)']);
-      let messages = e.message.split(' ');
-      let decodedError = _interface.parseError(messages[2].replace(',', ''));
-      requiredGas = decodedError.args[0].toNumber();
+      if (e.data) {
```
jurgenwerk, line 422: For clarity I would add a couple of comments on why the getting the error out is different in some cases (chains?). Maybe also provide examples of responses or the actual examples of strings that we are parsing here.

jurgenwerk (approved):
LGTM!

packages/cardpay-sdk/sdk/scheduled-payment-module.ts
FadhlanR: Here's the response example.

Sokol:
`{
    "jsonrpc": "2.0",
    "error": {
        "code": -32015,
        "message": "Reverted 0x292a9e2f351e79a0b468c93c4cd0a8560e4b50e9"
    },
    "id": 5
}`

Goerli and Mumbai:
`{
    "jsonrpc": "2.0",
    "id": 5,
    "error": {
        "code": 3,
        "data": "0x08c379a000000000000000000000000000000000000000000000000000000000000000200000000000000000000000000000000000000000000000000000000000000014292a9e2f351e79a0b468c93c4cd0a8560e4b50e9000000000000000000000000",
        "message":"execution reverted: )*�/5y��h�<LШVKP�"
    }
}`

habdelra: oh yeah--this is super annoying. more specifically it is geth vs nethermind vs open ethereum. each one deals with revert data differently

habdelra: If you are curious, this is how we dealt with this annoyance in the relay server code. the relay server uses these revert binary messages for the exact same purposes as we do (to estimate gas).
https://github.com/cardstack/card-protocol-relay-service/blob/master/safe.py#L420-L449

FadhlanR: I think I need to create a specific function to handle this. In our case, we need two functions to extract estimate gas and safe address.

PR #3240, opened 2022-09-20
[CS-4620]: Convert RangedNumberPicker to .gts

danibonilha:
This PR migrates the RangedNumberPicker to .gts template, unfortunately powerSelect seems to have types defined only inside the addon folder, but, importing the component from there doesn't seem to work, and we do need import from `'ember-power-select/components/power-select'` to get the component to render, so I add a ts-disabled rule for this case.


backspace (approved):
That’s unfortunate about the types. But this looks good 🎉

lukemelia:
@danibonilha I pushed a commit that lets us take advantage of PowerSelect's (imperfect) types.

lukemelia (approved):

danibonilha:
> @danibonilha I pushed a commit that lets us take advantage of PowerSelect's (imperfect) types.

Cool, thanks. I'm adding a new commit to remove the types from globals then

PR #3242, opened 2022-09-21
Add cron to lambda

tintinthong:
This adds an additional code path to the lambda to listen to a cron event (cloudwatch event) that is triggered every 1 day. 
- it looks at s3 and subgraph and figures out the unsubmitted roots 
- submits the unsubmitted roots
- theres `min_scan_block` of ~2 days back to avoid lambda timeouts

infra pr: [here](https://github.com/cardstack/infra/pull/258)

### DIFF
--- a/packages/reward-root-submitter/reward_root_submitter/utils.py
+++ b/packages/reward-root-submitter/reward_root_submitter/utils.py
@@ -1,0 +1,168 @@
+import json
+import logging
+import re
+from typing import Any, Callable, Dict, TypedDict
+
+import pandas as pd
+import requests
+from cloudpathlib import AnyPath
+from web3 import Web3
+
+from .config import Config
+
+DEFAULT_MAX_PAST_BLOCKS = 34560  # 2 days (1 block every 5s)
+DEFAULT_SUBGRAPH_PAGINATE_SIZE = 100
+
+
+class SubgraphQuery(TypedDict):
+    name: str
+    query: str
+
+
+def merkle_root_submissions_query(
+    skip: int, blockNumber_gt=0, **filters
+) -> SubgraphQuery:
+    query = """{{
+        merkleRootSubmissions(
+            skip: {skip},
+            orderBy: blockNumber,
+            where: {{ blockNumber_gt: {blockNumber_gt} }}
+        )
+        {{
+            rewardProgram {{
+                id
+            }}
+            paymentCycle
+        }}
+    }}""".format(
+        **{"skip": skip, "blockNumber_gt": blockNumber_gt, **filters}
+    )
+    return {"name": "merkleRootSubmissions", "query": query}
+
+
+def query_subgraph(
+    config: Config,
+    skip: int,
+    blockNumber_gt: int,
+    query: Callable[[int, int, Dict[str, Any]], SubgraphQuery],
+):
+    print(f"Skipping {skip}")
+    subgraph_url = config.subgraph_url
+    subgraph_query = query(skip, blockNumber_gt)
+    r = requests.post(
+        subgraph_url,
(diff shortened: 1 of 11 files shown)


tintinthong:
Screenshot 2022-09-22 at 18 52 59

Kinda not happy with this. s3 (`get_roots_s3`) is taking longer than subgraph(`get_roots_subgraph`. Actually, I wonder whether using the raw `list_objects` or `list_objects_v2` would be better call and then only getting the parquet file. I wonder the difference in speed -- gonna try that

tintinthong:
> Screenshot 2022-09-23 at 21 00 25
> 
> Kinda not happy with this. s3 (`get_roots_s3`) is taking longer than subgraph(`get_roots_subgraph`. Actually, I wonder whether using the raw `list_objects` or `list_objects_v2` would be better call and then only getting the parquet file. I wonder the difference in speed -- gonna try that
> 
> Note: I thought it was faster b4.Prev I used`iterdir` cos Im using `glob` now.

Yep looks like the `glob` pattern was slowing it down. Added a note to suggest next dev to never use it.

tintinthong:
I didn't add test given it was a diffing operation and I think there is little value add in mocking subgraph fixture

packages/reward-root-submitter/reward_root_submitter/utils.py
```diff
@@ -0,0 +1,143 @@
...
+    return {"name": "merkleRootSubmissions", "query": query}
+
+
+def query_subgraph(
+    config: Config,
+    skip: int,
+    blockNumber_gt: int,
+    query: Callable[[int, int, Dict[str, Any]], SubgraphQuery],
+):
+    print(f"Skipping {skip}")
```
IanCal, line 58: can we swap this to a logging.info?

packages/reward-root-submitter/reward_root_submitter/utils.py
```diff
@@ -0,0 +1,143 @@
...
+    query: str
+
+
+def merkle_root_submissions_query(
+    skip: int, blockNumber_gt=0, **filters
+) -> SubgraphQuery:
+    query = """{{
+        merkleRootSubmissions(
+            skip: {skip},
+            orderBy: blockNumber,
```
IanCal, line 37: I guess it might be by default but can we have the order being explicitly asc?

packages/reward-root-submitter/reward_root_submitter/utils.py
```diff
@@ -0,0 +1,143 @@
...
+        data = json_data["data"]
+        return data[subgraph_query["name"]]
+
+
+def get_roots_subgraph(config: Config, min_scan_block=0):
+    paginate_size = 100
+    res = []
+
+    i = 0
+    while c := query_subgraph(
```
IanCal, line 76: Looks good - I wondered about trying to avoid explicit pagination and update the block number we're filtering on but the end condition gets really awkward, I think this pagination is cleaner.

packages/reward-root-submitter/reward_root_submitter/utils.py
```diff
@@ -0,0 +1,143 @@
...
+):
+    print(f"Skipping {skip}")
+    subgraph_url = config.subgraph_url
+    subgraph_query = query(skip, blockNumber_gt)
+    r = requests.post(
+        subgraph_url,
+        json={"query": subgraph_query["query"]},
+    )
+    if r.ok:
+        json_data = json.loads(r.text)
```
IanCal, line 66: There's an `r.json()`

IanCal (approved):
A few style things mostly - overall change looks good to me. Happy to merge with a few tweaks without another review.

PR #3243, opened 2022-09-21
hub: DB migrate improvements

jurgenwerk:
This PR is a suggestion to change the behaviour of the `yarn db:migrate` command to be more Rails-like, where `rails db:migrate` will migrate all pending migrations, and `rails db:rollback` will revert the latest migration only. I find this a good default practice and wish to apply it here as well. 

So this translates to: `yarn db:migrate` (will migrate all pending) and `yarn db:migrate down` (will only rollback the last one). 

Also, `migrate` was added to the scripts so we can use it to generate migrations, for example `yarn migrate create add user address to scheduled payments`


habdelra (approved):
:+1:

backspace (approved):

PR #3244, opened 2022-09-22
Update to freestyle beta release to support usage.gts

lukemelia:


backspace (approved):
Here’s a side-by-side of `main` and this branch in same-width windows, are these differences something to be concerned about? I don’t see why the header would change.

Boxel Components 2022-09-22 09-23-28

The `:example` block has line breaks that match those in `usage.gts`, unlike on `main`, so that’s an improvement. 🎉

lukemelia:
@backspace There have been some other tweaks to freestyle in this release which account for the header changes. I am pleasantly surprised by the improved example block. That was inadvertent.

PR #3254, opened 2022-09-22
boxel: Add ControlPanel

backspace:
This is a component meant to support the functionality seen here:

Network Default to ETH then Create Safe - MVP - Zeplin 2022-09-23 12-48-58

Since the indentation there is inconsistent, I chose to align the body with the icon:

Boxel Components 2022-09-23 12-50-02

This is easily changed if it should be otherwise.

You can see it [here](https://boxel.stack.cards/boxel/control-panel-cs-4565/#/docs?s=Components&ss=%3CBoxel%3A%3AControlPanel%3E).

### DIFF
--- a/packages/boxel/addon/components/boxel/control-panel/index.gts
+++ b/packages/boxel/addon/components/boxel/control-panel/index.gts
@@ -53,9 +53,6 @@ const Item: TemplateOnlyComponent<ItemSignature> =
 </template>;
 
 export default class ControlPanel extends Component<Signature> {
-  get jortle(): number {
-    return 3;
-  }
 
   <template>
     <nav class="boxel-control-panel">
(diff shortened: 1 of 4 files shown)


packages/boxel/addon/components/boxel/control-panel/index.gts
```diff
@@ -0,0 +1,71 @@
...
+    <div class="boxel-control-panel__item-body">
+      {{yield}}
+    </div>
+  </section>
+</template>;
+
+export default class ControlPanel extends Component<Signature> {
+  get jortle(): number {
+    return 3;
+  }
```
Aierie, line 58:
```suggestion
```

packages/boxel/addon/components/boxel/control-panel/index.gts
```diff
@@ -0,0 +1,71 @@
...
+
+  <template>
+    <nav class="boxel-control-panel">
```
Aierie, line 61: I don't think we can guarantee that this is used as a `nav`. There are various screens that look like they offer actions, not navigation.

packages/boxel/addon/components/boxel/control-panel/index.gts
backspace: lol oops thanks

packages/boxel/addon/components/boxel/control-panel/index.gts
backspace: ya, changed now, thanks

PR #3255, opened 2022-09-22
[CS-4566]: Add BoxelDropdown and refactor DropdownBtn

danibonilha:
This PR adds the BoxelDropdown, which is a wrapper from ember-basic-dropdown in order to expose a limited api. Also:

- Refactors DropdownButton to use BoxelDropdown and adds named blocks 
- Partially types MenuItem to be used on files that glint checks
- Adds a global type `ArgsFromComponent` that extract the Args from an old glimmerComponent api (BasicDropdown doesn't exports its type, so this was the way to get it, without just copying and pasting).
- Migrates dropdown usage to .gts

PS: BasicDropdown has some weirdness on its import, that's why its type is on globals for now, and it wasn’t migrated to .gts yet, the plan is to investigate that next.


backspace (approved):
Successful extraction, well done!

This is not something you’re responsible for since `DropdownButton` already worked this way, but I’m thinking about why we have a dropdown that doesn’t work at all with the keyboard. Is this something we want? Maybe @Aierie has opinions.

But as you inherited this existing problem, I think this is good to merge 😀 we can revisit if we decide we want keyboard interactivity.

danibonilha:
> Successful extraction, well done!
> 
> This is not something you’re responsible for since `DropdownButton` already worked this way, but I’m thinking about why we have a dropdown that doesn’t work at all with the keyboard. Is this something we want? Maybe @Aierie has opinions.
> 
> But as you inherited this existing problem, I think this is good to merge 😀 we can revisit if we decide we want keyboard interactivity.

Thanks! Yeah I agree with you that's annoying to not be able to use the tab key to interact with the component, now as far as being able to "search" by key, I think it's more of a select behavior than a dropdown.

backspace:
I was vague but I meant even arrow keys, I expect to be able to navigate and choose/execute that way 🤓

PR #3256, opened 2022-09-23
Write the parameters into the parquet metadata

IanCal:
In testing this was possible the easiest way was to actually add the code.

The parquet files can have metadata added to the schema as a whole, the restriction is that the key and value both must be bytes or coercible to bytes. This means when reading them back out the key used for lookup must be b"run_parameters" and the contents then need to be decoded from json (example in the test).

Q on naming, is "run_parameters" clear? Parameters felt too vague.

### DIFF
--- a/packages/cardpay-reward-programs/cardpay_reward_programs/main.py
+++ b/packages/cardpay-reward-programs/cardpay_reward_programs/main.py
@@ -57,7 +57,7 @@ def run_reward_program(
     payment_list = rule.run(
         parameters["run"]["payment_cycle"], parameters["run"]["reward_program_id"]
     )
-    tree = PaymentTree(payment_list.to_dict("records"))
+    tree = PaymentTree(payment_list.to_dict("records"), run_parameters=parameters)
     table = tree.as_arrow()
     write_parquet_file(AnyPath(output_location), table)
 
--- a/packages/cardpay-reward-programs/pdm.lock
+++ b/packages/cardpay-reward-programs/pdm.lock
@@ -1,6 +1,6 @@
 [[package]]
 name = "aiohttp"
-version = "3.8.1"
+version = "3.8.3"
 requires_python = ">=3.6"
 summary = "Async http client/server framework (asyncio)"
 dependencies = [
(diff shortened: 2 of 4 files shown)


packages/cardpay-reward-programs/cardpay_reward_programs/main.py
```diff
@@ -57,7 +57,7 @@ def run_reward_program(
     payment_list = rule.run(
         parameters["run"]["payment_cycle"], parameters["run"]["reward_program_id"]
     )
-    tree = PaymentTree(payment_list.to_dict("records"))
+    tree = PaymentTree(payment_list.to_dict("records"), run_parameters=parameters)
```
tintinthong, line 60: I am fine with "run_parameters". 

When I think, I tend to refer metadata as something that occurs after a run whereas I tend to refer to parameters as something that occurs before a run. 

Since, "run_parameters" can be considered as something reproducible before a run. I prefer it over something e.g. "run_metadata". It's also clear since you are extracting "run_parameters" from the metadata of parquet file.

packages/cardpay-reward-programs/tests/test_add_metadata.py
```diff
@@ -0,0 +1,53 @@
...
+
+import pyarrow.parquet as pq
+from cardpay_reward_programs.payment_tree import PaymentTree
+from cardpay_reward_programs.utils import write_parquet_file
+from cloudpathlib import AnyPath
+
+
+def test_adds_metadata_to_payment_tree():
+    # Values don't matter here, but this is representative of nested data
+    metadata = {
```
tintinthong, line 12: Do we want to group the metadata into several components 

Keep "run_parameters" in a separate key?? In case, we want any additive changes to metadata.

tintinthong (approved):
LGTM in general. I like this strategy of one file per folder. And the use of metadata from parquet. I made a few comments.

packages/cardpay-reward-programs/pdm.lock
```diff
@@ -637,19 +434,14 @@ version = "1.0"
 requires_python = ">=2.7"
 git = "https://git@github.com/cardstack/merklelib"
 ref = "master"
-revision = "1613ccc2602230a2c6f6ce853fb6506e6055e452"
+revision = "af849ae7151077597cda94e87ff931b5b77c899a"
```
tintinthong, line 437: I noticed there were some errors with the test. So I decided to update the lock file. This update should fix the test error where they state "Collections.Iterable" not found

IanCal: Thanks!

packages/cardpay-reward-programs/tests/test_add_metadata.py
IanCal: Yeah the run_parameters are under their own key, so we can add more things if required. I partly did that because I think pandas and other tools tend to add some of their own metadata so I think this can help us avoid clashes in the future.

packages/cardpay-reward-programs/cardpay_reward_programs/main.py
IanCal: That distinction makes sense to me, thanks for going through it.

PR #3257, opened 2022-09-26
sdk, cli: Add scheduled payment

jurgenwerk:
This PR introduces a new approach for scheduling payments (one-time and recurring). Instead of having the client pass a signature to the hub which is responsible for submitting a transaction (previous approach, can be seen in this closed PR: https://github.com/cardstack/cardstack/pull/3231), the implementation changed to a way where `schedulePayment` SDK call orchestrates *the whole flow*, which makes the process easier to reason about:

1. Receiving all the params needed for a scheduled payment
2. Creating the scheduled payment hash by calling the pure function on the module contract
3. Getting the signature and submitting the transaction to add the scheduled payment to the blockchain (i.e. storing the `spHash` in the safe's module contract) 
4. Saving the scheduled payment in the crank (hub), making sure it's persisted before the on-chain flow begins
5. Updating the scheduled payment in the crank once tx hash is known, which triggers the async worker in the hub that waits until transaction is mined
5. Waiting until the transaction is mined by polling the hub and checking if creation block is written
6. Making it available to the client to pass in the `scheduledPaymentId` to resume waiting in case of timeouts or client restarts

The flow is depicted by this diagram (only the SDK part - the hub part will come in a separate PR):
image

### DIFF
--- a/packages/cardpay-cli/scheduled-payment/schedule-payment.ts
+++ b/packages/cardpay-cli/scheduled-payment/schedule-payment.ts
@@ -1,51 +1,141 @@
 import { Argv } from 'yargs';
-import { getConstant, getSDK } from '@cardstack/cardpay-sdk';
+import { getSDK } from '@cardstack/cardpay-sdk';
 import { getEthereumClients, getConnectionType, NETWORK_OPTION_ANY } from '../utils';
 import { Arguments } from 'yargs';
 
 export default {
-  command: 'schedule-payment <safeAddress> <safeModuleAddress> <gasTokenAddress> <spHash>',
-  describe: ' ',
+  command:
+    'schedule-payment <safeAddress> <moduleAddress> <tokenAddress> <amount> <payeeAddress> <fixedUSDFee> <feePercentage> <executionGas> <maxGasPrice> <gasTokenAddress> <salt> <payAt> <recurringDayOfMonth> <recurringUntil>',
+  describe:
+    "Schedules a one-time or a recurring payment. This creates a record in the crank (hub) and stores the scheduled payment hash in the safe's scheduled payment module contract. The crank is responsible for executing the payment at the scheduled time.",
   builder(yargs: Argv) {
     return yargs
       .positional('safeAddress', {
         type: 'string',
         description: 'The address of the safe that will fund the scheduled payment',
       })
-      .positional('safeModuleAddress', {
+      .positional('moduleAddress', {
         type: 'string',
         description: 'The address of the scheduled payment safe module',
+      })
+      .positional('tokenAddress', {
+        type: 'string',
+        description: 'The address of the token to be transferred',
+      })
+      .positional('amount', {
+        type: 'string',
+        description: 'The amount of tokens to be transferred',
+      })
+      .positional('payeeAddress', {
+        type: 'string',
+        description: 'The address of the payee (recipient)',
+      })
+      .positional('fixedUSDFee', {
+        type: 'number',
+        description: 'Fixed USD fee (e.g. 0.25)',
+      })
+      .positional('percentageFee', {
+        type: 'number',
+        description: 'Percentage fee (e.g. 5%, 0.05)',
+      })
+      .positional('executionGas', {
+        type: 'number',
+        description: 'The gas limit to execute scheduled payment',
+      })
+      .positional('maxGasPrice', {
+        type: 'string',
+        description: 'Maximum gas price (in the smallest units of gas token)',
       })
       .positional('gasTokenAddress', {
         type: 'string',
         description: 'The address of the gas token',
       })
-      .positional('spHash', {
+      .positional('salt', {
         type: 'string',
-        description: 'Keccak hash of the scheduled payment params',
+        description: 'Arbitrary string to make SP unique',
+      })
+      .positional('payAt', {
+        type: 'number',
+        description:
+          'Unix UTC time in seconds that represents the point in time when the payment should be executed. Used for one-time scheduled payments. Should be an empty string when recurringDayOfMonth and recurringUntil are set',
+      })
+      .positional('recurringDayOfMonth', {
+        type: 'number',
+        description:
+          'Day of the month on which the payment will be made recurringly (range: 1-31). Used for recurring scheduled payments. In case the month has less than days than the value provided, the payment will me made on the last day of the month. Should be an empty string when payAt is set.',
+      })
+      .positional('recurringUntil', {
+        type: 'number',
+        description:
+          'Unix UTC time in seconds that represents the point in time when the recurring payment should be stopped. Used for recurring scheduled payments. Should be an empty string when payAt is set',
       })
       .option('network', NETWORK_OPTION_ANY);
   },
   async handler(args: Arguments) {
-    let { network, safeAddress, gasTokenAddress, safeModuleAddress, spHash } = args as unknown as {
+    let {
+      network,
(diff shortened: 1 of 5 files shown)


packages/cardpay-cli/scheduled-payment/schedule-payment.ts
```diff
@@ -1,51 +1,141 @@
...
+      })
+      .positional('payAt', {
+        type: 'number',
+        description:
+          'The unix UTC time in seconds. Used for one-time scheduled payments. Should be an empty string when recurringDayOfMonth and recurringUntil are set',
+      })
+      .positional('recurringDayOfMonth', {
+        type: 'number',
+        description:
+          'Day of the month on which the payment will be made recurringly (range: 1-31). Used for recurring scheduled payments. In case the month has less than days than the value provided, the payment will me made on the last day of the month. Should be an empty string when payAt is set.',
```
jurgenwerk, line 65: We originally planned to have this param be limited in the 1-28 range but I think @habdelra's comment here makes sense: https://github.com/cardstack/cardstack/pull/3220#discussion_r970995395

packages/cardpay-sdk/sdk/scheduled-payment-module.ts
```diff
@@ -669,16 +671,189 @@ export default class ScheduledPaymentModule {
...
+    feePercentage: number,
+    executionGas: number,
+    maxGasPrice: string,
+    gasTokenAddress: string,
+    salt: string,
+    payAt: number | null,
+    recurringDayOfMonth?: number | null,
+    recurringUntil?: number | null,
+    onScheduledPaymentCreate?: (scheduledPaymentId: string) => unknown
+  ): Promise<{ scheduledPaymentId: string; promise: Promise<void> }>;
```
habdelra, line 717: since we pass a callback to get the scheduled payment ID as a parameter, probably we don't also need to return it in the response. when I originally proposed returning the schedule payment ID as a response, i didn't actually realize that we were also providing it to callers via a callback as well. i think we can just pick a single approach for returning this value to the callers (such that they don't have to wait for the txn to be mined).

jurgenwerk: Oh, I forgot to delete this return type! It should be `Promise<void>`. Will change this. I chose to provide the scheduled payment id in a callback so that it's immediately available to the caller once the record gets created in the hub, without waiting for the transaction to be created.

packages/cardpay-sdk/sdk/scheduled-payment/utils.ts
```diff
@@ -0,0 +1,13 @@
+import { hubRequest, poll } from '../utils/general-utils';
+
+export async function waitUntilSchedulePaymentTransactionMined(
+  hubRootUrl: string,
+  scheduledPaymentId: string,
+  authToken: string
+) {
+  return poll(
+    () => hubRequest(hubRootUrl, `api/scheduled-payments/${scheduledPaymentId}`, authToken, 'GET'),
+    (response: any) => response.data.attributes['creation-block-number'] != null,
```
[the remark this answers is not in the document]
jurgenwerk: Maybe we can add an error field to the scheduled_payments table and let the background worker update it in case of a revert. The poll can then check for that too.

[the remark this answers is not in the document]
habdelra: good idea

packages/cardpay-sdk/sdk/scheduled-payment-module.ts
jurgenwerk: Fixed.

packages/cardpay-sdk/sdk/scheduled-payment/utils.ts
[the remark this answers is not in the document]
jurgenwerk: Added

packages/cardpay-sdk/sdk/scheduled-payment-module.ts
```diff
@@ -680,16 +682,189 @@ export default class ScheduledPaymentModule {
...
+        { fixedUSD: feeFixedUSD, percentage: feePercentage },
+        executionGas,
+        maxGasPrice,
+        gasTokenAddress,
+        salt,
+        payAt!
+      );
+    }
+
+    let signature = await this.generateSchedulePaymentSignature(safeAddress, moduleAddress, gasTokenAddress, spHash);
```
FadhlanR, line 803: I think we can remove this line from this function and let the `schedulePaymentOnChain` function generate the signature.

I also think the `generateSchedulePaymentSignature` function can directly receive nonce, estimate, and payload as parameters, so the `gasEstimate` and `getNextNonceFromEstimate` won't be called twice in this flow.

alex-cardstack (approved):

packages/cardpay-sdk/sdk/scheduled-payment-module.ts
jurgenwerk: Great catch! Will fix it

packages/cardpay-sdk/sdk/scheduled-payment-module.ts
```diff
@@ -653,48 +671,194 @@ export default class ScheduledPaymentModule {
...
+    authToken: string,
+    scheduledPaymentId: string,
     safeAddress: string,
     moduleAddress: string,
     gasTokenAddress: string,
     spHash: string
   ) {
-    let contract = new this.web3.eth.Contract(ScheduledPaymentABI as AbiItem[], moduleAddress);
-    let payload = await contract.methods.schedulePayment(spHash).encodeABI();
+    let txnParams = await this.generateSchedulePaymentTxParams(safeAddress, moduleAddress, gasTokenAddress, spHash);
```
FadhlanR, line 686: I think we can put this part on the `scheduledPaymentOnChain` function as well.

jurgenwerk: You mean like this? (it's already there): https://github.com/cardstack/cardstack/pull/3257/files#diff-d85a97b0da51ca637fc08ca1b5f782d0d6bdc949ccf414465ab0236d607db756R891

FadhlanR: Yes, so I think this line can be removed?

jurgenwerk: I want to get the signature and the rest of the params before submitting the data to the hub. So in case anything fails with `generateSchedulePaymentTxParams`, we won't be saving anything to the hub. That's why I'm generating this data earlier and passing it in.

FadhlanR: But I don't see you could achieve that by putting `generateSchedulePaymentTxParams` here. If you want to make sure the `generateSchedulePaymentTxParams` work before submitting to the hub, you should put it in the `schedulePayment` before post payment data to the hub. And I think it's unnecessary because you already had the error catch to delete the data to the hub if the `schedulePaymentOnChain` failed.

jurgenwerk: Oh yeah I see now, I should have placed `generateSchedulePaymentTxParams` call higher in the call chain. For defensive purposes I still think it's more robust to attempt that call before the POST rather than rely on the error handler cleanup

FadhlanR (approved):
LGTM, thanks!

PR #3258, opened 2022-09-26
sdk & cli: Migrate from web3 to ethers for scheduled payment commmands

FadhlanR:
In this PR I did some updates on scheduled payment SDK & CLI to use `ethers.js` instead of `web3`.


jurgenwerk (commented):
Can you describe a little how will this change benefit us? 

Will non-schedule payment functions also work after this change? For example I had to revert to using '@truffle/hdwallet-provider' locally order to prevent the "No account" error when using the hub auth command.

packages/cardpay-sdk/providers/json-rpc-provider.ts
```diff
@@ -0,0 +1,170 @@
+import { BigNumber, logger, providers } from 'ethers';
```
jurgenwerk, line 1: Can we import this file from ethers somehow? I suppose it's copied from https://github.com/ethers-io/ethers.js/blob/master/packages/providers/src.ts/json-rpc-provider.ts

FadhlanR:
@jurgenwerk The main purpose of this PR is to solve the problem that occurs when we use the `web3` library. The problem is about `Account not found` and `Chain not found`. These two problems become a dilemma because we can solve one of them but trigger the other to happen. That's why we decided to move from `web3` to `ethers`.

I made changes that are compatible with the previous version which means non-scheduled payment functions should work fine.

packages/cardpay-sdk/providers/json-rpc-provider.ts
FadhlanR: I updated the way this provider handles error messages because we use error messages to calculate safe addresses and estimate execution gas.

FadhlanR:
I have tested these changes using the `ethersMnemonic`, `mnemonic`, and `walletConnect` options.

packages/cardpay-sdk/providers/json-rpc-provider.ts
jurgenwerk: In that case I think it would be good to add a comment with that the file was copied and modified. Or even better, commit the copied file and then commit the change so it's clearer what we copied and what we changed

FadhlanR: Added comments and commits provider modifications

jurgenwerk (approved):
There are a lot of changes here and it's not easy for me to review this just by reading this new code. I also am not sure what to do to thoroughly test this locally. Given the current SDK is broken in several places without what this PR promises to do, I'd like to get this merged so I approve this and trust it was tested well 👍

alex-cardstack (commented):
LGTM

PR #3259, opened 2022-09-26
boxel: Fix ControlPanel contextual yield

backspace:
This fixes [a lint failure](https://github.com/cardstack/cardstack/actions/runs/3128935827/jobs/5077520166). I don’t know why it didn’t show up in #3254.

I also don’t understand the SSR Percy diff, how could that be related? 🧐


lukemelia:
The percy diff doesn't seem like it could be related to this change. It seems more likely related to my changes to the StyledQRCode component on Friday.

lukemelia (approved):

PR #3262, opened 2022-09-26
Convert Boxel::CoverArt to .gts

lukemelia:


packages/boxel/addon/components/boxel/cover-art/usage.gts
```diff
@@ -0,0 +1,94 @@
...
+          @name="spacingMultiplier"
+          @defaultValue={{1}}
+          @min={{0}}
+          @max={{1.5}}
+          @step={{0.1}}
+          @description="Optional adjustment for how closely the covers are displayed"
+          @value={{this.spacingMultiplier}}
+          @onInput={{fn (mut this.spacingMultiplier)}}
+        />
+      </:api>
```
backspace, line 91: I can’t see why but in the preview deployment this slider shows the value below:

Boxel Components 2022-09-26 15-35-48

backspace (approved):

packages/boxel/addon/components/boxel/cover-art/usage.gts
lukemelia: This was missing an actual default value I believe, which I specified, so that must be the cause.

PR #3265, opened 2022-09-26
boxel: Add SelectableTokenAmount

backspace:
This is meant to work like this design:

image

It’s adapted from `TokenAmount` but needs a lot of work still.

### DIFF
--- a/packages/boxel/addon/components/boxel/input/selectable-token-amount/index.css
+++ b/packages/boxel/addon/components/boxel/input/selectable-token-amount/index.css
@@ -1,0 +1,67 @@
+.boxel-selectable-token-amount-input-group {
+  --input-height: 3.75rem; /* 60px */
+  --input-icon-size: var(--boxel-icon-lg);
+  --input-symbol-width: 7rem;
+  --input-icon-space: var(--boxel-sp-xs);
+  --input-symbol-space: var(--boxel-sp-sm);
+
+  position: relative;
+  width: 100%;
+  font-family: var(--boxel-font-family);
+  font-size: 1.25rem;
+  line-height: calc(27 / 20);
+  letter-spacing: var(--boxel-lsp-xs);
+}
+
+.boxel-selectable-token-amount-input-group__input {
+  height: var(--input-height);
+  padding-left: calc(var(--input-icon-size) + var(--input-icon-space) * 2);
+  padding-right: calc(var(--input-symbol-width) + var(--input-symbol-space) * 2);
+  font: inherit;
+  letter-spacing: inherit;
+}
+
+.boxel-selectable-token-amount-input-groxup__icon {
+  position: absolute;
+  width: var(--input-icon-size);
+  height: var(--input-height);
+  left: var(--input-icon-space);
+  top: 0;
+  user-select: none;
+}
+
+.boxel-selectable-token-amount-input-group .ember-basic-dropdown {
+  display: contents;
+}
+
+.boxel-selectable-token-amount-input__select {
+  position: absolute;
+  max-width: var(--input-symbol-width);
+  height: var(--input-height);
+  display: flex;
+  align-items: center;
+  top: 0;
+  right: var(--input-symbol-space);
+  user-select: none;
+}
+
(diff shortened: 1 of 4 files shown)


packages/boxel/addon/components/boxel/input/selectable-token-amount/index.css
```diff
@@ -0,0 +1,64 @@
...
+
+.boxel-input-selectable-token-amount .ember-power-select-status-icon {
+  background: url("/@cardstack/boxel/images/icons/caret-down.svg") no-repeat;
+  width: 11px;
+  height: 9px;
+  display: inline-block;
+}
+
+.boxel-input-selectable-token-amount .ember-basic-dropdown {
+  display: contents;
```
Aierie, line 29: I couldn't find an element that matches this rule. Could we remove it?

packages/boxel/addon/components/boxel/input/selectable-token-amount/index.css
```diff
@@ -0,0 +1,64 @@
...
+.boxel-input-selectable-token-amount__icon {
+  width: var(--boxel-icon-sm);
+  height: var(--boxel-icon-sm);
+}
+
+.boxel-input-selectable-token-amount__select[aria-disabled="true"] {
+  opacity: 0.5;
+}
+
+.boxel-input-selectable-token-amount__select[aria-disabled="true"] .ember-power-select-selected-item {
```
Aierie, line 62: The cursor on the caret is still "pointer" when disabled. I think there's a Boxel rule that is targeting disabled buttons this way:

```css
button:hover:not(:disabled),
[role="button"]:hover:not(:disabled) {
  cursor: pointer;
}
```

We could change it to also include `[aria-disabled="true"]`?

```css
button:hover:not(:disabled), 
[role="button"]:hover:not(:disabled, [aria-disabled="true"]) {
  cursor: pointer;
}
```

In which case we wouldn't need this rule, I think.

packages/boxel/addon/components/boxel/input/selectable-token-amount/index.css
```diff
@@ -0,0 +1,64 @@
...
+  height: var(--boxel-input-selectable-token-amount-height);
+  display: flex;
+  align-items: center;
+  top: 0;
+  right: 0;
+  user-select: none;
+  padding-right: var(--boxel-input-selectable-token-amount-horizontal-padding);
+}
+
+.boxel-input-selectable-token-amount__select .boxel-input-selectable-token-amount__dropdown-item.boxel-select__item,
```
Aierie, line 43: We've been doing this for a while, and it works for this PR but may be interesting to consider in the context of cards, this style needs to be globally-scoped because it is applied to a wormhole... shadow DOM usage might mean we need to rethink our default dropdown rendering method?

packages/boxel/addon/components/boxel/input/selectable-token-amount/index.css
backspace: yes!

packages/boxel/addon/components/boxel/input/selectable-token-amount/index.css
backspace: much better, thank you!

packages/boxel/addon/components/boxel/input/selectable-token-amount/index.css
```diff
@@ -0,0 +1,64 @@
...
+  --boxel-input-selectable-token-amount-horizontal-padding: var(--boxel-sp-sm);
+  --boxel-input-selectable-token-amount-input-font-size: var(--boxel-font-size);
+
+  position: relative;
+  width: 100%;
+  font-family: var(--boxel-font-family);
+  font-size: var(--boxel-input-selectable-token-amount-input-font-size);
+}
+
+.boxel-input-selectable-token-amount .boxel-input-selectable-token-amount__input {
```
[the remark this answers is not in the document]
backspace: ya… attaching an action to `onChange` that adjusts the input’s right padding seems like it would work. BUT I would like to defer this until it’s actually needed, I think for the initial use case it won’t be…???!?!?!

PR #3267, opened 2022-09-27
Convert Boxel::CardPicker to .gts

lukemelia:
- Also, avoid super-narrow menu when choosing with a card already selected


packages/boxel/addon/components/boxel/card-picker/index.css
```diff
@@ -15,6 +15,7 @@
 
 .boxel-card-picker__dropdown {
   position: static;
+  width: var(--boxel-card-picker-width);
```
backspace, line 18: I agree that the ultra-narrowness needs to be addressed but maybe this makes the button look a bit strange when a card has already been chosen? (new on left)

Boxel Components 2022-09-27 09-49-38

backspace (approved):

packages/boxel/addon/components/boxel/card-picker/index.css
lukemelia: Yes, I agree that this ugliness is... ugly. I decided not to spend more time on it for now since it has yet to be integrated into our production apps or designs.

backspace: ya makes sense!

PR #3269, opened 2022-09-27
Convert Boxel::DropTarget to .gts

lukemelia:


packages/boxel/addon/components/boxel/drop-target/usage.gts
```diff
@@ -0,0 +1,29 @@
+import Component from '@glimmer/component';
+import { tracked } from '@glimmer/tracking';
+import FreestyleUsage from 'ember-freestyle/components/freestyle/usage';
+import BoxelDropTarget from './index';
+//@ts-expect-error glint does not think this is consumed-but it is consumed in the template https://github.com/typed-ember/glint/issues/374
+import { fn, array } from '@ember/helper';
+
+export default class DropUsage extends Component {
+  @tracked state = 'rest';
```
danibonilha, line 9: nit: we could skip a line here

packages/boxel/addon/components/boxel/drop-target/index.gts
```diff
@@ -0,0 +1,33 @@
...
+
+export interface Signature {
+  Element: HTMLDivElement;
+  Args: {
+    state?: string;
+  };
+  Blocks: EmptyObject;
+}
+
+const DropTarget: TemplateOnlyComponent<Signature> = <template>
```
danibonilha, line 18: I was wondering if there's some sort of pattern for closing and opening template tags.

for example,  add `<template>`  the way it is right after the `=` sign, or add it after skipping a line.

danibonilha (approved):
LGTM, small nit

packages/boxel/addon/components/boxel/drop-target/usage.gts
lukemelia: 👍🏼

packages/boxel/addon/components/boxel/drop-target/index.gts
lukemelia: This format is new enough that there are not community standards yet AFAIK. I have been doing <template> after the equal sign for template only components and after a blank line for class components.

PR #3273, opened 2022-09-28
hub: add endpoints for scheduled payments 

jurgenwerk:
This PR adds POST/GET/PATCH/DELETE endpoints for scheduled payments, with a specialty in the PATCH endpoint which updates `creation-transaction-hash` and spawns a background worker that will wait until the transaction is done and update `creation-block-number` once the transaction has been mined. 

The consumer of these endpoints is the SDK. Please check PR https://github.com/cardstack/cardstack/pull/3257 for more info on how this flow works. 

One potential problem ➡️ we identified an edge case where the SDK call to schedule a payment could only succeed to the point of successfully calling POST, but no further (due to various system errors that could occur). In that case, we'd end up with a scheduled payment saved in the hub which has no `creation-transaction-hash`. That means we don't know if it has been mined or if was even submitted as a transaction on the blockchain. We need a watchdog process that continuously looks for these cases and tries to update their `creation-transaction-hash` (and spawn the waiter task) by looking at the blockchain event logs for the schedule payment event and filtering by `spHash`. In case no transaction could be found that way, we should remove that record from the hub. 

This edge case will be addressed as an automated process in a separate PR but it seems low priority for now.

### DIFF
--- a/packages/cardpay-sdk/sdk/scheduled-payment-module.ts
+++ b/packages/cardpay-sdk/sdk/scheduled-payment-module.ts
@@ -735,8 +735,8 @@ export default class ScheduledPaymentModule {
     recurringUntil?: number | null,
     onScheduledPaymentCreate?: (scheduledPaymentId: string) => unknown
   ) {
-    let hubAuth = await getSDK('HubAuth', this.web3);
-    let hubRootUrl = await hubAuth.getHubUrl();
+    let hubAuth = await getSDK('HubAuth', this.web3, 'http://localhost:3000');
+    let hubRootUrl = 'http://localhost:3000'; //await hubAuth.getHubUrl();
     let authToken = await hubAuth.authenticate();
 
     let safeAddress: string;
--- a/packages/cardpay-sdk/sdk/scheduled-payment-module.ts
+++ b/packages/cardpay-sdk/sdk/scheduled-payment-module.ts
@@ -735,8 +735,8 @@ export default class ScheduledPaymentModule {
     recurringUntil?: number | null,
     onScheduledPaymentCreate?: (scheduledPaymentId: string) => unknown
   ) {
-    let hubAuth = await getSDK('HubAuth', this.web3, 'http://localhost:3000');
-    let hubRootUrl = 'http://localhost:3000'; //await hubAuth.getHubUrl();
+    let hubAuth = await getSDK('HubAuth', this.web3);
+    let hubRootUrl = await hubAuth.getHubUrl();
     let authToken = await hubAuth.authenticate();
 
     let safeAddress: string;
--- a/packages/hub/db/migrations/20220921073021534_alter-scheduled-payment-fields-2.ts
+++ b/packages/hub/db/migrations/20220921073021534_alter-scheduled-payment-fields-2.ts
@@ -1,0 +1,24 @@
+import { MigrationBuilder } from 'node-pg-migrate';
+const TABLE = 'scheduled_payments';
+
+export async function up(pgm: MigrationBuilder): Promise<void> {
+  pgm.addColumns(TABLE, {
+    user_address: { type: 'text', notNull: true },
+    creation_transaction_error: { type: 'text' },
+    cancellation_transaction_error: { type: 'text' },
+  });
+
+  pgm.createIndex(TABLE, 'user_address');
+
+  pgm.alterColumn(TABLE, 'fee_fixed_usd', { type: 'numeric', notNull: true });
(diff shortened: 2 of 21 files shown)


packages/cardpay-sdk/sdk/scheduled-payment-module.ts
```diff
@@ -735,8 +735,8 @@ export default class ScheduledPaymentModule {
     recurringUntil?: number | null,
     onScheduledPaymentCreate?: (scheduledPaymentId: string) => unknown
   ) {
-    let hubAuth = await getSDK('HubAuth', this.web3);
-    let hubRootUrl = await hubAuth.getHubUrl();
+    let hubAuth = await getSDK('HubAuth', this.web3, 'http://localhost:3000');
```
habdelra, line 738: how will this API work when called from the CLI with this hard coded hub endpoint?

jurgenwerk: Good catch - forgot to remove this argument.

habdelra (approved):

packages/hub/db/migrations/20220921073021534_alter-scheduled-payment-fields-2.ts
```diff
@@ -0,0 +1,24 @@
...
+  });
+
+  pgm.createIndex(TABLE, 'user_address');
+
+  pgm.alterColumn(TABLE, 'fee_fixed_usd', { type: 'numeric', notNull: true });
+  pgm.alterColumn(TABLE, 'fee_percentage', { type: 'numeric', notNull: true });
+}
+
+export async function down(pgm: MigrationBuilder): Promise<void> {
+  pgm.dropColumn(TABLE, 'user_address');
```
backspace, line 18: is a corresponding `dropIndex` needed first?

packages/hub/node-tests/routes/scheduled-payments-test.ts
```diff
@@ -0,0 +1,561 @@
...
+        },
+      })
+      .set('Accept', 'application/vnd.api+json')
+      .set('Authorization', 'Bearer abc123--def456--ghi789')
+      .set('Content-Type', 'application/vnd.api+json')
+      .expect(422)
+      .expect({
+        errors: [
+          {
+            detail: 'is required',
```
backspace, line 59: Maybe it’s redundant because it’s in `source.pointer` but the name `detail` suggests to me that it should say _what_ is required 🤓

packages/hub/routes/scheduled-payments.ts
```diff
@@ -0,0 +1,208 @@
...
+    let scheduledPaymentId: string = ctx.params.scheduled_payment_id;
+
+    let scheduledPayment = await prisma.scheduledPayment.findFirstOrThrow({
+      where: {
+        id: scheduledPaymentId,
+        userAddress,
+      },
+    });
+
+    if (scheduledPayment.userAddress !== userAddress) {
```
backspace, line 188: Is this possible, considering that `userAddress` is in the database query?

packages/hub/tasks/scheduled-payment-on-chain-creation-waiter.ts
```diff
@@ -0,0 +1,95 @@
...
+      where: { id: payload.scheduledPaymentId },
+    });
+
+    if (scheduledPayment.creationBlockNumber) {
+      // Did we somehow spawn this task after it already updated creationBlockNumber previously?
+      return Sentry.captureException(
+        new Error('Task run for scheduled payment that already has a creation block number'),
+        {
+          tags: {
+            action: 'scheduled-payment-task',
```
backspace, line 24: You could set up an “Alert” on Sentry to get email notifications if/when these errors are encountered, then you don’t have to be checking Sentry manually

alert-web-team - cardstack - hub - Sentry 2022-09-28 12-28-48

backspace (approved):

packages/hub/routes/scheduled-payments.ts
jurgenwerk: You are correct, good catch! I should replace `findFirstOrThrow` with `findFirst` and use `if (scheduledPayment) {} else {}`

packages/hub/db/migrations/20220921073021534_alter-scheduled-payment-fields-2.ts
jurgenwerk: When you drop column in a table, all indexes and constraints which use this column will be also dropped silently

PR #3277, opened 2022-09-28
[CS-4626]: Refactor PowerSelect components to use BoxelSelect

danibonilha:
This PR replaces PowerSelect with BoxelSelect on components used in boxel, the webclient migration to BoxelSelect will happen on a diff PR. It also fixes the CardPicker "Change Card" layout.

### DIFF
--- a/packages/boxel/addon/components/boxel/card-picker/index.css
+++ b/packages/boxel/addon/components/boxel/card-picker/index.css
@@ -1,4 +1,3 @@
-/* stylelint-disable max-line-length */
 .boxel-card-picker {
   --boxel-card-picker-width: 24rem;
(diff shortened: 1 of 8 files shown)


lukemelia (approved):
Nice! Good to see that the styles got simpler, too!

backspace (approved):
The card picker placeholder text has become bold but who cares since it’s not in use 😆

Boxel Components 2022-09-28 16-26-45

danibonilha:
> The card picker placeholder text has become bold but who cares since it’s not in use 😆
> 
> Boxel Components 2022-09-28 16-26-45

ya, I noticed that, but thought it's ok, when there's usage we can change it

PR #3285, opened 2022-09-29
Simplify glint config

lukemelia:
- Now that all of boxel addon has been converted to glint, we can simplify the3 includes in consuming projects


backspace (approved):
🥳

danibonilha (approved):
👏

PR #3286, opened 2022-09-29
[Refactor]: Replace PowerSelect components w/ Boxel Select on web-client

danibonilha:
This PR replaces PowerSelect components w/ Boxel Select on web-client for two components, SafeChooser and BalanceChooser, they both have similar behaviors, I thought  they could be a single component and I'm pretty sure we can simplify the styles more, but I guess I need more web experience to do so, I've cracked my head a bit trying it, but some styles were not being overridden, anyways, this approach is the most secure and since it's on the web-client I didn't want to risk it.

### DIFF
--- a/packages/web-client/app/components/card-pay/balance-chooser-dropdown/index.css
+++ b/packages/web-client/app/components/card-pay/balance-chooser-dropdown/index.css
@@ -1,5 +1,6 @@
 .balance-chooser-dropdown {
   --field-height: 3.75rem;
+  --boxel-select-selected-color: transparent;
 
   position: relative;
   min-height: var(--field-height);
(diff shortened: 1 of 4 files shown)


packages/web-client/app/components/card-pay/balance-chooser-dropdown/index.css
```diff
@@ -28,11 +29,13 @@
...
 
-.balance-chooser-dropdown .ember-basic-dropdown-content.ember-basic-dropdown--transitioning-in {
+.balance-chooser-dropdown
+  .ember-basic-dropdown-content.ember-basic-dropdown--transitioning-in {
   animation: drop-fade-below var(--boxel-transition);
 }
 
-.balance-chooser-dropdown .ember-basic-dropdown-content.ember-basic-dropdown--transitioning-out {
+.balance-chooser-dropdown
+  .ember-basic-dropdown-content.ember-basic-dropdown--transitioning-out {
```
lukemelia, line 38: The underlying ember-basic-dropdown is leaking through our abstraction here. Maybe we should consider introducing CSS variables for this?

lukemelia (approved):

packages/web-client/app/components/card-pay/balance-chooser-dropdown/index.css
danibonilha: Will do on next PR, I noticed we have this animation spread across the components

PR #3287, opened 2022-09-30
[CS-4680]: Move scattered around select animations to single component

danibonilha:
Now that we have a single select component we can have a default animation within it, this PRs moves the drop-fade anim to the select component, and removes it from global styles. Let me know if there's any plan to use this on globals in this case I can revert this change. Also adds a default cursor behavior.

One thing I notice though is that the animation for some components, works on the first opening, but not for the subsequent ones, maybe @Aierie can help investigating this. it happens on production too, so it was not introduced by these changes.


Aierie:
The cause of animations being broken is [this](https://github.com/cibernox/ember-basic-dropdown/issues/583#issuecomment-758079663) issue - an incorrect assumption around timing of removal of the dropdown element is leading to stale state and preventing any future opening (or closing) animations. On my machine, this seems to only happen with `@renderInPlace=true`.

Aierie (approved):
Looks good! 🧹

lukemelia (approved):

PR #3291, opened 2022-10-04
Reward rollover

IanCal:
I can keep adding tests and going through this but thought it worth bringing up for starting to review/answer questions sooner rather than later.

This covers changes to both the scheduler and the rewards.


packages/cardpay-reward-programs/tests/rules/test_rollover.py
```diff
@@ -0,0 +1,119 @@
...
+        # The second run should account for the previous one
+        run_parameters = {
+            "reward_program_id": "0x0885ce31D73b63b0Fcb1158bf37eCeaD8Ff0fC72",
+            "payment_cycle": 200,
+            "previous_output": first_cycle_output / "results.parquet",
+            "rewards_subgraph_location": "s3://partitioned-graph-data/data/staging_rewards/0.0.1/",
+        }
+        payment_list = rollover_rule.get_payments(**run_parameters).to_dict("records")
+        tree = PaymentTree(payment_list)
+        assert len(payment_list) == 1  # Still only one payee
```
tintinthong, line 116: Nit pick. But perhaps 2 payees to test with is better.

packages/cardpay-reward-scheduler/reward_scheduler/main.py
```diff
@@ -54,7 +60,7 @@
 
 def safe_run(reward_program):
     try:
-        reward_program.run_all_payment_cycles()
+        reward_program.run_all_rules()
```
tintinthong, line 63: 👍

packages/cardpay-reward-scheduler/reward_scheduler/utils.py
```diff
@@ -31,61 +14,17 @@ def get_latest_details(config_location):
...
-                    "data.parquet",
-                )
-            )
-    return files
-
-
-def get_files(config_location, table, min_partition=None, max_partition=None):
-    file_list = get_partition_files(
-        AnyPath(config_location), table, min_partition, max_partition
+    table_metadata = config_location.joinpath(
```
tintinthong, line 20: Isn't the use of `table_metadata` here confusing. Some word "path" would help

packages/cardpay-reward-scheduler/reward_scheduler/reward_program.py
```diff
@@ -204,22 +212,43 @@ def run(self, payment_cycle, rule):
...
+    def run_all_rules(self) -> None:
         """
         Run all rules for all payment cycles that can be processed at this time
         """
         logging.info(f"Running all for {self.reward_program_id}")
         if self.is_locked():
             logging.info(f"Reward program {self.reward_program_id} is locked, skipping")
             return
+
+        self.update_processed()
```
tintinthong, line 245: Question: Given we `load_submitted_from_subgraph_extracction`, is it possible we can completely remove `update_processed` function which relies on subgraph??

IanCal: Yeah we could - only impact is latency. Subgraph gives us the data pretty much immediately. 

I'll have a check to see how much it would add because removing a dependency on subgraph completely sounds nice.

IanCal:
Those comments are really useful thanks, the changes make sense so I'll make those, always good to sort naming.

tintinthong (approved):
Nothing to add LGTM

PR #3294, opened 2022-10-05
Add boxel dependency and necessary config and deps to scheduled-payments-client

lukemelia:


packages/scheduled-payments-client/ember-cli-build.js
```diff
@@ -6,11 +6,73 @@ process.env.EMBROIDER_REBUILD_ADDONS = '@cardstack/boxel';
...
+          },
+          {
+            prefixIds: false,
+          },
+          {
+            uniqueID: svgoUniqueId,
+          },
+        ],
+      },
+    },
```
danibonilha, line 33: I've tried to add most of the boxel stuff, but the requirements were not clear, maybe we should think about documenting it, or create some sort of template.

packages/scheduled-payments-client/package.json
```diff
@@ -70,11 +72,13 @@
     "ember-cli-typescript": "^5.1.1",
     "ember-cli-typescript-blueprints": "^3.0.0",
     "ember-fetch": "^8.1.1",
+    "ember-focus-trap": "^1.0.1",
```
danibonilha, line 75: I don't know if there's some ember limitation, but shouldn't boxel, handled all required deps in it's package.json, so we don't need to individually add the package on each client ?

packages/scheduled-payments-client/app/templates/application.hbs
```diff
@@ -1,6 +1,7 @@
 {{page-title 'ScheduledPaymentsClient'}}
 
-{{! TODO: Remove once we have components }}
-<GlintTestComponent />
```
danibonilha, line 4: we can delete this from components folder then

danibonilha (approved):

packages/scheduled-payments-client/package.json
lukemelia: You are generally right. There is a funky embroider/addon bug that necessitates this.

lukemelia: https://github.com/embroider-build/embroider/issues/1175

packages/scheduled-payments-client/ember-cli-build.js
lukemelia: Agreed. It is pretty esoteric right now, mostly with the webpack processing requirements.

PR #3296, opened 2022-10-05
tools: Add layout

backspace:
It looks like this:

Cardstack Safe Tools 2022-10-06 16-52-55

Here’s the design:

Network Default to ETH then Create Safe - MVP - Zeplin 2022-10-06 16-55-21

I think this is a fine place to stop, we can fill in the various missing parts and add functionality (“Connect Wallet”) incrementally.


packages/safe-tools-client/app/templates/application.hbs
```diff
@@ -1,7 +1,48 @@
...
+      style={{
+        css-var
+        boxel-org-header-background-color='var(--boxel-dark)'
+        boxel-org-header-logo-size='var(--boxel-sp-xl)'
+        boxel-org-header-logo-position='right'
+        boxel-org-title-title-font='300 var(--boxel-font-size-lg) var(--boxel-font-family)'
+        boxel-org-title-subtitle-font='300 var(--boxel-font-size-xs) var(--boxel-font-family)'
+        boxel-org-title-letter-spacing='var(--boxel-lsp-sm)'
+        boxel-org-title-text-transform='none'
+      }}
```
danibonilha, line 19: Is there a specific reason to use the style property, instead of using class ?

packages/safe-tools-client/app/templates/application.hbs
```diff
@@ -1,7 +1,48 @@
...
+          'Split' (link route='split') icon='tab-icon-split' inactive=true
+        )
+      }}
+      @spread={{true}}
+      style={{
+        css-var
+        boxel-tab-bar-background-color='var(--boxel-dark)'
+        boxel-tab-bar-border-bottom='0px'
+        boxel-tab-bar-color-active='var(--boxel-cyan)'
+        boxel-tab-bar-font='600 var(--boxel-font-sm)'
```
danibonilha, line 40: ditto

packages/safe-tools-client/app/css/app.css
```diff
@@ -1,8 +1,3 @@
-body {
-  background-color: var(--boxel-purple-600);
+.safe-tools-header .boxel-org-title__subtitle {
+  margin-top: calc(-1 * var(--boxel-sp-xs));
 }
```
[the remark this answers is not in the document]
Aierie: I think it's fine?  I would use a magic number (`-11px`) since this is an adjustment, we don't need the Boxel variable dependency.

packages/safe-tools-client/app/templates/application.hbs
Aierie: Personal preference - I'd go with styles in templates for first-class Boxel CSS variables so that it's clearer what needs to change when Boxel components change.

backspace: I don’t really like it in templates but that’s the usual way we do it so that’s what I stuck with 🥴

PR #3300, opened 2022-10-06
refactor dropdown to use modifier for bindings

Aierie:
Refactored dropdown to provide a modifier that attaches appropriate attributes and event listeners via its `<:trigger>` named block - this modifier is meant to be bound to the trigger component. Also refactored it to not render the default dropdown trigger that `ember-basic-dropdown` provides, because the default dropdown trigger is a button in itself and can lead to nested interactives.

This means that we can now do:

```
<BoxelDropdown>
  <:trigger as |bindings|>
    <BoxelButton {{bindings}}>
        Trigger
    </BoxelButton>
  </:trigger>
  <:content as |dd|>
    {{!--Your content here--}}
  </:content>
</BoxelDropdown>
```

Without nested interactives, and we don't need to introduce `@triggerComponent`.

### DIFF
--- a/packages/boxel/tests/integration/components/boxel/dropdown-test.ts
+++ b/packages/boxel/tests/integration/components/boxel/dropdown-test.ts
@@ -42,6 +42,8 @@ module('Integration | Component | Dropdown', function (hooks) {
 
   // This test is important because event handlers are assigned assuming default HTML button behaviour
   // In particular this includes "free" click events generated from spacebar and enter keys and also touch events
+  // Click events from enter and space are not tested because synthetic keydown events don't generate a click event
+  // https://github.com/emberjs/ember-test-helpers/issues/1054
   test('it errors if it receives a non-button element', async function (assert) {
     let lastError: Error;
     Ember.onerror = (e: Error) => {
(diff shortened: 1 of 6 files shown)


lukemelia (approved):
Nice solution!

packages/boxel/tests/integration/components/boxel/dropdown-test.ts
```diff
@@ -29,45 +35,77 @@ module('Integration | Component | Dropdown', function (hooks) {
     `);
   });
 
-  test('it renders dropdown trigger, but not dropdown content at initialization', async function (assert) {
-    assert.dom(TRIGGER).isVisible().containsText('Trigger');
-    assert.dom(DROPDOWN_CONTENT).doesNotExist();
+  hooks.afterEach(function () {
+    // reset Ember.onerror
+    Ember.onerror = onerror;
```
backspace, line 40: [`setupOnerror`](https://github.com/emberjs/ember-test-helpers/blob/master/API.md#setuponerror) and `resetOnerror` could slightly simplify this 🤓

packages/boxel/tests/integration/components/boxel/dropdown-test.ts
```diff
@@ -29,45 +35,77 @@ module('Integration | Component | Dropdown', function (hooks) {
...
+  // no click event is generated with synthetic keydown
+  // eslint-disable-next-line qunit/no-commented-tests
+  // test('it can open and close on enter', async function (assert) {
+  //   await focus(TRIGGER);
+  //   // https://w3c.github.io/uievents/#fixed-virtual-key-codes
+  //   await triggerKeyEvent(TRIGGER, 'keydown', 13);
+  //   assert.dom(DROPDOWN_CONTENT).isVisible();
+  //   await triggerKeyEvent(TRIGGER, 'keydown', 13);
+  //   assert.dom(DROPDOWN_CONTENT).doesNotExist();
+  // });
```
backspace, line 107: Should we keep these? 🤔

backspace (approved):
Agreed! 🎉

packages/boxel/tests/integration/components/boxel/dropdown-test.ts
Aierie: Yep, done, also moved the rendering error test to its own module so that the `resetOnError` call in afterEach doesn't end up being a red herring for other tests.

packages/boxel/tests/integration/components/boxel/dropdown-test.ts
Aierie: Seemed to be easier to comment at the rendering error test about the synthetic keydown events not generating clicks, so I've done that instead.

PR #3301, opened 2022-10-06
Dropdown button label and icon separation

Aierie:
To merge after #3300 - replace `@button` with `@label`, separate `@label` and `@icon` so that screenreaders are more likely to get a descriptive message.


backspace (approved):
Important intervention, thank you!

PR #3303, opened 2022-10-07
Update ember-template-imports

backspace:
3.1.2 includes the fix for backticks and slashes breaking template compilation. This updates the Boxel components that had a workaround for this to put the template beneath the class getters and functions.


backspace:
Uncertainty: VSCode is showing template errors

cardstack 2022-10-07 11-17-48

However I’ve been seeing things like this a lot while working with `gts` files… is there a mismatch somewhere?

danibonilha (approved):

backspace:
thanks for checking!! I think that’s for a future PR 😝

PR #3304, opened 2022-10-10
[CS-4711]: Add schedule collapsible panel

danibonilha:
This PR adds the schedule collapsible banner, I added all the text in the component, since as of today we are only aware of this banner, if we get another one, we can extract the strings to be customizable, or even have an collapsible banner and yield it's content.

PS: I'm not 100% sure about my html tag choices, so feel free to comment on that 😅 

I've tried to add a smoother animation, but it only worked on the first tap, similar to the dropdown bug we have, so I switched the type of transition to "fake" an animation.

I've added a couple of CSS comments to keep track of my "choices", but let me know if there's any pattern to avoid that.

https://user-images.githubusercontent.com/20520102/195135251-1a151268-161c-4c29-bbb0-ecf8d0ee3c61.mov


packages/safe-tools-client/app/components/schedule-collapse-panel/index.gts
```diff
@@ -0,0 +1,91 @@
...
+              {{svgJar feature.iconName role='listitem'}}
+              <h4>{{feature.title}}</h4>
+              <p>{{feature.description}}</p>
+            </li>
+          {{/each}}
+        </ul>
+      </div>
+      <div class='collapse-panel__details-overlay'>
+        <h3>Start using this payment scheduler dApp:</h3>
+        <ol type='1'>
```
Aierie, line 78: If we're removing the list styles, we can omit `type='1'`

packages/safe-tools-client/app/components/schedule-collapse-panel/index.css
```diff
@@ -0,0 +1,136 @@
...
+
+/* Style ol numbers */
+.collapse-panel__details-overlay ol li:before {
+  content: counter(item);
+  color: var(--boxel-teal);
+  padding-right: var(--boxel-sp);
+  font-weight: 800;
+}
+
+.collapse-panel__details-overlay ol li h4 {
```
Aierie, line 130: I think using a class to target a specific element is preferable when we have control of the template, it makes it clear which things we're targeting.

type selectors can match things that we don't intend to match, especially as a component evolves.

packages/safe-tools-client/app/components/schedule-collapse-panel/index.gts
```diff
@@ -0,0 +1,91 @@
...
+interface Signature {
+  Element: HTMLElement;
+  Args: { 
+    noCollapse?: boolean
+    };
+  Blocks: EmptyObject;
+}
+
+const ScheduleCollapsePanel: TemplateOnlyComponent<Signature> = <template>
+  <details class='collapse-panel' ...attributes open={{@noCollapse}}>
```
Aierie, line 56: Right now, `@noCollapse` doesn't prevent collapse of this panel. In `web-client`, `@noCollapse` works by hiding the `summary`, which is something we could consider. A more involved (but more correct) alternative is to render the container as a noninteractive element if `@noCollapse` is `true`.

Properly implementing `@noCollapse` could be deferred to another PR, and we could remove the `@noCollapse` argument in the meantime, since it doesn't really do what its name suggests.

packages/safe-tools-client/app/components/schedule-collapse-panel/index.css
```diff
@@ -0,0 +1,136 @@
...
+  --collapse-panel-border-radius: 20px;
+  --collapse-panel-summary-height: 80px;
+  --collapse-panel-summary-anim-bottom-margin: 8px;
+
+  background: url('/images/backgrounds/schedule-panel.svg') no-repeat bottom/cover white;
+  border-radius: var(--collapse-panel-border-radius);
+  overflow: hidden;
+}
+
+.collapse-panel__summary {
```
Aierie, line 11:
```
.collapse-panel__summary {list-style: none}
.collapse-panel__summary::-webkit-details-marker {display: none; }
```

this should reset and prevent the summary markers from showing on most browsers - they're currently showing on Safari.

packages/safe-tools-client/app/components/schedule-collapse-panel/index.css
```diff
@@ -0,0 +1,136 @@
...
+  transition: margin-bottom var(--boxel-transition);
+}
+
+/* Adds margin to fake animation */
+.collapse-panel[open] .collapse-panel__summary {
+  margin-bottom: var(--collapse-panel-summary-anim-bottom-margin);
+}
+
+.collapse-panel__summary-collapsed {
+  display: inherit;
```
Aierie, line 27: We need to specify `display: flex; width: 100%;` explicitly because Safari is fussy.

Aierie:
> This looks good, I appreciate the subtle animation 🎉

Agree, the animation is a nice touch!

packages/safe-tools-client/app/components/schedule-collapse-panel/index.gts
```diff
@@ -0,0 +1,91 @@
...
+       iconName: 'punctual'
+  },
+];
+
+interface Signature {
+  Element: HTMLElement;
+  Args: { 
+    noCollapse?: boolean
+    };
+  Blocks: EmptyObject;
```
lukemelia, line 52: I've been told that we can leave `Blocks` off in this situation, rather than specifying it as `EmptyObject`

packages/safe-tools-client/app/components/schedule-collapse-panel/index.css
```diff
@@ -0,0 +1,136 @@
...
+.collapse-panel__details-overlay ol li:before {
+  content: counter(item);
+  color: var(--boxel-teal);
+  padding-right: var(--boxel-sp);
+  font-weight: 800;
+}
+
+.collapse-panel__details-overlay ol li h4 {
+  display: inline;
+}
```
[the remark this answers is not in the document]
danibonilha: ya, I couldn't find a way to handle that, but I'll look into hanging indent

packages/safe-tools-client/app/components/schedule-collapse-panel/index.gts
danibonilha: I think the name of the argument is confusing, the idea for this, is to actually have an option to first render already expanded, instead of the default behavior where it starts collapsed, I'll change the arg name to reflect that.

packages/safe-tools-client/app/components/schedule-collapse-panel/index.css
danibonilha: Thanks for checking that, I totally forgot to test on safari 🤦‍♀️

packages/safe-tools-client/app/components/schedule-collapse-panel/index.css
danibonilha: I thought it was okay for this case, since it's inside the .collapse-panel__details-overlay, but I understand it might be safer to use a class

packages/safe-tools-client/app/components/schedule-collapse-panel/index.css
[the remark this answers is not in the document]
danibonilha: Fixed it:
image

PR #3305, opened 2022-10-11
Cs 4684 sdk update get proofs with explanations

tintinthong:
- **sdk methods now require reward program id**. This is required because each reward program is assumed to have a different explanation block. If no reward program is given, then we would not be able to find the explanation block
- **backward compatible**. The old sdk should be able to query the new reward indexer like normal (since we are only adding columns in the database `explanationId` and `explanationData`
- **added explanationTemplate and explanationData to the proof**. The idea here is to fill in explanationData into the templates given in the UI either the web client or the wallet. 
  - As you can see below, there may be situations where the template run off, ie `rollover_amount` is missing. The UI should handle for the case where the imputation can't be done correctly.  
  - The explanationData essentially doesn't change in the dbs. But the template can change. So in this case, we can update the template onchain to not include `rollover_amount`. 

```
      proof: 0x
      leaf: 0x000000000000000000000000ab20c80fcc025451a3fc73bb953aae1b9f6409490000000000000000000000000000000000000000000000000000000001a468560000000000000000000000000000000000000000000000000000000001a468560000000000000000000000000000000000000000000000000000000001b045d60000000000000000000000000000000000000000000000000000000000000001000000000000000000000000159ade032073d930e85f95abbab9995110c43c7100000000000000000000000000000000000000000000000000000000000000e00000000000000000000000000000000000000000000000000000000000000040000000000000000000000000b0427e9f03eb448d030be3ebc96f423857ceeb2f0000000000000000000000000000000000000000000000008ac7230489e80000
      balance: 10 CARD.CPXD
      token: 0xB0427e9F03Eb448D030bE3EBC96F423857ceEb2f (CARD.CPXD)
      isValid: true 
      gasFees: No Gas Estimates
      explanationTemplate: Earned {amount} {token} for a one-time aidrop. {rollover_amount} tokens rolled over from unclaimed previous cycles 
      explanationData: 
         {
    "amount": "10000000000000000000",
    "token": "0xB0427e9F03Eb448D030bE3EBC96F423857ceEb2f"
} 
```

PS will update docs after review

You can look at 0xab20c80fcc025451a3fc73bB953aaE1b9f640949 reward program id for staging (there is an issue with 0x0885ce31D73b63b0Fcb1158bf37eCeaD8Ff0fC72)


packages/cardpay-cli/rewards/utils.ts
```diff
@@ -17,12 +17,13 @@ export function displayRewardTokenBalance(tokenBalances: WithSymbol<RewardTokenB
...
 export function displayRewardProgramInfo(rewardProgramInfo: RewardProgramInfo): void {
-  let { rewardProgramId, rewardProgramAdmin, locked, rule, tokenBalances } = rewardProgramInfo;
+  let { rewardProgramId, rewardProgramAdmin, locked, blob, tokenBalances, programExplainer } = rewardProgramInfo;
   console.log(`
   rewardProgramId : ${rewardProgramId}
   rewardProgramAdmin : ${rewardProgramAdmin}
   locked : ${locked}
-  rule : ${rule ? rule : 'No rule'}
+  ruleBlob : ${blob ? blob : 'No rule blob'}
+  programExplainer : ${programExplainer ? programExplainer : 'No rule blob'}
```
paulinhapenedo, line 26: nit: shouldn't this be "No programExplainer" instead?

packages/cardpay-cli/rewards/claimable-proofs.ts
```diff
@@ -88,6 +88,9 @@ function displayProof(o: WithSymbol<Proof> | WithSymbol<ClaimableProof>, rewardP
       gasFees: ${
         rewardPool.isClaimableProof(o) ? fromWei(o.gasEstimate.amount) + ' ' + o.tokenSymbol : 'No Gas Estimates'
       }
+      explanationTemplate: ${o.explanationTemplate ?? '**Not specified**'} 
+      explanationData: 
+         ${JSON.stringify(o.explanationData, null, 4) ?? '**Not specified**'}
```
paulinhapenedo, line 93: Having `**Not specified**` with the asterisks is proposital as some sort of markdown? Will the rewards program have the ability to add Markdown to their explanations?

oidoug: This fn only prints to console, so it doesn't matter that much the format used. Maybe the name `displayProof` does not convey that properly. I would suggest something like `logProof`.

packages/cardpay-sdk/sdk/reward-manager/base.ts
```diff
@@ -32,13 +37,20 @@ import { ZERO_ADDRESS } from '../constants';
...
 import { query } from '../utils/graphql';
 import { Signer } from 'ethers';
+import { get } from 'lodash';
 
 export interface RewardProgramInfo {
   rewardProgramId: string;
   rewardProgramAdmin: string;
   locked: boolean;
-  rule: string;
+  blob: string;
```
oidoug, line 46: Just out of curiosity, what does this blob attribute will hold?

oidoug (commented):
It looks good but it's hard for me to say if it's correct or not, I'm not familiarized enough.

packages/cardpay-sdk/sdk/reward-manager/base.ts
tintinthong: It is hex encoded version of the did string. For example, the blob for 0xab20c80fcc025451a3fc73bB953aaE1b9f640949 equals 0x6469643a63617264737461636b3a317236705438593943416d4a5a7066624858394d5362577763393639313538623466306334323461

packages/cardpay-cli/rewards/claimable-proofs.ts
tintinthong: I can remove the asterisk anyway. I renamed displayProof to logProof

packages/cardpay-reward-programs/scripts/read_parquet.py
```diff
@@ -0,0 +1,15 @@
+import pandas as pd
+import typer
+from cloudpathlib import AnyPath
+
+
+def read_parquet(reward_program_id: str, payment_cycle: int):
```
IanCal, line 6: Would be nice to default to staging but allow either staging or prod. Not a blocker.

packages/cardpay-sdk/sdk/reward-manager/base.ts
```diff
@@ -1055,9 +1070,28 @@ The owner of reward safe ${safeAddress} is ${rewardSafeOwner}, but the signer is
...
+    const content = await resolveDoc(did);
+    return content;
+  }
+
+  getProgramExplainer(rule: RuleJson) {
+    return get(rule, 'explanation.en.program', undefined);
+  }
+
+  getClaimExplainer(rule: RuleJson, explanationId: string) {
+    return get(rule, 'explanation.en.claim.' + explanationId, undefined);
```
IanCal, line 1092: We can default to en, but if we add this as at least an optional argument we don't have to modify the sdk in the future to support multiple locales.

IanCal (approved):
LGTM, nothing major but a couple of small changes noted I think would help over time.

IanCal (changes requested):
(ignore)

IanCal (approved):

PR #3306, opened 2022-10-11
Cs-4683 api Expose rule query

tintinthong:


tintinthong:
I realise this pr is not needed

tintinthong:
I realise this pr is not needed. We can just use the resolver for this to fetch s3. No access needed

PR #3307, opened 2022-10-11
hub: Scheduled payment executor

jurgenwerk:
This PR adds a service that will receive a scheduled payment, try to execute it, and track its blockchain transaction status in the db by creating scheduled payment attempts.

ℹ️ The service that will figure out which payment to pass to the executor, and at which time, will be added in a separate PR.

The execution involves:

1. Creating a scheduled payment attempt
2. Submitting the transaction to the blockchain
3. Waiting for the transaction to be mined in an async task
4. Updating the attempt with status, which starts with "in progress", and ends with "failed" or "succeeded"

I tested this using a combination of the CLI command for creating a scheduled payment, and the hub's console, like so:

### Scheduling a payment
CLI (scheduling a one-time payment from my safe on Sokol to my EOA):

```
yarn cardpay scheduled-payment schedule-payment 0x15356cc1C1DeCeDdcdE588937bB245C4f9D64F36 0x93E408e119FA77fF5C246FCFb9B55582897dd698 0x8F4fdA26e5039eb0bf5dA90c3531AeB91256b56b 10000000000000000000 0xeBCC5516d44FFf5E9aBa2AcaeB65BbB49bC3EBe1 0 0 120810 22000000000 0x8F4fdA26e5039eb0bf5dA90c3531AeB91256b56b salt_123 1665396689 "" "" --network sokol --mnemonic '...'
```

### Executing
Hub (running in `yarn console`):

```
scheduledPaymentId = "7c5da8d4-cf52-40d4-89e1-31f8b138d2e2" // output of the CLI command above
prismaManager = await container.lookup('prisma-manager');
prisma = await this.prismaManager.getClient();
scheduledPayment = await prisma.scheduledPayment.findFirst(scheduledPaymentId);

executor = await container.lookup('scheduled-payment-executor');
await executor.executePayment(scheduledPayment);
``` 

The result of the above is a successful scheduled payment attempt ([blockscount transation](https://blockscout.com/poa/sokol/tx/0xcd4db40cf19c901d2153799fad0331f539aeba5a348cf51d9f48fbfafebcd7a5)).

### Next steps
For our system to automatically execute payments when they are due, we will likely need a task that will be spawned periodically, can be every 5 minutes or so (can be done with cron), fetch the payments using the ScheduledPaymentFetcher service (there is already a method that will return payments that are due to be paid now) and pass them to the ScheduledPaymentExecutor. This is filed under a separate ticket.

### To discuss
Since the scheduled payment module contract only allows the crank to execute the payment, we need to sign the transaction using the crank's private key. For that purpose I added the `hubPrivateKey` environment variable.

❓ **Which key should we use for staging and production and who should be responsible for generating and keeping these keys?** After we have the crank key defined, we should add them to the [config contract](https://github.com/cardstack/cardstack-module-scheduled-payment/blob/main/contracts/ScheduledPaymentModule.sol#L59)


packages/hub/services/scheduled-payments/executor.ts
```diff
@@ -0,0 +1,157 @@
+import { inject } from '@cardstack/di';
+import { ScheduledPayment } from '@prisma/client';
+import { isBefore, subDays } from 'date-fns';
+import shortUUID from 'short-uuid';
+import { nowUtc } from '../../utils/dates';
+import config from 'config';
+import { Wallet } from 'ethers';
+import { JsonRpcProvider } from '@cardstack/cardpay-sdk';
+
+export let supportedChains = [
```
jurgenwerk, line 10: Should this go to the SDK?

packages/hub/services/scheduled-payments/executor.ts
```diff
@@ -0,0 +1,157 @@
...
+  {
+    id: 1,
+    name: 'ethereum',
+    rpcUrl: 'https://eth-mainnet.public.blastapi.io',
```
jurgenwerk, line 14: I got these addresses from https://chainlist.org.

packages/hub/services/scheduled-payments/executor.ts
```diff
@@ -0,0 +1,157 @@
...
+  prismaManager = inject('prisma-manager', { as: 'prismaManager' });
+  scheduledPaymentFetcher = inject('scheduled-payment-fetcher', { as: 'scheduledPaymentFetcher' });
+  cardpay = inject('cardpay');
+  web3 = inject('web3-http', { as: 'web3' });
+  workerClient = inject('worker-client', { as: 'workerClient' });
+
+  async executePayment(scheduledPayment: ScheduledPayment) {
+    let rpcUrl = supportedChains.find((chain) => chain.id === scheduledPayment.chainId)?.rpcUrl;
+    let provider = new JsonRpcProvider(rpcUrl, scheduledPayment.chainId);
+    let signer = new Wallet(config.get('hubPrivateKey'));
```
jurgenwerk, line 53: This is the crank's private key. We'll have to generate the private key for staging and production and adjust the infra, and configure the crank's address in the config contract. Is there an existing private key we can use?

packages/hub/services/scheduled-payments/executor.ts
```diff
@@ -0,0 +1,157 @@
...
+          amount.toString(),
+          payeeAddress,
+          Number(feeFixedUsd),
+          Number(feePercentage),
+          Number(executionGasEstimation),
+          String(maxGasPrice),
+          gasTokenAddress,
+          salt,
+          payAt!.getTime() / 1000,
+          String(maxGasPrice), // TODO: this should be the real gas price in gas token - we need to convert that to native token
```
jurgenwerk, line 122: Noted in the PR's description - we'll have a sdk method for this, and it should be used here instead of `maxGasPrice`

packages/hub/services/scheduled-payments/fetcher.ts
```diff
@@ -5,7 +5,7 @@ import { convertDateToUTC } from '../../utils/dates';
 
 export default class ScheduledPaymentsFetcherService {
   prismaManager = inject('prisma-manager', { as: 'prismaManager' });
-  validForDays = 10; // TODO: Keep this in sync with the value in the config contract
+  validForDays = 3;
```
jurgenwerk, line 8: We can read this from the config contract. But if we don't anticipate this will change we can also leave it hardcoded for now.

packages/hub/tasks/scheduled-payment-on-chain-execution-waiter.ts
```diff
@@ -0,0 +1,85 @@
...
+      await scheduledPaymentModule.executeScheduledPayment(paymentAttempt.transactionHash); // Will wait until mined. If already mined, will return immediately
+      await prisma.scheduledPaymentAttempt.update({
+        where: { id: paymentAttempt.id },
+        data: { status: 'succeeded', endedAt: nowUtc() },
+      });
+    } catch (error: any) {
+      // waitUntilTransactionMined will return "Transaction took too long to complete" in case it wasn't mined in 60 minutes.
+      // In this case we want to restart the task and wait some more. We could throw an error here so that the worker would restart the task, but
+      // due to the exponential-backoff it could take some time before the waiting can start again. That's why we just spawn a new task directly.
+      if (error.message.includes('took too long')) {
```
jurgenwerk, line 35: This check can fall apart in case someone changes the error copy in the SDK 😬 I guess a more proper solution would be to wrap the error in the SDK in a new error type, e.g `TimeoutError`

packages/hub/services/scheduled-payments/executor.ts
jurgenwerk: Oh, I see we already have network related configs in https://github.com/cardstack/cardstack/blob/main/packages/cardpay-sdk/sdk/constants.ts

I'll add the rpc urls there.

packages/hub/node-tests/services/scheduled-payments/executor-test.ts
```diff
@@ -0,0 +1,136 @@
+// import shortUUID from 'short-uuid';
+import { registry, setupHub } from '../../helpers/server';
+// import crypto from 'crypto';
```
backspace, line 3: Might as well remove this and the above comments?

backspace (approved):

packages/hub/node-tests/services/scheduled-payments/executor-test.ts
jurgenwerk: Ah sure, forgot to remove these. Will remove

jurgenwerk:
@habdelra @FadhlanR @alex-cardstack this is now ready for review.

packages/cardpay-sdk/sdk/utils/general-utils.ts
```diff
@@ -327,18 +327,24 @@ export function isTransactionHash(candidate: string): boolean {
...
+        },
+        (err, hash) => {
+          if (err) {
+            reject(err);
+          } else {
+            resolve(hash);
+          }
+        }
+      )
+      .catch(reject);
```
habdelra, line 346: The sematics of this are different. now. previously this would return as soon as a txn hash was available, now i think it will wait for a consensus before it returns. I don't think we want that. rather we should only wait until a txnHash is available. (when the `transactionHash` event is emitted). check out the docs for more info: https://web3js.readthedocs.io/en/v1.8.0/web3-eth.html#eth-sendtransaction-return

FadhlanR: I reverted these changes because we migrated from `web3` to `ethers.js` for the scheduled payment module.

packages/hub/tasks/scheduled-payment-on-chain-creation-waiter.ts
```diff
@@ -1,20 +1,23 @@
...
     let prisma = await this.prismaManager.getClient();
-
-    let scheduledPaymentModule = await this.cardpay.getSDK('ScheduledPaymentModule', this.web3.getInstance());
     let scheduledPayment = await prisma.scheduledPayment.findFirstOrThrow({
       where: { id: payload.scheduledPaymentId },
     });
 
+    let rpcUrl = supportedChains.find((chain) => chain.id === scheduledPayment.chainId)?.rpcUrl as string;
+    let provider = new JsonRpcProvider(rpcUrl, scheduledPayment.chainId);
+    let scheduledPaymentModule = await this.cardpay.getSDK('ScheduledPaymentModule', provider);
```
FadhlanR, line 19: I also implemented JsonRPCProvider here instead of web3.

packages/hub/tasks/scheduled-payment-on-chain-execution-waiter.ts
```diff
@@ -20,7 +21,12 @@ export default class ScheduledPaymentOnChainExecutionWaiter {
       return; // Such attempt should not be waited for
     }
 
-    let scheduledPaymentModule = await this.cardpay.getSDK('ScheduledPaymentModule', this.web3.getInstance());
+    let scheduledPayment = await prisma.scheduledPayment.findFirstOrThrow({
+      where: { id: paymentAttempt.scheduledPaymentId },
+    });
+    let rpcUrl = supportedChains.find((chain) => chain.id === scheduledPayment.chainId)?.rpcUrl as string;
+    let provider = new JsonRpcProvider(rpcUrl, scheduledPayment.chainId);
+    let scheduledPaymentModule = await this.cardpay.getSDK('ScheduledPaymentModule', provider);
```
FadhlanR, line 29: Also here.

habdelra (approved):
thx @FadhlanR looks good!

PR #3310, opened 2022-10-11
ci(safe-tools-client): Add github actions to deploy safe-tools-client

pcjun97:
Similar to deployment of `web-client`.
Uses native aws commands `s3 sync` and `cloudfront create-invalidation` instead of adding them to ember's `deploy.js`.


.github/actions/deploy-safe-tools-client/action.yml
```diff
@@ -0,0 +1,65 @@
...
+      working-directory: packages/safe-tools-client
+      run: yarn build:safe-tools-client --environment ${{ inputs.environment }}
+
+    - name: Configure AWS credentials
+      uses: aws-actions/configure-aws-credentials@v1
+      with:
+        role-to-assume: ${{ env.AWS_ROLE_ARN }}
+        aws-region: us-east-1
+
+    - name: Sync content to S3
```
lukemelia, line 49: I would prefer to use ember-cli-deploy here. It is standard in the ember world and allows developers to add additional plugins to adjust the deploy process. Happy to help with this if you want.

lukemelia (changes requested):

.github/actions/deploy-safe-tools-client/action.yml
pcjun97: Added the ember deploy plugins and updated the action and script.
I am unfamiliar with the tools though, so definitely appreciate the help to review them thoroughly!

.github/workflows/push-main.yml
```diff
@@ -450,3 +456,27 @@ jobs:
...
+        with:
+          path: |
+            node_modules
+            packages/*/node_modules
+            packages/firebase-functions/*/functions/node_modules
+          key: ${{ runner.os }}-node-modules-${{ hashFiles('yarn.lock') }}
+          restore-keys: |
+            ${{ runner.os }}-node-modules-
+
+      - name: Deploy web-client
```
lukemelia, line 478:
```suggestion
      - name: Deploy safe-tools-client
```

packages/safe-tools-client/config/deploy.js
```diff
@@ -0,0 +1,26 @@
...
+    },
+    cloudfront: {
+      objectPaths: ['/*'],
+    },
+  };
+
+  if (deployTarget === 'staging') {
+    ENV.build.environment = 'production';
+    ENV.s3.bucket = 'cardstack-safe-tools-client-staging';
+    ENV.cloudfront.distribution = 'PLACEHOLDER';
```
lukemelia, line 16: Do we know our distribution IDs and/or we make them available as environment variables to Github Actions?

lukemelia (changes requested):

packages/safe-tools-client/config/deploy.js
backspace: They’re [hardcoded](https://github.com/cardstack/cardstack/blob/main/packages/web-client/config/deploy.js#L111) for web-client (and ssr-web [previews](https://github.com/cardstack/cardstack/blob/main/packages/ssr-web/config/deploy.js#L77))

lukemelia (approved):
Looks good! 🚀

PR #3311, opened 2022-10-11
[CS-4724]: Add ScheduleCollpsePanel to safe-tools schedule tab

danibonilha:
I chose flexbox layout, but I'm open to suggestions if grid, or any other display is preferable.

@Aierie and I found a contrast issue with the details tag on axe-core, the contrast is great, but axe-core is unable to "see" it and it throws an error, so we had disable the accessibility check for collapse-panel, we thought about updating the lib, but then we would need to fix all the other issues, which we don't have bandwidth too rn.

image

image


backspace (approved):
This looks great. We might end up wanting to use CSS Grid but we can change to that if it does become necessary

PR #3312, opened 2022-10-11
Use parquet metadata file for reward rules

IanCal:
This removes the partitioning calculation code from the reward rules, and swaps to duckdb registration. You can "register" any dataset or dataframe as a table, then query it like it's just present. This means no string interpolation in the queries, we can depend on any number of tables, and importantly for testing we only need to override this registration.

This also removes the need to pull files locally, as pyarrow should read the _metadata file then inspect only the right partitions using filter pushdowns.

As this was a bit of refactoring, I've removed the rules we aren't actually using, and one of the tests doesn't really have the right hook any more.

#Cs-4153


.github/workflows/manual-cardpay-reward-program-rules.yml
```diff
@@ -13,10 +13,7 @@ on:
         required: true
         options:
           - FlatPayment
-          - MinSpend
           - RetroAirdrop
-          - WeightedUsage
-          - MinOtherMerchantsPaid
```
tintinthong, line 19: 👍

tintinthong (approved):
👏

PR #3315, opened 2022-10-12
Remove "Actions only visible to you" from ActionChin

lukemelia:


packages/boxel/addon/components/boxel/action-chin/usage.gts
```diff
@@ -235,6 +236,13 @@ export default class ActionChinUsage extends Component<Signature> {
         @value={{this.stepNumber}}
         @onInput={{fn (mut this.stepNumber)}}
       />
+      <Args.Bool
+        @name="hideLock"
+        @description="When true, suppress the 'Actions only visible to you' message and lock icon."
+        @defaultValue={{false}}
+        @value={{this.hideLock}}
+        @onInput={{fn (mut this.hideLock)}}
+      />
```
backspace, line 245: This needs to be consumed in the invocation 🤓

packages/boxel/addon/components/boxel/action-chin/index.gts
```diff
@@ -151,7 +153,7 @@ export default class ActionChin extends Component<Signature> {
       {{/if}}
       {{!-- in any state except: disabled + has a step, we want the lock shown --}}
       {{!-- template-lint-disable simple-unless --}}
-      {{#unless (and @disabled @stepNumber)}}
+      {{#unless (or (and @disabled @stepNumber) @hideLock)}}
```
backspace, line 156:
```suggestion
      {{#unless (or @hideLock (and @disabled @stepNumber))}}
```

This seems slightly easier to understand to me. Does the explanation two lines above need to be updated?

backspace (changes requested):
I feel like the lock should default to _not_ show as that seems more generally-applicable but that would be a far-reaching change.

packages/boxel/addon/components/boxel/action-chin/usage.gts
backspace: In the [preview](https://boxel.stack.cards/boxel-tweaks/#/docs?s=Components&ss=%3CBoxel%3A%3AActionChin%3E) checking the box does nothing:

Boxel Components 2022-10-12 14-54-54

lukemelia:
Revamped PR to simply remove the text and icon

backspace (approved):

PR #3316, opened 2022-10-12
Add @value arg to ToggleButtonGroup to facilitate data-down/actions-up

lukemelia:


backspace (approved):
Thanks for addressing this.

PR #3318, opened 2022-10-12
Boxel::ActionContainer now yields a Section component

lukemelia:
- Section provides standard padding
- Each Section takes an optional @title
- PR removes @prompt since it is no longer necessary

### DIFF
--- a/packages/boxel/addon/components/boxel/action-container/section/title/index.gts
+++ b/packages/boxel/addon/components/boxel/action-container/section/title/index.gts
@@ -1,0 +1,44 @@
+import type { TemplateOnlyComponent } from '@ember/component/template-only';
+import cn from '@cardstack/boxel/helpers/cn';
+import cssUrl from "@cardstack/boxel/helpers/css-url";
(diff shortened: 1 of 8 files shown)


packages/boxel/addon/components/boxel/action-container/section/title/index.gts
```diff
@@ -0,0 +1,44 @@
...
+import './index.css';
+
+
+export interface Signature {
+  Element: HTMLElement;
+  Args: {
+    dataTestId?: string;
+    title?: string;
+    icon?: string;
+    imgUrl?: string;
```
backspace, line 17: Should there be a usage example that shows `icon` and `imgUrl`?

backspace (approved):

packages/boxel/addon/components/boxel/action-container/section/title/index.gts
lukemelia: Good idea. Added.

PR #3319, opened 2022-10-13
sdk: calculate current gas price in provided token

jurgenwerk:
Ticket: CS-4702

This will be used in the scheduled payment executor to calculate current gas price in the specified gas token before executing the payment. The scheduled payment module contract will revert if this price exceeds the max gas price provided when creating the scheduled payment. 

The standard value from the gas station is used to get the current gas price in native token (wrapped ETH in this case). Then, the Exchange contract (which comes with the scheduled payment module) is queried which returns the exchange rates in USD. 

I tested this in Görli network for the DAI gas token (`0x026D3013A36A248f6624abC9FEfb76723Bd87561`) and got the following result: `36316263014200`. This means the current gas price in Görli was ~0.000036 DAI per gas unit.

### DIFF
--- a/packages/cardpay-sdk/contracts/addresses.ts
+++ b/packages/cardpay-sdk/contracts/addresses.ts
@@ -100,6 +100,7 @@ const GOERLI = {
   multiSendCallOnly: '0x40A2aCCbd92BCA938b02010E17A5b8929b49130D',
   moduleProxyFactory: '0x00000000000DC7F163742Eb4aBEf650037b1f588',
   metaGuard: '0xe2847462a574bfd43014d1c7BB6De5769C294691',
+  wrappedNativeToken: '0xB4FBF271143F4FBf7B91A5ded31805e42b2208d6', // WETH
 };
 const MUMBAI = {
   gnosisSafeMasterCopy: '0xd9Db270c1B5E3Bd161E8c8503c55cEABeE709552',
@@ -112,6 +113,7 @@ const MUMBAI = {
   multiSendCallOnly: '0x40A2aCCbd92BCA938b02010E17A5b8929b49130D',
   moduleProxyFactory: '0x00000000000DC7F163742Eb4aBEf650037b1f588',
   metaGuard: '0xe2847462a574bfd43014d1c7BB6De5769C294691',
+  wrappedNativeToken: '0x9c3C9283D3e44854697Cd22D3Faa240Cfb032889', // WMATIC
 };
 
 const MAINNET = {
@@ -123,6 +125,7 @@ const MAINNET = {
   scheduledPaymentConfig: '',
   scheduledPaymentExchange: '',
   scheduledPaymentModule: '',
+  wrappedNativeToken: '0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2', // WETH
 };
 const GNOSIS = {
   gnosisProxyFactory_v1_2: '0x76E2cFc1F5Fa8F6a5b3fC4c8F4788F0116861F9B',
@@ -181,7 +184,13 @@ const addresses: {
   xdai: GNOSIS,
 });
 
-export type AddressKeys = keyof typeof SOKOL | keyof typeof KOVAN | keyof typeof MAINNET | keyof typeof GNOSIS;
+export type AddressKeys =
+  | keyof typeof SOKOL
+  | keyof typeof KOVAN
+  | keyof typeof MAINNET
+  | keyof typeof GNOSIS
+  | keyof typeof GOERLI
+  | keyof typeof MUMBAI;
 
 export default addresses;
 
--- a/packages/cardpay-sdk/sdk/constants.ts
+++ b/packages/cardpay-sdk/sdk/constants.ts
@@ -82,6 +82,7 @@ const MAINNET = {
   name: 'Ethereum Mainnet',
   // check https://docs.tokenbridge.net/eth-xdai-amb-bridge/about-the-eth-xdai-amb for the finalization rate
   ambFinalizationRate: '20',
+  relayServiceURL: 'https://relay-ethereum.cardstack.com/api',
 };
 const GNOSIS = {
   apiBaseUrl: 'https://blockscout.com/xdai/mainnet/api',
--- a/packages/cardpay-sdk/index.ts
+++ b/packages/cardpay-sdk/index.ts
@@ -59,3 +59,4 @@ export { default as JsonRpcProvider } from './providers/json-rpc-provider';
 export { default as Web3Provider } from './providers/web3-provider';
 export { MIN_PAYMENT_AMOUNT_IN_SPEND as MIN_PAYMENT_AMOUNT_IN_SPEND__PREFER_ON_CHAIN_WHEN_POSSIBLE } from './sdk/do-not-use-on-chain-constants';
 export { protocolVersions } from './contracts/addresses';
+export { gasPriceInToken } from './sdk/utils/conversions';
--- a/packages/cardpay-sdk/package.json
+++ b/packages/cardpay-sdk/package.json
@@ -13,6 +13,8 @@
     }
   },
   "dependencies": {
+    "@ethersproject/contracts": "^5.7.0",
+    "@ethersproject/providers": "^5.7.1",
     "@ethersproject/solidity": "^5.7.0",
     "@truffle/hdwallet-provider": "^1.5.0",
     "@trufflesuite/web3-provider-engine": "^15.0.13-1",
@@ -21,6 +23,7 @@
     "@types/lodash": "^4.14.168",
     "@types/node-fetch": "^2.5.10",
     "@types/url-parse": "^1.4.4",
+    "@uniswap/sdk": "^3.0.3",
     "bignumber.js": "^9.0.1",
     "bn.js": "^5.2.0",
     "eth-sig-util": "^3.0.1",
--- a/packages/cardpay-sdk/sdk/utils/conversions.ts
+++ b/packages/cardpay-sdk/sdk/utils/conversions.ts
@@ -1,0 +1,45 @@
+/*global fetch */
+
+import { Fetcher, Route, Price } from '@uniswap/sdk';
+import { getAddressByNetwork } from '../../contracts/addresses';
+import { getConstantByNetwork } from '../constants';
+import JsonRpcProvider from '../../providers/json-rpc-provider';
+import { networkName } from './general-utils';
+import BN from 'bn.js';
+import { BaseProvider } from '@ethersproject/providers';
+
+async function tokenPairRate(provider: JsonRpcProvider, token1Address: string, token2Address: string): Promise<Price> {
+  let network = await provider.getNetwork();
+  let token1 = await Fetcher.fetchTokenData(network.chainId, token1Address, provider as unknown as BaseProvider);
+  let token2 = await Fetcher.fetchTokenData(network.chainId, token2Address, provider as unknown as BaseProvider);
+
+  let pair = await Fetcher.fetchPairData(token1, token2, provider as unknown as BaseProvider);
+
+  let route = new Route([pair], token2);
+
+  return route.midPrice; // How many "token 1" we can get for one "token 2" in Uniswap
+}
+
+export async function gasPriceInToken(provider: JsonRpcProvider, tokenAddress: string): Promise<BN> {
+  let network = await networkName(provider);
+
+  // We use the wrapped native token address because the native token doesn't have an address in Uniswap.
+  // The price of the wrapped native token, such as WETH, is the same as the price of the native token.
+  let rate = await tokenPairRate(provider, tokenAddress, getAddressByNetwork('wrappedNativeToken', network));
+
+  // Gas station will return current gas price in native token in wei.
+  let gasStationResponse = await fetch(`${getConstantByNetwork('relayServiceURL', network)}/v1/gas-station/`, {
+    method: 'GET',
+    headers: {
+      'Content-Type': 'application/json',
+      Accept: 'application/json',
+    },
+  });
+
+  let gasPriceInNativeTokenInWei = new BN((await gasStationResponse.json()).standard);
+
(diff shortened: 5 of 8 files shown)


packages/cardpay-sdk/contracts/addresses.ts
```diff
@@ -100,6 +100,7 @@ const GOERLI = {
   multiSendCallOnly: '0x40A2aCCbd92BCA938b02010E17A5b8929b49130D',
   moduleProxyFactory: '0x00000000000DC7F163742Eb4aBEf650037b1f588',
   metaGuard: '0xe2847462a574bfd43014d1c7BB6De5769C294691',
+  wrappedNativeToken: '0xB4FBF271143F4FBf7B91A5ded31805e42b2208d6', // WETH
```
jurgenwerk, line 103: I'm not sure if this is the correct place for storing this address, but we need it to look up the usd exchange rate on uniswap for the gas price in native token.

packages/cardpay-sdk/sdk/constants.ts
```diff
@@ -56,6 +57,7 @@ const GOERLI = {
   nativeTokenName: 'Ethereum',
   name: 'Goerli',
   relayServiceURL: 'https://relay-goerli.staging.stack.cards/api',
+  rpcUrl: 'https://rpc.ankr.com/eth_goerli',
```
FadhlanR, line 60: can we get the RPC URL from hub config?

packages/cardpay-sdk/sdk/utils/general-utils.ts
```diff
@@ -381,3 +387,26 @@ export async function wait(ms = 1000) {
...
+  });
+
+  let gasPriceInNativeToken = (await gasStationResponse.json()).standard;
+
+  let exchangeAddress = getAddressByNetwork('scheduledPaymentExchange', network);
+  let wrappedNativeTokenAddress = getAddressByNetwork('wrappedNativeToken', network);
+  const provider = new JsonRpcProvider(getConstantByNetwork('rpcUrl', network));
+  let contract = new EthersContract(exchangeAddress, ExchangeABI, provider);
+
+  let gasTokenInUsd = (await contract.exchangeRateOf(gasTokenAddress)).value;
```
FadhlanR, line 408: If `gasTokenAddress == usdTokenInExchangeContract` we don't need to call the `exchangeRateFunction`, define `gasTokenInUsd = 1e18`

habdelra:
I'm very confused how wrapped ETH has anything to do with this, could you please explain your rationale more?

habdelra:
Also it is good to be precise in terms of what exactly the gas station is returning--it's not an absolute value, rather it is a rate. The gas station returns the price per unit of gas in gwei. so the amount you got back is the price for a single unit of gas.  so you got back 0.000036 DAI / unit of gas right?

packages/cardpay-sdk/contracts/addresses.ts
habdelra: so the relay server does the the price oracle conversion work internally in order to calculate the amount of token to reimburse the relay's EOA. it's a little more involved than just talking to the uniswap contract--there is a bit of math to do on top of that. We should take a look at how the relay server does this and do the javascript equivalent.

jurgenwerk: @habdelra my rationale for using the WETH address is that I wanted to use the `exchangeRateOf(tokenAddress)` to get the native coin usd rate, but for that I needed the token address. But since native coin doesn't have an address I chose WETH as an alternative. But I'm really not sure if this makes sense. Will take a look at how the relayer is doing conversion, thanks!

packages/cardpay-sdk/sdk/constants.ts
jurgenwerk: Is there a reason why that would be better? I'd prefer if the sdk wouldn't rely on the hub to get this config

packages/cardpay-sdk/contracts/addresses.ts
jurgenwerk: It's coming from this contract: https://github.com/cardstack/cardstack-module-scheduled-payment/blob/main/contracts/Exchange.sol

I saw @FadhlanR added it and I thought I could use it for this conversion

jurgenwerk:
@habdelra right, I meant price per gas unit.

packages/cardpay-sdk/contracts/addresses.ts
jurgenwerk: Do you think we should port https://github.com/5afe/safe-relay-service/blob/master/safe_relay_service/tokens/models.py#L121?

habdelra: hmmm, i think there is more work to do then to make this work with a Uniswap contracts (perhaps that is what @FadhlanR is working on now). The `Exchange` contract had a bunch of cardpay protocol baggage though. it might be easier to just talk directly to uniswap contracts instead of having to go thru the `Exchange` contract.

habdelra: > Do you think we should port https://github.com/5afe/safe-relay-service/blob/master/safe_relay_service/tokens/models.py#L121?

No, i think _this_ is what you need to port for the uniswap oracle to work: https://github.com/safe-global/safe-eth-py/blob/master/gnosis/eth/oracles/oracles.py#L77-L467

packages/cardpay-sdk/sdk/utils/general-utils.ts
```diff
@@ -381,3 +387,26 @@ export async function wait(ms = 1000) {
...
+  let wrappedNativeTokenInUsd = (await contract.exchangeRateOf(wrappedNativeTokenAddress)).value;
+
+  return wrappedNativeTokenInUsd.div(gasTokenInUsd).mul(gasPriceInNativeToken);
```
habdelra, line 411: keep in mind these are _really_ large numbers we are dealing with, because they are in `wei`. in the past I have had issues with order of operations concerning dividing very large numbers that result in loss of precision. It might be better to first perform your multiplication before the divide so that you are dividing a very large number by another very large number instead of dividing a normal size number by a very large number.

packages/cardpay-sdk/sdk/utils/general-utils.ts
```diff
@@ -381,3 +387,26 @@ export async function wait(ms = 1000) {
...
+export async function currentGasPrice(network: string, gasTokenAddress: string) {
+  let gasStationResponse = await fetch(`${getConstantByNetwork('relayServiceURL', network)}/v1/gas-station/`, {
+    method: 'GET',
+    headers: {
+      'Content-Type': 'application/json',
+      Accept: 'application/json',
+    },
+  });
+
+  let gasPriceInNativeToken = (await gasStationResponse.json()).standard;
```
habdelra, line 401: these are in units of `gwei` not `gwei`. I'm not seeing where you are doing a `gwei` to `wei` conversion (that will result in a number that is off by 9 decimal points!)

packages/cardpay-sdk/sdk/utils/general-utils.ts
habdelra: we should be using BigNumber (or BigInteger) though right? don't use a native javascript number for any of this math.

packages/cardpay-sdk/contracts/addresses.ts
FadhlanR: @habdelra From your explanation, I think you mentioned the card pay exchange contract. But what @jurgenwerk  referring to here is the exchange contract that I've made for the scheduled payment module and already used Uniswap V3 as source data to get the USD rate of the token.

I think the Idea in this PR is to use that Exchange contract to get the conversion rate between two tokens (in this case `native token to X token`). Because the `exchangeRateOf(tokenAddress)` function only provides conversion to USD, we need to have conversion rates from `native token to USD` and `X token to USD`. With that conversion rates, we can get a conversion rate from `native token to X token = native token to USD / X token to USD`. 

And we need the `wrappedNativeTokenAddress` here because the `exchangeRateOf` function receives the token address as a parameter and inside that function, there's a process to get the `uniswap pool address` based on the defined USD token address and token address.

packages/cardpay-sdk/sdk/utils/general-utils.ts
FadhlanR: right

packages/cardpay-sdk/sdk/constants.ts
FadhlanR: I think we can follow the same approach that is used by other SDK functions, instead of initializing the provider in the SDK (in this case in your `currentGasPrice` function) we can let the client initialize the provider and pass it as a parameter to `currentGasPrice(provider: JsonRPCProvider, network: string, gasTokenAddress: string)`. In this way, the SDK doesn't need to care about the RPC url.

jurgenwerk: That makes sense. Thanks, will adjust this

packages/cardpay-sdk/sdk/utils/general-utils.ts
jurgenwerk: No, I think this is in `wei`. If you check the response from https://relay-goerli.staging.stack.cards/api/v1/gas-station/, you can see the current value for standard is `2896112618`. If that was in gwei, then the gas price would be 2+ ethers which is improbable. Will document this and change naming so it is clear on which units we are operating

packages/cardpay-sdk/sdk/constants.ts
jurgenwerk: Fixed

jurgenwerk:
I also just occurred to me that this implementation is flawed where we calculate the ratio between the 2 tokens because tokens can have different amount of decimals. For example, WETH has 18 decimals, USDT and USDC have 6.

packages/cardpay-sdk/contracts/addresses.ts
jurgenwerk: Agree with @FadhlanR that this were my thoughts behind this implementation. 

I'm wondering if we can use the uniswap sdk and fetch the price of gas token in weth like here: https://docs.uniswap.org/sdk/2.0.0/guides/pricing

This way we can avoid converting both tokens to usd rate first, and get a better handle on the token specifics such as number of decimals.

habdelra: thanks for the clarification @FadhlanR :+1: 

@jurgenwerk yeah, that's exactly what i was getting at--if we go to purely uniswap based token prices, is there even a reason to have any extra onchain logic like the exchange contract. and that's awesome that there is actually already an SDK for this so no need to reverse engineer what gnosis did.

habdelra:
good point! yeah, for the cardpay protocol I normalized all token prices to 8 decimals in the oracle adapter contracts so that i didn't have to deal with going between different decimals. I think moving to the uniswap SDK like you mentioned, might make that easier too (also it looks like you can request the amount of significant digits in the price)

packages/cardpay-sdk/contracts/addresses.ts
jurgenwerk: @habdelra uniswap has 2 SDK versions - one is v2 and the other v3 (https://github.com/Uniswap/v2-sdk/, https://github.com/Uniswap/v3-sdk/). @FadhlanR's Exhange contract mentioned above is using v3, but the v3 sdk says it's in alpha (check the v3 repo description) and that it might change significantly between patch versions. 

Do you think it's okay to use the v2 sdk for now, or should we keep versions consistent between the contract and the sdk? 

Just for demonstration purposes, here's how this looks using the v2 uniswap sdk:

```
export async function currentGasPriceV2(provider: JsonRpcProvider, gasTokenAddress: string) {
  let network = await provider.getNetwork();
  let gasToken = await Fetcher.fetchTokenData(network.chainId, gasTokenAddress, provider);
  let wrappedNativeToken = await Fetcher.fetchTokenData(
    network.chainId,
    getAddressByNetwork('wrappedNativeToken', network.name),
    provider
  );

  let pair = await Fetcher.fetchPairData(gasToken, wrappedNativeToken, provider);

  let route = new Route([pair], wrappedNativeToken);

  return route.midPrice.toSignificant(6);
}
```

habdelra: uniswap V2 is like the gold standard for liquidity pools. I suggest using that (that's also what the relay server uses if that helps)

jurgenwerk:
I completely rewrote the implementation–I removed querying our Exchange contract and used Uniswap SDK v2 instead. 

I tested this against göerli and mainnet. 

My questions during testing were:

### Is this conversion correct?

I made a practical test on mainnet using DAI (`0x6B175474E89094C44Da98b954EedeAC495271d0F`) as gas token (stablecoin pegged to the dollar). 


```
// network mainnet, current standard gas price in wei: 11496872890
let gasPriceInNativeToken = 11496872890
let daiTokenAddress = "0x6B175474E89094C44Da98b954EedeAC495271d0F"
let gasPriceInProvidedToken = await gasPriceInToken(provider, daiTokenAddress)
=> 15055934235899

gasPriceInProvidedToken / gasPriceInNativeToken 
=> 1.309,56 // at this time this is the dollar price for 1 eth, so this confirms the rate we calculated 
```

### Is this conversion respecting the different amount of decimals of tokens in comparison?

DAI token has 18 decimals, USDC has 6 decimals. Given they are both stablecoins the gas cost should be approximately the same.

```
let daiTokenAddress = "0x6B175474E89094C44Da98b954EedeAC495271d0F"
let usdcTokenAddress = "0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48"

await gasPriceInToken(provider, daiTokenAddress)
=> 14832540213353

await gasPriceInToken(provider, usdcTokenAddress)
=> 14831441860304

// gas prices are very similar for both stablecoins, so this confirms the calculation respects decimal differences

```

habdelra (approved):
This looks really amazing! the implementation is so much smaller than what you had before. thanks for the conversation rate walk thru in the comments too!

PR #3320, opened 2022-10-14
[CS-4657]: Create SetupSafeModal UI

danibonilha:
This PR adds the SetupSafeModal to the schedule tab, in order to apply the correct designs, I've tweaked some things on the ActionContainer, so it yields the ActionChin, giving it more flexibility to use the component instead of relying on the complete/incomplete states, also add color customization to the check-circle icon.
I've also split the ticket into logic and UI, to unblock the boxel changes so others can use,  this is UI only.
Some values are mocked, and some behaviors are still TBD, I've added some TODO on the code, to keep track of that.

https://user-images.githubusercontent.com/20520102/195949092-0156b5d8-962e-4c77-81b5-d854e9bc24d0.mov


packages/safe-tools-client/app/components/setup-safe-modal/index.css
```diff
@@ -0,0 +1,34 @@
+.setup-safe-modal__section {
+  padding: 50px 60px 50px 50px;
```
Aierie, line 2: (NO CHANGE REQUIRED) This implementation (which I think is correct) brings up an interesting point re: Boxel's spacing units that I've been thinking about for a while. Not for action in this PR, more generally about the length primitives we're using and how they should interact with user font size preferences.

We use `rem`-based spacing with `--boxel-sp-` variables since this helps things to scale evenly with user preferences for font size, but this is [not always desirable](https://medium.com/@sascha.wolff/dont-use-rem-em-for-paddings-margins-and-more-94e19026b000). I think it's relevant when the spacing communicates text/other grouping, but for things like padding it just reduces the available amount of space for content (that is already taking up more space because of increased font size).

packages/safe-tools-client/app/components/setup-safe-modal/index.gts
```diff
@@ -0,0 +1,80 @@
...
+          {{/if}}
+          {{! TODO: What's the behavior for not having enough gas ? }}
+        </Section>
+        <ActionChin @state={{'default'}}>
+          <:default as |a|>
+            {{! TODO: Provisioning should be handled on this component or be a param to be handled on a diff controller ? }}
+            <a.ActionButton @loading={{@loading}}>
+              Provision
+            </a.ActionButton>
+            <a.CancelButton {{on 'click' @onClose}} @disabled={{@loading}}>
```
Aierie, line 70: Why disable this over removing it from the UI when loading?

packages/safe-tools-client/app/components/setup-safe-modal/index.gts
```diff
@@ -0,0 +1,80 @@
...
+          <p>In this step, you create a safe equipped with a module to schedule
+            payments.
+            <br />What you need to know:
+          </p>
+          <ul>
+            <li>Your payment safe is used to fund your scheduled payments. </li>
+            <li>You are the owner of the safe. Only the safe owner can schedule
+              payments.
+            </li>
+            <li>Once you have scheduled payments, the module (source code here)
```
Aierie, line 42: Is `module (source code here)` intended to be a link to some open-sourced code?

packages/boxel/addon/components/boxel/action-container/index.gts
```diff
@@ -46,24 +47,27 @@ const ActionContainer: TemplateOnlyComponent<Signature> = <template>
...
+            {{@completeActionLabel}}
+          </m.ActionButton>
+        </:memorialized>
+        <:default as |a|>
+          <a.ActionButton @disabled={{@disabled}} {{on "click" (optional @onClickButton)}}>
+            {{@incompleteActionLabel}}
+          </a.ActionButton>
+        </:default>
+      </BoxelActionChin>
+    {{/if}}
```
Aierie, line 69: I wonder if we can remove this part and the associated arguments altogether. The `ActionContainer` component has no current consumers that use it, and leaving this in runs a risk of having two `ActionChin`s.

packages/boxel/public/images/icons/icon-check-circle-ht.svg
```diff
@@ -1 +1 @@
-<svg height="16" viewBox="0 0 16 16" width="16" xmlns="http://www.w3.org/2000/svg"><circle cx="8" cy="8" fill="#00ebe5" r="8"/><path d="m11.108 6-4.887 4.887-2.221-2.221" fill="none" stroke="#000" stroke-linecap="round" stroke-linejoin="round" stroke-width="2" transform="translate(.313)"/></svg>
\ No newline at end of file
+<svg height="16" viewBox="0 0 16 16" width="16" xmlns="http://www.w3.org/2000/svg"><circle cx="8" cy="8" fill="var(--icon-color, var(--boxel-cyan))" r="8"/><path d="m11.108 6-4.887 4.887-2.221-2.221" fill="none" stroke="var(--icon-stroke, var(--boxel-dark))" stroke-linecap="round" stroke-linejoin="round" stroke-width="2" transform="translate(.313)"/></svg>
```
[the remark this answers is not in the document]
danibonilha: oh ok, I guess the one .svg I've used as an example had boxel vars as default, but I'll revert it

packages/safe-tools-client/app/components/setup-safe-modal/index.gts
danibonilha: I guess i didn't give much thought to it  😅, hiding it, seems more user friendly.

packages/boxel/addon/components/boxel/action-container/index.gts
danibonilha: I tend to agree with you, I'll check if the team agrees with us.

lukemelia: +1 rip it out!

backspace (changes requested):
There’s some alignment confusion when looking at the designs but I think for consistency we can left-align this button with the text paragraph, while it looks right-aligned in the design I _believe_ it’s meant to align with the “SAFE” header. But because the designs show various alignments, I just chose to left-align everything with the header icon.

Cardstack Safe Tools 2022-10-17 10-35-42

packages/safe-tools-client/app/controllers/schedule.ts
```diff
@@ -0,0 +1,21 @@
...
+export default class Schedule extends Controller {
+  @tracked isSetupSafeModalOpen = false;
+
+  @action openSetupSafeModal() {
+    this.isSetupSafeModalOpen = true;
+  }
+
+  @action closeSetupSafeModal() {
+    this.isSetupSafeModalOpen = false;
+  }
```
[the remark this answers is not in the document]
danibonilha: I saw this set example and I found it very confusing to read, because the value is created on the controller, but then the template is the one "modifying", it is more concise, but it has mixed logic. If it's some type of pattern and ember required behavior I can switch to it.

danibonilha:
Left alignment update:

image

packages/safe-tools-client/app/controllers/schedule.ts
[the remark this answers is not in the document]
backspace: It’s a pattern but definitely not mandatory 🤓

backspace (approved):

PR #3322, opened 2022-10-17
[Chore]: Update github actions to node 16 migration

danibonilha:
Github actions only supports node: >16 now, and volta v1 was using node 14, this PR updates the volta actions to use the latest version alongside with checkout actions.

### DIFF
--- a/.github/actions/check-secrets/action.yml
+++ b/.github/actions/check-secrets/action.yml
@@ -15,7 +15,7 @@ runs:
       run: |
         sudo apt-get update && sudo apt-get install awscli jq --fix-missing
 
-    - uses: volta-cli/action@v1
+    - uses: volta-cli/action@v4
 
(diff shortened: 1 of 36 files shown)


habdelra:
thanks for this! we were having the same problem in the runtime project and I was super confused what was going on

backspace (approved):

backspace:
I think this still [uses Node 14](https://github.com/cardstack/cardstack/actions/runs/3236210091/jobs/5301700487#step:7:7) but that’s fine as that’s what we [tell Volta to use](https://github.com/cardstack/cardstack/blob/e45103f9d01e48930733883ac749284c0a863179/package.json#L116).

This is tangential to your updates but I wonder whether we might as well use Node 16, which has an LTS now. It would likely be faster for people who have Apple Silicon.

PR #3323, opened 2022-10-18
tools: Add connect modal

backspace:
This is UI-only, it doesn’t actually connect to wallets. It’s currently based on #3320 to make use of its `ActionContainer::Section`. It needs further cleanup as it was not an easy extraction from web client.


packages/safe-tools-client/app/helpers/truncate-middle.ts
```diff
@@ -0,0 +1,23 @@
...
+    return input;
+  } else if (input) {
+    return `${input.substring(0, startLength)}...${input.substring(
+      input.length - endLength
+    )}`;
+  }
+  return '';
+}
+
+export default helper<Signature>(truncateMiddle);
```
lukemelia, line 23: I think it is time to extract an ember-helpers addon package to let us share this rather than copy/pasting it

backspace: 👍🏻 makes sense, will do, thanks

packages/safe-tools-client/app/templates/application.hbs
```diff
@@ -18,9 +18,7 @@
         boxel-org-title-text-transform='none'
       }}
     >
-      <Boxel::Button @kind='primary'>
-        Connect Wallet
-      </Boxel::Button>
+      <ConnectButton @onConnect={{set this 'isShowingConnectModal' true}} />
```
Aierie, line 21: Does the control panel's wallet connection button also trigger opening of the `NetworkConnectModal`?

packages/safe-tools-client/package.json
```diff
@@ -89,6 +94,7 @@
     "eslint-plugin-qunit": "^7.3.1",
     "file-loader": "^6.2.0",
     "loader.js": "^4.7.0",
+    "macro-decorators": "^0.1.2",
```
Aierie, line 97: I don't think we need this dependency for this PR. Could we remove it?

packages/safe-tools-client/app/components/network-connect-modal/index.gts
```diff
@@ -0,0 +1,197 @@
...
+              <a.ActionButton {{on "click" this.connect}} data-test-mainnet-connect-button>
+                Connect Wallet
+              </a.ActionButton>
+            </:default>
+            <:inProgress as |i|>
+              <i.ActionStatusArea class="network-connect-modal__in-progress-logo" @icon={{concat
+                this.radioWalletProviderId "-logo" }} style={{cssVar status-icon-size="2.5rem" }}>
+                <BoxelLoadingIndicator class="network-connect-modal__loading-indicator" @color="var(--boxel-light)" />
+                <div class="network-connect-modal__waiting-status">
+                  Waiting for you to connect your {{!-- network-display-info "layer1" "conversationalName" --}} wallet...
```
Aierie, line 182: Layer 1 doesn't seem to be a user-facing concept in this app. Should we remove the commented text?

packages/safe-tools-client/app/components/network-connect-modal/index.gts
```diff
@@ -0,0 +1,197 @@
...
+              @disabled={{@isConnecting}}
+              @checkedId={{@currentWalletProviderId}}
+              @hideRadio={{true}}
+              class='network-connect-modal__wallet-group'
+              data-test-wallet-selection as |option|
+            >
+              {{#let option.data as |item|}}
+                <option.component
+                  @name='wallet-provider-selection'
+                  @onChange={{fn (optional @changeWalletProvider) item.id}}
```
Aierie, line 142: We should remember to remove the `optional` later... though with this implementation, it seems possible to add logic for updating the selected item within this component, since it is a local radio selection that only affects external/global state when "Connect Wallet" is pressed.

packages/safe-tools-client/app/components/network-connect-modal/index.gts
```diff
@@ -0,0 +1,197 @@
...
+
+          <Section @title="Connect your wallet">
+            <BoxelRadioInput
+              @name="wallet-provider"
+              @orientation="horizontal"
+              @spacing="default"
+              @groupDescription='Select a wallet to connect to'
+              @items={{this.walletProviders}}
+              @disabled={{@isConnecting}}
+              @checkedId={{@currentWalletProviderId}}
```
Aierie, line 134: `@currentWalletProviderId` seems like it should also be local state.

packages/safe-tools-client/app/templates/application.hbs
backspace: I’ve replaced that button now, thank you

packages/safe-tools-client/app/components/network-connect-modal/index.gts
backspace: ya sure hehe done

packages/safe-tools-client/app/components/network-connect-modal/index.gts
backspace: 👍🏻 I’ve made this an internal property now

packages/safe-tools-client/app/helpers/truncate-middle.ts
backspace: I opened 4768 to track this.

PR #3327, opened 2022-10-18
hub: Execute scheduled payments every 5 minutes

jurgenwerk:
Ticket: CS-4380

Still in progress and untested.


FadhlanR:
I will create another PR to main.

PR #3329, opened 2022-10-18
[safe-tools]: Update setup safe modal UI 

danibonilha:
This PR covers some cases that were left behind on the previous one.

- Adds source code link 
- Conditionally handles  not/enough balance UI
- Refactors modal opening to use ember set helper
- Prevents close behavior while provisioning a safe and adds `ing`
- Adds createSafe mock function to fake loading
- Disables provision if there's not enough balance

Not enough balance UI:
image

Provisioning UI:

https://user-images.githubusercontent.com/20520102/196521933-d2e42f94-7414-44af-a642-f253b83ecf08.mov


packages/safe-tools-client/app/components/setup-safe-modal/index.gts
```diff
@@ -39,44 +54,51 @@ export default class SetupSafeModal extends Component<Signature> {
...
             </li>
-            <li>Once you have scheduled payments, the module (source code here)  {{! TODO: Add link }}
+            <li>Once you have scheduled payments, the module
+              <a
+                href='https://github.com/cardstack/cardstack-module-scheduled-payment'
+                target='_blank'
+                rel='external'
+              >
+                (source code here)
+              </a>
```
backspace, line 64: I see that the design has it this way, I wonder whether we should make it visible as a link though 🤔 like via text colour or permanent underline.

packages/safe-tools-client/app/components/setup-safe-modal/index.gts
```diff
@@ -39,44 +54,51 @@ export default class SetupSafeModal extends Component<Signature> {
...
+          <div class='safe-setup-modal__section-wallet-info'>
+            {{svgJar
+              (if this.notEnoughBalance 'icon-x-circle-ht' 'icon-check-circle-ht')
+              class='safe-setup-modal__section-icon'
+              style=(cssVar
+                icon-color=(if this.notEnoughBalance 'var(--boxel-red)' 'var(--boxel-green)')
+              )
+            }}
+            <p>
+              Your wallet has {{if this.notEnoughBalance 'in'}}sufficient funds to cover the estimated gas cost.
```
backspace, line 81: We’re in a bit of a strange state as we get started with this application as we don’t really have tests, but once conditionals start to appear that’s when I usually want to see them. But since there’s no way to make `notEnoughBalance` be true programatically yet, we’ll just need to remember to write tests for this when there’s sufficient test infrastructure to mock wallet characteristics 🤓

packages/safe-tools-client/app/components/setup-safe-modal/index.gts
```diff
@@ -12,22 +17,32 @@ interface Signature {
...
+
+  @action async createSafe(): Promise<void> { 
+     //TODO: use sdk method createSafeWithModuleAndGuard
+    this.provisioning = true;
+    setTimeout(() => {
+      console.log('Create safe')
+      this.provisioning = false;
+      this.args.onClose();
+    }, 2000); 
+  }
```
backspace, line 39: This pattern of toggling a flag to indicate when something is in progress is one reason to use an [Ember Concurrency task](http://ember-concurrency.com/docs/tutorial/refactor) as `createSafe.isRunning` would be taken care of automatically. (You could also then use `timeout` for the currently-fake timer.)

packages/safe-tools-client/app/components/setup-safe-modal/index.gts
```diff
@@ -12,22 +17,32 @@ interface Signature {
...
+
   <template>
-    <BoxelModal
+    <BoxelModal 
       @size='medium'
-      @isOpen={{@isOpen}}
-      @onClose={{@onClose}}
-    >
+      @isOpen={{@isOpen}} 
+      @onClose={{if this.provisioning (noop) @onClose}}>
```
backspace, line 45: The cursor on the overlay suggests that clicking it will dismiss the modal when provisioning is in progress:

Cardstack Safe Tools 2022-10-18 17-22-53

This makes me think maybe we need a first-class option for `Boxel::Modal` that makes it not dismissible vs this workaround-ish `noop` 🤔

backspace (approved):
I have some suggestions but nothing blocking, looks good!

packages/safe-tools-client/app/components/setup-safe-modal/index.gts
lukemelia: Provisioning will require a signing request, so I think we will want to have some other states in the action chin during this similar to what we do in web-client.

packages/safe-tools-client/app/components/setup-safe-modal/index.gts
```diff
@@ -39,44 +54,51 @@ export default class SetupSafeModal extends Component<Signature> {
...
-                'icon-check-circle-ht'
-                class='safe-setup-modal__section-icon'
-              }}
-              <p>
-                Your wallet has sufficient funds to cover the estimated gas cost.
-              </p>
-            </div>
-          {{/if}}
-          {{! TODO: What's the behavior for not having enough gas ? }}
+          <b>Estimated gas cost: {{this.gasCost}}</b>
```
Aierie, line 71: Might be helpful to also provide the wallet address that is going to own the safe? Since this lives in a modal, and it can be troublesome for users to confirm what wallet address is connected.

packages/safe-tools-client/app/components/setup-safe-modal/index.gts
Aierie: Agree that it should be visible as a link! I can work on this in a separate PR.

packages/safe-tools-client/app/components/setup-safe-modal/index.gts
Aierie: Is creation of a new safe is necessary for each chain that the user intends to use? If so, would it also be good to include the network here and to highlight that point?

packages/safe-tools-client/app/components/setup-safe-modal/index.gts
danibonilha: got it, will take into consideration once the integration starts

packages/safe-tools-client/app/components/setup-safe-modal/index.gts
danibonilha: Agreed, will add a ticket for it

packages/safe-tools-client/app/components/setup-safe-modal/index.gts
```diff
@@ -12,22 +17,32 @@ interface Signature {
...
+
+  @tracked provisioning = false;
+
+  @action async createSafe(): Promise<void> { 
+     //TODO: use sdk method createSafeWithModuleAndGuard
+    this.provisioning = true;
+    setTimeout(() => {
+      console.log('Create safe')
+      this.provisioning = false;
+      this.args.onClose();
```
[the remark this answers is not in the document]
danibonilha: totally,  I was thinking about handling with some sort of toast/banner

packages/safe-tools-client/app/components/setup-safe-modal/index.gts
danibonilha: I guess the address will be already visible in the top right corner once we have it connected, but it's ok to add to the modal too. Now the second question I'm not aware of.

Aierie: Ya, address might be visible in the top right, though it will probably have poorer contrast and also be less prioritized visually because of the overlay

PR #3330, opened 2022-10-19
Update arguments of scheduled payment CLI

FadhlanR:
This update is based on @lukemelia 's feedback on the SP CLI demo video. Instead of using a blank string to differentiate a one-time payment and a recurring payment. I updated `payAt`, `recurringDayOfMonth`, and `recurringUntil` from position arguments to option arguments.


packages/cardpay-sdk/sdk/scheduled-payment-module.ts
```diff
@@ -341,10 +341,13 @@ export default class ScheduledPaymentModule {
...
-    payAt: number | null,
-    recurringDayOfMonth: number | null,
-    recurringUntil: number | null
+    payAt?: number | null,
+    recurringDayOfMonth?: number | null,
+    recurringUntil?: number | null
   ): Promise<number> {
+    if (payAt == null && recurringDayOfMonth == null && recurringUntil == null)
+      throw new Error('When payAt is null, recurringDayOfMonth and recurringUntil must have a value');
+
```
lukemelia, line 350: Should we guard against the opposite case as well? i.e. a user specifies payAt as well as one or both the recurring options? This input doesn't make sense and should be an error.

lukemelia (approved):

PR #3331, opened 2022-10-19
sdk & cli: Add function to estimate safe creation gas

FadhlanR:
Adding function to SP module SDK to estimate the safe creation `gas limit`.


alex-cardstack (approved):
LGTM

PR #3332, opened 2022-10-19
[CS-4762]: Adds option to disable overlay interaction on BoxelModal

danibonilha:
This PR is a follow-up from my previous one, it adapts BoxelModal to new behavior and also update the setup-safe-modal to use new arg.


lukemelia (approved):
Looks fine. Minor style thing: I prefer to name booleans with an "is" prefix. e.g. `isProvisioning` and `isOverlayDismissalDisabled`

PR #3333, opened 2022-10-19
Add css var for aligning BoxelField label text

lukemelia:


packages/boxel/addon/components/boxel/field/index.css
```diff
@@ -1,4 +1,6 @@
 .boxel-field {
+  --boxel-field-label-align: top;
```
backspace, line 2: `top` isn’t supported for `align-items`, should it be [`normal`](https://developer.mozilla.org/en-US/docs/Web/CSS/align-items#formal_definition), which is the default?

backspace (changes requested):

packages/boxel/addon/components/boxel/field/index.css
lukemelia: d'oh, good catch

backspace (approved):

PR #3334, opened 2022-10-19
[CS-4658]: Add safe info to control panel

danibonilha:
This PR adds the mocked UI for when a user has a safe. I've created a data structure to show the safe info, but that's flexible if we need to change it, also, I'm not sure about ember patterns to split concerns etc, but it felt cleaner to have the get safeButton function in order to render what we need, lmk if there's a diff pattern you'd like me to follow. It adds a flag for the deposit modal, which I'll be working on next.

PS: address truncation will be added once we have the full address and/or helper

image

### DIFF
--- a/packages/safe-tools-client/app/templates/schedule.hbs
+++ b/packages/safe-tools-client/app/templates/schedule.hbs
@@ -22,9 +22,9 @@
       {{/if}}
       <Boxel::Button
         @kind='primary'
-        {{on 'click' (set this this.safeButton.modalFlag true)}}
+        {{on 'click' this.onSafeButtonClick}}
       >
-        {{this.safeButton.label}}
+        {{this.safeButtonLabel}}
       </Boxel::Button>
     </cp.Item>
   </Boxel::ControlPanel>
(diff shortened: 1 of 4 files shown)


packages/safe-tools-client/app/templates/schedule.hbs
```diff
@@ -7,15 +7,24 @@
...
+      {{else}}
+        <p class='safe-tools__dashboard-schedule-control-panel-safe-info-create'>
+          You need to have a safe which contains funds that would be deducted
+          automatically when schedule payments are due.
+        </p>
+      {{/if}}
       <Boxel::Button
         @kind='primary'
-        {{on 'click' (set this 'isSetupSafeModalOpen' true)}}
+        {{on 'click' (set this this.safeButton.modalFlag true)}}
```
lukemelia, line 25: This level of indirection in the template feels like a step too far. Instead make an `onSafeButtonClick` action in the controller and move the assignment logic there.

lukemelia (changes requested):

lukemelia (approved):

PR #3336, opened 2022-10-19
Update Node to 16

backspace:
I confirmed that [this Ember CLI bug](https://github.com/ember-cli/ember-cli/pull/9987) was happening. I updated `web-client` to Ember CLI 4.8.0 but then tests wouldn’t run:

Monosnap 2022-10-19 17-47-37

The code in question:

Monosnap 2022-10-19 17-48-05

The Ember version is being incorrectly detected to be below 3.27, as seen in [this thread](https://discuss.emberjs.com/t/embroider-dependencysatisfies-gives-wrong-answer-for-ember-source/19464/4). It’s true that `@cardstack/core` has 3.24 but the error persisted even when I updated that to use 3.28 🤔

Possible next step: updating all Ember applications to 4.8. (But not `cardhost` and `core`?)


backspace:
Thanks for looking into this @ef4, did you have any luck with the `nohoist` attempts?

I’m curious about the pnpm fix, but how major of an undertaking would switching to that be?

jurgenwerk:
Were there any new developments made that would allow us to upgrade node to 16?

PR #3341, opened 2022-10-20
tools: Add support for connecting

backspace:
This allows you to connect via Metamask or Wallet Connect:

Cardstack Safe Tools 2022-10-24 14-41-12

The `Wallet` service provides a bare minimum interface to build onto, I’ve incorporated #3363 so the banner is collapsed when the wallet is connected.

Tests are forthcoming, I’m submitting this for review so the service can be used in parallel work. Tests with mock Metamask and Wallet Connect behaviours will drive these improvements:
* modal should indicate connection in progress
* and it should hide when connection is successful
* reconnection happens on application load, need in progress and failure handling
* “Connect” button should be disabled until a provider is chosen

Review note: the new files in `utils` are mostly copied from `web-client`. They will likely be rearranged somewhat to facilitate the above-mentioned automated tests.


packages/safe-tools-client/app/components/network-connect-modal/index.gts
```diff
@@ -130,17 +91,17 @@ class NetworkConnectModal extends Component<Signature> {
...
+              @checkedId={{this.wallet.providerId}}
               @hideRadio={{true}}
               class='network-connect-modal__wallet-group'
               data-test-wallet-selection as |option|
             >
               {{#let option.data as |item|}}
                 <option.component
                   @name='wallet-provider-selection'
-                  @onChange={{fn (mut this.walletProviderId) item.id}}
+                  @onChange={{fn (mut this.wallet.providerId) item.id}}
```
Aierie, line 104: Probably doesn't need to be in this PR, but I think `providerId` is trying to do two things at once, and they should be separated:

1. It's representing the UI state of the radio buttons. This should live in this component, close to the UI, and get passed to `WalletService.connect` as an argument. 

2. It's telling us which wallet provider is currently connected. This should be derived from a property of `ChainConnectionManager` (We probably should have done this in layer 1 chain in web-client, maybe we could make `strategy` a tracked property?).

backspace: I’ve added a property local to the component, thanks. I think the direction of #2 makes sense and I’ll try to include it in forthcoming work 😀

PR #3350, opened 2022-10-21
tools: Fix deployment name in Discord message

backspace:
When I deployed from this branch:

#releases-internal | Cardstack - Discord 2022-10-21 15-33-31


danibonilha (approved):
Nicee, I was like, why did it say web-client 🤣

PR #3354, opened 2022-10-24
Execute scheduled payments every 5 minutes

FadhlanR:
Ticket: [CS-4380](https://linear.app/cardstack/issue/CS-4380)

I have tested the scheduled payment execution worker that Matic developed.

### DIFF
--- a/packages/hub/services/crank-nonce-lock.ts
+++ b/packages/hub/services/crank-nonce-lock.ts
@@ -42,26 +42,19 @@ export default class CrankNonceLock {
   async withNonce(chainId: number, cb: (nonce: BN) => Promise<any>) {
     let prisma = await this.prismaManager.getClient();
 
-    //10s timeout is to wait
-    //until the transaction sent to the RPC node
-    //don't wait the transaction to be mined
-    let timeout = 10;
-    await prisma.$executeRaw`SET lock_timeout TO '${timeout}s'`;
-    await prisma.$executeRaw`SELECT pg_advisory_lock(${this.CRANK_LOCK_KEY})`;
-
     //Make database transaction here
     //because if the execution failed the nonce should be rollback
-    try {
-      return await prisma.$transaction(
-        async (tx) => {
-          let nonce = await this.getNonce(chainId, tx);
-          return await cb(nonce);
-        },
-        { maxWait: 3000, timeout: timeout * 1000 }
-      );
-    } finally {
-      await prisma.$executeRaw`SELECT pg_advisory_unlock(${this.CRANK_LOCK_KEY})`;
-    }
+    return await prisma.$transaction(
+      async (tx) => {
+        await tx.$executeRaw`SELECT pg_advisory_xact_lock(${this.CRANK_LOCK_KEY})`;
+        let nonce = await this.getNonce(chainId, tx);
+        return await cb(nonce);
+      },
+      //10s timeout is to wait
+      //until the transaction sent to the RPC node
+      //don't wait the transaction to be mined
+      { maxWait: 3000, timeout: 10000 }
+    );
   }
 }
 


packages/cardpay-sdk/sdk/utils/conversions.ts
```diff
@@ -35,9 +31,16 @@ export async function gasPriceInToken(provider: JsonRpcProvider, tokenAddress: s
       Accept: 'application/json',
     },
   });
-
   let gasPriceInNativeTokenInWei = new BN((await gasStationResponse.json()).standard);
 
+  // We use the wrapped native token address because the native token doesn't have an address in Uniswap.
+  // The price of the wrapped native token, such as WETH, is the same as the price of the native token.
+  let wrappedNativeToken = getAddressByNetwork('wrappedNativeToken', network);
+  if (tokenAddress === wrappedNativeToken) {
```
FadhlanR, line 39: I added this line to avoid errors when trying to convert wrapped native token using uniswap sdk.

packages/hub/services/scheduled-payments/executor.ts
```diff
@@ -51,6 +52,25 @@ export default class ScheduledPaymentsExecutorService {
     return gasPriceInToken(provider, gasTokenAddress);
   }
 
+  async executeScheduledPayments() {
+    // TODO: take a only a few, can be 10, to not overwhelm this method
+    // TODO: randomize the order
+    let scheduledPayments = await this.scheduledPaymentFetcher.fetchScheduledPayments();
+
+    // Currently we do one by one, but we can do in parallel when we'll have a lot of scheduled payments
```
habdelra, line 60: actually i wonder about this--the crank's EOA nonce could get out of sync if we do parallel on chain calls. The relay server actually uses redis to obtain a lock on the EOA as part of executing a txn on chain specifically so the nonce associated with the EOA will not get out of sync. I think eventually we will probably want to employ something similar to make sure that on chain calls are always performed sequentially.

habdelra: Here you can see an example of this--any txn sent from the relay EOA must use an `EthereumNonceLock` (which is backed by redis in this case)

habdelra: the danger here is that if you have a nonce collision (because 2 concurrent tasks use the same nonce), only one of the txns will be considered valid--and it will be up to the miner that makes the block that determines this. all the other txn with the same nonce will be invalid since the ethereum protocol does not allow mulitple txn's for the same address to have the same nonce

FadhlanR: @habdelra is it possible to add Redis for the hub? to implement this solution?

habdelra: we could also just use the DB we already have. either redis or a postgres DB could be used to facilitate a locking mechanism. I would think using the existing postgres DB is probably a lot easier since we alread have that service configured in all our envs.

FadhlanR: Alright, I'll use postgres DB

lukemelia: Agreed on using postgres with a transaction for this.

packages/hub/services/scheduled-payments/executor.ts
```diff
@@ -51,6 +52,25 @@ export default class ScheduledPaymentsExecutorService {
     return gasPriceInToken(provider, gasTokenAddress);
   }
 
+  async executeScheduledPayments() {
+    // TODO: take a only a few, can be 10, to not overwhelm this method
+    // TODO: randomize the order
```
jurgenwerk, line 57: It's unfortunate that Prisma API [doesn't support randomizing](https://github.com/prisma/prisma/discussions/5886#discussioncomment-2162265) the results. I suppose  we'll need to convert the implementation in ` fetchScheduledPayments` to raw sql to achieve this.

packages/hub/services/scheduled-payments/executor.ts
jurgenwerk: Could PG's [advisory locks](https://www.postgresql.org/docs/current/functions-admin.html#FUNCTIONS-ADVISORY-LOCKS) come handy here? They are useful for application defined locking

FadhlanR: I have added the nonce lock solution

packages/hub/services/scheduled-payments/executor.ts
FadhlanR: I think we could order the scheduled payment by the last attempt.

packages/hub/config/structure.sql
```diff
@@ -1320,6 +1336,14 @@ ALTER TABLE ONLY graphile_worker.migrations
     ADD CONSTRAINT migrations_pkey PRIMARY KEY (id);
 
 
+--
+-- Name: account_nonces account_nonces_pkey; Type: CONSTRAINT; Schema: public; Owner: postgres
+--
+
+ALTER TABLE ONLY public.account_nonces
+    ADD CONSTRAINT account_nonces_pkey PRIMARY KEY (id);
```
habdelra, line 1344: i think you actually have a multi-column uniqueness constraint--specifically the `account_address`, `chain_id`, `nonce` columns form a tuple that should be unique

packages/hub/config/structure.sql
```diff
@@ -1527,6 +1551,13 @@ ALTER TABLE ONLY public.wyre_prices
 CREATE INDEX jobs_priority_run_at_id_locked_at_without_failures_idx ON graphile_worker.jobs USING btree (priority, run_at, id, locked_at) WHERE (attempts < max_attempts);
 
 
+--
+-- Name: account_nonces_account_address_chain_id_unique_index; Type: INDEX; Schema: public; Owner: postgres
+--
+
+CREATE UNIQUE INDEX account_nonces_account_address_chain_id_unique_index ON public.account_nonces USING btree (account_address, chain_id);
```
habdelra, line 1558: i think you actually have a multi-column uniqueness constraint--specifically the account_address, chain_id, nonce columns form a tuple that should be unique. it looks like the nonce column is missing from this constraint.

packages/hub/services/nonce-lock.ts
```diff
@@ -0,0 +1,79 @@
...
+
+        return nextNonce;
+      },
+      { maxWait: 3000, timeout: 10000 }
+    );
+  }
+
+  async withNonce(accountAddress: string, chainId: number, cb: (nonce: BN) => Promise<any>) {
+    let prisma = await this.prismaManager.getClient();
+    let nonce = await this.getNonce(accountAddress, chainId);
```
habdelra, line 59: i think i see a problem here--you should actually block until the txn is submitted on chain. this assumes that the txn will succeed. Consider the fact that 2 txns call `withNonce` at exactly the same time. this will assign one of the txns a nonce and the other txn the next nonce. however, i think if the txn reverts, then you don't want to advance the nonce (probably we should look that up to be sure). you don't actually know that until you you send the txn. while the txn is being sent, i don't think you can assign a nonce, and any other calls to `withNonce` probably need to continue to block (via an `await`) until you know that the txn doesn't revert.

packages/hub/services/scheduled-payments/executor.ts
habdelra: I've used advisory locks in the past, and yes, they are created for situations like these. in this case there only ever is one crank EOA, so i think a single lock is all you need (as opposed to a solution that had multiple crank EOA's in which case each crank would need a lock)

packages/hub/services/nonce-lock.ts
```diff
@@ -0,0 +1,79 @@
...
+    let prisma = await this.prismaManager.getClient();
+    return await prisma.$transaction(
+      async (tx) => {
+        await tx.$executeRaw`SELECT * FROM account_nonces WHERE account_address = ${accountAddress} AND chain_id = ${chainId} FOR UPDATE`;
+
+        let accountNonce = await tx.accountNonce.findUnique({
+          where: { accountAddress_chainId: { accountAddress, chainId } },
+        });
+        let one = new BN(1);
+        let nextNonce = accountNonce?.nonce ? new BN(accountNonce.nonce.toString()).add(one) : new BN(0);
```
habdelra, line 22: we should probably ask ethers.js for the current nonce for the EOA and add one to that right? after all we are trying to reflect what is happening in ethereum--ethereum is the source of truth not us. if the nonce we get back is one that is already in the DB then we know something is wrong (probably related to my other comments around what happens when a txn reverts and a nonce is not advanced).

habdelra: oh wait never min. the API i'm thinking of is from web3.js. in web3 you can ask an EOA's current nonce (`web3.eth.getTransactionCount(address)`), but in ethers.js, the transaction count is only known _after_ you send the txn, which you determine on line 27 below.

packages/hub/services/nonce-lock.ts
```diff
@@ -0,0 +1,79 @@
...
+            where: { id: accountNonce.id },
+            data: {
+              nonce: BigInt(nextNonce.toString()),
+              updatedAt: now,
+            },
+          });
+        } else {
+          await tx.accountNonce.create({
+            data: {
+              id: shortUUID.uuid(),
```
habdelra, line 41: postgres can make this field value for us--we should not have to specify it.

packages/hub/services/nonce-lock.ts
```diff
@@ -0,0 +1,79 @@
...
+        if (accountNonce) {
+          await tx.accountNonce.update({
+            where: { id: accountNonce.id },
+            data: {
+              nonce: BigInt(nextNonce.toString()),
+              updatedAt: now,
+            },
+          });
+        } else {
+          await tx.accountNonce.create({
```
habdelra, line 39: instead of a separate `create` and `update`, we should use an `upsert`--that will do a create or update in a single call: https://www.prisma.io/docs/reference/api-reference/prisma-client-reference#upsert

packages/hub/services/nonce-lock.ts
```diff
@@ -0,0 +1,79 @@
...
+            data: {
+              id: shortUUID.uuid(),
+              accountAddress,
+              chainId,
+              nonce: BigInt(nextNonce.toString()),
+              createdAt: now,
```
habdelra, line 45: this field should be auto populated by postgres instead of us having to provide a value. postgres can use a `now()` function (i believe it is called) to make this value automatically when the row is created

packages/hub/services/nonce-lock.ts
```diff
@@ -0,0 +1,79 @@
...
+              updatedAt: now,
+            },
+          });
+        }
+
+        return nextNonce;
+      },
+      { maxWait: 3000, timeout: 10000 }
```
habdelra, line 53: i'm curious why we have just a long timeout--this is just talking to the DB right? won't the defaults suffice here?

packages/hub/services/nonce-lock.ts
FadhlanR: I was implementing it like that, but then I thought the timeout of the database transaction would be too long because it needs to wait until the execution transaction publishes on the blockchain. That's why I put error handling to reset the nonce to zero in case the execution transaction is failed to publish (https://github.com/cardstack/cardstack/blob/e9ebe31f131a91481b81201ef254ff308dd9022d/packages/hub/services/nonce-lock.ts#L63).

But you're right. I'll include the execution transaction in that database transaction scope.

packages/hub/config/structure.sql
```diff
@@ -818,6 +818,22 @@ CREATE TABLE graphile_worker.migrations (
 
 ALTER TABLE graphile_worker.migrations OWNER TO postgres;
 
+--
+-- Name: account_nonces; Type: TABLE; Schema: public; Owner: postgres
+--
+
+CREATE TABLE public.account_nonces (
+    id uuid NOT NULL,
+    account_address text NOT NULL,
```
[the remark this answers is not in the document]
FadhlanR: So we will have a table that only handles the nonce for crank address? And in that case, this table will only have one row because instead of inserting a new row when the nonce of the crank increases, we could update the value of the nonce column on the existing row.

My idea on this table is to have a nonce lock feature for any accounts, not only for the crank. And I'm not in the idea of having multiple crank addresses because this nonce lock feature is not aimed for that. To do that, we could have another table or mapping from this table to `scheduled_payment_attempts` table. But it's not necessary for now.

packages/hub/services/scheduled-payments/executor.ts
jurgenwerk: It looks like sorting by child records [is not supported](https://stackoverflow.com/questions/67930989/prisma-order-by-relation-has-only-count-property-can-not-order-by-relation-fie) by Prisma as well 🫠 I just tried adding `orderBy: { scheduledPaymentAttempts: { endedAt: 'asc' } },` and there's a type mismatch.

FadhlanR: Or it could be filtered after fetching to the database.

jurgenwerk: You mean by loading all the scheduled payments that are due and then filtering/ordering? That could work while we have a low number of scheduled payments

packages/hub/services/nonce-lock.ts
FadhlanR: updated

packages/hub/services/nonce-lock.ts
FadhlanR: updated

packages/hub/services/nonce-lock.ts
FadhlanR: updated

packages/hub/services/nonce-lock.ts
FadhlanR: updated

packages/hub/services/nonce-lock.ts
FadhlanR: updated, now the timeout with 10 seconds makes sense because the DB transaction needs to wait until the execution transaction is published.

packages/hub/config/structure.sql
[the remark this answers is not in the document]
habdelra: the the idea of nonce locking is necessary for the EOA _sending_ the txn. and I believe the only EOA sending txn's is the crank. in terms of the safe address that the crank is executing txns for, i don't see how that is relevant at all to the nonce locking--it seems like superfluous data... if that is necessary to track, I suggest using a different table.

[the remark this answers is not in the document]
habdelra: yeah--that's a great name that explains the context for the table very well

packages/hub/services/scheduled-payments/executor.ts
FadhlanR: updated, used advisory locks instead of pessimistic locks

packages/hub/config/structure.sql
FadhlanR: updated

packages/hub/config/structure.sql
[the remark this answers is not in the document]
FadhlanR: renamed

packages/hub/services/scheduled-payments/fetcher.ts
```diff
@@ -19,44 +21,74 @@ export default class ScheduledPaymentsFetcherService {
...
+    let results: [{ id: string; started_at: Date }] = await prisma.$queryRaw`SELECT 
+        "scheduled_payments"."id" AS "id",
+        "last_failed_payment_attempts"."started_at"
+      FROM 
+        "scheduled_payments"
+      LEFT JOIN (SELECT "scheduled_payment_attempts"."scheduled_payment_id",
+                  MAX("scheduled_payment_attempts"."started_at") AS "started_at"
+            FROM "scheduled_payment_attempts"
+            WHERE "scheduled_payment_attempts"."status" = 'failed'
+            GROUP BY "scheduled_payment_id") AS "last_failed_payment_attempts"
```
FadhlanR, line 33: @jurgenwerk I joined scheduled payments with the last failed payment attempts. So we can order scheduled payments that are due based on the last failed payment attempts and prioritize the scheduled payment with no payment attempts.

packages/hub/services/crank-nonce-lock.ts
```diff
@@ -0,0 +1,65 @@
...
+
+  private async getNonce(chainId: number, dbTransaction: Prisma.TransactionClient) {
+    let crankNonce = await dbTransaction.crankNonce.findUnique({
+      where: { chainId: chainId },
+    });
+    let one = new BN(1);
+    let nextNonce = crankNonce?.nonce ? new BN(crankNonce.nonce.toString()).add(one) : new BN(0);
+    let nonceTTL = Number(config.get('nonceTTL'));
+    let now = nowUtc();
+    if (nextNonce.cmp(one) <= 0 || (crankNonce && addSeconds(crankNonce.updatedAt, nonceTTL) <= now)) {
```
jurgenwerk, line 22: I think `nextNonce.lte(one)` is a bit simpler to say

packages/hub/services/crank-nonce-lock.ts
```diff
@@ -0,0 +1,65 @@
...
+    try {
+      return await prisma.$transaction(
+        async (tx) => {
+          let nonce = await this.getNonce(chainId, tx);
+          return await cb(nonce);
+        },
+        { maxWait: 3000, timeout: 10000 }
+      );
+    } finally {
+      await prisma.$executeRaw`SELECT pg_advisory_unlock(${this.CRANK_LOCK_KEY})`;
```
jurgenwerk, line 56: What will happen if the process exits before reaching this block, e.g. the app is restarted? Is the crank in a deadlock?

packages/hub/services/crank-nonce-lock.ts
```diff
@@ -0,0 +1,65 @@
...
+    await prisma.$executeRaw`SELECT pg_advisory_lock(${this.CRANK_LOCK_KEY})`;
+    //Make database transaction here
+    //because if the execution failed the nonce should be rollback
+    try {
+      return await prisma.$transaction(
+        async (tx) => {
+          let nonce = await this.getNonce(chainId, tx);
+          return await cb(nonce);
+        },
+        { maxWait: 3000, timeout: 10000 }
```
jurgenwerk, line 53: Can you explain why this is needed and on what these numbers are based on?

packages/hub/services/scheduled-payments/fetcher.ts
```diff
@@ -19,44 +21,74 @@ export default class ScheduledPaymentsFetcherService {
...
+          AND "scheduled_payments"."pay_at" > ${startOfDay(subDays(nowUtc, this.validForDays))}
+          AND "scheduled_payments"."pay_at" <= ${nowUtc}
+          AND (
+            (
+              "scheduled_payments"."recurring_day_of_month" IS NULL 
+              AND (
+                "scheduled_payments"."id"
+              ) NOT IN (
+                SELECT 
+                  "t0"."id"
```
jurgenwerk, line 48: Is this SQL generated by a script? It doesn't look written by hand. `t0`, `j0` etc sound too cryptic, let's use human readable code. Also you can drop the "s, for example.  `"scheduled_payments"."recurring_day_of_month"` can be simplified to `scheduled_payments.recurring_day_of_month`

packages/hub/services/scheduled-payments/fetcher.ts
```diff
@@ -19,44 +21,74 @@ export default class ScheduledPaymentsFetcherService {
...
+                  (
+                    "j0"."status" = 'in_progress'
+                    AND "t0"."id" IS NOT NULL
+                  )
+              )
+            )
+          )
+        )
+        ORDER BY CASE WHEN  "last_failed_payment_attempts"."started_at" IS NULL THEN 0 ELSE 1 END
+        LIMIT 10;`;
```
jurgenwerk, line 83: I'd make this configurable via a param

packages/hub/services/scheduled-payments/fetcher.ts
```diff
@@ -19,44 +21,74 @@ export default class ScheduledPaymentsFetcherService {
...
+                WHERE 
+                  (
+                    "j0"."status" = 'in_progress'
+                    AND "t0"."id" IS NOT NULL
+                  )
+              )
+            )
+          )
+        )
+        ORDER BY CASE WHEN  "last_failed_payment_attempts"."started_at" IS NULL THEN 0 ELSE 1 END
```
jurgenwerk, line 82: Did you mean `ORDER BY CASE WHEN  last_failed_payment_attempts.started_at IS NULL THEN 0 ELSE started_at END`?

packages/hub/services/scheduled-payments/fetcher.ts
jurgenwerk, line 33: Would it be better to say "last_failed_payment_attempt" instead of "last_failed_payment_attempts"? Given it's just one.

jurgenwerk (changes requested):

packages/hub/prisma/schema.prisma
```diff
@@ -330,6 +330,16 @@ model ScheduledPayment {
   @@map("scheduled_payments")
 }
 
+model CrankNonce {
+  id        String   @id @default(dbgenerated("'894820e7-e24f-48b3-826f-3239996626cb'::uuid")) @db.Uuid
```
habdelra, line 334: hard coded uuid again?

packages/hub/services/scheduled-payments/fetcher.ts
```diff
@@ -19,44 +21,74 @@ export default class ScheduledPaymentsFetcherService {
...
+        LIMIT 10;`;
+
+    // Retrieve scheduledPayment prisma object based on filtered id
+    return prisma.scheduledPayment.findMany({ where: { id: { in: results.map((result) => result.id) } } });
```
jurgenwerk, line 86: This could be done in a subquery together with the query above.

packages/hub/db/migrations/20221025140643069_create-crank-nonces.ts
```diff
@@ -0,0 +1,18 @@
+import { MigrationBuilder } from 'node-pg-migrate';
+import { uuid } from 'short-uuid';
+
+const TABLE = 'crank_nonces';
+
+export async function up(pgm: MigrationBuilder): Promise<void> {
+  pgm.createTable(TABLE, {
+    id: { type: 'uuid', primaryKey: true, default: uuid() },
+    chain_id: { type: 'integer', notNull: true, unique: true },
```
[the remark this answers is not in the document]
FadhlanR: I already made `chainId` a unique field, and as far as I know, a unique field will be indexed.

https://dba.stackexchange.com/questions/2783/does-making-a-field-unique-make-it-indexed

packages/hub/services/crank-nonce-lock.ts
FadhlanR: I put the unlock process in the final block to ensure that the advisory lock will be unlocked no matter what happens in the execution process.

packages/hub/services/crank-nonce-lock.ts
FadhlanR: - `maxWait`: The maximum amount of time the Prisma Client will wait to acquire a transaction from the database. The default value is 2 seconds.
- `timeout:` The maximum amount of time the interactive transaction can run before being canceled and roll back. The default value is 5 seconds.

I increased the value of `maxWait` and `timeout` because we need to wait for the execution transaction to be published, and that process may take more than 5 seconds.

packages/hub/services/scheduled-payments/fetcher.ts
FadhlanR: Some of them I got from the previous prisma query. I will rename it.

packages/hub/config/structure.sql
```diff
@@ -857,6 +857,21 @@ CREATE TABLE public.cards (
 
 ALTER TABLE public.cards OWNER TO postgres;
 
+--
+-- Name: crank_nonces; Type: TABLE; Schema: public; Owner: postgres
+--
+
+CREATE TABLE public.crank_nonces (
+    id uuid DEFAULT '894820e7-e24f-48b3-826f-3239996626cb'::uuid NOT NULL,
```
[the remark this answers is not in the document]
FadhlanR: It's generated from the migration script. I'll update the default value of this field as per Matic's comment.

packages/hub/services/scheduled-payments/fetcher.ts
FadhlanR: Is there a way to automatically map the DB result to the Prisma object? I did this because the naming fields of `scheduled_payments` table and `scheduledPayment` Prisma object are different. For example, `chain_id` and `chainId`. Another solution is to use `AS` in the raw query so we can rename the field, but I think it's not easy to maintain because if we add another field or rename the existing field to `scheduled_payments` table, we always need to update that raw query.

packages/hub/services/scheduled-payments/fetcher.ts
FadhlanR: No, it will not works because started_at is a `timestamp without timezone` we need an integer on that `ELSE` scope.

packages/hub/main.ts
```diff
@@ -97,6 +97,8 @@ import ScheduledPaymentSerializer from './services/serializers/scheduled-payment
 import ScheduledPaymentValidator from './services/validators/scheduled-payment';
 import ScheduledPaymentsRoute from './routes/scheduled-payments';
 import ScheduledPaymentsExecutorService from './services/scheduled-payments/executor';
+import EthersProvideer from './services/ethers-provider';
```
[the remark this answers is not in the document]
FadhlanR: updated

packages/hub/routes/scheduled-payments.ts
```diff
@@ -147,9 +147,13 @@ export default class ScheduledPaymentsRoute {
...
 
-        await this.workerClient.addJob('scheduled-payment-on-chain-creation-waiter', {
-          scheduledPaymentId: scheduledPayment.id,
-        });
+        try {
+          await this.workerClient.addJob('scheduled-payment-on-chain-creation-waiter', {
+            scheduledPaymentId: scheduledPayment.id,
+          });
+        } catch (e) {
+          console.log(e);
```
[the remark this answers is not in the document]
FadhlanR: Reverted this updates

packages/hub/services/crank-nonce-lock.ts
```diff
@@ -0,0 +1,65 @@
...
+  ethersProvider = inject('ethers-provider', { as: 'ethersProvider' });
+  CRANK_LOCK_KEY = 1;
+
+  private async getNonce(chainId: number, dbTransaction: Prisma.TransactionClient) {
+    let crankNonce = await dbTransaction.crankNonce.findUnique({
+      where: { chainId: chainId },
+    });
+    let one = new BN(1);
+    let nextNonce = crankNonce?.nonce ? new BN(crankNonce.nonce.toString()).add(one) : new BN(0);
+    let nonceTTL = Number(config.get('nonceTTL'));
```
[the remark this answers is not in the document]
FadhlanR: updated

packages/hub/services/crank-nonce-lock.ts
FadhlanR: updated

packages/hub/services/scheduled-payments/fetcher.ts
FadhlanR: Updated, make it more readable `ORDER BY CASE WHEN last_failed_payment_attempt.started_at IS NULL THEN ${convertDateToUTC(new Date(0))} ELSE last_failed_payment_attempt.started_at END`

packages/hub/services/scheduled-payments/fetcher.ts
FadhlanR: renamed

packages/hub/config/structure.sql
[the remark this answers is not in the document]
FadhlanR: updated

packages/hub/prisma/schema.prisma
FadhlanR: updated

packages/hub/services/scheduled-payments/fetcher.ts
FadhlanR: updated

packages/hub/services/scheduled-payments/fetcher.ts
FadhlanR: updated

packages/hub/db/migrations/20221025140643069_create-crank-nonces.ts
```diff
@@ -0,0 +1,18 @@
+import { MigrationBuilder } from 'node-pg-migrate';
+import { uuid } from 'short-uuid';
+
+const TABLE = 'crank_nonces';
+
+export async function up(pgm: MigrationBuilder): Promise<void> {
+  pgm.createTable(TABLE, {
+    id: { type: 'uuid', primaryKey: true, default: uuid() },
```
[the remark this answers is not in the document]
FadhlanR: updated

packages/hub/services/crank-nonce-lock.ts
jurgenwerk: Does this mean that transaction has to be published in under 10 seconds, otherwise the transaction will fail? Is it normal this should happen under 10s, even on mainnet?

packages/hub/services/crank-nonce-lock.ts
jurgenwerk: The process can terminate during execution for many reasons (server crash, app crash) after claiming the lock and before reaching the `finally` block. Does that mean someone will have to manually unlock it on app restart?

packages/hub/services/scheduled-payments/fetcher.ts
```diff
@@ -8,55 +8,92 @@ export default class ScheduledPaymentsFetcherService {
   validForDays = 3;
 
   // This query fetches scheduled payments that are due to be executed now.
+  // Will prioritize scheduled payments with no payment attempts.
+  // After that scheduled payments with the earlier falied payment attempts.
```
jurgenwerk, line 12: falied typo

packages/hub/services/scheduled-payments/fetcher.ts
```diff
@@ -8,55 +8,92 @@ export default class ScheduledPaymentsFetcherService {
...
   // This query mostly relies on the payAt field, which is set to the time when the payment should be executed.
   // In case of one-time payments, payAt is set only initially, and then never changed.
   // In case of recurring payments, payAt will update on every successful payment (to 1 month in the future).
   // It is also important to note that validForDays tells us for how many days the payment is still valid to retry in
   // case it is failing for some reason.
 
-  async fetchScheduledPayments(): Promise<ScheduledPayment[]> {
+  async fetchScheduledPayments(limit = 10): Promise<ScheduledPayment[]> {
     let prisma = await this.prismaManager.getClient();
     let nowUtc = convertDateToUTC(new Date());
```
jurgenwerk, line 22: We now have a `nowUtc()` function in utils/dates.

packages/hub/services/crank-nonce-lock.ts
FadhlanR: No, it means the database transaction will be acquired by this process for 10 seconds. If publishing the transaction to the blockchain takes more than 10 seconds, the database transaction will be released/rollback.

We only wait until the transaction is published, not mine, so it must be lower than 10 seconds.

packages/hub/services/scheduled-payments/fetcher.ts
```diff
@@ -8,55 +8,92 @@ export default class ScheduledPaymentsFetcherService {
...
+                    scheduled_payment_attempts1.status = 'in_progress'
+                    AND scheduled_payments1.id IS NOT NULL
+                  )
+              )
+            )
+          )
+        )
+        ORDER BY 
+          CASE WHEN last_failed_payment_attempt.started_at IS NULL 
+          THEN ${convertDateToUTC(new Date(0))}
```
jurgenwerk, line 86: You can use `utcNow()`. But I'm not sure I understand why sorting by an evaluated date value. Shouldn't this be sorting on a database field?

Please add ASC/DESC to make it clearer.

packages/hub/services/scheduled-payments/fetcher.ts
jurgenwerk: Ok I see the issue. It's a pity how unflexible Prisma is for nontrivial queries.

packages/hub/services/scheduled-payments/fetcher.ts
FadhlanR: I'm not trying to get the current date time. What I'm trying to do is insert the oldest date time into the scheduled payment with no payment attempts. So I can sort scheduled payments by started_at of the last failed payment attempt.

jurgenwerk (commented):
It would be really great to have a test for:

1. New sorting additions to the fetcher query
2. `executeScheduledPayments()` method

packages/hub/services/crank-nonce-lock.ts
jurgenwerk: >  If publishing the transaction to the blockchain takes more than 10 seconds, the database transaction will be released/rollback.

This is the part I am worried about. What makes us think it shouldn't take more than 10 seconds to publish the transaction? What about mainnet? Or when the network is really congested? 

p.s. whatever conclusion comes from this debate please add a comment in the code explaining it so the next person can see why/how we got these numbers.

habdelra: the act of sending a txn is a normal web request. these generally have responses in less than 1 second. A server that takes 10 seconds to respond is incredibly degraded.

habdelra: keep in mind that sending a txn is not the same as waiting for a txn to be mined (aka consensus). you can send a txn to any RPC node--so network congestion in the normal sense doesn't have any impact here since we are talking about a  decentralized system. there is not like a single RPC node that is clogged up when the network traffic is high. Albiet, we might be unlucky enough to be talking to an RPC node that is over burdended by a lot of clients--which is something that can happen even when network congestion for the protocol is low.

packages/hub/services/scheduled-payments/fetcher.ts
habdelra: (@jurgenwerk `new Date(0)` is jan 1st, 1970)

packages/hub/services/crank-nonce-lock.ts
jurgenwerk: 👍 Thanks, I'm fine with that then.

packages/hub/services/scheduled-payments/fetcher.ts
habdelra: is this so that nulls are sorted at the end? i think that there is a command you can use to tell postgres to place nulls at the end (or the beginning). i vaguely recall having to do this in the past

habdelra: Here it is:
```
ORDER BY somevalue DESC NULLS LAST
```
https://stackoverflow.com/questions/7621205/sort-null-values-to-the-end-of-a-table

packages/hub/services/scheduled-payments/fetcher.ts
```diff
@@ -1,62 +1,101 @@
...
+                    (
+                      scheduled_payment_attempts1.status = 'succeeded'
+                      OR scheduled_payment_attempts1.status = 'in_progress'
+                    ) 
+                    AND scheduled_payments1.id IS NOT NULL
+                  )
+              )
+            ) 
+            OR (
+              scheduled_payments.recurring_day_of_month IN (${Prisma.join(validRecurringDays, ',')})
```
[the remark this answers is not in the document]
FadhlanR: I don't think we have `payAt` data for recurring payment

packages/hub/services/scheduled-payments/fetcher.ts
FadhlanR: updated

packages/hub/services/scheduled-payments/fetcher.ts
FadhlanR: updated

packages/hub/services/crank-nonce-lock.ts
FadhlanR: Added comment to the code

packages/hub/services/crank-nonce-lock.ts
FadhlanR: I added this line of code `await prisma.$executeRaw`SET lock_timeout TO '${timeout}s'`;` to have a timeout for locking, is this good enough to avoid deadlock?

packages/hub/services/scheduled-payments/fetcher.ts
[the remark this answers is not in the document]
jurgenwerk: `payAt` is used  for recurring payments as well: https://github.com/cardstack/cardstack/blob/main/packages/hub/services/scheduled-payments/fetcher.ts#L14

It just occurred to me that we are missing setting the `payAt` at the end of the payment execution. Similar to what we do here: https://github.com/cardstack/cardstack/blob/main/packages/hub/routes/scheduled-payments.ts#L98

packages/hub/services/crank-nonce-lock.ts
jurgenwerk: I think so!

packages/hub/services/scheduled-payments/fetcher.ts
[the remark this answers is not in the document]
FadhlanR: updated

jurgenwerk (approved):

PR #3355, opened 2022-10-24
[CS-4662][safe-tools]: Add deposit modal

danibonilha:
This PR adds the deposit modal. The initial setup without scheduled payments and the case where all balances are enough coved, are handled. It's missing the tokens list. I'm opening as a draft so we can discuss the args naming and data structure. I'm not toosure if tokenToCover it's a meaningful enough name. I've used the `flashCopiedConfirmation` from the input usage, if there's other places if this kind of copied banner, we can extract it to be a component. Also I believe it will be nice to share the tokens icon mapping from web, probably on a next PR ?

image

image


packages/safe-tools-client/app/components/deposit-modal/index.gts
```diff
@@ -0,0 +1,128 @@
...
+      this.isShowingCopiedConfirmation = false;
+    } , 1000)
+  }
+
+  <template>
+    <BoxelModal @size="small" @isOpen={{@isOpen}} @onClose={{@onClose}}>
+      <BoxelActionContainer as |Section ActionChin|>
+        <Section @title="Deposit Instructions" class="deposit-modal__section">
+          <p>
+            it is the user's responsibility to ensure that sufficient funds are
```
backspace, line 54:
```suggestion
            It is the user’s responsibility to ensure that sufficient funds are
```

The design is inconsistent re the typographer apostrophe but we might as well be consistent and use it 🤓 

Schedule Payment - Deposit Modal - MVP - Zeplin 2022-10-24 09-27-59

backspace (approved):

packages/safe-tools-client/app/components/deposit-modal/index.gts
danibonilha: wow, good eye, I'll be more attentive to use `’`

PR #3357, opened 2022-10-24
[CS-4662]: Add token list on deposit modal

danibonilha:
This PR, adds the token list on the deposit modal when the user has scheduled payments. it also handles multiple tokens balances.

variations

both

Safe multiple tokens:
image

### DIFF
--- a/packages/safe-tools-client/app/components/deposit-modal/index.css
+++ b/packages/safe-tools-client/app/components/deposit-modal/index.css
@@ -31,7 +31,7 @@
 }
 
 .deposit-modal__section p.deposit-modal__section-funds-info-balance {
-  margin: 0;
+  margin: var(--boxel-sp-xxs) 0;
 }
 
 .deposit-modal__section-funds-info-icon {
--- a/packages/safe-tools-client/app/components/deposit-modal/index.css
+++ b/packages/safe-tools-client/app/components/deposit-modal/index.css
@@ -43,7 +43,7 @@
 }
 
 .deposit-modal__section-funds-info-token {
-  font-size: 14px;
+  font-size: var(--boxel-font-size-sm);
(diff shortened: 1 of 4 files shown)


packages/safe-tools-client/app/components/deposit-modal/index.css
```diff
@@ -12,23 +12,49 @@
...
 .deposit-modal__section-funds-info-icon {
   --icon-color: var(--boxel-dark);
 
   flex: 0.1;
   width: 24px;
   height: 24px;
 }
+
+.deposit-modal__section-funds-info-token {
+  font-size: 14px;
```
backspace, line 46: In cases where the design specifies something that isn’t covered by the `--boxel-font-size…` properties, I tend toward using one of those vs specifying it numerically, just so we have more consistent appearance in the applications. `--boxel-font-size-sm` is 13px which seems close enough to me 😆

packages/safe-tools-client/app/controllers/schedule.ts
```diff
@@ -14,18 +14,32 @@ export default class Schedule extends Controller {
...
+      tokens: [
+        {
+          address: 'eth',
+          name: 'Ethereum',
+          symbol: 'ETH',
+          decimals: 18,
+          balance: 1, //in ETH
+        },
+        {
+          address: '0x6B...1d0F', // TODO: check when address is hex, to truncate on template
```
backspace, line 29: Is this the token contract address? I know this is static/placeholder, I’m wondering if this is something we’ll really be supporting 😯

danibonilha: Not sure, was just trying to follow a pattern,  I think that it's confusing to show token address and then safe address, maybe we should be showing the nativeToken ? 🤔

PR #3352, opened 2022-10-24
Update indexer and api to use explanation

tintinthong:
This indexes the explanation id and explanation data.

**Example output**

Look at `explanationId` and `explanationData`

```
[
  {
    "rootHash": "0x4ae8a3bf1ba836cd297318afbdff53b877334beb66beee294f9f85f6136fba3a",
    "paymentCycle": 27551782,
    "tokenAddress": "0xB0427e9F03Eb448D030bE3EBC96F423857ceEb2f",
    "payee": "0x159ADe032073d930E85f95AbBAB9995110c43C71",
    "proofArray": [],
    "rewardProgramId": "0xab20c80fcc025451a3fc73bB953aaE1b9f640949",
    "amount": "10000000000000000000",
    "leaf": "0x000000000000000000000000ab20c80fcc025451a3fc73bb953aae1b9f6409490000000000000000000000000000000000000000000000000000000001a468260000000000000000000000000000000000000000000000000000000001a468260000000000000000000000000000000000000000000000000000000001b045a60000000000000000000000000000000000000000000000000000000000000001000000000000000000000000159ade032073d930e85f95abbab9995110c43c7100000000000000000000000000000000000000000000000000000000000000e00000000000000000000000000000000000000000000000000000000000000040000000000000000000000000b0427e9f03eb448d030be3ebc96f423857ceeb2f0000000000000000000000000000000000000000000000008ac7230489e80000",
    "validFrom": 27551782,
    "validTo": 28329382,
    "explanationId": "flat_payment",
    "explanationData": {
      "amount": "10000000000000000000",
      "token": "0xB0427e9F03Eb448D030bE3EBC96F423857ceEb2f"
    }
  },
  {
    "rootHash": "0x0443109c61fd5325eab4a3c71bd65c5620f1f876dc957267983c95ce2aad5978",
    "paymentCycle": 27551788,
    "tokenAddress": "0xB0427e9F03Eb448D030bE3EBC96F423857ceEb2f",
    "payee": "0x159ADe032073d930E85f95AbBAB9995110c43C71",
    "proofArray": [],
    "rewardProgramId": "0xab20c80fcc025451a3fc73bB953aaE1b9f640949",
    "amount": "10000000000000000000",
[...]

PR #3360, opened 2022-10-25
Extract transfer details from leaf

IanCal:
First step towards removing the indexer from knowing what's inside the rewards, so that it can serve up proofs for other things present in the merkle tree.

I changed the SDK slightly to get the token address and amount from the leaf rather than the API.

Removes an explicit error being thrown on other tokenTypes, instead just returning blank amounts/token addresses so it doesn't break existing dbs.


tintinthong:
This PR can go in first before https://github.com/cardstack/cardstack/pull/3352

IanCal:
> This PR can go in first before #3352

Thanks - happy to do either way if there's any conflicts, could be easier to fix here. Can merge whichever is sorted first.

IanCal:
PDM failure is happening on main too :/

tintinthong (approved):

PR #3361, opened 2022-10-25
Abi updating script for python packages

IanCal:
A bunch of extra files as it copies all abis. Main change is to add a script in the root scripts folder that reads all the latest abis from the SDK and copies them into the abi folders in the three python packages that need them. Minor tweaks after to read the kebab case names.

It'll sync everything with:

`yarn codegen:python-package-abis`

Am not so happy about the eval in there to get the json structure back out of the typescript file, but that's how it's done with the SDK codegen scripts.


IanCal:
Q: I had to add an extra line to disable the linting in the script about "extraneous modules" but I'm not sure if that's raising an actual issue.

tintinthong (approved):
Looks good! I added the missing scheduler abis

PR #3364, opened 2022-10-25
Introduce ember-shared addon to share truncate-middle...

lukemelia:
...and future shared helpers, components, etc.

It is implemented as a v2 addon with a sibling test-app


packages/web-client/package.json
```diff
@@ -161,6 +161,7 @@
     "extends": "../../package.json"
   },
   "dependencies": {
+    "@cardstack/ember-helpers": "^0.0.1",
```
backspace, line 164: It used to matter with Ember CLI whether things go in `dependencies` or `devDependencies`, in `safe-tools-client` it’s different from this. But maybe it doesn’t matter anymore?

Should this use the same lockstep version as the rest of the packages?

backspace:
All the `web-client` tests having disappeared is reminiscent of [the Node 16 branch](https://github.com/cardstack/cardstack/pull/3336#issuecomment-1284657435) though I don’t know if it’s truly related 🤔

backspace (approved):
Other than the lint errors and test problems, this looks good to me. Will the `test-app` be added to CI elsewhere?

`helpers` being in the name doesn’t mean we should only use it for that, right? I have been wanting to share other things between applications, like utility functions, maybe some wallet code, even configuration?

packages/web-client/package.json
lukemelia: >It used to matter with Ember CLI whether things go in dependencies or devDependencies, in safe-tools-client it’s different from this. But maybe it doesn’t matter anymore?

For apps, I don't think it matters. For addons, it definitely matters.

>Should this use the same lockstep version as the rest of the packages?

Ah, good point.

lukemelia:
> Other than the lint errors and test problems, this looks good to me. Will the test-app be added to CI elsewhere?

Yes, working on this now.

> helpers being in the name doesn’t mean we should only use it for that, right? I have been wanting to share other things between applications, like utility functions, maybe some wallet code, even configuration?

Agreed. Can you think of a better name?

lukemelia:
> All the web-client tests having disappeared is reminiscent of https://github.com/cardstack/cardstack/pull/3336#issuecomment-1284657435 though I don’t know if it’s truly related 🤔

It is because the v2 addon format needs to be built before being consumed. I'm working out the best way to do that in CI.

backspace:
> Can you think of a better name?

`ember-shared`? But `ember-helpers` is surely fine too, I just wanted to confirm 🤓

package.json
```diff
@@ -43,20 +43,24 @@
     "codegen:subgraph-sokol": "yarn --cwd packages/cardpay-subgraph codegen-sokol",
     "codegen:subgraph-xdai": "yarn --cwd packages/cardpay-subgraph codegen-xdai",
     "build:hub": "yarn --cwd packages/hub build",
+    "build:ember-shared": "yarn --cwd packages/ember-shared/addon build",
+    "build_if_missing:ember-shared": "yarn --cwd packages/ember-shared/addon build_if_missing",
```
ef4, line 47: An alternative to this would be to put

```
"scripts": {
  "prepare": "build-if-missing"
}
```

in `packages/ember-shared/addon/package.json`.

This will do the build the first time you `yarn install` the monorepo.

lukemelia: Ah, that seems much better. Thanks.

lukemelia: @ef4 "prepare" didn't work but "postinstall" does. I did see a bit int he yarn docs warning away from postinstall though, so I thought I'd check with you if you had something else in mind.

PR #3366, opened 2022-10-26
[CS-4656][safe-tools]: Add network switcher UI

danibonilha:
This PR adds the network switcher on the control-panel, it uses mocked network info that should be replaced with sdk info.
I've added a new caret icon, because the css approaches to change the svg color while using it as a background image, doesn't seemed to work for this icon, I think it's because its color comes from stroke property and not fill one. 

https://user-images.githubusercontent.com/20520102/198088483-7e6afee9-b3a5-4235-aa4c-0d3b526f8dda.mov

### DIFF
--- a/packages/safe-tools-client/app/css/schedule.css
+++ b/packages/safe-tools-client/app/css/schedule.css
@@ -59,7 +59,26 @@
 }
 
 .temporary-form-container {
-  display:flex;
-  justify-content:center;
-  margin-top:var(--boxel-sp-xl);
+  display: flex;
+  justify-content: center;
+  margin-top: var(--boxel-sp-xl);
+}
+
+.safe-tools__dashboard-network-selector {
+  display: inline-flex;
+  align-items: center;
+}
+
+.safe-tools__dashboard-network-selector.boxel-select .boxel-select__item {
+  font-weight: 600;
+  font-size: var(--boxel-font-size-sm);
+}
+
+.safe-tools__dashboard-network-selector .ember-power-select-status-icon {
+  background: url('/@cardstack/boxel/images/icons/caret-down-cyan.svg')
+    no-repeat;
+  width: 11px;
+  height: 9px;
+  margin-top: 5px;
+  display: inline-block;
 }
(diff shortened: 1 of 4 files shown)


packages/safe-tools-client/app/css/schedule.css
```diff
@@ -59,7 +59,26 @@
...
+}
+
+.safe-tools__dashboard-network-selector .ember-power-select-status-icon {
+  background: url('/@cardstack/boxel/images/icons/caret-down-cyan.svg')
+    no-repeat;
+  width: 11px;
+  height: 9px;
+  margin-top: 5px;
+  display: inline-block;
+}
```
backspace, line 84: Locally I was able to get this working with unmodified `caret-down` thanks to [this article](https://codepen.io/noahblon/post/coloring-svgs-in-css-background-images):

Cardstack Safe Tools 2022-10-26 12-38-24

```
  background-color: var(--boxel-cyan);
  -webkit-mask-image: url('/@cardstack/boxel/images/icons/caret-down.svg');
  -webkit-mask-repeat: no-repeat;
  mask-image: url('/@cardstack/boxel/images/icons/caret-down.svg');
  mask-repeat: no-repeat;
```

backspace (approved):

packages/safe-tools-client/app/css/schedule.css
danibonilha: 🤦‍♀️  I didn't use the --webkit prefix, I'll try to apply it

backspace: ya… I hate the prefix mess, people sometimes use CSS preprocessing for this but we have yet to enter that realm 😯

lukemelia:
@danibonilha @backspace Is it just me, or do you think the current selection should be one of the options in the dropdown?

danibonilha:
> @danibonilha @backspace Is it just me, or do you think the current selection should be one of the options in the dropdown?

Giving that is not sequential and it's a small set, I think it's ok to not have the current option on the dropdown, since it's already selected anyways, I don't see a reason to re-select the same item. Unless there's some web pattern i'm not familiar with. Initially I've added all the options, and it felt 'weird'. But i'm ok with adding it if we decide to.

This is what looks like with the selected option:
image

PR #3368, opened 2022-10-26
Add boxel-component blueprint

lukemelia:
This blueprint allows a developer to generate a new boxel component (index.gts), a
CSS file, a usage.gts file, and an integration test. Run it using
`ember generate boxel-component your-component-name`.

### DIFF
--- a/packages/boxel/.eslintrc.js
+++ b/packages/boxel/.eslintrc.js
@@ -42,6 +42,7 @@ module.exports = {
         './tests/dummy/config/**/*.js',
         './generate-thumbs.js',
         './lib/**/*.js',
+        './blueprints/*/index.js',
       ],
       parserOptions: {
         sourceType: 'script',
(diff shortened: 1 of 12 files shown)


packages/boxel/.eslintrc.js
```diff
@@ -56,7 +57,7 @@ module.exports = {
         {},
         require('eslint-plugin-node').configs.recommended.rules,
         {
-          // add your custom rules and overrides for node files here
+          'node/no-unpublished-require': 'off',
```
backspace, line 60: Could this be applied more narrowly?

backspace (approved):
🤩

packages/boxel/.eslintrc.js
lukemelia: Probably, but we disable this in our other packages, so the consistency seems like a good move.

danibonilha (approved):
This is awesome 🤩

PR #3374, opened 2022-10-27
[CS-4798]: Add action to disconnect wallet

danibonilha:
Adds the action to disconnect a wallet. I thought it made more sense for it to leave in the walletService, but let me know if the pattern is to keep in the component. The UI also reflects in case we disconnect from Metamask.

https://user-images.githubusercontent.com/20520102/198363594-cf9ca406-b464-4fad-a8a9-7af1dbae7a8c.mov

PS. the close action will be added on a diff PR

### DIFF
--- a/packages/safe-tools-client/app/components/connect-button/index.gts
+++ b/packages/safe-tools-client/app/components/connect-button/index.gts
@@ -1,6 +1,5 @@
 import Component from '@glimmer/component';
 import { inject as service} from '@ember/service';
-import { action } from '@ember/object';
 import BoxelButton from '@cardstack/boxel/components/boxel/button';
 import cn from '@cardstack/boxel/helpers/cn';
 import or from 'ember-truth-helpers/helpers/or';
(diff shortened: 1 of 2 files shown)


backspace (approved):
Removing the action in the component makes sense!

I noticed that after I disconnected I was not able to connect again. I also noticed that I can’t connect with Wallet Connect anymore! But these are likely problems that predate this PR.

danibonilha:
> Removing the action in the component makes sense!
> 
> I noticed that after I disconnected I was not able to connect again. I also noticed that I can’t connect with Wallet Connect anymore! But these are likely problems that predate this PR.

The wallet connect issue, I guess it's related with having the hub running or not, I was also having issues, and I replaced the url with the staging one. Now the reconnect issue I didn't experience it, but let's see if still happens on my next PRs

backspace:
oh yes, that makes sense, I forgot about Hub 😬 thanks

PR #3375, opened 2022-10-27
[CS-4802][safe-tools]: Connect modal features

danibonilha:
This PR

Completes CS-4802: Adds state to track progress on modal
Completes CS-4803: Hides the modal on a successful connection
Completes CS-4804: Disables Connect Wallet button if no provider is selected
Completes CS-4809: Handles cancel action. It adds a isConnecting flag, to track the connection instead of using the taskRunner since the task might still be occurring once it's dependent on the user and not on us.


backspace (approved):
This is great, maybe I got a little too granular with these issues 😆

PR #3376, opened 2022-10-28
Hub CLI command failures are captured to sentry and alert web team

Aierie:
CS-4775

### DIFF
--- a/packages/hub/cli.ts
+++ b/packages/hub/cli.ts
@@ -3,8 +3,17 @@ import { hideBin } from 'yargs/helpers';
 import { printCompilerError } from '@cardstack/core/src/utils/errors';
 import dotenv from 'dotenv';
 import { commands } from './cli/index';
+import * as Sentry from '@sentry/node';
+import config from 'config';
+import initSentry from './initializers/sentry';
 
 dotenv.config();
+
+let sentryEnabled: boolean = config.get('sentry.enabled');
+if (sentryEnabled) {
+  initSentry();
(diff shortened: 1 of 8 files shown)


packages/hub/cli.ts
```diff
@@ -17,9 +26,27 @@ yargs(hideBin(process.argv))
       console.log(yargs.help());
     }
     if (err) {
+      console.log('\n💡 While running:');
+      console.log(yargs.help());
```
Aierie, line 30: I can't find a good way to get information about the failing command in a `yargs.fail` handler, so using `yargs.help` for that, it is a bit verbose but knowing the failing command seems more important. Choosing not to log arguments in case a command ever requires secrets.

packages/hub/cli.ts
```diff
@@ -33,20 +34,19 @@ yargs(hideBin(process.argv))
...
-          process.exit(1);
-        });
-    } else {
-      // eslint-disable-next-line no-process-exit
-      process.exit(1);
+      let sentryClient = Sentry.getCurrentHub()?.getClient();
+
+      if (sentryClient) {
+        // Make sure that we send all exceptions before exiting
+        let result = await sentryClient.close(5000);
```
Aierie, line 41: We're waiting 5 seconds for Sentry to send its exceptions, which feels like it should be enough for most CLI commands

habdelra (approved):

backspace (approved):
This is great 🤩

PR #3382, opened 2022-10-31
[CS-4814]: Refactor network constants and setup test suite on sdk

danibonilha:
The network constants had loose typing and to add a new network we needed to create an object then remember to add to the consts types, constants object, networksIds, networks. In summary there were 4-5 places to be updated. This PR refactors the constants structure, so every-time a new network is added, typescript will alert which fields are required. 
Also:

- Move some recurring constants into their own object to be share across the main and test net 
- Makes some fields required for every network
- Adds chainId for every network 
- Updated `getConstant` and `getConstantbyNetwork` to return correct type of constant based on the key value 
- Add helper to create networkIds object
- Setup test suite on cardpay sdk 
- Adds tests for network constants 
- Fixes tsc error on the sdk
- Removes some constants that were not used on any project like `balanceCheckerContractAddress`
- Adds polygon network

There's still room for improvement like identifying testnets, or making required fields based on the network type if it's layer1 or layer2 etc, or adding new properties like `shortName`.

One breaking change though is that networks object, will return `100: gnosis` instead of previous `100: xdai`.  But I've searched the code and all cases seems to be covered already, so we shouldn't have any issues.

Here you can see, typescript warns us that the new network is missing chainId, name and hubUrl
image

### DIFF
--- a/packages/cardpay-sdk/.eslintrc.cjs
+++ b/packages/cardpay-sdk/.eslintrc.cjs
@@ -21,4 +21,9 @@ module.exports = {
       },
     },
   ],
+  env: {
+    commonjs: true,
+    node: true,
+    mocha: true,
+  },
 };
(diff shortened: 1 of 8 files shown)


habdelra (approved):
:+1:

PR #3384, opened 2022-11-01
[CS-4816]: Add supportedNetworks helpers on sdk

danibonilha:
This PR adds the supported networks and a helper to get the correct config, it also replaces the schedule payment usage. The same function will be used on the safe-tools to match the networks.

### DIFF
--- a/packages/cardpay-sdk/sdk/constants.ts
+++ b/packages/cardpay-sdk/sdk/constants.ts
@@ -2,6 +2,7 @@ import Web3 from 'web3';
 import invert from 'lodash/invert';
 import { networkName } from './utils/general-utils';
 import JsonRpcProvider from '../providers/json-rpc-provider';
+import { pullAll } from 'lodash';
 
 export const ZERO_ADDRESS = '0x0000000000000000000000000000000000000000';
 
@@ -12,6 +13,12 @@ export const CARDWALLET_SCHEME = 'cardwallet';
 type NestedKeyOf<Obj extends object> = {
   [K in keyof Obj]: Obj[K] extends object ? NestedKeyOf<Obj[K]> : K;
 }[keyof Obj];
+
+export const ethereumNetworks = ['mainnet', 'goerli', 'kovan'];
+export const gnosisNetworks = ['gnosis', 'sokol', 'xdai'];
+export const polygonNetworks = ['polygon', 'mumbai'];
+
+export const deprecatedNetworks = ['kovan', 'xdai'];
 
 const testHubUrl = {
   hubUrl: 'https://hub-staging.stack.cards',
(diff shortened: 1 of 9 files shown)


FadhlanR (approved):
LGTM!

packages/cardpay-sdk/sdk/network-config-utils.ts
```diff
@@ -0,0 +1,12 @@
...
+import { HubConfigResponse } from './hub-config';
+
+export const getWeb3ConfigByNetwork = (config: HubConfigResponse, network: string | number) => {
+  const networkName = (typeof network === 'number' ? networkIds[network] : network) as string;
+
+  if (ethereumNetworks.includes(networkName)) return config.web3.ethereum;
+  if (gnosisNetworks.includes(networkName)) return config.web3.gnosis;
+  if (polygonNetworks.includes(networkName)) return config.web3.polygon;
+
+  return { rpcNodeWssUrl: '', rpcNodeHttpsUrl: '' };
```
jurgenwerk, line 11: Would it be better to throw an error here? I imagine it's undesirable to end up with empty config in case the network is unsupported

packages/cardpay-sdk/sdk/constants.ts
```diff
@@ -13,6 +14,12 @@ type NestedKeyOf<Obj extends object> = {
   [K in keyof Obj]: Obj[K] extends object ? NestedKeyOf<Obj[K]> : K;
 }[keyof Obj];
 
+export const ethereumNetworks = ['mainnet', 'goerli', 'kovan'];
+export const gnosisNetworks = ['gnosis', 'sokol', 'xdai'];
```
jurgenwerk, line 18: Gnosis is the same as xdai (xdai renamed to gnosis). I would remove it from the array so that we don't confuse anyone thinking these are two separate networks. Maybe a better approach is to rename the name from "xdai" to "gnosis" before it enters any mapping/filtering functions?

packages/cardpay-sdk/sdk/constants.ts
```diff
@@ -13,6 +14,12 @@ type NestedKeyOf<Obj extends object> = {
...
+export const polygonNetworks = ['polygon', 'mumbai'];
+
+export const deprecatedNetworks = ['kovan', 'xdai'];
```
jurgenwerk, line 21: Can we maybe remove this and let the `getWeb3ConfigByNetwork` throw an error in case we want to get config for an unsupported network (i.e kovan (and I think soon Sokol as well))?

packages/cardpay-sdk/sdk/network-config-utils.ts
danibonilha: Sure, that would be the next step, my goal first was just to use the same logic between both tools, but I guess I can already handle as an error here

packages/cardpay-sdk/sdk/constants.ts
danibonilha: The main dependency today is the wallet, so I'll migrate the wallet to reference only gnosis first, then I'll circle back to this file and remove all xdai reference

jurgenwerk (approved):
Looking good!

PR #3385, opened 2022-11-02
[CS-4801] Initial testing strategy for scheduled payments

alex-cardstack:
CS-4801

This needs a bit of cleanup but I think this is a good strategy so I'm opening a draft PR now in order to get some feedback before writing some more tests.

I spent some time messing with a strategy for injecting a custom mock, but then I found a really useful npm module [eth-testing](https://github.com/VGLoic/eth-testing) which supports mocking out metamask and walletconnect, and allows convenient high-level mocking along with low-level RPC interception, I think it should be able to meet all the needs we have for testing the scheduled payments frontend.


lukemelia (commented):
This looks pretty useful. I noticed in the eth-testing README, they are calling `testingUtils.clearAllMocks()` after each test. Perhaps we should do that as well?

alex-cardstack:
Ah well spotted, I added that to a cleanup hook

packages/safe-tools-client/app/utils/chain-connection-manager.ts
```diff
@@ -232,6 +243,12 @@ export class ChainConnectionManager {
 }
 
 class ConcreteStrategyFactory implements ConnectionManagerStrategyFactory {
+  mockProvider?: TestWalletConnectProvider;
+
+  setMockProvider(mockProvider: TestWalletConnectProvider) {
+    this.mockProvider = mockProvider;
+  }
+
```
danibonilha, line 251: I find it a bit odd to have "test" code inside the production code, is it not possible to mock the walletConnect and/or metamask providers ? Or maybe we could expose the web3 on wallet service so we could set a fake provider on each tests ?

danibonilha: Or if a package like this one would help https://github.com/rsksmart/mock-web3-provider

packages/safe-tools-client/app/utils/chain-connection-manager.ts
```diff
@@ -461,23 +493,27 @@ class WalletConnectConnectionStrategy extends ConnectionStrategy {
...
-        [networkIds[this.networkSymbol]]:
-          hubConfigResponse.web3.ethereum.rpcNodeWssUrl,
-      },
-      // based on https://github.com/WalletConnect/walletconnect-monorepo/blob/7aa9a7213e15489fa939e2e020c7102c63efd9c4/packages/providers/web3-provider/src/index.ts#L47-L52
-      connector: new CustomStorageWalletConnect(connectorOptions, chainId),
-    });
+    let provider;
+    if (this.mockProvider) {
+      provider = this.mockProvider as TestWalletConnectProvider;
+    } else {
```
danibonilha, line 499: Maybe we could take leverage of the Environments, and if it's the test env we skip the walletConnect one

packages/safe-tools-client/app/utils/chain-connection-manager.ts
alex-cardstack: So, the `mockProvider` in this spinnet is already a package very similar to the the one you linked - that one is cypress specific but the one I'm using is more general, it's called [eth-testing](https://github.com/VGLoic/eth-testing)

The issue here is injecting that mock provider into the right place so that it's used.

Background: This file is quite convoluted already, it makes use of a strategy factory pattern so it's quite hard to actually inject it into the right place, I think this code has been inherited and copypasted and could be cleaned up a bit.

The approach I have landed on is a medium between the first really hacky solution (put the mockProvider on the `window` object), and a really complex one I made that had a whole new strategy factory and ended up adding a lot of bloated code.

Then I noticed that this file is already switching based on the ember environment to inject a mock: https://github.com/cardstack/cardstack/blob/8b09163a0fbcceb56f277af216acc0929defdd73/packages/safe-tools-client/app/utils/chain-connection-manager.ts#L96-L99

So I figured the least confusing thing would be to follow the conventions for the file so at least there weren't two different non-ideal ways of doing it in the same file :)

I can definitely make this a lot better by restructuring the connection manager to simplify it and make it easier to inject mocked providers into it, along with just being easier to follow in general - I'm happy to do that and I think it would be a better result, the only question is if that is a yak that I should shave right now or not.

alex-cardstack: Btw I think that probably the more idiomatic ember solution here, for both places where the mocks are used, are to use services - a localStorage service and an ethereumProvider service, and then in the tests inject mock versions of those using the ember testing tooling. That is what I would do if writing from scratch I think.

alex-cardstack:
@danibonilha @lukemelia I refactored and I think there's a better implementation using dependency injection now, I also removed the ConnectionStrategyFactory and made it a simple method instead to reduce complexity

alex-cardstack:
@danibonilha I merged with your upstream changes, I think this should be good to go

danibonilha (approved):
it looks good to me, nice cleanup and typings 🔥 , just one small comment on the chainChainged event. But giving my knowledge of ember maybe it's worth a double check by @lukemelia or @backspace.

packages/safe-tools-client/tests/acceptance/wallet-connection-test.ts
```diff
@@ -0,0 +1,54 @@
...
+      assert.dom(
+        '.boxel-radio-option__input boxel-radio-option__input--hidden-radio boxel-radio-option__input--checked'
+      );
+
+      await click('[data-test-mainnet-connect-button]');
+
+      this.mockWalletConnect.mockConnectedWallet([FAKE_WALLET_CONNECT_ACCOUNT]);
+      this.mockWalletConnect.mockAccountsChanged([FAKE_WALLET_CONNECT_ACCOUNT]);
+
+      await click('.network-connect-modal__close-button'); // FIXME: I don't think this click should be necessary
```
backspace, line 48: Agreed, I just didn’t want to get too deep into adding functionality like this until we could have tests around it, so thanks!

packages/safe-tools-client/app/utils/events.ts
```diff
@@ -1,9 +1,9 @@
-/* eslint-disable @typescript-eslint/ban-types */
+import { VoidCallback } from './types';
```
backspace, line 1: 🎉

backspace (approved):

PR #3388, opened 2022-11-03
subgraph: Index safes for safe tools by adding a new subgraph

jurgenwerk:
This new package is a subgraph that will support our new safe tools project, starting with the scheduled payments tool.

### Why we need this

We need a way to get the following data for an authenticated user in the safe tools client:
1. safe address
2. scheduled payment module address

These two addresses are (among other data) necessary to register a scheduled payment. Normally, the user would first provision a safe + sp module within the tool's interface. Then, one of the ways of getting this info is by indexing the [ScheduledPaymentModuleSetup](https://github.com/cardstack/cardstack-module-scheduled-payment/blob/main/contracts/ScheduledPaymentModule.sol#L17) event which is emitted after the scheduled payment module is added to the safe. 

I already deployed this to my personal [thegraph.com](https://thegraph.com) account and I was able to test it by manually creating safes using `yarn cardpay scheduled-payment create-safe` command in goerli and mumbai networks. Shortly after confirmed transaction (around 10s later) the safe appeared in the subgraph query result in the following form:

```
{
  "data": {
    "safes": [
      ...,
      {
        "id": "0xd0b5a6e7fb6c515e75918d29e015d6bcba74d919", // safe address
        "owner": "0xf4e4c6bfd83e4d0e7a5627b631800487f4a154fd", // user's EOA
        "spModule": "0x0eedecf526f2efb6b44637fd9feb86faaa51700b" // scheduled payment module address
      }
    ]
  }
}
```

You can try this yourself [in the subgraph GraphQL query pad](https://api.thegraph.com/subgraphs/name/jurgenwerk/cardstack-safe-tools-mumbai/graphql?query=%7B%0A++safes+%7B%0A++++id%0A++++owner%0A++++spModule%0A++%7D%0A%7D) (the link already includes the query). 

The sync time feels fast enough but we should test this more, also on mainnet, and see if it continues to be good.

### Why are we not including this in the existing card pay subgraph

Because we want to:

1. Set a new direction for writing subgraph mappings where we only use the event params and not make any eth calls to other contracts (which is currently used in cardpay subgraph mappings). This will make our subgraph code more future proof and ready for streaming features like [Firehose](https://forum.thegraph.com/t/introducing-the-firehose/2329)
2. Try using a hosted version to see if it's a viable option (performant enough) to reduce the maintenance burden of hosting it ourselves

### How will this be used

By the SDK - there will be a function to pull safe + sp module address for the authenticated user. 

### TODO (help me answer these questions)

- Would be good to use Cardstack's account in [thegraph.com](thegraph.com) to deploy this (I'll need admin access to Cardstack github repo, or somebody else should create the account and pass the access keys). I'm also fine if this stays in my personal account for a while
- What happens if the safe owner transfers the ownership to somebody else? I suppose we should listen for that event and modify the safe's ownership in the safe entity. I'm currently researching how to do that but if you have a hint that would be great.


packages/safe-tools-subgraph/networks.json
```diff
@@ -0,0 +1,12 @@
+{
+  "goerli": {
+    "ScheduledPaymentModule": {
+      "startBlock": 7870552
```
jurgenwerk, line 4: This is when https://github.com/cardstack/cardstack-module-scheduled-payment/commit/7d285c2bebdd942e94cecc5e1fee4dbce89c9029 was deployed

packages/safe-tools-subgraph/networks.json
```diff
@@ -0,0 +1,12 @@
...
+    }
+  },
+  "mumbai": {
+    "ScheduledPaymentModule": {
+      "startBlock": 28111696
```
jurgenwerk, line 9: This is when https://github.com/cardstack/cardstack-module-scheduled-payment/commit/7d285c2bebdd942e94cecc5e1fee4dbce89c9029 was deployed

habdelra:
yeah i think it's a good idea to use cardstack's account for the graph. i'll grant you admin access

habdelra:
I'd suggest using the `packages/cardpay-subgraph/tsconfig.json`, the `packages/cardpay-subgraph/eslint.rc` and the `packages/cardpay-subgraph/src/global.d.ts` in this project so you can get the typechecks and linting to work.

packages/safe-tools-subgraph/src/mapping.ts
```diff
@@ -0,0 +1,10 @@
+/* eslint-disable node/no-unsupported-features/es-syntax */
+import { ScheduledPaymentModuleSetup } from '../generated/ScheduledPaymentModule/ScheduledPaymentModule';
+import { Safe } from '../generated/schema';
+
+export function handleScheduledPaymentModuleSetup(event: ScheduledPaymentModuleSetup): void {
+  let safe = new Safe(event.params.owner.toHex());
+  safe.owner = event.transaction.from;
```
habdelra, line 7: the subgraph doesn't have an understanding of the checksum address format, so you might get trolled by this when querying the subgraph. to deal with this I made a util that can perform this for you, this way the addresses in the subgraph are normalized. Here's the util: https://github.com/cardstack/cardstack/blob/main/packages/cardpay-subgraph/src/utils.ts#L228-L245

packages/safe-tools-subgraph/src/mapping.ts
```diff
@@ -0,0 +1,10 @@
...
+  safe.spModule = event.params.moduleAddress;
```
habdelra, line 8: ditto on the safe address

packages/safe-tools-subgraph/schema.graphql
```diff
@@ -0,0 +1,5 @@
+type Safe @entity {
+  id: ID!
+  owner: Bytes! # address
+  spModule: Bytes! # address
```
habdelra, line 4: same as above

packages/safe-tools-subgraph/build/ScheduledPaymentModule/abis/ScheduledPaymentModule.json
```diff
@@ -0,0 +1,1080 @@
+[
```
habdelra, line 1: let's not commit build artifacts

packages/safe-tools-subgraph/package.json
```diff
@@ -0,0 +1,13 @@
+{
+  "name": "safe-tools-subgraph",
+  "scripts": {
+    "codegen": "graph codegen",
+    "build": "graph build",
+    "deploy:goerli": "graph deploy --product hosted-service jurgenwerk/cardstack-safe-tools-goerli",
+    "deploy:mumbai": "graph deploy --product hosted-service jurgenwerk/cardstack-safe-tools-mumbai"
```
habdelra, line 7: might want a "create" script too

packages/safe-tools-subgraph/schema.graphql
```diff
@@ -0,0 +1,5 @@
+type Safe @entity {
+  id: ID!
+  owner: Bytes! # address
```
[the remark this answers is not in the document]
habdelra: Oh also a safe can actually have mulitple owners, and a single own can own multiple safes, so probably you want a "join table" like we did in the cardpay subgraph

habdelra:
Also, to answer your other question, yes, we definitly need to listen for safe ownership transfers in order to update the subgraph. More precisely, it's supporting the `AddedOwner` and `RemovedOwner` safe events. That's something that we were doing in the cardpay-subgraph here for this very reason https://github.com/cardstack/cardstack/blob/195a1d264f03fc75fd16b866280f68992939d7de/packages/cardpay-subgraph/src/mappings/gnosis-safe.ts#L19-L72

habdelra:
I mentioned it in the comments, but its worth calling out here separately the relationship of owners to safes is a many to many relationship, which can be a bit tricky to get right in a document database like the graph. usually the most straight forward way to model this is via a 3rd entity that acts like a "join table". here is a bit of documentation around this: https://thegraph.com/docs/en/developing/creating-a-subgraph/#many-to-many-relationships.

You can see this in our cardpay subgraph for safes <-> accounts in the `SafeOwner` entity here. This allows us to effectively map a many to many relationship between `Account` and `Safe`

habdelra:
One thing I found that is really helpful is to actually capture the events as their own entities--so like an `OwnerChange` entity in the code example above for handling adding and removing owners. I did this because it helps for debugging and auditing, so that if I need to know _why_ the safe entity has a particular state, i can review all the events that led up to that state, as well as trace back the events to a transaction hash.

packages/safe-tools-subgraph/build/ScheduledPaymentModule/abis/ScheduledPaymentModule.json
habdelra: (you can use `git rm --cached` to remove files that are being ignored)

packages/safe-tools-subgraph/src/mapping.ts
jurgenwerk: Thanks for describing the issue! I ported over the checksum code.

packages/safe-tools-subgraph/schema.graphql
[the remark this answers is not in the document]
jurgenwerk: Added the join entity.

packages/safe-tools-subgraph/package.json
jurgenwerk: Added.

jurgenwerk:
@habdelra thanks for a great review. I rewrote the code to include the join relationship (SafeOwner) and added handling of the added and removed owners. 

I also deployed this using Cardstack's account to Goerli network: https://thegraph.com/hosted-service/subgraph/cardstack/safe-tools-goerli

I tested this by creating the safe + sp module using the CLI. Seconds later, the safe appeared in the query:

image

I also tested adding and removing owners. I did that in the Gnosis Safe dApp. I confirmed the ownership changes were tracked succesfully:

image

scripts/typecheck-all.sh
```diff
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-for name in $(find packages -name node_modules -prune -o -name 'cardpay-subgraph' -prune -o -name 'tsconfig.json' -print); do
+for name in $(find packages -name node_modules -prune -o -name '*-subgraph' -prune -o -name 'tsconfig.json' -print); do
```
jurgenwerk, line 3: Was getting a lot of typecheck errors (can bee seen here - https://github.com/cardstack/cardstack/actions/runs/3436424201/jobs/5729924858)  so I followed this pattern where we skip the checking for subgraph related packages.

habdelra (approved):

PR #3390, opened 2022-11-04
hub: Add data integrity checks for scheduled payments

jurgenwerk:
Ticket: CS-4383

This PR introduces the first batch of checks that will be run periodically (by Checkly, which will send alerts to our Discord) and will tell us if there are problematic scheduled payments that need manual inspection or troubleshooting. 

Here are 7 checks that tell us:

1. Are there scheduled payments without creation transaction hash? It means something went wrong with the SDK call to create a scheduled payment.
2. Are there scheduled payments without creation block number? It means the worker that waits for the mined creation transaction isn't functioning or the transaction is stuck 
3. Same as (1) but for cancelation
4. Same as (2) but for cancelation
5. Are there scheduled payments that should be attempted but they weren't? It could indicate the executor is not picking them up
6. Are there scheduled payments without tx hash? It means something went wrong when trying to submit the transaction
7. Are there attempts that are stuck? It could mean the gas is too low to compete or something else went wrong with it - needs inspection.

The next task which will come in a separate PR will be to introduce a data integrity checks route which will respond with a result of these checks. Then we can configure Checkly to periodically query `data-integrity-checks/scheduled-payments` and send an alert if the status is degraded.

### DIFF
--- a/packages/hub/main.ts
+++ b/packages/hub/main.ts
@@ -97,6 +97,7 @@ import ScheduledPaymentSerializer from './services/serializers/scheduled-payment
 import ScheduledPaymentValidator from './services/validators/scheduled-payment';
 import ScheduledPaymentsRoute from './routes/scheduled-payments';
 import ScheduledPaymentsExecutorService from './services/scheduled-payments/executor';
+import DataIntegrityChecksScheduledPayments from './services/data-integrity-checks/scheduled-payments';
 
 //@ts-ignore polyfilling fetch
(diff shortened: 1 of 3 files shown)


lukemelia (approved):
LGTM. Is the code to trigger this to run already written or coming in a separate PR?

jurgenwerk:
@lukemelia thanks! It's gonna come in a separate PR so it's easier to review.

PR #3392, opened 2022-11-04
[CS-4824]: Adds support to different chains

danibonilha:
This PR removes the hardcoded ethereum setup and networks.

- Adds a list of supported network from sdk
- Adds alert in case wallet's user is  on unsupported network and disconnects wallet
- Updates the UI to reflect wallet's network

This PR DOES NOT cover: 
- Switching the network from the dapp.
- Conditional rendering networks by environment (right now all supported nets are in the dropdown)

The network selector starts with gnosis chain, and once a ethereum wallet connects, the value is updated:

https://user-images.githubusercontent.com/20520102/199979877-757505df-5efb-481a-9638-8b970312ce26.mov

### DIFF
--- a/packages/cardpay-sdk/sdk/network-config-utils.ts
+++ b/packages/cardpay-sdk/sdk/network-config-utils.ts
@@ -1,7 +1,7 @@
-import { supportedChains, Networks, networks, supportedChainsArray } from './constants';
+import { supportedChains, Network, networks, supportedChainsArray } from './constants';
 import { HubConfigResponse } from './hub-config';
 
-type Networkish = string | number | Networks;
+type Networkish = string | number | Network;
 
 const convertChainIdToName = (network: Networkish) => (typeof network === 'number' ? networks[network] : network);
 
--- a/packages/cardpay-sdk/sdk/network-config-utils.ts
+++ b/packages/cardpay-sdk/sdk/network-config-utils.ts
@@ -1,4 +1,4 @@
-import { supportedChains, Networks, networks } from './constants';
+import { supportedChains, Networks, networks, supportedChainsArray } from './constants';
 import { HubConfigResponse } from './hub-config';
 
 type Networkish = string | number | Networks;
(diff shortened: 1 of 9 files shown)


packages/safe-tools-client/app/services/network.ts
```diff
@@ -0,0 +1,48 @@
...
+  networks,
+  supportedChainsArray,
+} from '@cardstack/cardpay-sdk';
+
+const DEFAULT_NETWORK = 'mainnet' as const; // TODO: add default based on env
+
+export default class Network extends Service {
+  @tracked chainId = getConstantByNetwork('chainId', DEFAULT_NETWORK);
+  @tracked name = getConstantByNetwork('name', DEFAULT_NETWORK);
+  @tracked symbol: Networks = DEFAULT_NETWORK;
```
lukemelia, line 18: If these three values always change together, perhaps it would be better to have a single tracked `Network`-type value and convenience getters if desired.

packages/safe-tools-client/app/services/network.ts
```diff
@@ -0,0 +1,48 @@
...
+import {
+  getConstantByNetwork,
+  Networks,
+  networks,
+  supportedChainsArray,
+} from '@cardstack/cardpay-sdk';
+
+const DEFAULT_NETWORK = 'mainnet' as const; // TODO: add default based on env
+
+export default class Network extends Service {
```
lukemelia, line 15: I suggest naming this class `NetworkService` to avoid conflicts with the Network type. I think that confusion has already crept in -- e.g. the argument type for the `onSelect` method below.

packages/cardpay-sdk/sdk/network-config-utils.ts
```diff
@@ -1,4 +1,4 @@
-import { supportedChains, Networks, networks } from './constants';
+import { supportedChains, Networks, networks, supportedChainsArray } from './constants';
 import { HubConfigResponse } from './hub-config';
 
 type Networkish = string | number | Networks;
```
lukemelia, line 4: Maybe outside the scope of this PR, but I find it weird that this type is plural.

lukemelia (changes requested):
Looks good overall. I left a few comments for you to consider before merging.

packages/cardpay-sdk/sdk/network-config-utils.ts
danibonilha: Agreed, I'll add a ticket to rename that

packages/safe-tools-client/app/services/network.ts
danibonilha: Ya, my first line of thought was that this.network is already the common denominator and that's why I was using Network as a type, but I'm ok with renaming it

PR #3394, opened 2022-11-07
Set default rollover_amount to 0

tintinthong:


IanCal (changes requested):
Change makes sense - the original implementation in an earlier commit looked right to be but I guess wasn't given the change - can you add a rollover test for this? It seems like the more obvious solution is subtly incorrect, so would be good to avoid someone changing it in the future.

IanCal (approved):

PR #3396, opened 2022-11-07
subgraph: Add missing param in the CreatePrepaidCard handler

jurgenwerk:
Adds a missing parameter in the event signature. The event that gets emitted is here: https://github.com/cardstack/card-pay-protocol/blame/f98e599c005747e8c1311a9af2bb4e278fcc9dc7/contracts/PrepaidCardManager.sol#L55

Will probably need to regenerate the ABI before the deploy.


jurgenwerk:
Not worth fixing this. We are sunsetting the feature.

PR #3397, opened 2022-11-07
Improve typing of dynamic method lookup for keyboard handlers

lukemelia:
Thanks @alex-cardstack for the assist!

### DIFF
--- a/packages/boxel/addon/components/boxel/input/time/keyboard.ts
+++ b/packages/boxel/addon/components/boxel/input/time/keyboard.ts
@@ -63,11 +63,8 @@ abstract class BaseKeyboardHandler implements KeyboardHandler {
 
     const isValidMethodKey = (
       methodName: string
-    ): methodName is ValidStringKey => {
-      return getMethodNames(this as Record<string, unknown>).includes(
-        methodName
-      );
-    };
+    ): methodName is ValidStringKey =>
+      getMethodNames(this).includes(methodName);
 
     let handlerMethodName = `on${eventCode}`;
     if (isValidMethodKey(handlerMethodName)) {
--- a/packages/boxel/addon/components/boxel/input/time/keyboard.ts
+++ b/packages/boxel/addon/components/boxel/input/time/keyboard.ts
@@ -1,3 +1,4 @@
+import { getMethodNames } from '@cardstack/boxel/utils/reflection';
 import { DropdownAPI } from '../../dropdown';
 
 export type KeyboardFocus = 'none' | 'hour' | 'minute' | 'meridian';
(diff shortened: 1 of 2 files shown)


lukemelia:
@alex-cardstack our work type-checked but didn't actually work because `Object.keys` does not include keys on prototypes. I pushed a commit to address this but couldn't figure out how to do it without one unfortunate cast to `Record<string, unknown>`. Maybe you have an idea?

packages/boxel/addon/components/boxel/input/time/keyboard.ts
```diff
@@ -59,6 +58,27 @@ abstract class BaseKeyboardHandler implements KeyboardHandler {
     this.component.setMeridian('pm');
   }
 
+  private handlerMethodFromEventCode(eventCode: string) {
+    type ValidStringKey = Extract<keyof this, string>;
+
+    const isValidMethodKey = (
+      methodName: string
+    ): methodName is ValidStringKey => {
+      return getMethodNames(this as Record<string, unknown>).includes(
```
danibonilha, line 67: nit: you could omit the return keyword

packages/boxel/addon/utils/reflection.ts
```diff
@@ -0,0 +1,20 @@
...
+    .concat(deepFunctions(Object.getPrototypeOf(obj)) || []);
+};
+
+const distinctDeepFunctions = (obj: Record<string, unknown>) =>
+  Array.from(new Set(deepFunctions(obj)));
+
+export const getMethodNames = (obj: Record<string, unknown>): string[] =>
+  distinctDeepFunctions(obj).filter(
+    (name: string) => name !== 'constructor' && !~name.indexOf('__')
+  );
```
danibonilha, line 20: you could use a generic type here, then you don't need to cast it: 
```ts
export const getMethodNames = <Obj extends Record<keyof Obj, unknown>>(
  obj: Obj
): string[] =>
```

packages/boxel/addon/utils/reflection.ts
```diff
@@ -0,0 +1,20 @@
+const isGetter = (obj: unknown, name: string): boolean =>
+  !!(Object.getOwnPropertyDescriptor(obj, name) || {}).get;
+
+const isFunction = (obj: Record<string, unknown>, name: string): boolean =>
```
danibonilha, line 4: Maybe we could create a type with  `Record<string, unknown>` to reuse it

packages/boxel/addon/utils/reflection.ts
```diff
@@ -0,0 +1,20 @@
+const isGetter = (obj: unknown, name: string): boolean =>
```
danibonilha, line 1: isn't this object also `Record<string, unknown>` ?

packages/boxel/addon/utils/reflection.ts
lukemelia: Yay, this is the concept I was missing!

danibonilha (approved):
LGTM

PR #3400, opened 2022-11-08
hub: Add data integrity checks route

jurgenwerk:
Ticket: CS-4671

This PR adds a route which responds with the provided data integrity check (currently we only have the scheduled-payments check).

This route has to be protected and since Checkly will consume it, I thought it would be a good idea to repurpose the `checkly.webhookSecret` config which we already have configured for Checkly webhooks. That way when we'll configure Checkly checks in the infra, we can include this query parameter to the call and authorize it.

### DIFF
--- a/packages/hub/routes/data-integrity-checks.ts
+++ b/packages/hub/routes/data-integrity-checks.ts
@@ -19,7 +19,14 @@ export default class DataIntegrityChecksRoute {
   }
 
   async get(ctx: Koa.Context) {
-    if (ctx.query.secret !== config.get('checkly.webhookSecret')) {
+    // The middleware converts header names to lower case
+    let authorization = ctx.headers['x-data-integrity-checks-authorization'] as string;
+
+    // Checkly's webhookSecret is used in checkly-webhook.ts for authorizing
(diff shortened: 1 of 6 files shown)


packages/hub/routes/data-integrity-checks.ts
```diff
@@ -0,0 +1,59 @@
...
+      as: 'dataIntegrityCheckScheduledPayments',
+    }
+  );
+
+  constructor() {
+    autoBind(this);
+  }
+
+  async get(ctx: Koa.Context) {
+    if (ctx.query.secret !== config.get('checkly.webhookSecret')) {
```
lukemelia, line 22: Can we put the secret somewhere besides the URL, like a header for example? I'm concerned about the secret getting logged.

lukemelia (changes requested):
All looks good except secret use. See my comment.

packages/hub/routes/data-integrity-checks.ts
jurgenwerk: Nice catch and I agree. I moved the authorization to a header.

lukemelia (approved):

PR #3403, opened 2022-11-08
Update action environments to Node 16

devenes:
🚨 Node 12 has an end of life on April 30, 2022.

The GitHub Actions workflow gives the following warning while running the action:

> Node.js 12 actions are deprecated. For more information see: github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12.

This PR updates the default runtime to [node16](https://github.blog/changelog/2021-12-10-github-actions-github-hosted-runners-now-run-node-js-16-by-default/), rather then node12.

You can see more here: https://github.com/actions/setup-node/compare/v2.5.1...v3.0.0

This is supported on all Actions Runners v2.285.0 or later.

Signed-off-by: Enes <ahmedenesturan@gmail.com>

### DIFF
--- a/.github/actions/discord-message/action.yml
+++ b/.github/actions/discord-message/action.yml
@@ -13,5 +13,5 @@ inputs:
     required: true
 
 runs:
-  using: 'node12'
+  using: 'node16'
   main: 'index.js'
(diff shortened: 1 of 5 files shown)


devenes:
Hello, any comments? @jurgenwerk

PR #3404, opened 2022-11-08
format scientific amount for explainability

tintinthong:


IanCal (approved):
LGTM - do we also need to do that for the rollover amount?

PR #3413, opened 2022-11-09
Update JSONRpcProvider error handling

FadhlanR:
Ticket: CS-4846

**Problem**

Error when extracting the calculation result of a new safe address. The problem is that the Metamask throws a different error structure than we expected.

Metamask error structure:
`{
    "code": -32603,
    "message": "Internal JSON-RPC error.",
    "data": {
        "code": 3,
        "data": "0x08c379a0000000000000000000000000000000000000000000000000000000000000002000000000000000000000000000000000000000000000000000000000000000149bd4b4946eabc8a64d198b2fc95866208d1acce5000000000000000000000000",
        "message": "execution reverted: �Դ�n�ȦM\u0019�/�Xf �\u001a��"
    }
}`

Expected error structure:
`{
     "code": 3,
    "data": "0x08c379a0000000000000000000000000000000000000000000000000000000000000002000000000000000000000000000000000000000000000000000000000000000149bd4b4946eabc8a64d198b2fc95866208d1acce5000000000000000000000000",
   "message": "execution reverted: �Դ�n�ȦM\u0019�/�Xf �\u001a��"
}`

**Solution**

I updated the way of JSONRpcProvider handles errors. The code will try to find the expected error structure using the `spelunk` function.

### DIFF
--- a/packages/cardpay-sdk/providers/json-rpc-provider.ts
+++ b/packages/cardpay-sdk/providers/json-rpc-provider.ts
@@ -5,13 +5,17 @@ import { Networkish } from '@ethersproject/networks';
 
 const errorGas = ['call', 'estimateGas'];
 
+function isExpectedMessage(message: string) {
+  return message.toLocaleLowerCase().match('reverted') || message.match(/0x.+/);
+}
+
 function spelunk(value: any): null | { message: string; data: string } {
   if (value == null) {
     return null;
   }
 
   // These *are* the droids we're looking for.
-  if (typeof value.message === 'string' && value.message.match('reverted') && isHexString(value.data)) {
+  if (typeof value.message === 'string' && (isExpectedMessage(value.message) || isHexString(value.data))) {
     return { message: value.message, data: value.data };
   }
 
@@ -28,7 +32,11 @@ function spelunk(value: any): null | { message: string; data: string } {
 
   // Might be a JSON string we can further descend...
   if (typeof value === 'string') {
-    return spelunk(JSON.parse(value));
+    try {
+      return spelunk(JSON.parse(value));
+    } catch (e) {
+      return null;
+    }
   }
 
   return null;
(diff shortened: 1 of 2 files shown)


danibonilha (commented):
I've cherry picked your commits into this [dani/safe-tools-sdk-connect-test](https://github.com/cardstack/cardstack/tree/dani/safe-tools-sdk-connect-test), and still experiencing the same errors:
image

This is a wallet connect through  metamask using goerli network

packages/cardpay-sdk/sdk/utils/signing-utils.ts
```diff
@@ -305,7 +305,7 @@ async function signTypedDataWithWeb3(web3: Web3, account: string, data: any): Pr
     let payload = {
       jsonrpc: '2.0',
       method: 'eth_signTypedData',
-      params: [account, data],
+      params: [account, JSON.stringify(data)],
```
danibonilha, line 308: The wallet connect using cli, worked great 👍

habdelra:
@danibonilha it's not _actually_ an error. the idea is that we are using ethereum errors to communicate gas usage. We want to measure gas for a executing a transaction without actually changing the state of the blockchain. in this case the only way we can really do that is to run the txn for real--but we throw at the very end so that the txn that we are running for real doesn't actually change the state of the block chain, and in the error message we encapsulate the amount of gas used up to the point the error was thrown. This particular approach is how the relay server actually measures gas for safe transactions. We are using the same approach here as well. The challenge here is that the error body is actually unique to the RPC server software that emits the error. sadly there is not a standard here. this same kind of testing logic to parse the error body for different RPC servers exists int eh relay server too.

packages/cardpay-sdk/providers/json-rpc-provider.ts
```diff
@@ -5,13 +5,17 @@ import { Networkish } from '@ethersproject/networks';
...
+}
+
 function spelunk(value: any): null | { message: string; data: string } {
   if (value == null) {
     return null;
   }
 
   // These *are* the droids we're looking for.
-  if (typeof value.message === 'string' && value.message.match('reverted') && isHexString(value.data)) {
+  if (typeof value.message === 'string' && (isExpectedMessage(value.message) || isHexString(value.data))) {
```
habdelra, line 18: isn't the problem that you also need to check `value.data.data`--from your description, it sounded like the revert data was nested in an extra `data` object?

habdelra: or wait never mind--i see that spelunk is recursive, so it will eventually reach the right place.

PR #3414, opened 2022-11-09
tools: Restore missing icons

backspace:
This was presumably caused by `ember-svg-jar` version mismatches across packages. Comparing the navigation tab icons in the [preview deployment](https://toolsrestore-icons-cs-4852.safe-tools-preview.cardstack.com/schedule) to those in the [production deployment](https://safe-tools.cardstack.com/schedule) confirms the fix.

Some uses of `aria-hidden` were being swallowed, with the update it became necessary to change them from `=true` to `="true"` which matches the [specification](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-hidden).


danibonilha (approved):
Nice!

PR #3418, opened 2022-11-10
tools: Add Percy

backspace:
This only has [one snapshot](https://percy.io/Cardstack-1/-cardstack-safe-tools-client/builds/22828411) but that seems fine at this point. Having this would have caught the disappeared icons that #3414 restored.


danibonilha (approved):
🔥

PR #3420, opened 2022-11-10
ci: Disable deploying hub-event-listener in staging

pcjun97:


Aierie (commented):
Given that we aren't actively developing this service at the moment and that there isn't a suitable testnet, it seems right to take it down since it's wasting resources and being a distraction in our monitoring channel.

lukemelia (approved):

PR #3421, opened 2022-11-10
[Feat]: Calculate gas estimate and balance on `Create Safe` click

danibonilha:
This PR refactor somethings and adds sdk integration to handle gas estimate.

- Moves setupSafe modal to `CreateSafeButton`, so it handles the presentation logic and information.
- Adds a fix to emit the chainChanged for either number/string
- Adds function to get wallet balance 
- Calculates balance vs gasCost and reflects on the UI.

PS. the sdk error fix has not been merged yet, so it's expect to have an error while trying to  fetch the gasCost

This is an example of a mocked successful balance/gasCost:

https://user-images.githubusercontent.com/20520102/201115899-600f93da-7a1c-43a7-b52e-0dacbc2fc82d.mov

### DIFF
--- a/packages/safe-tools-client/app/components/setup-safe-modal/index.gts
+++ b/packages/safe-tools-client/app/components/setup-safe-modal/index.gts
@@ -1,14 +1,13 @@
 import Component from '@glimmer/component';
 import { on } from '@ember/modifier';
 import not from 'ember-truth-helpers/helpers/not';
-import { tracked } from '@glimmer/tracking';
-import { action } from '@ember/object';
-
+import or from 'ember-truth-helpers/helpers/or';
 
 import { svgJar } from '@cardstack/boxel/utils/svg-jar';
 import cssVar from '@cardstack/boxel/helpers/css-var';
 import BoxelModal from '@cardstack/boxel/components/boxel/modal';
 import BoxelActionContainer from '@cardstack/boxel/components/boxel/action-container'
+import BoxelLoadingIndicator from '@cardstack/boxel/components/boxel/loading-indicator'
 
 import './index.css';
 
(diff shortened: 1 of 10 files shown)


backspace:
Now that there’s test infrastructure, does it seem possible to use it for this? It’s worrisome to me to get deeper into implementations without that 😳

danibonilha:
> Now that there’s test infrastructure, does it seem possible to use it for this? It’s worrisome to me to get deeper into implementations without that 😳

I'd vouch to have at least one flow working e2e with the sdk before spending much time writing a test, as I have a suspicion our connection providers are not working properly and there's still lot's of TODO's on this create safe component, so if it's ok, I'll be adding the tests after finishing the integration

packages/safe-tools-client/app/components/setup-safe-modal/index.gts
```diff
@@ -68,30 +61,36 @@ export default class SetupSafeModal extends Component<Signature> {
...
             <ac.ActionButton
-              @loading={{this.provisioning}}
-              disabled={{this.notEnoughBalance}}
-              {{on 'click' this.createSafe}}
+              @loading={{@provisioning}}
+              disabled={{or this.notEnoughBalance @isLoadingGasInfo}}
+              {{on 'click' @onProvisionClick}}
             >
-              Provision{{if this.provisioning 'ing'}}
+              Provision{{if @provisioning 'ing'}}
```
lukemelia, line 91: ActionChin has an in-progress state that would be suitable to use while the provisioning is occurring.

lukemelia (approved):
One suggestion, otherwise good progress here.

packages/safe-tools-client/app/components/setup-safe-modal/index.gts
danibonilha: wil be applying that on my next pr which will handle the safe provisioning 🙂

PR #3423, opened 2022-11-10
tools: Fix lint error

backspace:
As seen here:
https://github.com/cardstack/cardstack/actions/runs/3440097244/jobs/5738191147


danibonilha:
It seems this issue was fixed on #3426

PR #3424, opened 2022-11-10
[CS-4862]: Fix another metamask error on estimating gas

danibonilha:
After some investigation I noticed the error I was getting from metamask, had one more level of encapsulation and a key called `originalError`. This PR adds a verification on spelunk to go into deeper levels until it finds data that is hex.
I've aded tests to the known error structures as of today, so we can make sure more levels are working.


packages/cardpay-sdk/providers/json-rpc-provider.ts
```diff
@@ -14,8 +17,10 @@ function spelunk(value: any): null | { message: string; data: string } {
     return null;
   }
 
+  const isDataHexString = isHexString(value.data);
+
   // These *are* the droids we're looking for.
-  if (typeof value.message === 'string' && (isExpectedMessage(value.message) || isHexString(value.data))) {
+  if (isDataHexString || (isExpectedMessage(value.message) && isDataHexString)) {
```
danibonilha, line 23: My wondering if we need this OR condition, or if just checking `isExpectedMessage(value.message) && isDataHexString` is enough

FadhlanR: Sokol doesn't throw error data with a hex value. The hex value is included in the error message.

`{
  message: 'Reverted 0x3756e7636b415b81342e52c75eb03c931941c450',
  data: undefined
}`

FadhlanR: I think we could have a condition like this. `isExpectedMessage(value.message) && (isDataHexString || value.data == undefined)`

danibonilha: Done

FadhlanR (approved):
LGTM, thanks

PR #3425, opened 2022-11-10
tools: Add reconnection tests

backspace:
I’m seeing this error periodically 🤔 

tests 2022-11-10 15-43-47


jurgenwerk:
Is this still a problem / something we want to consider finishing?

backspace:
I think it would be good to have tests for reconnection but I can’t follow up on this at the moment 😳

jurgenwerk:
Closing due to inactivity. The wallet connection mechanism matured in the meanwhile so it might not be necessary anymore.

PR #3428, opened 2022-11-11
Validate that user-specified gas token is a supported gas token

FadhlanR:
Ticket: CS-4858

### DIFF
--- a/packages/hub/node-tests/services/scheduled-payments/validator-test.ts
+++ b/packages/hub/node-tests/services/scheduled-payments/validator-test.ts
@@ -1,9 +1,52 @@
 import { ScheduledPayment } from '@prisma/client';
 import { Decimal } from '@prisma/client/runtime';
-import { setupHub } from '../../helpers/server';
+import { registry, setupHub } from '../../helpers/server';
+import { Networkish, TokenDetail } from '@cardstack/cardpay-sdk';
+
+class StubCardpaySDK {
+  getConstantByNetwork(name: string, network: Networkish) {
+    if (!network) {
+      throw new Error(`network can't be null`);
+    }
+
+    switch (name) {
+      case 'scheduledPaymentFeeFixedUSD':
+        return 0.25;
+      case 'scheduledPaymentFeePercentage':
+        return 0.1;
+      default:
+        throw new Error(`unsupported mock cardpay`);
+    }
+  }
+
+  fetchSupportedGasTokens(network: Networkish): TokenDetail[] {
+    if (!network) {
+      throw new Error(`network can't be null`);
+    }
+
+    return [
+      {
+        address: '0x52031d287Bb58E26A379A7Fec2c84acB54f54fe3',
+        name: 'CARD Token',
+        symbol: 'CARD.CPXD',
+        decimals: 18,
+      },
+      {
+        address: '0x26F2319Fbb44772e0ED58fB7c99cf8da59e2b5BE',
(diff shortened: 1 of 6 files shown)


packages/hub/services/validators/scheduled-payment.ts
```diff
@@ -89,23 +89,33 @@ export default class ScheduledPaymentValidator {
...
+        }
 
-    if (scheduledPayment.chainId && isSupportedChain(scheduledPayment.chainId)) {
-      let feeFixedUSD = this.cardpay.getConstantByNetwork('scheduledPaymentFeeFixedUSD', scheduledPayment.chainId) ?? 0;
-      if (Number(scheduledPayment.feeFixedUsd) < feeFixedUSD) {
-        errors.feeFixedUsd.push(`fee USD must be greater than or equal ${feeFixedUSD}`);
-      }
+        let feePercentage =
+          this.cardpay.getConstantByNetwork('scheduledPaymentFeePercentage', scheduledPayment.chainId) ?? 0;
+        let _feePercentage = Number(scheduledPayment.feePercentage);
```
jurgenwerk, line 104: I think introducing a _variable is a bit too noisy and unnecessary here. Can't we just use `feePercentage`?

packages/hub/services/validators/scheduled-payment.ts
```diff
@@ -89,23 +89,33 @@ export default class ScheduledPaymentValidator {
...
-        errors.feePercentage.push(`fee percentage must be between 0 and 1`);
-      } else if (_feePercentage < feePercentage) {
-        errors.feePercentage.push(`fee percentage must be greater than or equal ${feePercentage}`);
+        let supportedGasTokens = await this.cardpay.fetchSupportedGasTokens(scheduledPayment.chainId);
+        if (
+          !supportedGasTokens.find(
+            (supportedGasToken) => supportedGasToken.address === scheduledPayment.gasTokenAddress
+          )
+        ) {
+          errors.gasTokenAddress.push(`gas token is not supported`);
```
jurgenwerk, line 117: Perhaps it would be useful to add which tokens are supported in the message

packages/hub/node-tests/services/scheduled-payments/validator-test.ts
```diff
@@ -1,10 +1,53 @@
...
+        throw new Error(`unsupported mock cardpay`);
+    }
+  }
+
+  fetchSupportedGasTokens(network: Networkish) {
+    if (!network) {
+      throw new Error(`network can't be null`); 
+    }
+
+    return [
```
lukemelia, line 27: The shape of this return value is actually just what is in the `TokenDetail` interface. The example here is the data that the relay server returns but the implementation of fetchSupportedGasTokens returns a subset of it.

lukemelia (changes requested):

packages/hub/services/validators/scheduled-payment.ts
FadhlanR: updated

packages/hub/services/validators/scheduled-payment.ts
FadhlanR: updated

packages/hub/node-tests/services/scheduled-payments/validator-test.ts
FadhlanR: updated

lukemelia (approved):

PR #3432, opened 2022-11-11
[CS-4866]: Fix signing on metamask extension

danibonilha:
Using the metamask extension to estimate the gas doesn't work, the message to sign is never delivered to the wallet,  by taking a look [here](https://github.com/MetaMask/test-dapp/blob/main/src/index.js) it seems that In order to use stringified params, we need to use a new version of `eth_signTypedData`. This PR updates ethersProvider to use v4.

image

No sure though if this impacts other things, I've also tested using WalletConnect, cli  it works.


habdelra (approved):
this is a good catch, all the versions before 4 were experimental

FadhlanR (approved):

PR #3435, opened 2022-11-14
[CS-4751]: Integrate create safe modal with sdk

danibonilha:
This PR adds the ability to create a safe within the setup-safe-modal. Also updates the modal to use  inProgress btn and states.
It partially adds the fetchSafes function, i'm not 100% certain it's in the right place, I've just added it to make sure we only allow one safe to be created and once we have that, the UI is updated with the mocked information for now, until we properly handle the safe info, the same goes for the`initializeSafes` function, which should be replace with a ember-resources and it was added to handle first render and page refresh. Anyways the goal for this PR is to have a safe provisioned which we can check its existence against the subgraph.
it also converts the address toChecksumAddress, to avoid failing queries.

### DIFF
--- a/packages/safe-tools-client/app/components/create-safe-button/index.gts
+++ b/packages/safe-tools-client/app/components/create-safe-button/index.gts
@@ -66,7 +66,9 @@ export default class CreateSafeButton extends Component<Signature> {
 
   @action handleSafeCreation() {
     taskFor(this.scheduledPaymentsSdk.createSafe).perform()
-    .then(() => { 
+    .then(async () => { 
+      // Fetch from sdk or add task result manually ??
+      await this.wallet.fetchSafes();
       this.closeModal();
     }).catch((e) => {
       //TODO: handle error case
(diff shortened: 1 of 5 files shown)


backspace (approved):
I think having `fetchSafes` in the wallet service makes sense! I tried this on the preview deployment and it appeared to create a safe 🎉 although I’m not sure how to confirm that 😆

danibonilha:
> appeared

You can check if the safe was created by querying the safe-tools subgraph, replace `[network]` with the name of the network you created the safe:
https://api.thegraph.com/subgraphs/name/cardstack/safe-tools-[network]/graphql

Here's an [example on mumbai](https://api.thegraph.com/subgraphs/name/cardstack/safe-tools-mumbai/graphql?query=query%28%24account%3A+ID%21%29%7B%0A++account%28id%3A+%24account%29+%7B%0A++++safes+%7B%0A+++%09%09safe+%7B%0A+++%09%09++id%0A++++++++owners+%7B%0A++++++++++id%0A++++++++%7D%0A+++%09%09%7D%0A++++++%0A++++%7D%0A++%7D%0A%7D%0A)

Just replace the account query variable with your wallet address

PR #3436, opened 2022-11-15
Add endpoint for fetching a list of scheduled payments

jurgenwerk:
Ticket: CS-4374

This PR adds an endpoint to fetch all scheduled payments created by the user. It supports filtering by `payAt` which will allow the client to fetch future payments (to list which ones will be executed and when). 

It lacks pagination but we can add this later when we see the need and the clients supports it.

### DIFF
--- a/packages/hub/routes/scheduled-payments.ts
+++ b/packages/hub/routes/scheduled-payments.ts
@@ -51,6 +51,29 @@ export default class ScheduledPaymentsRoute {
     ctx.type = 'application/vnd.api+json';
   }
 
+  async list(ctx: Koa.Context) {
+    if (!ensureLoggedIn(ctx)) {
+      return;
+    }
+
+    let prisma = await this.prismaManager.getClient();
+
+    let minPayAt = new Date((ctx.query['filter[payAt][gt]'] as string) || 0);
+
+    let scheduledPayments = await prisma.scheduledPayment.findMany({
+      where: {
+        userAddress: ctx.state.userAddress,
+        payAt: {
+          gt: minPayAt,
+        },
+      },
+    });
+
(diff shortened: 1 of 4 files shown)


packages/hub/routes/scheduled-payments.ts
```diff
@@ -51,6 +51,29 @@ export default class ScheduledPaymentsRoute {
...
   }
 
+  async list(ctx: Koa.Context) {
+    if (!ensureLoggedIn(ctx)) {
+      return;
+    }
+
+    let prisma = await this.prismaManager.getClient();
+
+    let minPayAt = new Date((ctx.query['filter[payAt][gt]'] as string) || 0);
```
jurgenwerk, line 61: Regarding `filter[payAt][gt]` query param style for filtering by range - I didn't find an existing pattern for this in the code but it seemed like a good way to do it. JSON:API is agnostic about the filter param implementation.

backspace: maybe `pay-at` makes more sense? Since dasherised aligns with the format elsewhere.

backspace: I don’t have strong opinions but would it be more elegant to use a timestamp for the value vs an entire ISO date string? 🤔

backspace (approved):

packages/hub/routes/scheduled-payments.ts
jurgenwerk: Agree with pay-at, will make that change 👍

Regarding using the date string - I prefer that format over timestamps in query params because it's easier to debug and test when inspecting the URL (can quickly identify the date)

PR #3441, opened 2022-11-16
Update estimate function to create a new safe

FadhlanR:
With this approach, we can estimate creating a new safe without requiring a signature.


jurgenwerk (approved):
Nice! This will be a much nicer experience for users.

PR #3443, opened 2022-11-17
Update gas estimation to create a safe

FadhlanR:
I added the calculation to get the gas estimation in wei. The estimate function in SDK returns the estimation of `gasUsed`, but in the create safe modal we want to show the gas estimation in wei. So we need to do this calculation `gaUsed x gasPrice`.


packages/safe-tools-client/app/components/create-safe-button/index.gts
```diff
@@ -36,18 +37,20 @@ export default class CreateSafeButton extends Component<Signature> {
...
     try {
-      const [gasEstimate, nativeTokenBalance] = await Promise.all([
+      const [gasEstimate, nativeTokenBalance, currentGasPrice] = await Promise.all([
         this.scheduledPaymentsSdk.getCreateSafeGasEstimation(),
         this.wallet.fetchNativeTokenBalance(),
+        getCurrentGasPrice(this.wallet.network.chainId)
       ]);
 
+      const gasPrice = BigNumber.from(currentGasPrice?.standard.toString());
+      const gasEstimateInWei = gasEstimate?.mul(gasPrice);
```
danibonilha, line 47: Is there any reason to not return calculated gasPrice estimation from the sdk directly ? I mean, instead of using  `getCurrentGasPrice` and multiplying here, could the `createSafeWithModuleAndGuardEstimation` already return the correct value in wei, so we just match it against the balance ? That being the case, no change on the frontEnd is needed

packages/cardpay-sdk/sdk/utils/conversions.ts
```diff
@@ -8,6 +8,12 @@ import { networkName } from './general-utils';
 import BN from 'bn.js';
 import { BaseProvider } from '@ethersproject/providers';
 
+export interface GasPrice {
+  slow: BN,
+  standard: BN,
+  fast: BN
+}
+
```
danibonilha, line 16: supernit: if all the values have  the same type you could do something like:

```ts
type GasPrice = Record<'slow' | 'standard' | 'fast', BN>;
```

packages/safe-tools-client/app/components/create-safe-button/index.gts
FadhlanR: You're right, updated

packages/cardpay-sdk/sdk/utils/conversions.ts
FadhlanR: updated

jurgenwerk (approved):
Looking good! ⛽️

danibonilha (approved):
🚀

PR #3445, opened 2022-11-18
sdk, cli: Bug fixes

jurgenwerk:
Found two bugs while testing scheduling payments.

### DIFF
--- a/packages/cardpay-sdk/sdk/constants.ts
+++ b/packages/cardpay-sdk/sdk/constants.ts
@@ -204,7 +204,7 @@ export function getConstantByNetwork<K extends ConstantKeys>(name: K, network: N
   }
 
   let value = constants[network as Network][name];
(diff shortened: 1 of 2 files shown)


packages/cardpay-sdk/sdk/constants.ts
```diff
@@ -204,7 +204,7 @@ export function getConstantByNetwork<K extends ConstantKeys>(name: K, network: N
   }
 
   let value = constants[network as Network][name];
-  if (!value) {
+  if (value == null) {
```
jurgenwerk, line 207: This is a fix for when the value is 0.

I discovered this bug when I was requesting `scheduledPaymentFeeFixedUSD` which is specified 0 in goerli, for example. The code threw an error but it should return 0.

FadhlanR (approved):
Thanks!

PR #3447, opened 2022-11-18
[Networks]: Remove xdai, drop sokol/gnosis scheduler support and improve types

danibonilha:
This PR removes the xdai references from sdk, since the wallet has been migrated already. it also lists the correct networks the scheduler tool supports, and improves the constants typing. Now based on the network the constants are listed correctly while using the getByNetwork function, I've also dropped the option to use chainID on this function since we already have a helper to convert that if we want to, and using the chain number makes it harder to type, giving that the types are the name of the networks.


packages/cardpay-sdk/sdk/constants.ts
```diff
@@ -244,16 +235,17 @@ export const networkIds: Record<string, number> = Object.freeze(
...
+  name: K,
+  network: N
+): NetworkConstants<N>[K] {
+  // @ts-expect-error ts doesn't know how to handle indexed type, but function params and return already does that;
+  const value = constants[network as Network][name];
 
-  let value = constants[network as Network][name];
-  if (value == null) {
-    throw new Error(`Don't know about the constant '${name}' for network ${network}`);
+  if (!value && value !== 0) {
```
danibonilha, line 245: I've replaced this condition to explicitly show that we don't want to error on a 'zero' case

lukemelia (approved):
Thanks!

PR #3448, opened 2022-11-21
sdk: Update the crank when canceling a scheduled payment

jurgenwerk:
Ticket: CS-4377

This PR adjusts the process to cancel a scheduled payment: now it includes communication with the crank too. This is needed so that when the user cancels a scheduled payment, the scheduled payment record in the hub is updated - `canceledAt` gets set, and a worker that waits for the on chain transaction is spawned. The worker then updates `cancelationBlockNumber` which lets us know that the transaction has been mined and that the payment is canceled successfully in both the crank and on chain. 

I tested this on Goerli network using the following command:

`yarn cardpay scheduled-payment cancel-payment <SCHEDULED_PAYMENT_ID> --network goerli --ethersMnemonic '...'`

This SDK functionality can be used in the scheduled payment client like so: `await cancelScheduledPayment(scheduledPaymentId)`.

### DIFF
--- a/packages/cardpay-cli/scheduled-payment/cancel.ts
+++ b/packages/cardpay-cli/scheduled-payment/cancel.ts
@@ -1,49 +1,33 @@
 import { Argv } from 'yargs';
-import { getConstant, getSDK } from '@cardstack/cardpay-sdk';
+import { getSDK } from '@cardstack/cardpay-sdk';
 import { getEthereumClients, getConnectionType, NETWORK_OPTION_ANY } from '../utils';
 import { Arguments, CommandModule } from 'yargs';
 
 export default {
-  command: 'cancel <safeAddress> <moduleAddress> <gasTokenAddress> <spHash>',
-  describe: 'Cancel a scheduled payment',
+  command: 'cancel-payment <scheduledPaymentId>',
+  describe:
+    'Cancels a scheduled payment by updating it in the crank and removing the scheduled payment hash from the payment module contract',
   builder(yargs: Argv) {
     return yargs
-      .positional('safeAddress', {
+      .positional('scheduledPaymentId', {
         type: 'string',
-        description: 'The address of the safe that will fund the scheduled payment',
-      })
-      .positional('moduleAddress', {
-        type: 'string',
-        description: 'The address of scheduled payment module',
-      })
-      .positional('gasTokenAddress', {
-        type: 'string',
-        description: 'The address of gas token',
-      })
-      .positional('spHash', {
-        type: 'string',
-        description: 'Keccak hash of the scheduled payment params',
+        description: 'The id of the scheduled payment to be canceled (uuid of the record in the crank)',
       })
       .option('network', NETWORK_OPTION_ANY);
   },
   async handler(args: Arguments) {
-    let { network, safeAddress, moduleAddress, gasTokenAddress, spHash } = args as unknown as {
+    let { network, scheduledPaymentId } = args as unknown as {
       network: string;
-      safeAddress: string;
-      moduleAddress: string;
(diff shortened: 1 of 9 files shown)


packages/cardpay-cli/scheduled-payment/cancel.ts
```diff
@@ -1,49 +1,33 @@
...
-  describe: 'Cancel a scheduled payment',
+  command: 'cancel-payment <scheduledPaymentId>',
+  describe:
+    'Cancels a scheduled payment by updating it in the crank and removing the scheduled payment hash from the payment module contract',
   builder(yargs: Argv) {
     return yargs
-      .positional('safeAddress', {
+      .positional('scheduledPaymentId', {
         type: 'string',
         description: 'The address of the safe that will fund the scheduled payment',
```
FadhlanR, line 14: I think this description is not valid.

jurgenwerk: Fixed!

packages/hub/tasks/scheduled-payment-on-chain-cancelation-waiter.ts
```diff
@@ -0,0 +1,99 @@
...
+      return Sentry.captureException(
+        new Error('Task run for scheduled payment that already has a cancelation block number'),
+        {
+          tags: {
+            action: 'scheduled-payment-task',
+            scheduledPaymentId: scheduledPayment.id,
+          },
+        }
+      );
+    }
```
backspace, line 31: I saw assertions for some of these outcomes but not others, including this, is it worth having these exercised in tests?

packages/hub/tasks/scheduled-payment-on-chain-cancelation-waiter.ts
```diff
@@ -0,0 +1,99 @@
...
+        return; // We don't want to throw an error here because we don't want the task to be retried - the transaction was obviously reverted
+      }
+
+      // At this point, we don't know what error to expect. Let's throw the error so that the task will restart
+      Sentry.captureException(error, {
+        tags: {
+          action: 'scheduled-payment-task',
+          scheduledPaymentId: scheduledPayment.id,
+        },
+      });
```
backspace, line 88: You can make assertions to ensure things are captured to Sentry [like this](https://github.com/cardstack/cardstack/blob/a9e9b09cd2ec0e6ef5de354aae64536335b7e0ea/packages/hub/node-tests/tasks/subscribe-email-test.ts#L83-L90) if you like

packages/hub/routes/scheduled-payments.ts
```diff
@@ -215,6 +232,42 @@ export default class ScheduledPaymentsRoute {
 
     ctx.type = 'application/vnd.api+json';
   }
+
+  async cancel(ctx: Koa.Context) {
```
backspace, line 236: Is this wired up somewhere? I’d expect a corresponding change in `api-router`

jurgenwerk: Oops, forgot to remove this endpoint. We're only using PATCH in this case.

packages/hub/tasks/scheduled-payment-on-chain-cancelation-waiter.ts
jurgenwerk: Nice!

packages/hub/tasks/scheduled-payment-on-chain-cancelation-waiter.ts
jurgenwerk: Thanks - added a couple of test cases.

PR #3450, opened 2022-11-21
Add support for remotely hosted icons to Boxel::Input::SelectableToken

lukemelia:
The token lists we're using include PNG and SVG logo URLs. This change lets us use them rather than requiring a local SVG in all cases.

- Also, add a max height to the token list so that it will scroll.


backspace (approved):

danibonilha (approved):
Nice!

PR #3451, opened 2022-11-21
API adjustments and cleanup to Boxel token select components

lukemelia:
This is making our SelectableToken Typescript interface more consistent with the TokenList standard's naming.


backspace (approved):

backspace:
I see now that there are Glint errors as a result of this in `safe-tools-client` 😯

PR #3452, opened 2022-11-21
[safe-tools-client] Show correct token list based on network

lukemelia:


backspace (approved):
👍🏻 pending lint fixes of course! (why are they doubled 🤔)

jurgenwerk (approved):

PR #3454, opened 2022-11-22
hub: payment attempts endpoint

jurgenwerk:
Ticket: CS-4664

We'll use this endpoint to fetch payment attempts (past payments). 

image

TODO: 
- [x] Filter by status

### DIFF
--- a/packages/hub/routes/scheduled-payment-attempts.ts
+++ b/packages/hub/routes/scheduled-payment-attempts.ts
@@ -2,8 +2,6 @@ import Koa from 'koa';
 import autoBind from 'auto-bind';
 import { ensureLoggedIn } from './utils/auth';
 import { inject } from '@cardstack/di';
-import ScheduledPaymentValidator from '../services/validators/scheduled-payment';
-import ScheduledPaymentSerializer from '../services/serializers/scheduled-payment-serializer';
 import { ScheduledPayment, ScheduledPaymentAttempt } from '@prisma/client';
 
 export type ScheduledPaymentAttemptWithScheduledPayment = ScheduledPaymentAttempt & {
@@ -11,13 +9,6 @@ export type ScheduledPaymentAttemptWithScheduledPayment = ScheduledPaymentAttemp
 };
 
 export default class ScheduledPaymentAttemptsRoute {
-  scheduledPaymentValidator: ScheduledPaymentValidator = inject('scheduled-payment-validator', {
-    as: 'scheduledPaymentValidator',
-  });
-  scheduledPaymentSerializer: ScheduledPaymentSerializer = inject('scheduled-payment-serializer', {
-    as: 'scheduledPaymentSerializer',
-  });
-  web3 = inject('web3-http', { as: 'web3' });
   prismaManager = inject('prisma-manager', { as: 'prismaManager' });
   scheduledPaymentAttemptSerializer = inject('scheduled-payment-attempt-serializer', {
     as: 'scheduledPaymentAttemptSerializer',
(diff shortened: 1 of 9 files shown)


packages/hub/services/serializers/scheduled-payment-attempt-serializer.ts
```diff
@@ -0,0 +1,52 @@
...
+    as: 'scheduledPaymentSerializer',
+  });
+
+  serialize(
+    model: ScheduledPaymentAttemptWithScheduledPayment | ScheduledPaymentAttemptWithScheduledPayment[]
+  ): JSONAPIDocument {
+    if (Array.isArray(model)) {
+      return {
+        data: model.map((model) => this.serialize(model).data),
+        included: uniqBy(model, 'scheduledPayment.id').map((model) => {
```
jurgenwerk, line 18: Include scheduled payment as a related model in JSON:API fashion - https://jsonapi.org/format/#document-compound-documents

packages/hub/routes/scheduled-payment-attempts.ts
```diff
@@ -0,0 +1,102 @@
...
+import ScheduledPaymentSerializer from '../services/serializers/scheduled-payment-serializer';
+import { ScheduledPayment, ScheduledPaymentAttempt, ScheduledPaymentAttemptStatusEnum } from '@prisma/client';
+import { isValidDate } from '../utils/dates';
+
+export type ScheduledPaymentAttemptWithScheduledPayment = ScheduledPaymentAttempt & {
+  scheduledPayment: ScheduledPayment;
+};
+
+export default class ScheduledPaymentAttemptsRoute {
+  scheduledPaymentValidator: ScheduledPaymentValidator = inject('scheduled-payment-validator', {
```
FadhlanR, line 15: We can remove this property

packages/hub/routes/scheduled-payment-attempts.ts
```diff
@@ -0,0 +1,102 @@
...
+    as: 'scheduledPaymentValidator',
+  });
+  scheduledPaymentSerializer: ScheduledPaymentSerializer = inject('scheduled-payment-serializer', {
+    as: 'scheduledPaymentSerializer',
+  });
+  web3 = inject('web3-http', { as: 'web3' });
```
FadhlanR, line 21: This one also

packages/hub/routes/scheduled-payment-attempts.ts
```diff
@@ -0,0 +1,102 @@
...
+
+    let status: ScheduledPaymentAttemptStatusEnum | undefined = undefined;
+    let statusQuery = ctx.query['filter[status]'] as string;
+    if (['failed', 'succeeded', 'inProgress'].includes(statusQuery)) {
+      status = statusQuery as ScheduledPaymentAttemptStatusEnum;
+    }
+
+    let prisma = await this.prismaManager.getClient();
+
+    let attempts = (await prisma.scheduledPaymentAttempt.findMany({
```
FadhlanR, line 53: Do we need pagination?

packages/hub/routes/scheduled-payment-attempts.ts
jurgenwerk: Good catch. Removed.

jurgenwerk:
@FadhlanR I think you asked if we need pagination but your question disappeared (probably because I rebased). My opinion is that we will eventually need it but not very important for now until we get the MVP shipped. It would be a good idea to come up with a general pagination solution that we can reuse in many endpoints. I'll add a ticket for that.

lukemelia (approved):
Nice!

PR #3455, opened 2022-11-22
[CS-4908]: Update polygon/mumbai apiBaseUrl

danibonilha:
https://api-testnet.polygon.io/api doesn't work correctly, this PR replaces this old apiBaseUrl for mumbai/polygon with one from polygonscan

### DIFF
--- a/packages/cardpay-sdk/sdk/constants.ts
+++ b/packages/cardpay-sdk/sdk/constants.ts
@@ -196,7 +196,7 @@ const constants: {
     ...hubUrl,
     ...polygonNativeTokens,
     tokenList: polygonTokenList,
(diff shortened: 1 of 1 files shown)


paulinhapenedo (approved):
Works on the app! :shipit:

lukemelia (approved):

FadhlanR (approved):

danibonilha:
> Would it make sense to include these urls in our monitoring checks (Checkly)?

Not entirely sure, it depends whats the future of the wallet, this is basically to list the EOAs tokens, which is not a requirement, but it's nice to have, if we were to add I think it's ok, but it's not a incident

PR #3457, opened 2022-11-22
Begin acceptance test for scheduling a one-time payment

lukemelia:
While there is still more work to do before we can schedule a payment via the UI, there are enough changes in this PR that I think it is a good time for a review and merge.

Here's the mock and the current status:

Mock:

Payment Total - Information entered

As of this PR:

image


packages/safe-tools-client/app/components/schedule-payment-form-action-card/validator.ts
```diff
@@ -0,0 +1,75 @@
...
+  maxGasFee: 'normal' | 'high' | 'max' | undefined;
+}
+
+export default class SchedulePaymentFormValidator {
+  form: ValidatableForm;
+
+  @tracked isRecipientAddressInvalid = false;
+  @tracked recipientAddressErrorMessage = '';
+  @tracked isPaymentAmountInvalid = false;
+  @tracked paymentTokenErrorMessage = '';
```
backspace, line 31: Are these used anywhere? Maybe this is a copypaste relic 🤔

backspace (approved):
happy to see this 🎉

packages/safe-tools-client/app/components/schedule-payment-form-action-card/validator.ts
lukemelia: It is more aspirational :). Open question at the moment: besides disabling the button, what other "invalid" state do we want to show and when?

backspace: oh yes, having an invalid state without indicating what needs to change can be so frustrating!

PR #3459, opened 2022-11-23
Add task: Check reward roots

tintinthong:
This pr is a cron task that checks for new merkle root submissions and fires a job `process-reward-root` job if it there are new merkle roots. See `process-reward-root` pr first https://github.com/cardstack/cardstack/pull/3480. This task is only suitable for indexing the db from scratch (not realtime updates)

logic (similar to old implementation)
- check newly indexed reward programs from subgraph 
- for each reward program 
  - check if new merkle roots indexed in subgraph

Some points
- there is a `max_index` key that limits amount of roots to index per reward program 
- this task crontab runs every 10 minutes ~so it will be unsuitable for checking reward roots as the user is claiming (and can't go lower than minute). we need to find a more efficient way to index for user claiming but uses same `process-reward-root-job`. https://linear.app/cardstack/issue/CS-5024/implement-checkrewardrootsjob-using-event-listener. I think subscription handler goes back about 10000 (80 minutes??) blocks so we need this subgraph to fill up our dbs still~
- added a filter so that proofs which are already expired are not indexed. But root will be recalled as indexed


IanCal:
Broadly looks good, I think there's a risk if payment cycles are not always incrementing when they're submitted that something gets lost. The most common case for this would be that the root submission failed and we catch up on it later (e.g. 100, 200, 300 all fine, 400 fails due to RPC node issues, 500 fine, 600 fine, then we resubmit 400).

If we order by block number instead as it scans forwards this should always be either incrementing or the same (has to be >= rahter than >). We need to add logic that ignores if it's seen the payment cycle for that program before though, otherwise we can miss some during pagination. 

> This task is only suitable for indexing the db from scratch (not realtime updates)

Based on the ticket and looking at this, I think this is fine for the standard use. There may be a 10 minute delay before things appear but that seems reasonable. Nothing to change just a note for the future when we look back at the PR.

packages/hub/tasks/check-reward-roots.ts
```diff
@@ -0,0 +1,55 @@
...
+        );
+        let files: S3FileInfo[] = r2?.data?.merkleRootSubmissions.map((root) => {
+          return {
+            paymentCycle: root.paymentCycle,
+            rewardProgramId: root.rewardProgram.id,
+          };
+        });
+        for (const file of files) {
+          // Process 1 parquet (rewardProgramId, paymentCycle) file at one time
+          await this.workerClient.addJob('process-reward-root', file, {
```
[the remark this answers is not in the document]
tintinthong: If it fails, it should be present in `select * from graphile_worker.jobs` within the hub database.

tintinthong:
> Based on the ticket and looking at this, I think this is fine for the standard use. There may be a 10 minute delay before things appear but that seems reasonable. Nothing to change just a note for the future when we look back at the PR.

Agreed. I have dropped the ticket for using the subscription. This pr is sufficient to complete the work

packages/hub/tasks/check-reward-roots.ts
```diff
@@ -0,0 +1,57 @@
...
+            blockNumber: root.blockNumber,
+            s3FileInfo: {
+              paymentCycle: root.paymentCycle,
+              rewardProgramId: root.rewardProgram.id,
+            },
+          };
+        });
+        for (const payload of payloads) {
+          // Process 1 parquet (rewardProgramId, paymentCycle) file at one time
+          await this.workerClient.addJob('process-reward-root', payload, {
```
jurgenwerk, line 40: Can you describe what the expected and maximum range of `rewardProgramIds` and `payloads` is at every run?

jurgenwerk: Also, how long does `process-reward-root` task usually take?

tintinthong: For now, 

production has 2 reward programs

every 10 minutes, max 1000 jobs will be executed per reward program. So at a maximum of 2000 jobs in total

using my local computer, the first run 1000+23 jobs will be indexed within 2 minutes of the worker running. In total, 67 proofs are indexed. So pretty sparse. In fact, this also assumes that I do not even check for expiry, if I check for expiry its probably 0 proofs.  

Most of the 1023 records are just storing `reward_root_index` which just keeps the indexer knwoing which reward_root is submitted. 

The speed of process_reward_root will depend on how many rows of rewards are being given. For example,

paymentCycle: 23979900 has 45 proofs (this was staking). The indexing is almost instant.

tintinthong:
Som errors on indexing production data. Will post a fix

packages/hub/tasks/check-reward-roots.ts
jurgenwerk: Sounds like this won't bring our job processing to its knees for extended periods of time. Few more questions:

1. The first run currently takes a minute and then next runs before a new cycle won't have much work to do? How long does a cycle last?
2. The first run will blast the workers with work and it should be done in let's say minute like you benchmarked. I suppose this means any new work tasks that will enter the queue at this time will have to wait until a bunch reward tasks will be done? I think this can potentially be problematic - we have several tasks which we want to be started almost immediately, for example tasks of the "waiter" type, e.g `scheduled-payment-on-chain-creation-waiter`.  These tasks are coupled with feedback to user actions in the interface and we want to avoid having them enqueued for too long - they have to start as soon as possible. So my concern is that in the reward task task spike these other tasks will be delayed for a while. For this reason, have you considered introducing a separate queue for tasks that aren't critical to be performed immediately after enqueuing (I suppose reward tasks can be labelled as such)?

tintinthong: 1. The first run currently takes a minute and then next runs before a new cycle won't have much work to do? How long does a cycle last?

It depends on the rule. But in general we use a rule of thumb for staking. Staking rule runs every month once. Each payment cycle will represent a set of rewards for staking (ie set of reward_proofs).

packages/hub/tasks/check-reward-roots.ts
```diff
@@ -29,17 +29,17 @@ export default class CheckRewardRoots {
...
-            },
+            paymentCycle: root.paymentCycle,
+            rewardProgramId: root.rewardProgram.id,
           };
         });
         for (const payload of payloads) {
           // Process 1 parquet (rewardProgramId, paymentCycle) file at one time
           await this.workerClient.addJob('process-reward-root', payload, {
-            jobKey: payload.s3FileInfo.rewardProgramId + '-' + payload.s3FileInfo.paymentCycle,
+            jobKey: payload.rewardProgramId + '-' + payload.paymentCycle,
```
jurgenwerk, line 39: 👍

jurgenwerk (approved):

packages/hub/tasks/check-reward-roots.ts
tintinthong: Ahh I see you've approved. @jurgenwerk do you know if there is an easy way to test the queue priority

I am not too certain what the expected outcome is in terms of a test when it comes to graphile worker

tintinthong: Ok skipping test for this

PR #3460, opened 2022-11-23
[CS-4918]: Fix supportedChains, tests and cli bridge

danibonilha:
This PR fixes a couple of issues:
- SupportedChains were not returning correctly, `pullAll` method actually mutates the array, so gnosis was being remove, pullAll was replaced with `difference` and tests were added to guarantee.
- Sdk tests were not running on CI check and it turns out there was a couple of `.only` which made it harder to perceive this issue
- WalletConnect was not working on cardpay-cli for gnosis and sokol, since the bridge was outdated, so I've replaced with the public one(it would be better to have this bridge on a constant, but considering we'll be migrating to WC 2.0, not worth it).

### DIFF
--- a/package.json
+++ b/package.json
@@ -84,6 +84,7 @@
     "test:web-client:serve": "cd packages/web-client && ember test --serve --no-launch",
     "test:safe-tools-client": "cd packages/safe-tools-client && ember test",
     "test:safe-tools-client:percy": "cd packages/safe-tools-client && percy exec -- ember test",
+    "test:cardpay-sdk": "yarn --cwd packages/cardpay-sdk test",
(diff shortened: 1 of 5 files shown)


lukemelia (approved):
👍

lukemelia:
@danibonilha does this PR have enough config to ensure that it runs SDK tests on CI?

danibonilha:
> @danibonilha does this PR have enough config to ensure that it runs SDK tests on CI?

@lukemelia good catch, one commit to be pushed was missing 😅

FadhlanR (approved):
Looks good

PR #3461, opened 2022-11-28
Returns the range of gas prices in USD

FadhlanR:
In the payment scheduling form, we need to show the gas prices range in USD. I updated the `estimateGas` function in the scheduled payment module SDK to include the gas prices in USD by converting gas prices in Wei to USD.

### DIFF
--- a/packages/cardpay-cli/scheduled-payment/estimate-gas.ts
+++ b/packages/cardpay-cli/scheduled-payment/estimate-gas.ts
@@ -43,8 +43,11 @@ export default {
     );
 
     console.log(`Required gas: ${result.gas}`);
-    console.log(`Required gas in eth (slow): ${fromWei(result.gasRangeInWei.slow.toString(), 'ether')}`);
-    console.log(`Required gas in eth (standard): ${fromWei(result.gasRangeInWei.standard.toString(), 'ether')}`);
-    console.log(`Required gas in eth (fast): ${fromWei(result.gasRangeInWei.fast.toString(), 'ether')}`);
+    console.log(`Required gas in ETH (slow): ${fromWei(result.gasRangeInWei.slow.toString(), 'ether')}`);
+    console.log(`Required gas in ETH (standard): ${fromWei(result.gasRangeInWei.standard.toString(), 'ether')}`);
+    console.log(`Required gas in ETH (fast): ${fromWei(result.gasRangeInWei.fast.toString(), 'ether')}`);
+    console.log(`Required gas in USDT (slow): ${fromWei(result.gasRangeInUSD.slow.toString(), 'ether')}`);
+    console.log(`Required gas in USDT (standard): ${fromWei(result.gasRangeInUSD.standard.toString(), 'ether')}`);
+    console.log(`Required gas in USDT (fast): ${fromWei(result.gasRangeInUSD.fast.toString(), 'ether')}`);
   },
 } as CommandModule;
--- a/packages/cardpay-cli/scheduled-payment/estimate-gas.ts
+++ b/packages/cardpay-cli/scheduled-payment/estimate-gas.ts
@@ -46,8 +46,8 @@ export default {
     console.log(`Required gas in ETH (slow): ${fromWei(result.gasRangeInWei.slow.toString(), 'ether')}`);
     console.log(`Required gas in ETH (standard): ${fromWei(result.gasRangeInWei.standard.toString(), 'ether')}`);
     console.log(`Required gas in ETH (fast): ${fromWei(result.gasRangeInWei.fast.toString(), 'ether')}`);
-    console.log(`Required gas in USDT (slow): ${fromWei(result.gasRangeInUSD.slow.toString(), 'ether')}`);
-    console.log(`Required gas in USDT (standard): ${fromWei(result.gasRangeInUSD.standard.toString(), 'ether')}`);
-    console.log(`Required gas in USDT (fast): ${fromWei(result.gasRangeInUSD.fast.toString(), 'ether')}`);
+    console.log(`Required gas in USD (slow): ${fromWei(result.gasRangeInUSD.slow.toString(), 'ether')}`);
+    console.log(`Required gas in USD (standard): ${fromWei(result.gasRangeInUSD.standard.toString(), 'ether')}`);
+    console.log(`Required gas in USD (fast): ${fromWei(result.gasRangeInUSD.fast.toString(), 'ether')}`);
   },
 } as CommandModule;
--- a/packages/cardpay-sdk/sdk/scheduled-payment-module.ts
+++ b/packages/cardpay-sdk/sdk/scheduled-payment-module.ts
@@ -46,7 +46,7 @@ import {
 import BN from 'bn.js';
 import { Interface } from 'ethers/lib/utils';
 import JsonRpcProvider from '../providers/json-rpc-provider';
-import { getConstant, getConstantByNetwork, SchedulerCapableNetworks } from './constants';
+import { getConstant, getConstantByNetwork } from './constants';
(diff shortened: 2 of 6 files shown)


packages/cardpay-sdk/sdk/scheduled-payment-module.ts
```diff
@@ -1289,13 +1290,26 @@ export default class ScheduledPaymentModule {
     });
     let gasStationResponse = await getCurrentGasPrice(chainId);
     let gas = BigNumber.from((await gasEstimationResponse.json()).data?.attributes?.gas);
+
+    let tokenList = getConstantByNetwork('tokenList', network as SchedulerCapableNetworks)?.tokens;
+    let tokenUSD = tokenList.find((t) => t.symbol === 'USDT');
```
FadhlanR, line 1295: Which USD token can I use for the conversion?

packages/cardpay-sdk/sdk/utils/conversions.ts
```diff
@@ -63,3 +68,14 @@ export async function getCurrentGasPrice(chainId: number): Promise<GasPrice> {
     fast: BigNumber.from(gasPriceJson.fast),
   };
 }
+
+export async function getNativeWeiInToken(provider: JsonRpcProvider, tokenAddress: string): Promise<BigNumber> {
```
jurgenwerk, line 72: I am confused with the combination of this function's name and its implementation.  The name suggests it converts the native coin value in wei to the specified token, but there is a call inside to get the gas price. These are two different things and I feel this can be simplified or renamed for better clarity.

packages/cardpay-sdk/sdk/scheduled-payment-module.ts
```diff
@@ -1289,13 +1290,26 @@ export default class ScheduledPaymentModule {
     });
     let gasStationResponse = await getCurrentGasPrice(chainId);
```
jurgenwerk, line 1291: I think `gasPricesInNativeWei` would be a more informative name for this variable

packages/hub/node-tests/services/scheduled-payments/executor-test.ts
```diff
@@ -5,6 +5,7 @@ import ScheduledPaymentsExecutorService from '../../../services/scheduled-paymen
 import { setupStubWorkerClient } from '../../helpers/stub-worker-client';
 import BN from 'bn.js';
 import CrankNonceLock from '../../../services/crank-nonce-lock';
+import { BigNumber } from 'ethers';
```
jurgenwerk, line 8: Is adding `BigNumber` needed when we're already using `BN`?

packages/cardpay-sdk/sdk/scheduled-payment-module.ts
jurgenwerk: Is it important that the usd stablecoin is in our token list available for payments? Can we just hardcode it to USDT or USDC? The latter seems more regulated and thus a safer choice but they are both supposed to be pegged so it doesn't really matter, right?

lukemelia:
Overall, looking good. I agree with @jurgenwerk's comments.

Is this API intended to be the thing that powers the Normal/High/Max "max gas price" UI?

Screenshot 2022-11-28 at 8 55 16 AM

Will some transformation need to be applied to go from the API returned values to "max" values?

packages/cardpay-sdk/sdk/scheduled-payment-module.ts
habdelra: yeah I think USDC is probably a safer choice. well they are both stable coins, pegged to USD. but if the peg ever drops (like it did for Terra), then it will matter. so mostly it;s about choosing the token that seems less likely to loose its peg, which is becomes, which token is better collateralized and hence has less risk of becoming insolvent.

habdelra (approved):
Looks good pending Matic's review comments

packages/cardpay-sdk/sdk/utils/conversions.ts
```diff
@@ -63,3 +68,14 @@ export async function getCurrentGasPrice(chainId: number): Promise<GasPrice> {
...
+  let naticeWeiInUSD;
```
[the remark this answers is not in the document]
FadhlanR: updated

packages/cardpay-sdk/sdk/utils/conversions.ts
```diff
@@ -63,3 +68,14 @@ export async function getCurrentGasPrice(chainId: number): Promise<GasPrice> {
...
+  try {
+    naticeWeiInUSD = await gasPriceInToken(provider, tokenAddress, BigNumber.from(1));
+  } catch (e) {
+    naticeWeiInUSD = BigNumber.from(0);
```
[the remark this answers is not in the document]
FadhlanR: updated

packages/cardpay-sdk/sdk/scheduled-payment-module.ts
```diff
@@ -1289,13 +1290,26 @@ export default class ScheduledPaymentModule {
...
     let gas = BigNumber.from((await gasEstimationResponse.json()).data?.attributes?.gas);
+
+    let tokenList = getConstantByNetwork('tokenList', network as SchedulerCapableNetworks)?.tokens;
+    let tokenUSD = tokenList.find((t) => t.symbol === 'USDT');
+    let priceWeiInUSD = BigNumber.from(0);
+    if (tokenUSD) {
```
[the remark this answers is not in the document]

packages/cardpay-sdk/sdk/utils/conversions.ts
FadhlanR: updated

packages/cardpay-sdk/sdk/scheduled-payment-module.ts

packages/hub/node-tests/services/scheduled-payments/executor-test.ts
FadhlanR: reverted this update

packages/cardpay-cli/scheduled-payment/estimate-gas.ts
```diff
@@ -43,8 +43,11 @@ export default {
...
     console.log(`Required gas: ${result.gas}`);
-    console.log(`Required gas in eth (slow): ${fromWei(result.gasRangeInWei.slow.toString(), 'ether')}`);
-    console.log(`Required gas in eth (standard): ${fromWei(result.gasRangeInWei.standard.toString(), 'ether')}`);
-    console.log(`Required gas in eth (fast): ${fromWei(result.gasRangeInWei.fast.toString(), 'ether')}`);
+    console.log(`Required gas in ETH (slow): ${fromWei(result.gasRangeInWei.slow.toString(), 'ether')}`);
+    console.log(`Required gas in ETH (standard): ${fromWei(result.gasRangeInWei.standard.toString(), 'ether')}`);
+    console.log(`Required gas in ETH (fast): ${fromWei(result.gasRangeInWei.fast.toString(), 'ether')}`);
+    console.log(`Required gas in USDT (slow): ${fromWei(result.gasRangeInUSD.slow.toString(), 'ether')}`);
+    console.log(`Required gas in USDT (standard): ${fromWei(result.gasRangeInUSD.standard.toString(), 'ether')}`);
+    console.log(`Required gas in USDT (fast): ${fromWei(result.gasRangeInUSD.fast.toString(), 'ether')}`);
```
lukemelia, line 51:
```suggestion
    console.log(`Required gas in USD (slow): ${fromWei(result.gasRangeInUSD.slow.toString(), 'ether')}`);
    console.log(`Required gas in USD (standard): ${fromWei(result.gasRangeInUSD.standard.toString(), 'ether')}`);
    console.log(`Required gas in USD (fast): ${fromWei(result.gasRangeInUSD.fast.toString(), 'ether')}`);
```

lukemelia (approved):

PR #3462, opened 2022-11-28
safe client: List safe info and token balances in the safe section in the sidebar

jurgenwerk:
Ticket: CS-4770

This adds displaying the safe address and a list of token balances in the sidebar. They show up if the user has at least one safe with scheduled payment module enabled. Normally, users will have one safe with sp module if they go through the setup process, but they can also have multiple if they do it manually using the CLI (and eventually in the Gnosis safe dApp).

**This is how it looks when the user has one safe:**

image

**Multiple safes:**

image

image

**No safes:** 

image

### Balances of which tokens should we list?

1. Currently, we list tokens intended for transferring. Should we also list gas tokens?

Answer: We will list all supported tokens (they are stored as `tokenList` in the SDK)

### CSS

I feel that the classes that we use are sometimes uncomfortably long and confusing.

For example: `safe-tools__dashboard-schedule-control-panel-connect-button`

1. The name implies there exists a `safe-tools` block but there isn't one. Maybe we can drop this prefix?
2. `dashboard` and `control-panel` in the same class are confusing since they mean a very similar thing. Also, I'm wondering what is the dashboard supposed to be - only the side panel? If yes, I think we could reduce the class to something like `side-panel-connect-button`

### DIFF
--- a/packages/cardpay-sdk/sdk/utils/safe-utils.ts
+++ b/packages/cardpay-sdk/sdk/utils/safe-utils.ts
@@ -17,6 +17,7 @@ import { Contract, ethers, utils } from 'ethers';
 import { AddressZero } from '@ethersproject/constants';
 import { Interface, LogDescription } from 'ethers/lib/utils';
 import JsonRpcProvider from '../../providers/json-rpc-provider';
+import { ERC20ABI } from '../..';
 
 export interface EventABI {
   topic: string;
@@ -475,3 +476,19 @@ export function gasInToken(estimate: Estimate): BN {
 // The number is based off empirical observation of sentry errors; the baseGas difference is usually 12
 // https://github.com/cardstack/card-protocol-relay-service/blob/master/safe.py#L303-L316
 export const baseGasBuffer: BN = new BN('30');
+
+export async function getTokenBalancesForSafe(
+  provider: JsonRpcProvider,
+  tokenAddresses: string[],
+  safeAddress: string
+) {
+  return await Promise.all(
+    tokenAddresses.map(async (tokenAddress: string) => {
+      let token = new Contract(tokenAddress, ERC20ABI, provider);
+      let symbol = await token.symbol();
+      let balance = await token.callStatic.balanceOf(safeAddress);
+      let decimals = await token.decimals();
+      return { tokenAddress, symbol, balance, decimals };
+    })
+  );
+}
--- a/packages/safe-tools-client/app/controllers/schedule.ts
+++ b/packages/safe-tools-client/app/controllers/schedule.ts
@@ -1,10 +1,10 @@
 import ApplicationController from '@cardstack/safe-tools-client/controllers/application';
 import NetworkService from '@cardstack/safe-tools-client/services/network';
 import WalletService from '@cardstack/safe-tools-client/services/wallet';
-
+import TokensService from '@cardstack/safe-tools-client/services/tokens';
+import SafesService from '@cardstack/safe-tools-client/services/safes';
 import Controller, { inject as controller } from '@ember/controller';
-import { action } from '@ember/object';
-import { default as Owner } from '@ember/owner';
+
 import { inject as service } from '@ember/service';
 import { tracked } from '@glimmer/tracking';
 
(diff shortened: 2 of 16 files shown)


packages/cardpay-sdk/sdk/utils/safe-utils.ts
```diff
@@ -475,3 +476,19 @@ export function gasInToken(estimate: Estimate): BN {
...
 export const baseGasBuffer: BN = new BN('30');
+
+export async function getTokenBalancesForSafe(
+  provider: JsonRpcProvider,
+  tokenAddresses: string[],
+  safeAddress: string
+) {
+  return await Promise.all(
+    tokenAddresses.map(async (tokenAddress: string) => {
+      let token = new Contract(tokenAddress, ERC20ABI, provider);
```
jurgenwerk, line 487: I just realized we already have this in the SDK: https://github.com/cardstack/cardstack/blob/main/packages/cardpay-sdk/sdk/assets.ts#L27

Will adjust this to use the SDK function.

packages/safe-tools-client/app/controllers/schedule.ts
```diff
@@ -1,79 +1,58 @@
...
-      setTimeout(async () => {
-        if (this.wallet.isConnected) {
-          await this.wallet.fetchSafes();
-        }
-        this.isInitializingSafes = false;
-      }, 500);
-    } catch (e) {
-      console.log('Error initializing safes', e);
-    }
+    return await getTokenBalancesForSafe(
```
jurgenwerk, line 37: Just realized we'll need native token balance too. I'll use this SDK method: https://github.com/cardstack/cardstack/blob/main/packages/cardpay-sdk/sdk/assets.ts#L15

VitaliKirpu:
BTW, what if you have 10+ tokens?

lukemelia:
@jurgenwerk We should not show tokens with zero balance. We would ideally show all the tokens we support for the selected chain that have a non-zero, non-dust balance. You can find these tokens here: https://github.com/cardstack/cardstack/tree/main/packages/cardpay-sdk/token-lists

jurgenwerk:
@VitaliKirpu All tokens with non-negligible value will be listed.

jurgenwerk:
This is now ready for review.

jurgenwerk:
@lukemelia I think it would be nice to show native coin balance (e.g. ETH) even if its value is 0 or negligible. That way there will always be at least one balance shown so that the user gets a hint where the balances are listed.

packages/safe-tools-client/app/services/safes.ts
```diff
@@ -0,0 +1,146 @@
...
+  });
+
+  @use safesResource = resource(() => {
+    // because networkInfo is tracked, anytime the network switches,
+    // this resource will re-run
+    const chainId = this.network.networkInfo.chainId;
+    const address = this.wallet.address;
+
+    const state: SafeResourceState = new TrackedObject({
+      load: async () => {
```
jurgenwerk, line 128: I added this method so that we can reload this resource on demand (needed after provisioning a new safe).

VitaliKirpu:
What i am getting at is that if you have lots of tokens, then the view won't show all tokens properly. Is there a listview built into this? I am not part of the project, so I don't know how the view looks from the full view, but I can envision that becoming a problem if this is like a sidebar UI.

lukemelia:
> @lukemelia I think it would be nice to show native coin balance (e.g. ETH) even if its value is 0 or negligible. That way there will always be at least one balance shown and the user gets a hint where the balances are listed.
Agreed!

packages/safe-tools-client/app/services/safes.ts
```diff
@@ -0,0 +1,146 @@
...
+        if (!address) return;
+
+        state.isLoading = true;
```
lukemelia, line 131: I assume that lexical scope lets reference `state` here. Do you think it is clearer to do it that way or to use `this` within the `load` method?

lukemelia (approved):
Overall, the code looks good. A few thoughts:

* Could we add an acceptance test for this? @alex-cardstack may be able to help
* Is there a place in the UI where we can show and/or allow you to copy the full (non-truncated) safe address? I think this is an important part of many multi-tool workflows, such as checking safe state in a the Gnosis Safe app/dapp or a block explorer

packages/safe-tools-client/app/services/safes.ts
jurgenwerk: That'r right. It is kinda awkward this way but it seemed convenient. 

If we use `this`, then `this` will point to the safe service object and not to `state`. An alternative would be to use a non fat arrow function (`load: function () {...}`), followed by `state.load.bind(state);`. Then, `this.loading = true` would work as expected. Thoughts?

jurgenwerk:
@lukemelia regarding truncated addresses problem, I really agree with you! It's important to have a simple way of copying the address. My initial thought is why are we even truncating - can't we just show the full address? But the address can get really long:

image

One idea that I have is to have a small "Copy" button next to the truncated address. I'll bring this up in the design channel.

packages/safe-tools-client/app/services/safes.ts
lukemelia: I'm OK with either. I guess the lexical scoping is probably easier to understand.

lukemelia (approved):
OK to merge once green 👍

PR #3464, opened 2022-11-28
[CS-4903] Store selected network in localStorage

alex-cardstack:
Persists the selected network between reloads


lukemelia (approved):
Once the failing test is addressed, looks good to me.

PR #3467, opened 2022-11-29
Update Glint to 1.0.0 beta.1

backspace:
A surprising amount of errors remain, but some of this should be addressed by strategic application of `@glint-nocheck`, at least for what used to be `transform.exclude`.

A partial list of the overall pattern of updates here:
* [removal of `@ts-expect-error`](https://github.com/cardstack/cardstack/pull/3467/commits/d0e651c0b7ed0a635bb66493e015172cbfdea058) for helper imports that were falsely `no-unused-vars` (the motivation for this upgrade)
* `npx -p @glint/scripts auto-glint-nocheck  "app/**/*.hbs"` in Ember applications to block Glint from checking files that haven’t been converted
  * it’s unclear to me how route-controller templates can use strict mode…?
* updating `@typescript-eslint/*` packages to prevent a [spurious `no-unused-vars` error](https://stackoverflow.com/a/73593347) stemming from Typescript 4.8’s AST changing its handling of decorators
* conversion of some Ember tests to abandon the `hbs` style in favour of `<template>`
* changing `config.get(` to `config.get<string>(` (or whatever) in Hub


backspace:
There are two things left to address.

## `wave-player`

The first is [strange](https://github.com/cardstack/cardstack/actions/runs/3597667103/jobs/6059651116#step:7:69):

```
addon/components/boxel/wave-player/index.gts:180:9 - error TS2345: Argument of type 'HTMLCanvasElement' is not assignable to parameter of type 'Function'.
  Type 'HTMLCanvasElement' is missing the following properties from type 'Function': apply, call, bind, prototype, and 5 more.

180         {{didInsert this.setupPlayer}}
            ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

Why is it expecting `Function` when this is the function definition?

```
  async setupPlayer(canvas: HTMLCanvasElement): Promise<void> {
```

## `ember-basic-dropdown` and `ember-power-select` type imports

The second is that I hacked `packages/boxel/types/ember-basic-dropdown/components/basic-dropdown/index.d.ts` and `packages/boxel/types/ember-power-select/components/power-select/index.d.ts` with inlined types from their respective addons instead of trying to extract them from imports like [here](https://github.com/cardstack/cardstack/blob/04b41b2b8810f736e7d7ad51f6e51679bc6ce462/packages/boxel/types/ember-power-select/components/power-select/index.d.ts#L3).

I learned in the Ember Discord that [importing from `…/addon`](https://discord.com/channels/480462759797063690/484421406659182603/1047566070321532928) would cause typechecking for happen to the whole addon in `node_modules`, so I was seeing errors like this:

```
../../node_modules/ember-power-select/addon/components/power-select.hbs:149:35 - error TS2339: Property 'loadingMessage' does not exist on type 'PowerSelectArgs'.

149             @loadingMessage={{or @loadingMessage "Loading options..."}}
```

Things almost work with the types inlined but I don’t like having so much duplicated and slightly modified from the addon directories. I’d like to use the types from the addons directly but I have not succeeded in getting that to happens. I’ve tried two major approaches:

1. Using `patch-package` to make the types exported when needed and add the extra arguments (`verticalPosition` for instance). This left me with even more type errors but it’s likely I did something wrong.
2. Using `ArgsFromComponent` to get the args off `PowerSelect` and then extending that. With this I was getting `recursively references itself as a base type` and didn’t know how to resolve it.

Maybe another route would be to make PRs for the addons so we can expose the types as needed?

lukemelia:
re wave-player: totally weird, I went with the @glint-expect-error path

re: power-select/basic-dropdown, I decided to fix the copy-pasted types on the theory that if those addons ever see another release, it will be with good types.

jurgenwerk (approved):

PR #3468, opened 2022-12-01
Update estimation endpoint to avoid decimals error

FadhlanR:
Not all tokens have the same decimal point as the native tokens. For example, the decimal value of a USDC token is 6. Before this update, we expected the crank safe to have a balance of 1 Gwei for all tokens, which makes no sense for a token like USDC as it would be too large. So in this PR, we don't use 1 Gwei for all tokens, we check the decimals value first and after that, decide the amount of transfer token and gas token.

### DIFF
--- a/packages/hub/services/gas-estimation.ts
+++ b/packages/hub/services/gas-estimation.ts
@@ -2,9 +2,9 @@ import { inject } from '@cardstack/di';
 import { addDays, addMilliseconds } from 'date-fns';
 import { nowUtc } from '../utils/dates';
 import config from 'config';
-import { ethers, Wallet } from 'ethers';
+import { Wallet } from 'ethers';
 import { GasEstimationResultsScenarioEnum } from '@prisma/client';
-import { convertChainIdToName, SchedulerCapableNetworks } from '@cardstack/cardpay-sdk';
+import { convertChainIdToName } from '@cardstack/cardpay-sdk';
 import { supportedChains } from '@cardstack/cardpay-sdk';
 import { NotFound } from '@cardstack/core/src/utils/errors';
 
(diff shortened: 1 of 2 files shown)


packages/hub/services/gas-estimation.ts
```diff
@@ -39,24 +39,71 @@ export default class GasEstimationService {
...
+    let tokenList = this.cardpay.getConstantByNetwork('tokenList', networkName as SchedulerCapableNetworks);
+    let token = tokenList.tokens.find((t) => t.address === params.tokenAddress);
+    let gasToken = tokenList.tokens.find((t) => t.address === params.gasTokenAddress);
+    if (!token || !gasToken) {
+      throw Error('unknown token and gas token');
+    }
+    //Check the decimals of tokens
+    //to avoid error when converting USD tokens
+    //because USD token's decimals is lower than native token's decimals
+    let tokenAmount = ethers.utils.parseUnits('1', token.decimals >= 18 ? 'gwei' : 3);
```
jurgenwerk, line 99: Can you explain why 3? Can we simplify this, would this work? 🔽

```
let tokenAmount = ethers.utils.parseUnits('1', token.decimals);
let gasTokenAmount = ethers.utils.parseUnits('0.01', token.decimals);
```

packages/hub/services/gas-estimation.ts
```diff
@@ -39,24 +39,71 @@ export default class GasEstimationService {
...
+      throw Error(`tokenAddress and gasTokenAddress is required in ${params.scenario}`);
+    }
+
+    let provider = this.ethersProvider.getInstance(params.chainId);
+    let networkName = convertChainIdToName(params.chainId);
+    let tokenList = this.cardpay.getConstantByNetwork('tokenList', networkName as SchedulerCapableNetworks);
+    let token = tokenList.tokens.find((t) => t.address === params.tokenAddress);
+    let gasToken = tokenList.tokens.find((t) => t.address === params.gasTokenAddress);
+    if (!token || !gasToken) {
+      throw Error('unknown token and gas token');
```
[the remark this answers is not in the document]
FadhlanR: updated

packages/hub/services/gas-estimation.ts
FadhlanR: Updated like this:

`let tokenAmount = ethers.utils.parseUnits('1', token.decimals >= 18 ? 'gwei' : token.decimals);
let gasTokenAmount = ethers.utils.parseUnits('0.01', gasToken.decimals >= 18 ? 'gwei' : token.decimals);`

If we use `ethers.utils.parseUnits('1', token.decimals);` it will be too big to have 1 eth for the crank safe.

packages/hub/services/gas-estimation.ts
```diff
@@ -39,24 +39,71 @@ export default class GasEstimationService {
...
+    let token = tokenList.tokens.find((t) => t.address === params.tokenAddress);
+    let gasToken = tokenList.tokens.find((t) => t.address === params.gasTokenAddress);
+    if (!token) throw Error('unknown token');
+    if (!gasToken) throw Error('unknown gas token');
+
+    //Check the decimals of tokens
+    //to avoid error when converting USD tokens
+    //because USD token's decimals is lower than native token's decimals
+    let tokenAmount = ethers.utils.parseUnits('1', token.decimals >= 18 ? 'gwei' : token.decimals);
+    let gasTokenAmount = ethers.utils.parseUnits('0.01', gasToken.decimals >= 18 ? 'gwei' : token.decimals);
```
lukemelia, line 100: Would this work?

```suggestion
    let tokenAmount = ethers.utils.parseUnits('1', token.decimals);
    let gasTokenAmount = ethers.utils.parseUnits('0.01', token.decimals);
```

FadhlanR: There is a worst case where the user uses WETH as a transfer token and gas token. With execution gas around `140000`, the crank safe need to have WETH balance around `1500` WETH to estimate the payment execution.

`total token = transfer token amount + (transfer token amount x fee percentage) + (gas token amount x execution gas) + (gas token amount in USD)`

FadhlanR: The question is how much amount of each token (WETH, USDC, USDT, etc) do we want to deposit in the crank safe?

jurgenwerk: Just had a discussion with @FadhlanR, we realized that:

1. The `gasTokenAmount` is wrongly named, it is in fact `gasTokenPrice`
2. Both `tokenAmount`and `gasTokenPrice` should be as small as possible (so that we don't have to keep a non-trivial balance of tokens in the crank) given their token decimal amount. So in decimal terms, something like `1e-decimals`.

FadhlanR: Sorry for the confusion I made it simple by defining `token amount = 1000 and gas price = 100` it worked for all tokens.

jurgenwerk: Looks good to me. Can you please add a comment in the code explaining why the value of these numbers make sense? I think it's good to document these hardcoded values in case someone needs to tweak them later.

FadhlanR: Comment added

PR #3470, opened 2022-12-02
Field-level error states for scheduled payment form

lukemelia:
- Revealed once the user has interacted with the field

The border colors were not captured that well by this video, but you can get the basic idea, and try it out locally.

https://user-images.githubusercontent.com/353/205231573-61eede99-8447-4516-a60d-efe273692117.mov

### DIFF
--- a/packages/safe-tools-client/app/components/schedule-payment-form-action-card/validator.ts
+++ b/packages/safe-tools-client/app/components/schedule-payment-form-action-card/validator.ts
@@ -52,7 +52,7 @@ export default class SchedulePaymentFormValidator {
       }
     } else if (form.selectedPaymentType === 'monthly') {
       if (!form.paymentDayOfMonth) {
-        return 'must choose an payment day of month';
+        return 'must choose a payment day of month';
       }
       if (!form.monthlyUntil) {
         return 'must choose an until date';
--- a/packages/safe-tools-client/app/components/schedule-payment-form-action-card/validator.ts
+++ b/packages/safe-tools-client/app/components/schedule-payment-form-action-card/validator.ts
@@ -1,5 +1,4 @@
 import { SelectableToken } from '@cardstack/boxel/components/boxel/input/selectable-token';
-import { tracked } from '@glimmer/tracking';
 import { isAddress } from 'web3-utils';
 
 function isNumeric(str: unknown) {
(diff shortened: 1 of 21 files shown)


jurgenwerk (approved):
I appreciate the hand written validator. I always seem to struggle with 3rd party form validation libs.

packages/safe-tools-client/app/components/schedule-payment-form-action-card/validator.ts
```diff
@@ -39,37 +33,88 @@ export default class SchedulePaymentFormValidator {
...
+  get paymentTypeErrorMessage(): string {
     const { form } = this;
     if (form.selectedPaymentType === 'one-time') {
-      return !!form.paymentDate;
+      if (!form.paymentDate) {
+        return 'must choose a payment date and time';
+      }
+    } else if (form.selectedPaymentType === 'monthly') {
+      if (!form.paymentDayOfMonth) {
+        return 'must choose an payment day of month';
```
backspace, line 55:
```suggestion
        return 'must choose a payment day of month';
```

packages/safe-tools-client/app/components/schedule-payment-form-action-card/validator.ts
```diff
@@ -39,37 +33,88 @@ export default class SchedulePaymentFormValidator {
   get isValid() {
     return (
       this.isPaymentTypeValid &&
-      this.isRecipientValid &&
+      this.isRecipientAddressValid &&
       this.isAmountValid &&
       this.isGasTokenValid &&
       this.isMaxGasFeeValid
```
backspace, line 39: Thoughts re validation of dates? Currently I can choose a one-time payment in the past or an until in the past for recurring.

backspace (approved):

backspace:
This predates your PR as I see it [on `main`](https://safe-tools-staging.stack.cards/schedule) but I noticed this while trying it out, that after switching to recurring I can’t switch back to one-time:

cannot-switch-back-to-one-time

There are no console errors 🤔

packages/safe-tools-client/app/components/schedule-payment-form-action-card/validator.ts
lukemelia: Yeah, we should add validation for this, or a minimum DateTime to the picker.

lukemelia: captured this in Linear

lukemelia:
>This predates your PR as I see it [on main](https://safe-tools-staging.stack.cards/schedule) but I noticed this while trying it out, that after switching to recurring I can’t switch back to one-time

Thanks. I captured this in Linear.

PR #3471, opened 2022-12-02
Update conversion util

FadhlanR:
### Problem

Conversion util didn't work in polygon networks because Uniswap had not deployed their version 2 contracts, and we used those contracts to do the conversion.

### Solution

We could deploy Uniswap V2 contracts by ourselves, but we would have a different UniswapV2Factory address and a different init code hash than the one already used in `@uniswap/sdk`. So the solution is to make the UniswapV2Factory address and init code hash parameterizable.

### DIFF
--- a/packages/cardpay-sdk/sdk/constants.ts
+++ b/packages/cardpay-sdk/sdk/constants.ts
@@ -179,7 +179,7 @@ const constants: {
     scheduledPaymentFeeFixedUSD: 0.25,
     scheduledPaymentFeePercentage: 0.1, //10%
     subgraphURL: '',
-    uniswapPairInitCodeHash: '0x96e8ac4277198ff8b6f785478aa9a39f403cb768dd02cbee326c3e7da348845f',
+    uniswapPairInitCodeHash: '0x96e8ac4277198ff8b6f785478aa9a39f403cb768dd02cbee326c3e7da348845f', //Result of keccak256(type(UniswapV2Pair).creationCode) from deployed UniswapV2Factory
   },
   goerli: {
     ...testHubUrl,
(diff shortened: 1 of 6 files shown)


packages/cardpay-sdk/sdk/constants.ts
```diff
@@ -47,6 +47,7 @@ interface SchedulerCapableNetworkConstants {
   tokenList: TokenList;
   scheduledPaymentFeeFixedUSD: number;
   scheduledPaymentFeePercentage: number;
+  uniswapPairInitCodeHash: string;
```
jurgenwerk, line 50: Can you add a comment in the code explaining why we need this. I think it would be good to explain this special case and mention the patch

jurgenwerk (commented):
Does this mean we have to set up our own liquidity pools for pairs of tokens? How does the conversion rate get determined in this case?

habdelra:
Yes, we’ll need to setup our own liquidity pool, but only for the Mumbai testnet. We can leverage existing liquidity pools in other networks (provided they have the token pairs that we want).

packages/cardpay-sdk/sdk/constants.ts
FadhlanR: Comment added

packages/cardpay-sdk/sdk/constants.ts
```diff
@@ -178,6 +179,7 @@ const constants: {
     scheduledPaymentFeeFixedUSD: 0.25,
     scheduledPaymentFeePercentage: 0.1, //10%
     subgraphURL: '',
+    uniswapPairInitCodeHash: '0x96e8ac4277198ff8b6f785478aa9a39f403cb768dd02cbee326c3e7da348845f', //Result of keccak256(type(UniswapV2Pair).creationCode) from deployed UniswapV2Factory
```
jurgenwerk, line 182:
```suggestion
    uniswapPairInitCodeHash: '0x96e8ac4277198ff8b6f785478aa9a39f403cb768dd02cbee326c3e7da348845f', // Result of keccak256(type(UniswapV2Pair).creationCode) from deployed UniswapV2Factory. We had to deploy our own Uniswap contracts on some networks (for now Mumbai) because they were missing, and add the patches/@uniswap+sdk+3.0.3.patch to support custom token pairs (by providing factory address and token pair code hash)
```

lukemelia (approved):

PR #3477, opened 2022-12-05
Begin wiring schedule payment form to SDK

lukemelia:
This is not up to our usual standards of completeness but I want to propose that we review and merge as we continue working in order to avoid blocking other tasks.

I was able to schedule a payment on staging/goerli:

Screenshot 2022-12-07 at 4 59 34 PM

### DIFF
--- a/packages/cardpay-sdk/sdk/utils/general-utils.ts
+++ b/packages/cardpay-sdk/sdk/utils/general-utils.ts
@@ -10,6 +10,8 @@ import { AbiItem } from 'web3-utils';
 import { Operation } from './safe-utils';
 import { v4 } from 'uuid';
 import JsonRpcProvider from '../../providers/json-rpc-provider';
+import { providers } from 'ethers';
+
 import { BytesLike, Interface } from 'ethers/lib/utils';
 import { utils } from 'ethers';
 import { getResolver } from '@cardstack/did-resolver';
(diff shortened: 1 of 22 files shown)


alex-cardstack (approved):
LGTM

packages/safe-tools-client/app/components/schedule-payment-form-action-card/index.gts
```diff
@@ -104,8 +111,35 @@ export default class SchedulePaymentFormActionCard extends Component<Signature>
...
+    const salt = btoa(String.fromCharCode.apply(null, array));
+
+    await taskFor(this.scheduledPaymentsSdk.schedulePayment).perform(
+      currentSafe.address,
+      currentSafe.spModuleAddress,
+      this.paymentToken.address,
+      convertAmountToRawAmount(this.paymentAmount, this.paymentToken.decimals),
+      this.payeeAddress,
+      this.executionGas,
+      '15000000000', // TODO: this.maxGasFee,
```
jurgenwerk, line 132: Would be good to use `maxGasPrice` name so that it is consistent with the name in the SDK and the contract.

jurgenwerk (approved):

packages/cardpay-sdk/sdk/utils/general-utils.ts
```diff
@@ -48,11 +50,17 @@ export interface ErrorFragment {
   inputs: string[];
 }
 
+// We don't care if it's the patched version from the SDK or one instantiated directly from ethers.js
```
FadhlanR, line 53: We do care about it I think because unpatched JsonRpcProvider had a different way to handle a reverted message, it might cause some scheduler module functions doesn't work properly such as `estimateExecutionGas` and `createSafeEstimation`.

packages/safe-tools-client/app/services/scheduled-payments-sdk.ts
```diff
@@ -46,6 +52,73 @@ export default class SchedulePaymentSDKService extends Service {
...
+    gasTokenAddress: string,
+    salt: string,
+    gasPrice: string,
+    payAt?: number | null,
+    recurringDayOfMonth?: number | null,
+    recurringUntil?: number | null
+  ): TaskGenerator<number> {
+    const scheduledPayments: ScheduledPaymentModule =
+      yield this.getSchedulePaymentsModule();
+    const result = yield scheduledPayments.estimateExecutionGas(
```
FadhlanR, line 71: I don't think the safe tool client needs to call this SDK function.

lukemelia: 👍

PR #3479, opened 2022-12-06
safe client: switch metamask network 

jurgenwerk:
Ticket: CS-4938

This PR assures that when using Metamask, the network selected in the client is also selected in Metamask. Without this, fetching token balances will fail because we use eth calls for that (`balanceOf`). If selected network in the client and in metamask are different, there will be a mismatch between the network and safe/token addresses.

This is how it looks like when the user selects a different network from the dropdown. If user approves, the change is applied and safes / token balances will be refetched from the newly selected network:

image

And this is how it looks if the user tries to switch to a network that they don't have in Metamask yet:

image

### DIFF
--- a/packages/cardpay-sdk/sdk/constants.ts
+++ b/packages/cardpay-sdk/sdk/constants.ts
@@ -39,6 +39,7 @@ interface RequiredNetworkConstants {
   nativeTokenCoingeckoId: string;
   nativeTokenName: string;
   nativeTokenSymbol: string;
+  nativeTokenDecimals: number;
   subgraphURL: string;
   relayServiceURL: string;
 }
@@ -99,6 +100,7 @@ const ethNativeTokens = {
   nativeTokenAddress: 'eth',
   nativeTokenSymbol: 'ETH',
   nativeTokenName: 'Ethereum',
+  nativeTokenDecimals: 18,
 };
 
 const bridgedTokens = {
@@ -111,6 +113,7 @@ const polygonNativeTokens = {
   nativeTokenCoingeckoId: 'matic-network',
   nativeTokenSymbol: 'MATIC',
   nativeTokenName: 'Matic',
+  nativeTokenDecimals: 18,
 };
 
 const constants: {
@@ -143,6 +146,7 @@ const constants: {
     nativeTokenCoingeckoId: 'xdai',
     nativeTokenSymbol: 'XDAI',
     nativeTokenName: 'xDai',
+    nativeTokenDecimals: 18,
     name: 'Gnosis Chain',
     relayServiceURL: 'https://relay.cardstack.com/api',
     subgraphURL: 'https://graph.cardstack.com/subgraphs/name/habdelra/cardpay-xdai',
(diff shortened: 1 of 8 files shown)


packages/safe-tools-client/app/services/wallet.ts
```diff
@@ -84,6 +84,64 @@ export default class Wallet extends Service {
...
+    }
+
+    if (!window.ethereum) {
+      return;
+    }
+
+    const chainIdHex = `0x${chainId.toString(16)}`;
+
+    try {
+      //@ts-expect-error currentProvider does not match Web3Provider Property 'request' does not exist on type 'EthereumProvider'
```
jurgenwerk, line 99: https://github.com/cardstack/cardstack/pull/3479/files#diff-1cc2eaa548adbb2771ec9077d450bb484fd247fa7a3578c64734d42c70f6110dR12 is suposed to take care of this but the error still remains. I'm researching why.

packages/safe-tools-client/app/services/wallet.ts
```diff
@@ -84,6 +84,64 @@ export default class Wallet extends Service {
...
+        method: 'wallet_switchEthereumChain',
+        params: [{ chainId: chainIdHex }],
+      });
+    } catch (error) {
+      if (error.code === 4902) {
+        // Network not added in Metamask. We prompt the user to add it.
+        try {
+          // Fetching data from this remote URL because it's supoosed to have the most up-to-date data on the RPC urls
+          const chainDataResponse = await (
+            await fetch('https://chainid.network/chains.json')
```
jurgenwerk, line 110: Will change this to read from the SDK

packages/cardpay-sdk/sdk/constants.ts
```diff
@@ -47,6 +48,7 @@ interface SchedulerCapableNetworkConstants {
   tokenList: TokenList;
   scheduledPaymentFeeFixedUSD: number;
   scheduledPaymentFeePercentage: number;
+  publicRpcUrls: string[];
```
jurgenwerk, line 51: Needed for the case when the user doesn't have this network added in metamask yet, and we try to add it

packages/safe-tools-client/app/services/network.ts
```diff
@@ -50,7 +57,12 @@ export default class NetworkService extends Service {
       }))
       .filter(({ name }) => name !== this.name);
   }
-  @action onSelect(networkInfo: NetworkInfo) {
+
+  @action async onSelect(networkInfo: NetworkInfo) {
+    await this.wallet.switchNetwork(networkInfo.chainId);
+
+    if (this.networkInfo.chainId === networkInfo.chainId) return;
```
jurgenwerk, line 64: When switching the network, this action will be called twice - firstly from the dropdown and secondly when metamask changes network. In this case, we don't want to assign the same network twice. If we assign the same thing, safe info area will flicker because it will try to refetch token balances.

packages/safe-tools-client/app/services/wallet.ts
```diff
@@ -84,6 +84,64 @@ export default class Wallet extends Service {
...
+    const chainIdHex = `0x${chainId.toString(16)}`;
+
+    try {
+      //@ts-expect-error currentProvider does not match Web3Provider Property 'request' does not exist on type 'EthereumProvider'
+      await window.ethereum.request({
+        method: 'wallet_switchEthereumChain',
+        params: [{ chainId: chainIdHex }],
+      });
+    } catch (error) {
+      if (error.code === 4902) {
```
lukemelia, line 105: Make 4902 a constant named ERR_METAMASK_UNKNOWN_NETWORK

lukemelia (approved):
Good to go apart from minor comments

PR #3480, opened 2022-12-07
Cs 4872 process reward root job

tintinthong:
This pr creates the job to process a reward root
- the purpose is to index a single parquet file (rewardProgramId=xxx/paymentCycle=xxx) in s3. This corresponds to a single merkle root but can have multiple rewardees in one process
- the file is queried using s3 select which reads parquet file and serializes to a json format (automatically)
- this job is triggered by a recurring task `check-reward-roots` which you can see here https://github.com/cardstack/cardstack/pull/3459 -- its good place to see how everything ties in together
- the access to s3 bucket `(cardpay-<env>-reward-programs)` should be open and no role assuming is needed
- it also updates `reward_root_index` in a single transaction so the hub knows where indexing has left off

### DIFF
--- a/packages/hub/tasks/process-reward-root.ts
+++ b/packages/hub/tasks/process-reward-root.ts
@@ -65,20 +65,32 @@ export default class ProcessRewardRoot {
           ')'
         );
       });
-    if (rows.length > 0) {
-      const proofsQuery = `
+    const indexQuery = 'INSERT INTO reward_root_index( reward_program_id, payment_cycle ) VALUES (%L, %L);';
+    const indexSql = pgFormat(indexQuery, file.rewardProgramId, file.paymentCycle);
+    try {
+      if (rows.length > 0) {
+        const proofsQuery = `
       INSERT INTO reward_proofs(
         reward_program_id, payee, leaf, payment_cycle, proof, token_type,  valid_from, valid_to, explanation_id, explanation_data
       ) VALUES %s;
     `;
-      const indexQuery = 'INSERT INTO reward_root_index( reward_program_id, payment_cycle ) VALUES (%L, %L);';
-      const proofsSql = pgFormat(proofsQuery, rows.join());
-      const indexSql = pgFormat(indexQuery, file.rewardProgramId, file.paymentCycle);
-      await this.databaseManager.performTransaction(db, async () => {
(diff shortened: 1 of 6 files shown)


packages/hub/tasks/process-reward-root.ts
```diff
@@ -0,0 +1,186 @@
...
+  'rewardProgramID,paymentCycle,validFrom,validTo,tokenType,payee,root,leaf,proof,explanationId';
+
+const queryParquet = async (
+  s3Client: S3Client,
+  file: S3FileInfo,
+  expression?: SelectObjectContentCommandInput['Expression']
+): Promise<Proof[]> => {
+  let records: Proof[] = [];
+  const params: SelectObjectContentCommandInput = {
+    Bucket: 'cardpay-staging-reward-programs',
```
IanCal, line 115: This should be configurable, not sure what the right way of passing this in is.

packages/hub/tasks/process-reward-root.ts
```diff
@@ -0,0 +1,186 @@
...
+          const os = record.split('\n');
+          os.map((o) => {
+            if (o != '') {
+              const b: any = JSON.parse(o);
+              if (!('explanationData' in b)) {
+                b.explanationData = '{}';
+              }
+              // when proof array serialize to json, it becomes [{item:... }. {item: ...}]
+              // need to transform to array of hex strings
+              b.proof = b.proof.length > 0 ? b.proof.map((o: any) => '0x' + o.item) : ['0x'];
```
IanCal, line 149: I think the empty case should be [], otherwise the client will get a proof array with a single item in.

tintinthong: This has been the pattern we've been using actually. When there is a single reward, proof = ["0x"] on returned by the api. When there is two rewards proof= ["0xbla"]. The sdk handles it as an array in both instances. 

I renamed proof to proofBytes for clarity (and also bcos of prisma related issue on pluralization)

IanCal: oh ok, thanks!

tintinthong:
@IanCal I have resolved the comments as above. Is it good to put in?

tintinthong:
@IanCal I have updated the configurations. Thanks for reminding

PR #3481, opened 2022-12-07
Fix naming of explanation data to from_block and to_block & fix values

IanCal:
These were set as the start and end block for when the rule should be calculated. Instead the range should be from the start to the end of the payment cycle, and payment cycle refers to the last block. from/end seemed inconsistent so I've standardised on from/to.


tintinthong (approved):
LGTM

PR #3482, opened 2022-12-07
Add minDate attribute in Boxel::Calendar and Boxel::Input::Date

FadhlanR:
Add `minDate` attribute so the dates before this date will be disabled. With this attribute, we can block the dates in the past.

https://user-images.githubusercontent.com/12637010/206197260-7a601d2d-ed37-42bc-9e20-b25a01b96c06.mov


lukemelia (approved):
Excellent!

PR #3483, opened 2022-12-07
Fix deposit modal display of safe info

lukemelia:
- Glint should have caught this -- need to figure out why it isn't. Answer: it is because of our "include" glint config. Let's merge #3467 and then opt this template into glint checks


alex-cardstack (approved):
This is something that glint could catch, right?

PR #3484, opened 2022-12-08
subgraph: Add polygon and mainnet subgraphs for tracking safes with scheduled payment module

jurgenwerk:
This PR adds 2 subgraphs that we can use to fetch safes in the payments tool. One for ethereum mainnet, and the other one for polygon (we already have subgraphs for their testnets):

1. https://api.thegraph.com/subgraphs/name/cardstack/safe-tools-mainnet
2. https://api.thegraph.com/subgraphs/name/cardstack/safe-tools-polygon

## TODO
 
Once the scheduled payment module is deployed to Polygon, change the `startBlock` number to be the actual block number of when the scheduled payment module was deployed, and redeploy the polygon subgraph. The current `startBlock`is currently set to one of the yesterday's blocks.

### DIFF
--- a/packages/cardpay-sdk/sdk/constants.ts
+++ b/packages/cardpay-sdk/sdk/constants.ts
@@ -178,7 +178,7 @@ const constants: {
     chainId: 1,
     scheduledPaymentFeeFixedUSD: 0.25,
     scheduledPaymentFeePercentage: 0.1, //10%
-    subgraphURL: '',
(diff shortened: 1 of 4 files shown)


jurgenwerk:
@alex-cardstack can you share:

1. `startBlock` for scheduled payment module startBlock in Polygon
2.  Contract addresses for Mainnet and Polygon (an example with a full list of addresses: https://github.com/cardstack/cardstack/blob/main/packages/cardpay-sdk/contracts/addresses.ts#L107)

FadhlanR (approved):
Looks good!

PR #3485, opened 2022-12-08
Remove parquet map type to support s3 select

tintinthong:


packages/cardpay-reward-programs/cardpay_reward_programs/payment_tree.py
```diff
@@ -145,7 +145,7 @@ def as_arrow(self):
                 columns["proof"].append(self.get_hex_proof(leaf))
                 columns["explanationId"].append(explanation_id)
                 columns["explanationData"].append(
-                    [(k, str(v)) for k, v in payment["explanationData"].items()]
+                    json.dumps(payment["explanationData"], separators=(",", ":"))
```
IanCal, line 148: I think the cutom separators is just removing the spaces - if this is required for s3select can you add a comment and a test? That's the kind of subtle regression we could see at some point. If it works anyway it'd be good to use just json.dumps with its defaults.

IanCal (changes requested):

tintinthong:
Additionally, I have added a quick fix to the indexer so it supports the old AND new column type. This should prevent anything from breaking

packages/cardpay-reward-indexer/cardpay_reward_indexer/indexer.py
```diff
@@ -30,6 +31,7 @@ def index_for_program(self, db: Session, reward_program_id, storage_location):
             last_submitted_root_block_number = self.get_last_indexed_root_block_number(
                 db, reward_program_id
             )
+            last_submitted_root_block_number = 27715485
```
tintinthong, line 34: oh gosh

IanCal (approved):
lgtm :)

PR #3486, opened 2022-12-08
Upgrade Prisma to get rid of a bug in db transaction error handling

jurgenwerk:
Ticket: CS-4970

When a smart contract revert happens during executing a scheduled payment, for example `ExceedMaxGasPrice` revert error, it looks like the process fails later in Prisma with an odd error that looks like something is wrong with Prisma code (`Cannot create property 'clientVersion' on string 'ExceedMaxGasPrice 0x55e6314242b42a6e6ffa13d68656503af31e53575b837e4b57e8e2b4b397196f'`) 

Looks like the error was fixed here in a more recent version of Prisma: https://github.com/prisma/prisma/pull/15196. This upgrade fixes the issue.

### DIFF
--- a/packages/hub/package.json
+++ b/packages/hub/package.json
@@ -151,7 +151,7 @@
     "msw": "^0.40.0",
     "node-loader": "^2.0.0",
     "npm-run-all": "^4.1.5",
-    "prisma": "^4.1.0",
+    "prisma": "^4.7.1",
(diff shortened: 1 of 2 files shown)


FadhlanR (approved):

lukemelia (approved):
:shipit:

PR #3489, opened 2022-12-11
Fix pdm install. New requirement parser giving trouble

tintinthong:
Do not upgrade to `packaging=22.0` lib that uses new requirement parser

See here 
- https://github.com/pdm-project/pdm/issues/1558
- https://github.com/pdm-project/pdm/issues/1554 for workaround 

Also remember to downgrade on your local. If you use pipx you can use this https://github.com/pdm-project/pdm/issues/1558#issuecomment-1345444582


tintinthong:
I think new release of pdm hs solved this issue

https://github.com/cardstack
Screenshot 2022-12-15 at 21 31 50
/cardstack/pull/3489

PR #3490, opened 2022-12-12
Utilizing the `minDate` and `minValue` attributes to block the selection of times in the past.

FadhlanR:
Ticket: CS-4941

- [x] Users cannot select a payment date before the current date.
- [x] Users cannot select a payment-specific time that occurs less than one hour from the current time.
- [x] Users cannot select a monthly recurring limit date before the first day of the recurring payment period.

https://user-images.githubusercontent.com/12637010/207019107-1ee72f7e-b4bb-4831-8044-b6eaf7954eb4.mov


lukemelia (changes requested):
All the code looks good. 👍  The changes to yarn.lock are concerning, though. They seem unrelated to the new dependency.

PR #3493, opened 2022-12-13
safe client: past payments list

jurgenwerk:
Ticket: CS-4664

The goal is to have a list of transactions like this (from design):

image

It looks like this: 

image

This PR also includes authenticating with the hub. This is needed to fetch the scheduled payment attempts from the hub. The prompt for that shows up after the wallet gets connected:

image

To see some payments, you need to run the hub and create a scheduled payment and scheduled payment attempts. I inserted these into the db manually - let me know if you want to run this, I can send an sql dump for these records. 

TODO:
- [x] Figure out a better user experience for authenticating with the hub. Currently, the wallet will request a signature when a request is made to the hub to fetch the payment attempts and the user is not authenticated yet. Would be good to explain to the user why this sign request is needed. (@jurgenwerk)
- [x] Integration tests for payments list component (@jurgenwerk)
- [x] Add date filter (@FadhlanR)
- [x] Add status filter (@lukemelia)
- [x] Style status filter (@FadhlanR)
- [x] Add "View on Etherscan" button (@lukemelia)
- [x] Add style to the "Status" column (red/green/in progress(?) indicator) (@jurgenwerk)
- [x] Add style to `to`address
- [x] Convert amount from wei to decimal (@jurgenwerk)
- [x] Add token symbol next to the amount (using tokenAddress) (@jurgenwerk)
- [x] Acceptance tests for hub auth (port from web-client)
- [x] Filter the transactions by selected network
- [x] Read hub url from config
- [x] Handle case when there are no attempts yet (display a message)


jurgenwerk:
This is now ready for review.

jurgenwerk:
https://github.com/cardstack/cardstack/pull/3493/commits/64436be33c180892e49cb5ce47605dd78a8be50c breaks the build and causes it to timeout. I'm checking why.

jurgenwerk:
I couldn't figure out why that was happening so decided to revert the commit to make this PR mergeable. Will fix later.

FadhlanR (approved):
LGTM

packages/safe-tools-client/app/services/scheduled-payments.ts
```diff
@@ -0,0 +1,125 @@
+/* eslint-disable @typescript-eslint/no-non-null-assertion */
+import { hubRequest } from '@cardstack/cardpay-sdk';
+import config from '@cardstack/safe-tools-client/config/environment';
+import HubAuthenticationService from '@cardstack/safe-tools-client/services/hub-authentication';
+import Service, { inject as service } from '@ember/service';
+import { BN } from 'bn.js';
```
lukemelia, line 6: Use BigNumber.js

packages/safe-tools-client/app/services/scheduled-payments.ts
```diff
@@ -0,0 +1,125 @@
...
+    feePercentage: string;
+    tokenAddress: string;
+    gasTokenAddress: string;
+    payeeAddress: string;
+    payAt: string;
+    chainId: string;
+  };
+}
+
+export interface ScheduledPaymentResponseItem {
```
lukemelia, line 26: It might be nice if the hub published these types somehow for us to consume.

lukemelia (approved):

packages/safe-tools-client/app/components/hub-auth-modal/index.gts
```diff
@@ -0,0 +1,88 @@
...
+
+            <p>You only need to do this once per browser per device.</p>
+          </div>
+        </Section>
+
+        <ActionChin @state={{this.connectionState}}>
+          <:default as |a|>
+            <a.ActionButton {{on "click" this.authorize}}>
+              Authenticate
+            </a.ActionButton>
```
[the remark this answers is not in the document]
jurgenwerk: I think this should be mandatory and we shouldn't let user use the app without being authenticated. Otherwise there might be confusion around why the payments list is not loaded.

packages/safe-tools-client/app/components/hub-auth-modal/index.gts
```diff
@@ -0,0 +1,88 @@
...
+  @service declare network: NetworkService;
+  @service declare wallet: WalletService;
+  @tracked isAuthenticating = false;
+
+  get isModalOpen() {
+    return this.wallet.isConnected && this.hubAuthentication.isAuthenticated === false;
+  }
+
+  onClose() {
+    // noop, needed for the BoxelModal component API
```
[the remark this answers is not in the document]
jurgenwerk: Added.

packages/safe-tools-client/app/services/scheduled-payments.ts
jurgenwerk: Right! Actually BigNumber from Ethers.

PR #3496, opened 2022-12-15
Add date and status filter

FadhlanR:
The goal of this PR is to make the date and status filter like this:

Screen Shot 2022-12-15 at 15 27 35

So I added `Boxel::Input::FilterSelect` component and implemented it in the `payment-transactions-list`.

https://user-images.githubusercontent.com/12637010/207810597-a6bc7047-6e3c-40f6-bb2c-6d7b38f3eefe.mov

https://user-images.githubusercontent.com/12637010/207811032-e234e811-b151-47a2-b3f5-d582d147d299.mov

### DIFF
--- a/packages/boxel/addon/components/boxel/input/filter-select/index.css
+++ b/packages/boxel/addon/components/boxel/input/filter-select/index.css
@@ -1,0 +1,42 @@
+.boxel-input-filter-select {
+  --boxel-input-filter-select-border-color: var(--boxel-form-control-border-color);
+  --boxel-input-filter-select-border-radius: var(--boxel-border-radius-lg);
+
+  align-items: center;
+  border-radius: var(--boxel-input-filter-select-border-radius);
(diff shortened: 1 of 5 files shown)


jurgenwerk (commented):
I'm not sure this use case warrants creating a new Boxel component. Why not use BoxelSelect directly? It should be able to support this case. We could feed it an array of filter options, for example `[ { display: 'last 30 days', value: subDays(new Date, 30)}, ... ]`. And then you can set the `filter[started-at][gt]` and  `filter[status]` params in the hub request url, depending on the choice in the dropdown.

FadhlanR:
Doesn't need to create a new boxel component for this. I will commit directly to this branch `cs-4664-past-payments-list-2`

FadhlanR:
Doesn't need to create a new boxel component for this. I will commit directly to this branch cs-4664-past-payments-list-2

PR #3499, opened 2022-12-15
Add proofs table

tintinthong:
Migration to add proofs table in the db. This is the reward data that is indexed and made available to the sdk to check balances and list proofs

Previous [implementation](https://github.com/cardstack/cardstack/blob/0329cf7a2e81fb41ac2c5797c2269e2eeb2ad09b/packages/cardpay-reward-indexer/cardpay_reward_indexer/models.py#L7-L21)

Removed unnecessary fields  `tokenAddress` and  `amount` and added readily available available fields from parquet file `tokenType`


packages/hub/db/migrations/20221215134519797_reward-proofs.ts
```diff
@@ -0,0 +1,22 @@
+import { MigrationBuilder } from 'node-pg-migrate';
+
+const TABLE = 'reward_proofs';
+
+export async function up(pgm: MigrationBuilder): Promise<void> {
+  pgm.createTable(TABLE, {
```
jurgenwerk, line 6: Do we need any indexes on this table?

tintinthong: Ive added a compound index `payee, reward_program_id`

tintinthong: Thanks! Ive added compound index [payee, reward_program_id]

packages/hub/config/structure.sql
```diff
@@ -1202,6 +1202,26 @@ CREATE TABLE public.reservations (
...
 
+--
+-- Name: reward_proofs; Type: TABLE; Schema: public; Owner: postgres
+--
+
+CREATE TABLE public.reward_proofs (
+    reward_program_id text NOT NULL,
+    payee text NOT NULL,
+    leaf text NOT NULL,
+    payment_cycle integer NOT NULL,
```
IanCal, line 1213: Not vital now but if it doesn't cause issues some futureproofing here would be to change this to a numeric(78,0) type - that covers the full range of uint256 that the underlying types used in the contract can hold.

IanCal: Not quite sure how that plays with prisma, might need to explicitly set it as a decimal(76,0) 

[@db.Decimal(p, s)†
--](https://www.prisma.io/docs/reference/api-reference/prisma-schema-reference#decimal)

IanCal (commented):
Looks good - added a note around integer types

packages/hub/config/structure.sql
IanCal: Actually - we'd hit other issues in the processing setup if we were close to this at all. Let's keep it at integer.

PR #3501, opened 2022-12-16
Use @uniswap/v2-sdk instead of @uniswap/sdk

FadhlanR:
Ticket: CS-4958

We decided to use `@uniswap/v2-sdk` instead of `@uniswap/sdk`, so I implemented the conversion function using that package in this PR. I have tested it by comparing the required gas in USD result of estimate-gas CLI in this branch and the main branch.


jurgenwerk (approved):
Code looks good. Can you list a couple of examples where you tested if conversion results match? I vaguely remember someone was complaining in the github issue mentioned in the ticket the results from v2-sdk differ from the old sdk package.

PR #3504, opened 2022-12-19
Implement gas estimation

FadhlanR:
I implemented the gas fee fetching in this PR and used that process's results to schedule payment.


lukemelia (approved):
Thanks!

PR #3519, opened 2023-01-03
Add Endpoint: Add GET reward proofs 

tintinthong:


packages/hub/routes/reward-proofs.ts
```diff
@@ -0,0 +1,60 @@
...
+          id: proof.leaf,
+          type: 'reward-proofs',
+          attributes: proof,
+        };
+      });
+      ctx.status = 200;
+      ctx.body = { data };
+      ctx.type = 'application/vnd.api+json';
+    } catch (e) {
+      log.error('Failed to retrieve reward proofs', e);
```
jurgenwerk, line 51: Would be good to rethrow or respond with an error

packages/hub/routes/reward-proofs.ts
```diff
@@ -0,0 +1,60 @@
+import { inject } from '@cardstack/di';
+import Koa from 'koa';
+import autoBind from 'auto-bind';
+import Logger from '@cardstack/logger';
+import { isAddress } from 'web3-utils';
```
jurgenwerk, line 5: I think you can get this util from the web3 package.

`import Web3 from 'web3';`

`Web3.utils.isAddress`

packages/hub/package.json
```diff
@@ -110,6 +110,7 @@
     "web3": "1.5.2",
     "web3-core": "1.7.3",
     "web3-eth-contract": "^1.7.3",
+    "web3-utils": "^1.8.1",
```
jurgenwerk, line 113: Potentially not needed. See my other comment

jurgenwerk (approved):
Looks good, minor comments.

packages/hub/routes/reward-proofs.ts
tintinthong: Will throw an error. Good point

packages/hub/routes/reward-proofs.ts
tintinthong: Noted. Will follow suggestion

PR #3521, opened 2023-01-03
SDK: Cs 4878 update rewards sdk point to hub

tintinthong:


packages/cardpay-sdk/sdk/reward-pool/base.ts
```diff
@@ -154,31 +151,55 @@ export default class RewardPool {
...
+            }
+          } else {
+            res.push({
+              ...o,
+              tokenAddress: token,
+              amount: amount,
+              isValid,
+              explanationTemplate,
+            });
+          }
```
oidoug, line 202: I suppose this conditional could be rewritten as something like:
```
if (knownClaimed || !claimedLeafs.includes(o.leaf)) {
   res.push({
     ...o,
     tokenAddress: token,
     amount: amount,
     isValid,
     explanationTemplate,
});
```
Same for the one above.

oidoug (approved):
LGTM, left a comment below, I think the conditionals for adding to `res` could be simplified a bit.

tintinthong:
There are 77 proofs which are valid and active in previous indexer blue and the new hub db

packages/cardpay-sdk/sdk/reward-pool/base.ts
tintinthong: Yes! Thx

IanCal (approved):

packages/hub/package.json
```diff
@@ -110,6 +110,7 @@
     "web3": "1.5.2",
     "web3-core": "1.7.3",
     "web3-eth-contract": "^1.7.3",
+    "web3-utils": "^1.8.1",
```
jurgenwerk, line 113: Is this needed? I don't see any new imports

tintinthong: Cheers!

PR #3525, opened 2023-01-04
[CS-4986] Query balances from gnosis transaction service instead of eth calls

alex-cardstack:
CS-4986

### DIFF
--- a/packages/cardpay-sdk/sdk/utils/safe-utils.ts
+++ b/packages/cardpay-sdk/sdk/utils/safe-utils.ts
@@ -11,7 +11,7 @@ import { AbiItem } from 'web3-utils';
 import { getAddress } from '../../contracts/addresses';
 import GnosisSafeProxyFactoryABI from '../../contracts/abi/gnosis-safe-proxy-factory';
 import GnosisSafeABI from '../../contracts/abi/gnosis-safe';
-import { Transaction, isJsonRpcProvider } from './general-utils';
+import { Transaction, isJsonRpcProvider, nonNullable } from './general-utils';
 /* eslint-disable node/no-extraneous-import */
 import { Contract, ethers, utils } from 'ethers';
 import { AddressZero } from '@ethersproject/constants';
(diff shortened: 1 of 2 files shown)


packages/cardpay-sdk/sdk/utils/safe-utils.ts
```diff
@@ -477,18 +477,55 @@ export function gasInToken(estimate: Estimate): BN {
...
+          token: { symbol, decimals },
+        } = balanceInfo;
+
+        return { tokenAddress, symbol, balance: BigNumber.from(balance), decimals };
+      } else {
+        return null;
+      }
     })
-  );
+    .filter((e) => !!e) as BalanceSummary[];
```
lukemelia, line 530:
```suggestion
    .filter(Boolean) as BalanceSummary[];
```

lukemelia (approved):
Should we consider filtering the token balances against our supported tokens list from the tokenlist json or let the application do that?

packages/cardpay-sdk/sdk/utils/safe-utils.ts
danibonilha, line 530: if you use a function like this `const nonNullable = <T>(value: T): value is NonNullable<T> => !!value;`  typescript will know the value is valid and you won't need to cast anything.
then 
`.filter(nonNullable)`

it would be nice to have this as a util in the sdk, since we also do the same stuff on the wallet

packages/cardpay-sdk/sdk/utils/safe-utils.ts
```diff
@@ -477,18 +477,55 @@ export function gasInToken(estimate: Estimate): BN {
...
+      let balanceInfo = balances.find((b) => b.tokenAddress?.toLowerCase() === tokenAddress.toLowerCase());
+
+      if (balanceInfo?.token) {
+        let {
+          balance,
+          token: { symbol, decimals },
+        } = balanceInfo;
+
+        return { tokenAddress, symbol, balance: BigNumber.from(balance), decimals };
+      } else {
```
danibonilha, line 526: nit: there's no need to add the else here, since there's a return on the if, but I understand if you prefer to keep it.

danibonilha (approved):

packages/cardpay-sdk/sdk/utils/safe-utils.ts
alex-cardstack: That's so much better, I always feel guilty using `as` :)

packages/cardpay-sdk/sdk/utils/safe-utils.ts
alex-cardstack: It was only there because typescript was complaining not all code paths returned a value, even when I explicitly typed the anonymous function being OK to return undefined. But I refactored a different way to avoid it.

alex-cardstack:
> Should we consider filtering the token balances against our supported tokens list from the tokenlist json or let the application do that?

So this function takes an explicit list of tokens that it wants the balances of, which (presumably in most cases, definitely in this case) is already coming from a supported list. So as it's explicitly passed in, I think filtering here is counterintuitive - better to filter in the consuming application and allow getting balances that are present in the JSON but are not "supported" depending on the use case later.

PR #3526, opened 2023-01-04
ssr: Add Fastboot test for early failure detection

backspace:
This doesn’t actually have the test yet but the goal is to find when things like #3506 are needed before deployment.


jurgenwerk (approved):
Thanks for a great explanation. Looks good to me.

PR #3529, opened 2023-01-05
Show USD value of the token in Scheduled Payment Card

FadhlanR:
Ticket: CS-5051

Screen Shot 2023-01-05 at 15 52 11
*) The USD rate is stubbed on the test code, so it doesn't represent the real value.


jurgenwerk (commented):
I understand this was mostly copied from web-client but it involves a nontrivial amount of trickery in the helper <-> service coupling. I am not sure it should be the service's responsibility to know about its consumers (the helpers in this case). 

If I understand correctly, it was made that way so that we optimize data loading and we only refresh conversion data in the token usd value displays that are actually rendered? 

I am thinking about an alternative where we could have a tagless component and read the token's usd conversion rate from the service, and have *the component* periodically call the service to refresh only the conversion data it cares about. I believe that would be easier to reason about with less code perhaps.

Something like:

Usage:

`<TokenToUsd @tokenAddress="0x123" @amount=100>`

TokenToUsd component:

```
constructor() {
 ... 
  setInterval(function(){
    this.usdConversionService.refresh(this.args.tokenAddress); // in the refresh implementation, return early if data is already fresh
  }, 5000);
  
 // make sure timer stops on component teardown
}
get usdAmount() {
  return this.usdConversionService.rates[this.args.tokenAddress] * this.args.amount;
}
```
```
<template>
  {{this.usdAmount}} 
</template>
```

jurgenwerk:
Yeah I don't see a problem with that since Ember is performant enough. Wondering what @lukemelia thinks about this too. I'm fine with this if it was battle tested in the web client, just thinking about an alternative where we can reduce the number of ember-specific things (i.e. registering helpers and `waitForQueue`)

FadhlanR:
I see, so there is a cache in the service eventhough each components will call the refresh function regularly, it will be okay. Is my understanding right? @jurgenwerk

jurgenwerk:
For that I think we'd need to have a `conversionRateRefreshedAt` for each of the token addresses. So in the `refresh` method we'd check if the rate is still fresh otherwise we'd reload it.

lukemelia:
I agree it is worth being creative about how to simplify. But I think there should be caching that works when there are multiple instances of the component on different parts of the screen. Using a service without registration but with a well-configured ember-concurrency task might be a good approach.

packages/cardpay-sdk/sdk/utils/conversions.ts
```diff
@@ -123,3 +124,21 @@ export async function getNativeWeiInToken(provider: JsonRpcProvider, tokenAddres
   let rateAdjusted = adjustRate(rate);
   return new BN(rateAdjusted.numerator.toString()).div(new BN(rateAdjusted.denominator.toString()));
 }
+
+export async function getUsdConverter(
```
jurgenwerk, line 128: The output of this is the conversion rate, right? Let's rename to `getUsdConversionRate`

packages/safe-tools-client/app/components/token-to-usd/index.gts
```diff
@@ -0,0 +1,50 @@
...
+
+export default class TokenToUsd extends Component<Signature> {
+  @service('token-to-usd') declare tokenToUsdService: TokenToUsdService;
+  updateInterval: NodeJS.Timeout;
+
+  constructor(owner: unknown, args: Args) {
+    super(owner, args);
+
+    this.updateInterval = setInterval(() => {
+      console.log('here');
```
jurgenwerk, line 28: Probably forgot to delete this

packages/safe-tools-client/app/services/token-to-usd.ts
```diff
@@ -0,0 +1,41 @@
...
+    let now = new Date();
+    if(!this.convertersLastUpdate[tokenAddress] || addMilliseconds(this.convertersLastUpdate[tokenAddress], INTERVAL) < now) {
+      this.usdConverters[tokenAddress] = yield getUsdConverter(this.wallet.ethersProvider, tokenAddress);
+      this.usdConverters = this.usdConverters;
+
+      this.convertersLastUpdate[tokenAddress] = now;
+    }
+  }
+
+  toUsdFrom(tokenAddress: string, amount: BigNumber): BigNumber | undefined {
```
jurgenwerk, line 29: I think `toUsd` would be descriptive enough

jurgenwerk: Should we add some assurance here that the conversion rate is up to date?

packages/cardpay-sdk/sdk/utils/conversions.ts
FadhlanR: The output is a function to convert token amount to usd amount.

jurgenwerk: Ok, I see

packages/safe-tools-client/app/services/token-to-usd.ts
FadhlanR: Renamed the function. 

This function must be re-executed every time we get the latest value from `getUsdConverter`, so I don't think it is needed.

lukemelia (approved):

packages/safe-tools-client/app/components/token-to-usd/index.gts
```diff
@@ -0,0 +1,49 @@
...
+interface Signature {
+  Element: HTMLElement;
+  Args: Args
+}
+
+const INTERVAL = config.environment === 'test' ? 1000 : 60 * 1000;
+
+export default class TokenToUsd extends Component<Signature> {
+  @service('token-to-usd') declare tokenToUsdService: TokenToUsdService;
+  updateInterval: NodeJS.Timeout;
```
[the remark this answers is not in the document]
FadhlanR: Updated the type.

PR #3531, opened 2023-01-06
Update time windows

FadhlanR:
Ticket: CS-5803


packages/safe-tools-client/app/components/future-payments-list/index.gts
```diff
@@ -111,9 +124,10 @@ export default class FuturePaymentsList extends Component<Signature> {
         <Section @title="Future Payments" data-test-future-payments-list>
           <div class="future-payments-list__payments-section">
             <div class="future-payments-list__payments-section-time-brackets">
-              <ScheduledPaymentTimeBracket @title="next hour" @scheduledPayments={{this.nextHour}}/>
-              <ScheduledPaymentTimeBracket @title="next month" @scheduledPayments={{this.nextMonth}}/>
-              <ScheduledPaymentTimeBracket @title="next few months" @scheduledPayments={{this.nextFewMonthsPayments}}/>
+              <ScheduledPaymentTimeBracket @title="today" @scheduledPayments={{this.today}}/>
+              <ScheduledPaymentTimeBracket @title="tomorrow" @scheduledPayments={{this.tomorrow}}/>
+              <ScheduledPaymentTimeBracket @title="this month" @scheduledPayments={{this.thisMonth}}/>
+              <ScheduledPaymentTimeBracket @title="later" @scheduledPayments={{this.later}}/>
```
jurgenwerk, line 130: Will this display "Later" when all scheduled payments happen next month?

FadhlanR: Yes, in that case, the "Later" time window will be displayed.

jurgenwerk: I think showing "Later" string doesn't make sense when then there are no earlier other time windows displayed

packages/safe-tools-client/app/components/future-payments-list/index.gts
```diff
@@ -31,33 +32,45 @@ export default class FuturePaymentsList extends Component<Signature> {
   @service declare network: NetworkService;
   // @ts-expect-error loading services/scheduled-payments
   @service('scheduled-payments') declare scheduledPaymentsService: ScheduledPaymentsService;
+  @service declare date: DateService;
+  
+  get now(): Date {
+    return new Date(Number(this.date.now()));
```
jurgenwerk, line 38: We have a [clock service](https://github.com/cardstack/cardstack/blob/main/packages/hub/services/clock.ts) in the hub. Maybe follow that approach so that our time travel techniques are consistent

packages/safe-tools-client/tests/integration/future-payments-list-test.ts
```diff
@@ -52,36 +101,50 @@ class ScheduledPaymentsStub extends Service {
...
+  hooks.beforeEach(async function (this: TestContext) {
     this.owner.register('service:wallet', WalletServiceStub);
     this.owner.register('service:scheduled-payments', ScheduledPaymentsStub);
     this.owner.register(
       'service:hub-authentication',
       HubAuthenticationServiceStub
     );
+
+    dateService = this.owner.lookup('service:date') as FakeDateService;
+    await dateService.setNow(NOW.getTime());
```
jurgenwerk, line 141: Is redefining "now" really needed? If you use the real current timestamps the test should work anyway right?

packages/safe-tools-client/app/components/future-payments-list/index.gts
FadhlanR: Do you mean to hide all scheduled payments or change the label?

packages/safe-tools-client/tests/integration/future-payments-list-test.ts
jurgenwerk: Oh I see. Makes sense

lukemelia (approved):
Is the component assuming that the scheduled payments are pre-sorted? Is that a valid assumption?

packages/safe-tools-client/app/components/future-payments-list/index.gts
FadhlanR: Updated

PR #3540, opened 2023-01-10
safe tools: on payment cancel, remove scheduled payment from the list 

jurgenwerk:
This PR will reload the payments list after cancelling a scheduled payment. This is so that the canceled payment won't be in the "future payments" list anymore.

Another improvement is adding a message explaining that this action could take minutes (due to waiting for confirmation):
image


FadhlanR (approved):
Looks great!

PR #3545, opened 2023-01-11
safe tools client: Increase day of month recurring max range from 28 to 31

jurgenwerk:
To make it a bit more intuitive I think we can increase the range to 31. 

There's the obvious problem of months having different amount of days but we are handling this in the hub where the [calculateNextPayAt](https://github.com/cardstack/cardstack/blob/main/packages/hub/utils/scheduled-payments.ts#L6) function will adjust this for every month. So in case the user chooses 31st, the payment will be performed on the last day of months that have fewer days. Same for 30 and 29. I think this way is better so that we can allow users to execute payments on the actual last day of month and not on 28th as the maximum available day. 

This behaviour is pretty nicely covered in [tests](https://github.com/cardstack/cardstack/blob/main/packages/hub/node-tests/utils/scheduled-payments-test.ts).

image

### DIFF
--- a/packages/safe-tools-client/app/components/schedule-payment-form-action-card/ui/index.css
+++ b/packages/safe-tools-client/app/components/schedule-payment-form-action-card/ui/index.css
@@ -70,3 +70,13 @@
 .schedule-payment-form-action-card__in-progress-message {
   max-width: 20rem;
 }
+
+.schedule-payment-form-action-card__recurring-explanation {
+  font: var(--boxel-font-xs);
+  color: var(--boxel-purple-400);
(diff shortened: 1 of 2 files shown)


jurgenwerk:
I think it would be good to add an info note about this behaviour.

jurgenwerk:
@FadhlanR this is supported in the contracts right?

FadhlanR:
Yes it is supported @jurgenwerk

FadhlanR (approved):

lukemelia (approved):

lukemelia:
> I think it would be good to add an info note about this behaviour next to the dropdown.

Maybe it is something that could appear if you choose 29-31?

PR #3546, opened 2023-01-11
feat: Script to create relay servers gas tokens

pcjun97:


pcjun97:
The delete functions are left unimplemented since we might not want to automate removing the tokens or tickers. (In case there is a typo or an accidental removal of tokens in the tokenlist)

scripts/update-relay-servers-tokens.ts
```diff
@@ -0,0 +1,107 @@
+/* eslint-disable node/no-unsupported-features/es-syntax */
+/* eslint-disable node/no-unpublished-import  */
+/* eslint-disable node/no-missing-import */
+
+import { RelayAdmin, Token, PriceOracleTicker } from './relay/relay-admin.js';
+import { readFileSync } from 'fs';
+
+if (process.argv.length < 6) {
+  console.error(
+    'usage: yarn update-relay-tokens [tokenlist.json] [host] [username] [password]\n' +
```
lukemelia, line 10: Let's make the command `yarn update-relay-gas-tokens` and the script file name `update-relay-server-gas-tokens.ts`

lukemelia (approved):
Per our discussion, let's add deleting too

jurgenwerk (approved):

PR #3547, opened 2023-01-11
Improve messaging when errors occurr on scheduled payment form

alex-cardstack:
CS-5082

1. Show error if scheduling the payment fails for any reason:
image

2. Show error if gas estimation fails:
image

3. Show a loading indicator when gas estimation is in progress:
image


jurgenwerk:
I made some improvements:

1. Showing loading indication where the amounts will show up

telegram-cloud-photo-size-4-5909196540948364128-y

telegram-cloud-photo-size-4-5909196540948364129-y

2. Error display (+ show generic error instead of the raw error message)

telegram-cloud-photo-size-4-5909196540948364127-y

3. Show a better user message for `NotEnoughFundsForMultisigTx` case:
image

jurgenwerk:
@alex-cardstack @FadhlanR I'm confused why are we getting the `NotEnoughFundsForMultisigTx` in case when the safe has no ETH - I thought the relayer should pay for this?

lukemelia (approved):
👍

PR #3548, opened 2023-01-12
Fix usd converter

FadhlanR:
Ticket: CS-5111

Screenshots:

While loading the USD converter or USD converter is undefined:
Screen Shot 2023-01-12 at 19 07 52

After the USD converter is defined:
Screen Shot 2023-01-12 at 19 09 12

*) in goerli 1 ETH = 36886.9 USD

### DIFF
--- a/packages/cardpay-sdk/sdk/utils/conversions.ts
+++ b/packages/cardpay-sdk/sdk/utils/conversions.ts
@@ -71,7 +71,7 @@ export async function gasPriceInToken(provider: JsonRpcProvider, tokenAddress: s
   let network = await networkName(provider);
   let chainId = (await provider.getNetwork()).chainId;
 
-  // Gas station will return current gas price in native token in wei.
+  // Gas station will return current gas price in native units of the token.
   let gasPriceInNativeTokenInWei = new BN((await getGasPricesInNativeWei(chainId)).standard.toString());
 
   // We use the wrapped native token address because the native token doesn't have an address in Uniswap.
--- a/packages/cardpay-sdk/sdk/utils/conversions.ts
+++ b/packages/cardpay-sdk/sdk/utils/conversions.ts
@@ -136,7 +136,7 @@ export async function getUsdConverter(
     return (amountInWei: BigNumber) => amountInWei;
   }
 
-  let rate = await tokenPairRate(provider, tokenAddress, usdcTokenAddress);
+  let rate = await tokenPairRate(provider, usdcTokenAddress, tokenAddress);
   return (amountInWei: BigNumber) => {
     let rateAdjusted = adjustRate(rate);
     return amountInWei.mul(rateAdjusted.numerator.toString()).div(rateAdjusted.denominator.toString());
--- a/packages/cardpay-sdk/sdk/utils/conversions.ts
+++ b/packages/cardpay-sdk/sdk/utils/conversions.ts
@@ -132,7 +132,7 @@ export async function getUsdConverter(
   let network = await networkName(provider);
   let usdcTokenAddress = getAddressByNetwork('usdcToken', network);
 
-  if (usdcTokenAddress === tokenAddress) {
+  if (usdcTokenAddress.toLowerCase() === tokenAddress.toLowerCase()) {
     return (amountInWei: BigNumber) => amountInWei;
   }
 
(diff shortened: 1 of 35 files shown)


jurgenwerk (commented):
If a converter is undefined, isn't that indicative of a big problem? Meaning we should show an error? Or when is it undefined?

Edit: Oh I see, it's undefined during initial loading. Can we add a placeholder or a loading indicator?

packages/safe-tools-client/app/components/token-to-usd/index.gts
```diff
@@ -19,26 +21,28 @@ const INTERVAL = config.environment === 'test' ? 1000 : 60 * 1000;
...
   }
 
   willDestroy() {
     clearInterval(this.updateInterval);
   }
 
   <template>
-    {{this.usdAmount}}
+    {{#if this.usdAmount}}
+      $ {{(weiToDecimal this.usdAmount 18)}} USD
```
jurgenwerk, line 44: Is 18 correct in all cases? Don't some tokens have 6 decimals?

FadhlanR: Sorry, I pushed the wrong code. Updated.

FadhlanR:
@jurgenwerk I'll add the loading indicator

packages/cardpay-sdk/sdk/utils/conversions.ts
```diff
@@ -136,7 +136,7 @@ export async function getUsdConverter(
     return (amountInWei: BigNumber) => amountInWei;
   }
 
-  let rate = await tokenPairRate(provider, tokenAddress, usdcTokenAddress);
+  let rate = await tokenPairRate(provider, usdcTokenAddress, tokenAddress);
```
jurgenwerk, line 139: Was this the reason for this bug? Was the rate inverted?

FadhlanR: This is one of the reasons, the other one is that I was not converting to the biggest unit.

jurgenwerk: Ok. It's so weird we didn't notice this when we tested - I suppose because the rates are so odd on Goerli 🙈

lukemelia: I was definitely running into this yesterday, and super confused about whether it was a) a bug in the converter code, b) Goerli wackiness, c) test setup mistakes, or d) a bug in my code. Thanks for fixing this.

packages/safe-tools-client/app/components/token-to-usd/index.gts
```diff
@@ -23,22 +25,23 @@ export default class TokenToUsd extends Component<Signature> {
...
     clearInterval(this.updateInterval);
   }
 
   <template>
-    {{this.usdAmount}}
+    {{#if this.usdAmount}}
+      $ {{(weiToDecimal this.usdAmount @tokenDecimals)}} USD
+    {{/if}}
   </template>
 }
```
lukemelia, line 46: Unrelated to this PR, but I think this might be better as a helper than a component. What do you think?

packages/safe-tools-client/app/components/token-to-usd/index.gts
```diff
@@ -23,22 +27,30 @@ export default class TokenToUsd extends Component<Signature> {
...
 
   <template>
-    {{this.usdAmount}}
+    {{#if this.usdAmount}}
+      $ {{(weiToDecimal this.usdAmount @tokenDecimals)}} USD
+    {{else}}
+      <div class="token-to-usd-loading">
+        <BoxelLoadingIndicator class="token-to-usd-loading-indicator" @color="var(--boxel-purple-400)" />
+        <div class="network-connect-modal__waiting-status">
+          Converting to USD...
```
FadhlanR, line 50: @jurgenwerk Added the loading icon, maybe you have a better description for this.

FadhlanR: Screen Shot 2023-01-13 at 16 05 28

jurgenwerk: I think text is good but drop the spinner. To make it consistent with the loading indicator here:
image


https://github.com/cardstack/cardstack/pull/3547#issuecomment-1380130274

FadhlanR: Updated

packages/safe-tools-client/app/components/token-to-usd/index.gts
lukemelia: Hmm, I think this should be possible. Let's go ahead with this PR and we can consider revisiting the possible conversion to a helper separately.

lukemelia (approved):

PR #3551, opened 2023-01-12
Clarify references to native units of a token

lukemelia:
- We were previously using "wei" to mean "the native unit of a token", but it's meaning specific to 18 decimals


jurgenwerk (approved):
Thanks for clarifying this!

habdelra (approved):

PR #3557, opened 2023-01-13
Fix variable fee calculation

FadhlanR:
Ticket: CS-5117


packages/hub/node-tests/services/scheduled-payments/validator-test.ts
```diff
@@ -119,32 +119,6 @@ describe('ScheduledPaymentValidator', function () {
...
-    const scheduledPayment: Partial<ScheduledPayment> = {
-      chainId: 1, //Mainnet
-      feeFixedUsd: new Decimal(0.25),
-      feePercentage: new Decimal(2),
-    };
-
-    let errors = await subject.validate(scheduledPayment);
-    expect(errors.feePercentage).deep.equal(['fee percentage must be between 0 and 1']);
-  });
-
```
lukemelia, line 147: These validations could be replaced with a zero to 100 check.

lukemelia (approved):

PR #3559, opened 2023-01-16
Calculate fees and show in summary

lukemelia:
- [x] First merge #3562 and rebase


jurgenwerk (approved):
Code looks good to me. Thanks for plowing through this. I added a small CSS fix. 

In terms of UX I don't like that the execution plan section shows only *after* I enter all the fields in the form. From the first time user standpoint - If it was displayed right from the start with blank values I would for sure know the app will calculate all the payment amounts at the end which would give me more confidence to create my first scheduled payment in a safe and controllable way. But we can discuss this later as a future enhancement.

PR #3561, opened 2023-01-16
Add state of scheduled payment form when no compatible safe is present

FadhlanR:
Ticket: CS-5078

Screenshot 2023-01-16 at 22 16 18


packages/safe-tools-client/app/components/schedule-payment-form-action-card/ui/index.gts
```diff
@@ -295,51 +297,66 @@ export default class SchedulePaymentFormActionCardUI extends Component<Signature
           </div>
         </BoxelField>
       </Section>
-        <ActionChin @state={{this.actionChinState}}>
-          <:default as |ac|>
+      <ActionChin @state={{this.actionChinState}}>
+        <:default as |ac|>
+          {{#if @isSafesEmpty}}
```
jurgenwerk, line 302: If safes are empty, then display the Schedule Payment button? Isn't this the other way around as it should be?

packages/safe-tools-client/app/components/schedule-payment-form-action-card/ui/index.gts
```diff
@@ -295,51 +297,66 @@ export default class SchedulePaymentFormActionCardUI extends Component<Signature
...
-            <ac.ActionStatusArea @icon={{concat @walletProviderId "-logo" }} style={{cssVar status-icon-size="2.5rem"}}>
-              <BoxelLoadingIndicator
-                class="schedule-payment-form-action-card__loading-indicator"
-                @color="var(--boxel-light)"
-              />
-              <div class="schedule-payment-form-action-card__in-progress-message" data-test-in-progress-message>
-                {{@schedulingStatus}}
+          {{else}}
+            <ac.ActionStatusArea style={{cssVar status-icon-size="2.5rem"}}>
+              <div class="scheduled-payment-form-action-card__empty-safe-stage-icon">
```
jurgenwerk, line 312: Let's make it so that this markup is reusable for the other case where we will prompt the user to connect the wallet. Consider generalizing classes, something like:

```
<div class="schedule-payment-form-prerequisite-step">
  <div class="schedule-payment-form-prerequisite-step__icons">
    {{svgJar 'icon123' width="..." height="..."
      style=(cssVar icon-color='var(--boxel-teal)')
    }}
    {{svgJar ...}}
    {{svgJar ...}}
  </div>
  <div class="schedule-payment-form-prerequisite-step__text">
    Step 2 - Create a payments safe
  </div>
</div>
```

```
.schedule-payment-form-prerequisite-step {
  display: flex;
}

.schedule-payment-form-prerequisite-step__icons svg {
  margin-left: ...;
}

.schedule-payment-form-prerequisite-step__text {
  margin: auto;
  ... 
}
```

With this approach, we can simply reuse that for the other case ("Step 1 - First connect your wallet"), we can just change the copy and icons.

packages/safe-tools-client/tests/integration/schedule-payment-form-action-card-test.ts
```diff
@@ -316,6 +334,23 @@ module(
         .containsText('Cardstack charges $0.25 USD and 0.1%');
     });
 
+    test(`it shows up the create safe instruction if get empty safes`, async function (assert) {
```
jurgenwerk, line 337:
```suggestion
    test(`it displays the create safe instruction if user has no safes`, async function (assert) {
```

jurgenwerk (changes requested):
Would like a change so that the classes can be reused for the other case which I am currently working on ("Step 1 - connect your wallet"). 

Should we also disable the form when no safe is present?

jurgenwerk:
Can you also add a screenshot of how the sidebar looks in this state?

packages/safe-tools-client/app/components/schedule-payment-form-action-card/ui/index.gts
FadhlanR: Reused your code for that

FadhlanR:
This is the screenshot with the sidebar:
 
Screenshot 2023-01-17 at 21 59 00

lukemelia (approved):

jurgenwerk:
image

I think this step indicator arrow is missing?

jurgenwerk:
I think you can use the `@isActive` in the ControlPanel's item

FadhlanR:
Added step indicator

Screenshot 2023-01-17 at 22 52 51

jurgenwerk:
But should both Network and Safe have the indicator? Isn't just the Safe section that needs it?

FadhlanR:
@jurgenwerk According to the design both are needed.

Screenshot 2023-01-17 at 23 22 09

lukemelia:
> @jurgenwerk According to the design both are needed.

I think this is a case of the design being wrong. It doesn't make sense to have them both show as active.

FadhlanR:
> > @jurgenwerk According to the design both are needed.
> 
> I think this is a case of the design being wrong. It doesn't make sense to have them both show as active.

Removed

PR #3570, opened 2023-01-18
Restrict scheduler tool network options to the networks supported by the connected hub server

lukemelia:


FadhlanR:
Should we add a validation to the `onChainChanged` function?

packages/safe-tools-client/app/services/network.ts
```diff
@@ -47,14 +50,24 @@ export default class NetworkService extends Service {
...
-    return supportedChainsArray
-      .map((networkSymbol) => ({
-        name: getConstantByNetwork('name', networkSymbol),
-        chainId: getConstantByNetwork('chainId', networkSymbol),
-        symbol: networkSymbol,
-      }))
-      .filter(({ name }) => name !== this.name);
+  get supportedNetworks() {
+    // Initially return all the scheduler-capable networks the SDK supports, and then
+    // narrow it down after retrieving the networks supported by the hub.
```
jurgenwerk, line 55: Why is this step needed? Can we just use `remoteConfig.web3.schedulerNetworks` without using it to narrow down supportedChainsArray?

lukemelia: I did it this way to avoid a loading state. We start with all the options (available synchronously) and then narrow down to what the current hub supports.

jurgenwerk (approved):

lukemelia:
> Should we add a validation to the onChainChanged function?

@FadhlanR see my last commit for this. We still need more robust handling of unsupported networks, but this is something.

FadhlanR (approved):
It looks good. We can have more robust handling for network change later.

PR #3572, opened 2023-01-19
Add missing HUB_URL env var to safe tools deployments

jurgenwerk:
The issue is that safe tools client is making requests to the hub with a broken and very wrong url:

image

It turns out `HUB_URL` is missing in the environment vars, which makes `hubUrl` undefined (`hubUrl: process.env.HUB_URL` in environment.js

### DIFF
--- a/.github/workflows/manual-safe-tools-client.yml
+++ b/.github/workflows/manual-safe-tools-client.yml
@@ -20,6 +20,18 @@ jobs:
       - uses: ./.github/actions/setup-node-modules-cache
         with:
           yarn_lock_md5: ${{ hashFiles('yarn.lock') }}
+      - name: Set up env
+        env:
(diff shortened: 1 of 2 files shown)


.github/workflows/manual-safe-tools-client.yml
```diff
@@ -20,6 +20,18 @@ jobs:
       - uses: ./.github/actions/setup-node-modules-cache
         with:
           yarn_lock_md5: ${{ hashFiles('yarn.lock') }}
+      - name: Set up env
```
jurgenwerk, line 23: This is copied from the web client config: https://github.com/cardstack/cardstack/blob/main/.github/workflows/manual-web-client.yml#L23

lukemelia (approved):
🐿️

PR #3575, opened 2023-01-19
Add new Cardstack test tokens to goerli

lukemelia:


FadhlanR:
We also should update `usdcToken` value in `address.ts`.

PR #3576, opened 2023-01-20
safe tools: Fix wrongly formatted "pay at" filter

jurgenwerk:
Ticket: CS-5152

The area where we noticed this problem was the "Future payments" which kept showing payments that were already executed. The filtering mechanism for this is: "fetch payments with payAt larger than the current time". 

The bug that broke this came in 2 parts:
- Should be `filter[pay-at][gt]` and not `filter[payAt][gt]`
- Hub expects ISO formatted time values (e.g. `'2023-01-20T10:12:43.707Z'`). This is a more verbose alternative to unix timestamps but so much better for human readability and debugging.

### DIFF
--- a/packages/safe-tools-client/app/services/scheduled-payments.ts
+++ b/packages/safe-tools-client/app/services/scheduled-payments.ts
@@ -213,9 +213,7 @@ export default class ScheduledPaymentsService extends Service {
   ): Promise<ScheduledPayment[]> {
     let queryString = `filter[chain-id]=${chainId}`;
     if (minPayAt) {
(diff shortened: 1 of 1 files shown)


danibonilha (approved):
LGTM, i'm wondering though if there's any action we could take next to avoid this kind of issue, does the hub return an error message stating the filter key doesn't exist ?  Or maybe we could add a filter query builder utils which takes the available fields and return the correct string, idk, we can circle back if this issue reoccurs.

jurgenwerk:
Currently there is no such filter presence validation. I agree that would be a great addition.

PR #3578, opened 2023-01-23
hub: Ensure payments are executed at least 1min after payAt (defense against block time manipulation)

jurgenwerk:
This PR aims to address the issue in the scheduler tool where when the payment gets executed at the full hour and it fails, but then at the next run 5 minutes after (the crank triggers every 5 minutes looking for payments to execute), it succeeds:

image

We discovered this happens because there is a revert condition in the ScheduledPaymentModule smart contract: (https://github.com/cardstack/cardstack-module-scheduled-payment/blob/main/contracts/ScheduledPaymentModule.sol#L167) which reverts payment execution if it is attempted before 1 minute has passed from the `payAt` timestamp. This is for mitigating the effects of miners' block timestamp manipulation, which could result, if I understand correctly, in executing payments ahead of time, or bunching many transactions together to get a bigger mining reward in the time window that block time can be manipulated (which is around 1 minute judging by the comment in the smart contract code mentioned above).

The obvious solution to this is to delay the execution of scheduled payments by at least one minute, which is implemented as a change in this PR in the query that returns the payments that are due to be executed now. But this comes with a downside which we need to discuss if we are ok with it. 

The crank will trigger every 5 minutes (our current cron config), looking for scheduled payments to execute. Let's say the crank will fire at 11:00 and there is a scheduled payment with 11:00:00 `payAt` value. Previously, this payment would be attempted at this step but it would result in a failure due to the revert condition described above. But with the new implementation, this payment would be left out, and it would only be picked up at 11:05 when the crank triggers again. This means the payment would get executed at 11:05 when it was supposed to be executed 11:00. It means with this change, the crank executes payments with a 5 minute delay. 

This is not ideal – ideally we should execute payments with a minimal delay, which is 1 minute when conforming to the block manipulation time buffer mentioned earlier. I can think of a couple of alternatives (as an addition to this PR):

1. Making the crank fire every minute or every 2 minutes (but we then really need the exponential backoff for retries... which is what we are planning to do anyway)
2. Introducing an artificial 1 minute delay here: https://github.com/cardstack/cardstack/blob/main/packages/hub/tasks/execute-scheduled-payments.ts#L6 - only when trying to execute payments on full hour timestamps (because scheduled payments can only be scheduled on full hours)

### DIFF
--- a/packages/hub/node-tests/routes/data-integrity-checks-test.ts
+++ b/packages/hub/node-tests/routes/data-integrity-checks-test.ts
@@ -1,3 +1,4 @@
+import cryptoRandomString from 'crypto-random-string';
 import { subMinutes } from 'date-fns';
 import shortUuid from 'short-uuid';
 import { CREATION_WITHOUT_TX_HASH_ALLOWED_MINUTES } from '../../services/data-integrity-checks/scheduled-payments';
@@ -50,7 +51,7 @@ describe('GET /api/data-integrity-checks/scheduled-payments', async function () 
         feePercentage: '0',
         salt: '54lt',
         payAt: nowUtc(),
-        spHash: '0x123',
+        spHash: cryptoRandomString({ length: 10 }),
         chainId: 1,
         userAddress: '0x57022DA74ec3e6d8274918C732cf8864be7da833',
         creationTransactionHash: null,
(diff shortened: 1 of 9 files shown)


jurgenwerk:
We just had a discussion that it makes more sense to get rid of the time buffer requirement in the smart contract. This is because the miners have no real incentive to manipulate block timestamp in this case. The removal of that logic makes this PR irrelevant and we can execute the payment  immediately when it's due, without waiting.

PR #3580, opened 2023-01-24
Disable network switcher when using WalletConnect

FadhlanR:
Ticket: CS-5122

https://user-images.githubusercontent.com/12637010/214264268-5d71273a-7f80-4998-86e2-0b0826eafd35.mov

### DIFF
--- a/packages/safe-tools-client/app/templates/schedule.hbs
+++ b/packages/safe-tools-client/app/templates/schedule.hbs
@@ -40,7 +40,7 @@
           <span class="safe-tools__current-network-text">{{this.network.networkInfo.name}}</span>
           <Boxel::IconButton
             @icon="info"
-            aria-label="Special Button"
+            aria-label="Current Network"
             style={{css-var
               boxel-icon-button-width=0
               boxel-icon-button-height=0
--- a/packages/safe-tools-client/app/templates/schedule.hbs
+++ b/packages/safe-tools-client/app/templates/schedule.hbs
@@ -25,7 +25,7 @@
       />
     </cp.Item>
     <cp.Item @title='Network' @icon='network'>
-      {{#if this.wallet.isConnected}}
+      {{#if (and this.wallet.isConnected (eq this.wallet.providerId "metamask"))}}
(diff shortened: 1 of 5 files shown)


packages/safe-tools-client/app/templates/schedule.hbs
```diff
@@ -35,6 +35,20 @@
         >
           <div class={{cn itemCssClass safe-tools__current-network-option=(eq this.network.symbol network.symbol)}} data-test-selected-network>{{network.name}}</div>
         </Boxel::Select>
+      {{else if this.wallet.isConnected}}
+        <div class="safe-tools__current-network">
+          <span class="safe-tools__current-network-text">{{this.network.networkInfo.name}}</span>
+          <Boxel::IconButton
+            @icon="info"
+            aria-label="Special Button"
```
lukemelia, line 43: Should this be something like "Current Network"?

lukemelia (approved):

packages/safe-tools-client/app/templates/schedule.hbs
```diff
@@ -35,6 +35,20 @@
...
+          <span class="safe-tools__current-network-text">{{this.network.networkInfo.name}}</span>
+          <Boxel::IconButton
+            @icon="info"
+            aria-label="Current Network"
+            style={{css-var
+              boxel-icon-button-width=0
+              boxel-icon-button-height=0
+              icon-color='var(--boxel-light)'
+            }}
+            data-hover="You can change the network from your wallet. Supported networks: {{(array-join this.network.supportedNetworksName ',')}}"
```
danibonilha, line 49: Unfortunately this may not be true for wallet connect 2.0, I don't think there's an action to take now, but just a heads up to keep this on our radar

danibonilha (approved):

packages/safe-tools-client/app/templates/schedule.hbs
FadhlanR: We can remove it once wallet connect 2.0 is implemented.

PR #3582, opened 2023-01-24
Add copy button next to all safe address

FadhlanR:
Ticket: CS-4931

https://user-images.githubusercontent.com/12637010/214316700-ba4c6454-ccf6-4214-a67a-58c13bf58657.mov


jurgenwerk (commented):
While we're at it let's add this to all blockchain addresses we display, I think this is what it says in the ticket too. This means the wallet in the upper left section in the sidebar, recipients in scheduled payments, basically anywhere `truncateMiddle` is used

packages/safe-tools-client/app/components/safe-info/index.gts
```diff
@@ -27,6 +33,15 @@ interface Signature {
 }
 
 export default class SafeInfo extends Component<Signature> {
+  @tracked isShowingCopiedConfirmation = 0;
+
+  @action flashCopiedConfirmation() {
```
lukemelia, line 38: Consider converting to a concurrency task to get better async lifecycle management. i.e. shouldn't try to set execute `this.isShowingCopiedConfirmation = 0` after delay if component has been destroyed.

packages/safe-tools-client/app/components/safe-info/index.gts
```diff
@@ -38,16 +53,36 @@ export default class SafeInfo extends Component<Signature> {
...
+              boxel-icon-button-height="100%"
+              icon-color="var(--boxel-light)"
+            }}
+          {{on "click"
+                (copyToClipboard
+                  value=@currentSafe.address
+                  onCopy=this.flashCopiedConfirmation
+                )
+              }}
+        />
```
lukemelia, line 81: It is a little weird from an information architecture perspective to have this outside of the dropdown, but I think given the options it is OK.

FadhlanR:
@jurgenwerk Added to all blockchain addresses.


https://user-images.githubusercontent.com/12637010/214785581-3d3028b8-53fe-4efb-b182-f4f28b5750e2.mov

jurgenwerk:
The icon could be a bit smaller (otherwise it stands out too much), and in line with the text

packages/safe-tools-client/app/components/truncated-blockchain-address/index.gts
```diff
@@ -0,0 +1,73 @@
...
+  @action flashCopiedConfirmation() {
+    this.isShowingCopiedConfirmation = 1;
+    later(() => {
+      this.isShowingCopiedConfirmation = 0;
+    } , 1000)
+  }
+
+  <template>
+    {{#if @isCopyable}}
+      <div class='truncated-blockchain-address' {{didInsert this.registerListener}}>
```
jurgenwerk, line 42: This is a bit misleading - in both cases of `@isCopyable` the address will be truncated, but `truncated-blockchain-address` class is only used in the first case.

FadhlanR:
@jurgenwerk I updated the size of the icon to 14px.

Screenshot 2023-01-26 at 15 27 40

jurgenwerk (approved):

packages/safe-tools-client/app/components/truncated-blockchain-address/index.gts
```diff
@@ -0,0 +1,78 @@
...
+  @task *flashCopiedConfirmation() {
+    this.isShowingCopiedConfirmation = 1;
+    later(() => {
+      this.isShowingCopiedConfirmation = 0;
+    } , 1000)
+  }
+
+  <template>
+    {{#if @isCopyable}}
+      {{!@glint-expect-error "Passing this function should be valid but it is not as of 1.0.0-beta.1"}}
```
jurgenwerk, line 44: Why is it not valid?

FadhlanR: I am unsure about the error here, it's a weird error. I fixed it the same way as we did [here](https://github.com/cardstack/cardstack/pull/3467#issuecomment-1335834902)

PR #3583, opened 2023-01-24
hub: add exponential backoff to scheduled payment (re)attempts

jurgenwerk:
Currently, the crank triggers every 5 minutes and on each run and it will fetch scheduled payments that are due to be executed at that time, and try to execute them. This also includes failed payments. 

The problem is that we don't really want to reattempt payments every 5 minutes, but to add a progressively larger delay between attempts as it is usual with such mechanisms (so that we don't overwhelm our systems).

Currently, the strategy goes like this: If the payment fails, it can be attempted again after 5 minutes. And if it still fails, it can be retried after 60 minutes after the last failure. Then, after 360 minutes (6 hours), and then every 720 minutes (12 hours) until max attempts is reached, which is 8 attempts. There is no special formula how I got these numbers, it's just something that felt good to me, and it's hardcoded within the query and easy to change if we feel the need to do that.

### DIFF
--- a/packages/hub/services/scheduled-payments/fetcher.ts
+++ b/packages/hub/services/scheduled-payments/fetcher.ts
@@ -1,10 +1,10 @@
 import { inject } from '@cardstack/di';
 import { ScheduledPayment } from '@prisma/client';
 import { startOfDay, subDays } from 'date-fns';
-import { nowUtc } from '../../utils/dates';
 
 export default class ScheduledPaymentsFetcherService {
   prismaManager = inject('prisma-manager', { as: 'prismaManager' });
+  clock = inject('clock', { as: 'clock' });
   validForDays = 3;
 
   // This query fetches scheduled payments that are due to be executed now.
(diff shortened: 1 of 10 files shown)


packages/hub/services/scheduled-payments/fetcher.ts
```diff
@@ -17,30 +17,41 @@ export default class ScheduledPaymentsFetcherService {
...
-            GROUP BY scheduled_payment_id) AS last_failed_payment_attempt
+      LEFT JOIN (
+        SELECT
+          scheduled_payment_attempts.scheduled_payment_id,
+          MAX(scheduled_payment_attempts.started_at) AS started_at,
+          COUNT(*) as failed_attempts_count
+        FROM scheduled_payment_attempts
+        INNER JOIN scheduled_payments ON scheduled_payments.id = scheduled_payment_attempts.scheduled_payment_id
+        WHERE scheduled_payment_attempts.status = 'failed'
+          AND started_at >= scheduled_payments.pay_at
```
jurgenwerk, line 41: Only count failed attempts in the most recent schedule period (when the payment is recurring)

packages/hub/services/scheduled-payments/fetcher.ts
```diff
@@ -17,30 +17,41 @@ export default class ScheduledPaymentsFetcherService {
...
-          scheduled_payments.canceled_at IS NULL 
-          AND scheduled_payments.creation_block_number > 0 
-          AND scheduled_payments.pay_at > ${startOfDay(subDays(_nowUtc, this.validForDays))}
-          AND scheduled_payments.pay_at <= ${_nowUtc}
+          scheduled_payments.canceled_at IS NULL
+          AND scheduled_payments.creation_block_number > 0
+          AND scheduled_payments.pay_at > ${startOfDay(subDays(now, this.validForDays)).toISOString()}::timestamp
+          AND scheduled_payments.pay_at <= ${nowString}::timestamp
+          AND COALESCE(failed_attempts_count, 0) < 8
+          AND ${nowString}::timestamp >= (COALESCE(last_failed_payment_attempt.started_at, to_timestamp(0)) + (interval '1 minute' * (ARRAY[0, 5, 60, 360, 720, 720, 720, 720])[COALESCE(failed_attempts_count + 1, 1)]))
```
jurgenwerk, line 51: `failed_attempts_count + 1` because pg arrays start with 1! 🤯

The coalesce parts are there for addressing the situation where there is no last failed payment - in that case there are values in place that will always satisfy the `>=` condition.

packages/hub/services/scheduled-payments/fetcher.ts
```diff
@@ -61,8 +72,8 @@ export default class ScheduledPaymentsFetcherService {
               )
             ) 
             OR (
-              scheduled_payments.recurring_day_of_month > 0 
-              AND scheduled_payments.recurring_until >= ${_nowUtc}
+              scheduled_payments.recurring_day_of_month > 0
+              AND scheduled_payments.recurring_until >= ${nowString}::timestamp
```
jurgenwerk, line 76: To ensure UTC. We could use `now() at time zone 'utc'` but we want to have the possibility to mock current utc time for testing (using the clock service stub)

lukemelia (approved):
Looks good. Does the effective latest retry time work with our default valid-for-days config?

FadhlanR (approved):
LGTM

jurgenwerk:
@lukemelia nice catch - I added some logic to make the exponential backoff fit into the retry period (`validForDays`)

packages/hub/services/scheduled-payments/fetcher.ts
```diff
@@ -84,9 +100,16 @@ export default class ScheduledPaymentsFetcherService {
...
         LIMIT ${limit};`;
 
-    // Retrieve scheduledPayment prisma object based on filtered id
     return prisma.scheduledPayment.findMany({ where: { id: { in: results.map((result) => result.id) } } });
   }
+
+  calculateRetryBackoffsInMinutes() {
+    let fixedPart = [0, 5, 60, 360]; // Retry immediately, then after 5 minutes, 1 hour, 6 hours
+    let fixedPartSum = fixedPart.reduce((a, b) => a + b, 0);
+    let variablePartChunksCount = Math.floor((this.validForDays * 24 * 60 - fixedPartSum) / 720);
```
danibonilha, line 109: supernit: but it would be nice to add meaning to some of the values, like:
 ```ts
const  oneDayInMinutes = 24 * 60`;
const  (halfDayInMinutes or twelveHoursInMin) = 720; and then reuse this value 
...
```
or you could have one single object with all of this values
```ts
const time = {
    fiveMin: 5,
    oneHour: 60,
    sixHours: 360,
    oneDay: 24 * 60,
    twelveHours: 720
}
```

PR #3591, opened 2023-01-31
Use USDC to estimate the execution

FadhlanR:
Ticket: CS-5175

### DIFF
--- a/packages/hub/services/gas-estimation.ts
+++ b/packages/hub/services/gas-estimation.ts
@@ -4,15 +4,12 @@ import { nowUtc } from '../utils/dates';
 import config from 'config';
 import { Wallet } from 'ethers';
 import { GasEstimationResultsScenarioEnum } from '@prisma/client';
-import { convertChainIdToName } from '@cardstack/cardpay-sdk';
-import { supportedChains } from '@cardstack/cardpay-sdk';
+import { convertChainIdToName, getAddress, supportedChains } from '@cardstack/cardpay-sdk';
 import { NotFound } from '@cardstack/core/src/utils/errors';
 
 export interface GasEstimationParams {
   scenario: GasEstimationResultsScenarioEnum;
   chainId: number;
-  tokenAddress?: string;
-  gasTokenAddress?: string;
 }
 
 export default class GasEstimationService {
@@ -27,8 +24,6 @@ export default class GasEstimationService {
       where: {
         chainId: params.chainId,
         scenario: params.scenario,
-        tokenAddress: params.tokenAddress,
-        gasTokenAddress: params.gasTokenAddress,
       },
     });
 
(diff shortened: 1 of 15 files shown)


packages/hub/services/gas-estimation.ts
```diff
@@ -96,17 +78,18 @@ export default class GasEstimationService {
     // since transferAmount will be calculated with fee percentage.
     // And gasPrice is price per unit of gas, so it will be multiplied by the gas execution.
     let transferAmount = '1000';
-    let gasPrice = '100';
+    let gasPrice = '1';
     let gas;
+    let usdTokenAddress = await getAddress('usdStableCoinToken', provider);
```
jurgenwerk, line 83: Can you add a code comment where we explain that estimations will actually vary between different token implementations but we are using usdc to make things simpler.

jurgenwerk (approved):
Looks good, just minor requests for more explanations in comments

lukemelia (approved):
Once comments requested by @jurgenwerk are added, 👍

packages/hub/services/gas-estimation.ts
```diff
@@ -76,62 +67,43 @@ export default class GasEstimationService {
...
-    // For transferAmount we should consider the underflow error,
-    // since transferAmount will be calculated with fee percentage.
-    // And gasPrice is price per unit of gas, so it will be multiplied by the gas execution.
-    let transferAmount = '1000';
-    let gasPrice = '100';
+    // Estimating using transfer and gas amount are zero,
+    // so we don't need to deposit any amount to the safe owned by the crank.
+    // Using USDC token because USDC token doesn't throw any errors
+    // if transfer amount is zero.
+    // Other tokens might revert erorr if the transfer amount is zero.
```
jurgenwerk, line 78:
```suggestion
  // Payment execution can support multiple transfer and gas tokens, but for estimating the 
  // execution we can simply use the USDC token. This is because this token contract allows 
  // zero amounts for transfer, which is usually not the case for other token implementations.
  // Because of this feature, we don't have to deposit any USDC to the safe if we
  // set the transaction amounts to 0. If we used some other transfer/gas token for estimation,
  // then we would have to deposit an adequate amount of tokens to the safe in order for 
  // the estimation process to work.
```

FadhlanR:
Since we decided to use USDC to estimate the execution, it doesn't make sense to deposit the USDC in the safe owned by the crank because it will be a too high amount of USDC to deposit. Let's say we put the gas price as 1 and assume the minimum execution gas is 170000, so we need to deposit more than 170 USDC.

So I decided to do the estimation by defining the transfer amount and gas price as 0, so we don't need to deposit any USDC. And it worked because USDC doesn't have validation to make sure the transfer value is greater than 0, whereas other tokens might have that validation. Because of that, we can also use the user's safe to estimate the execution instead of the safe owned by the crank.

lukemelia (approved):
Will the zero token transfer approach work with CTSC? Do we need to make changes to the contract to allow it?

FadhlanR:
It already worked with CTSC.

PR #3599, opened 2023-02-02
Add missing config for HUB_DATABASE_URL

lukemelia:


lukemelia:
Deploying this to staging. If it resolves the issue, we'll update waypoint.prod.hcl and then merge

PR #3600, opened 2023-02-03
Fix max gas toggle to show different gas values

FadhlanR:
Ticket: CS-5222


packages/safe-tools-client/app/components/schedule-payment-form-action-card/ui/max-gas-toggle-button.gts
```diff
@@ -14,6 +15,14 @@ interface Signature {
 }
 
 export default class MaxGasToggleButton extends Component<Signature> {
+  get tokenQuantity(): TokenQuantity | undefined {
+    let value = this.args.maxGasDescriptions?.value;
+    if (this.args.name === 'normal') return value?.normal;
+    else if (this.args.name === 'high') return value?.high;
+    else if (this.args.name === 'max') return value?.max;
+    else return undefined;
+  }
```
danibonilha, line 24: nit: you could either omit all 'else s' since all of cases already  returns, or even better maybe, you could use the key of the value to simplify this function. e.g:

```ts
  get tokenQuantity(): TokenQuantity | undefined {
    let value = this.args.maxGasDescriptions?.value;
    return value?.[this.args.name]
  }
```

danibonilha (approved):
LGTM, just a nit way to rewrite the function, but feel free to ignore if it doesn't make sense to you.

lukemelia (approved):

PR #3603, opened 2023-02-06
CLI create claim-settlement safe (using paymentScheduler work)

tintinthong:
`claim-settlement safe = reward program safe`. This pr adds cli functions to 

 - create a safe with claim-settlement module and guard attached 
 - enable module on existing safe

```
 Safe deployed to: 0x0110d49dF753b293d85c8a2A3820EDfdfc178eE8                                                                          
 Claim settlement module deployed to: 0xE71F7445A69a2D09d43dE99BE72be79a2C0BD94b                                                      
 Meta guard deployed to: 0x0D6b08Bea6D7A1CAA15c5E32fFB722A3DCEdD9eF 
```

EDIT: For some reason, the guard transaction multisend is failing. This leads to the `setUp` command not running when configuring the module which makes the module unusable. Heres an example [transaction](https://goerli.etherscan.io/tx/0xc8be55a9020bce961b8b81bfbc8e7bdb6d15092bbc2e223794acb03c622da5a3#internal). But it works when I remove guard from the multisend -- still investigating. 

I have manually enabled below so as to not get blocked on my work

```
Claim settlement module enabled: 0x76cA4D29a9F7eb11E21C808E627A24e1eb41C7c5
```


packages/cardpay-sdk/sdk/utils/module-utils.ts
```diff
@@ -18,7 +18,8 @@ export interface SetupArgs {
 export async function deployAndSetUpModule(
   ethersProvider: JsonRpcProvider,
   masterCopy: Contract,
-  setupArgs: SetupArgs
+  setupArgs: SetupArgs,
+  salt = 'cardstack-sp-deploy-module'
```
IanCal, line 22: If we're adding a salt, I think this should not default to a specific module, but be added by the calling code. SP here is I assume scheduled payments?

packages/cardpay-sdk/sdk/safe-module.ts
```diff
@@ -0,0 +1,417 @@
...
+      AddressZero,
+      '0x',
+      AddressZero,
+      AddressZero,
+      '0',
+      AddressZero,
+      generateSaltNonce(this.salt)
+    );
+    let enableModuleTxs = await this.generateEnableModuleTxs(expectedSafeAddress, [from]);
+    let setGuardTxs = await this.generateSetGuardTxs(expectedSafeAddress);
```
FadhlanR, line 95: Do you need a meta guard for this module?

tintinthong: No. I don't. I can refactor it out. But I added it because I copied most of the code from payment scheduler It is causing the multisend to fail. Do you have any idea why the transaction failed 

If you see this address 0xE71F7445A69a2D09d43dE99BE72be79a2C0BD94b 
there are failures here https://goerli.etherscan.io/address/0xE71F7445A69a2D09d43dE99BE72be79a2C0BD94b#internaltx

tintinthong: Ive added a function interface that doesn't add metaguard here https://github.com/cardstack/cardstack/pull/3603/commits/74de51cb6fa252f7860325fee33a6b7f3663e9a8

tintinthong: Here is the succesful [transaction](https://goerli.etherscan.io/tx/0xe707b0149839d7881c5085f565f440d14e0a09e5171c9950c0d404bc59d25f99)

FadhlanR (approved):
Looks great!

packages/cardpay-sdk/sdk/utils/module-utils.ts
tintinthong: I resolve this [here](https://github.com/cardstack/cardstack/pull/3603/commits/67151fb15e51078fa729fe1fe464c659e0e49cab)

PR #3604, opened 2023-02-06
Include relay/crank balance checker in scheduled payments data check

jurgenwerk:
This PR adds some basic monitoring to see if our relayer and the crank have enough balance to fund scheduling tool transactions (scheduling and executing). If the balance is too low to pay for transactions gas, the scheduling or executing will stop working, so we need to make sure we are informed when topping up is needed. 

We are running a Checkly check that continuously makes requests to our data integrity checks endpoint, which includes scheduled payments check. This PR adds the balance check. If the check discovers the balance is low, we will get alarmed in our discord channels and also in PagerDuty (for production). 

The low balance threshold is currently set at 0.05 of native coin value (e.g ETH, MATIC). There will also be a process described where a person in charge will manually transfer the funds from the fee receiver to the crank/relayer when this automatic check will trigger an alarm. 

**To discuss** (I am working through these questions now but comments welcome)

Is 0.05 a good threshold for both relayer and the crank? Does that vary on different networks significantly? Ideally the threshold should be at least a multiplication of the gas cost. Would it be good to introduce separate thresholds for relayer and the crank?

### DIFF
--- a/packages/hub/services/data-integrity-checks/scheduled-payments.ts
+++ b/packages/hub/services/data-integrity-checks/scheduled-payments.ts
@@ -1,6 +1,10 @@
 import { inject } from '@cardstack/di';
+import config from 'config';
 import { subMinutes } from 'date-fns';
 import { nowUtc } from '../../utils/dates';
+import fetch from 'node-fetch';
+import { getConstantByNetwork } from '@cardstack/cardpay-sdk';
+import { BigNumber, ethers, Wallet } from 'ethers';
 
 export const CREATION_WITHOUT_TX_HASH_ALLOWED_MINUTES = 2;
 export const CREATION_UNMINED_ALLOWED_MINUTES = 3 * 60;
@@ -9,6 +13,15 @@ export const CANCELATION_UNMINED_ALLOWED_MINUTES = CREATION_UNMINED_ALLOWED_MINU
 export const UNATTEMPTED_ALLOWED_MINUTES = 24 * 60;
 export const ATTEMPTS_WITHOUT_TX_HASH_ALLOWED_MINUTES = 60;
 export const UNFINISHED_ATTEMPT_DURATION_ALLOWED_MINUTES = 24 * 60;
+
+// Relayer needs funds to pay for safe transactions gas (e.g creating safes, scheduling a payment).
+// Gas cost reimbursement for the relayer will be paid from the (specified) gas token balance in the user's safe.
+const LOW_RELAYER_FUNDS_THRESHOLD = ethers.utils.parseEther('0.05');
+
+// Crank (i.e the hub) needs funds to pay for scheduled payment execution transactions gas.
+// This gets reimbursed to the fee receiver using platform defined fees collected from the
+// scheduled payment (for example $0.25 USD and 0.1% of the transaction amount)
+const LOW_CRANK_FUNDS_THRESHOLD = LOW_RELAYER_FUNDS_THRESHOLD;
 
 function addToMessages(collection: any, message: string, messages: string[]) {
   if (collection.length == 0) return;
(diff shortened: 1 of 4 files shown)


packages/hub/services/data-integrity-checks/scheduled-payments.ts
```diff
@@ -10,6 +14,15 @@ export const UNATTEMPTED_ALLOWED_MINUTES = 24 * 60;
 export const ATTEMPTS_WITHOUT_TX_HASH_ALLOWED_MINUTES = 60;
 export const UNFINISHED_ATTEMPT_DURATION_ALLOWED_MINUTES = 24 * 60;
 
+// Relayer needs funds to pay for safe transactions gas (e.g creating safes, scheduling a payment).
```
jurgenwerk, line 17: Note to self: remove "creating safes" – that actually comes from the user's wallet!

packages/hub/services/data-integrity-checks/scheduled-payments.ts
```diff
@@ -10,6 +14,15 @@ export const UNATTEMPTED_ALLOWED_MINUTES = 24 * 60;
...
+// Gas cost reimbursement for the relayer will be paid from the (specified) gas token balance in the user's safe.
+const LOW_RELAYER_FUNDS_THRESHOLD = ethers.utils.parseEther('0.05');
+
+// Crank (i.e the hub) needs funds to pay for scheduled payment execution transactions gas.
+// This gets reimbursed to the fee receiver using platform defined fees collected from the
+// scheduled payment (for example $0.25 USD and 0.1% of the transaction amount)
```
FadhlanR, line 23: And also gets gas reimbursement for the execution process.

FadhlanR (approved):

jurgenwerk:
Per our discussion yesterday we decided a good threshold for now would be something like expected users (500) * 50 transactions/month * avg transaction gas cost. 

However this sounds like a huge amount for networks with high gas fees, for example Ethereum. In that case the avg gas fee per tx has been around [0.6 USD](https://ycharts.com/indicators/ethereum_average_transaction_fee#:~:text=Ethereum%20Average%20Transaction%20Fee%20is,61.40%25%20from%20one%20year%20ago) which is around 0.0003 ETH (avg gas fee for an Ethereum transaction). If we calculate the threshold, we get: 500 * 50 * 0.0003 = 7.5 ETH. 

In case of Polygon, the average gas fee per tx seems to be around 0.03 MATIC. Plugging that into the formula we get 500 * 50 * 0.03 = 750 MATIC

I think maintaining ~7.5 ETH and ~750 MATIC balance for both crank and relayer sounds quite a lot for now when the user activity is still very low. 

I would suggest to simply hardcode values for supported network and revisit them later if scaling is needed.

1. Ethereum (mainnet and goerli): 0.3 ETH 
2. Polygon: 30 MATIC 

This would still suffice for ~1000 transactions once we get a warning.

packages/hub/services/data-integrity-checks/scheduled-payments.ts
```diff
@@ -152,12 +166,53 @@ export default class DataIntegrityChecksScheduledPayments {
...
+        );
+      }
+    }
+  }
+
+  async getRelayerFunderBalance(networkName: string): Promise<BigNumber> {
+    let relayerUrl = getConstantByNetwork('relayServiceURL', networkName);
+
+    let responseText = await (await fetch(`${relayerUrl}/v1/about`)).text();
+    let relayerFunderPublicKey = JSON.parse(responseText).settings.SAFE_FUNDER_PUBLIC_KEY;
```
jurgenwerk, line 205: Note to self: SAFE_FUNDER_PUBLIC_KEY is wrong, SAFE_TX_SENDER_PUBLIC_KEY is correct

PR #3607, opened 2023-02-06
Add helper text to scheduled payment recipient form

lukemelia:


packages/safe-tools-client/app/components/schedule-payment-form-action-card/ui/index.gts
```diff
@@ -234,6 +234,7 @@ export default class SchedulePaymentFormActionCardUI extends Component<Signature
             data-test-payee-address-input
             @value={{@payeeAddress}}
             @invalid={{this.isPayeeAddressInvalid}}
+            @helperText="Any, including EOA and safe. Use caution sending to exchange deposit addresses."
```
jurgenwerk, line 237: Should we add that "When sending to an exchange address, make sure it's an ERC20 deposit address"? @FadhlanR  is that correct to say?

lukemelia: I'm concerned this is already too wordy.

FadhlanR: @jurgenwerk I think that information should be provided by the exchange.

FadhlanR (approved):

packages/safe-tools-client/app/components/schedule-payment-form-action-card/ui/index.gts
jurgenwerk: Exhanges give you an option to generate deposit addresses that can be ERC20 or something else. So in this case I believe the user needs to generate an ERC20 deposit address in the exchange if they want to transfer it there.

PR #3610, opened 2023-02-07
Fix scheduled payment fees

FadhlanR:
Ticket: CS-5227

Executed payment with fees: https://goerli.etherscan.io/tx/0x99c0721a468fac67e310f61c269e8f7527b75a0ade1341a5eb581c9f856708bd

### DIFF
--- a/packages/cardpay-sdk/sdk/scheduled-payment-module.ts
+++ b/packages/cardpay-sdk/sdk/scheduled-payment-module.ts
@@ -5,7 +5,7 @@ import MetaGuardABI from '../contracts/abi/modules/meta-guard';
 import ScheduledPaymentABI from '../contracts/abi/modules/scheduled-payment-module';
 import ScheduledPaymentConfigABI from '../contracts/abi/modules/scheduled-payment-config';
 import { getAddress } from '../contracts/addresses';
-import { isAddress } from 'web3-utils';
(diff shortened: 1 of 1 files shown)


packages/cardpay-sdk/sdk/scheduled-payment-module.ts
```diff
@@ -1384,4 +1339,19 @@ export default class ScheduledPaymentModule {
     const scheduledPaymentConfig = new Contract(configModuleAddress, ScheduledPaymentConfigABI, this.ethersProvider);
     return await scheduledPaymentConfig.validForDays();
   }
+
+  private composeFees(
+    feeFixedUSD: number,
+    feePercentage: number
+  ): { fixedUSD: { value: string }; percentage: { value: string } } {
```
danibonilha, line 1346: nit: typescript is smart enough to infer the return of this function, but if we do want to enforce it, I'd suggest to extract this type to improve readability.

danibonilha (approved):
Looks good, and props on creating the composeFee function. I've just added a nit comment.

packages/cardpay-sdk/sdk/scheduled-payment-module.ts
FadhlanR: I choose to remove it. Thanks

PR #3617, opened 2023-02-08
safe client: Reload token balances: continuously, on schedule, on cancel

jurgenwerk:
When user schedules or cancels a payment, their gas token balance decreases and we want this to be reflected in the sidebar, so we want to reload token balances when these actions are performed.

There is also a case where the safe has a change in token balances independently of the actions taken in this app. For this we introduce continuous refreshing of the balances - it refreshes every 30 seconds.

### DIFF
--- a/packages/safe-tools-client/app/services/safes.ts
+++ b/packages/safe-tools-client/app/services/safes.ts
@@ -10,6 +10,9 @@ import WalletService from '@cardstack/safe-tools-client/services/wallet';
 import { action } from '@ember/object';
 import Service, { inject as service } from '@ember/service';
 import { tracked } from '@glimmer/tracking';
+import { TaskGenerator, timeout } from 'ember-concurrency';
+import { task } from 'ember-concurrency-decorators';
+import { taskFor } from 'ember-concurrency-ts';
 import { use, resource } from 'ember-resources';
 import { BigNumber } from 'ethers';
 import { TrackedObject } from 'tracked-built-ins';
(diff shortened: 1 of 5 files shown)


packages/safe-tools-client/app/services/safes.ts
```diff
@@ -45,6 +46,16 @@ export default class SafesService extends Service {
 
   @tracked selectedSafe?: Safe;
 
+  constructor(properties?: object | undefined) {
+    super(properties);
+
+    // We keep reloading the token balances so that they are up do date
+    // when user adds funds to the safe independently of the app
+    setInterval(() => {
```
lukemelia, line 54: Consider using a while loop with yield in an ember-concurrency task, so that the task is canceled when the service is torn town.

jurgenwerk: I was actually considering this but does a service ever get torn down? Aren't they just singletons that live until the app is running?

lukemelia: Yes, but not in tests, and that will drive you crazy with weird errors.

jurgenwerk: I see! Ok, will adjust it.

jurgenwerk: Added the change, please check.

lukemelia: Looks good. I left one small comment. Address if you agree and good to merge.

lukemelia (approved):

packages/safe-tools-client/app/services/safes.ts
lukemelia: @jurgenwerk I saw that there were still issues and I pushed a commit to your branch that fixes them.

jurgenwerk:
@lukemelia thank you for fixing the test situation! From that I have learned that `rawTimeout` will use the native timeout instead of the runloop, meaning the test helpers will not wait for it to complete

PR #3621, opened 2023-02-09
Use chainId because network name is unreliable; ethers doesn't recognize "polygon" and still uses "matic"

jurgenwerk:
Getting this error in production:

```
hub/server Unhandled error: Error: invalid network (argument="network", value="polygon", code=INVALID_ARGUMENT, version=providers/5.7.2)
    at Logger.makeError (/workspace/packages/hub/dist/hub.js:93630:23)
    at Logger.throwError (/workspace/packages/hub/dist/hub.js:93639:20)
    at Logger.throwArgumentError (/workspace/packages/hub/dist/hub.js:93642:21)
    at new BaseProvider (/workspace/packages/hub/dist/hub.js:95074:24)
    at new JsonRpcProvider (/workspace/packages/hub/dist/hub.js:98813:9)
    at new JsonRpcProvider (/workspace/packages/hub/dist/hub.js:665365:9)
    at EthersProvider.getInstance (/workspace/packages/hub/dist/hub.js:681443:16)
    at DataIntegrityChecksScheduledPayments.getRelayerFunderBalance (/workspace/packages/hub/dist/hub.js:680526:44)
```

Ethers.js still uses "matic" even though the network renamed to "polygon". We use "polygon" in our config so that makes instantiating providers using the name unreliable. It is better to use chain id because that doesn't change. 

Proof: https://github.com/ethers-io/ethers.js/blob/0bf53d7804109f5c0322d8c9a0c10d73abc84136/dist/ethers.js#L17925 (thanks @FadhlanR for finding this!)

### DIFF
--- a/packages/hub/services/data-integrity-checks/scheduled-payments.ts
+++ b/packages/hub/services/data-integrity-checks/scheduled-payments.ts
@@ -216,17 +216,19 @@ export default class DataIntegrityChecksScheduledPayments {
 
   async getRelayerFunderBalance(networkName: string): Promise<BigNumber> {
     let relayerUrl = getConstantByNetwork('relayServiceURL', networkName);
+    let chainId = getConstantByNetwork('chainId', networkName);
 
     let responseText = await (await fetch(`${relayerUrl}/v1/about`)).text();
     let relayerFunderPublicKey = JSON.parse(responseText).settings.SAFE_TX_SENDER_PUBLIC_KEY;
 
-    let provider = this.ethersProvider.getInstance(networkName);
+    let provider = this.ethersProvider.getInstance(chainId);
     return await provider.getBalance(relayerFunderPublicKey);
   }
 
   async getCrankBalance(networkName: string): Promise<BigNumber> {
+    let chainId = getConstantByNetwork('chainId', networkName);
     let crank = new Wallet(config.get('hubPrivateKey'));
-    let provider = this.ethersProvider.getInstance(networkName);
+    let provider = this.ethersProvider.getInstance(chainId);
     return await provider.getBalance(crank.address);
   }
 }
--- a/packages/hub/services/ethers-provider.ts
+++ b/packages/hub/services/ethers-provider.ts
@@ -2,7 +2,7 @@ import config from 'config';
 import { getWeb3ConfigByNetwork, JsonRpcProvider } from '@cardstack/cardpay-sdk';
 
 export default class EthersProvider {
-  getInstance(chainId: number | string) {
+  getInstance(chainId: number) {
     let rpcUrl = getWeb3ConfigByNetwork({ web3: config.get('web3') }, chainId).rpcNodeHttpsUrl;
     return new JsonRpcProvider(rpcUrl, chainId);
   }


jurgenwerk:
Actually, I'm not sure this will work given getWeb3ConfigByNetwork converts chain id back to network name 🤔

https://github.com/cardstack/cardstack/blob/main/packages/hub/services/ethers-provider.ts#L6
https://github.com/cardstack/cardstack/blob/main/packages/cardpay-sdk/sdk/network-config-utils.ts#L12

jurgenwerk:
Confirmed in the production console:

```
Hub > s.ethersProvider.getInstance(137).network.name
'matic'
Hub > s.ethersProvider.getInstance("polygon").network.name
Uncaught:
Error: invalid network (argument="network", value="polygon", code=INVALID_ARGUMENT, version=providers/5.7.2)
    at Logger.makeError (/workspace/packages/hub/dist/hub.js:93630:23)
    at Logger.throwError (/workspace/packages/hub/dist/hub.js:93639:20)
    at Logger.throwArgumentError (/workspace/packages/hub/dist/hub.js:93642:21)
    at new BaseProvider (/workspace/packages/hub/dist/hub.js:95074:24)
    at new JsonRpcProvider (/workspace/packages/hub/dist/hub.js:98813:9)
    at new JsonRpcProvider (/workspace/packages/hub/dist/hub.js:665365:9)
    at EthersProvider.getInstance (/workspace/packages/hub/dist/hub.js:681443:16) {
  reason: 'invalid network',
  code: 'INVALID_ARGUMENT',
  argument: 'network',
  value: 'polygon'
}
```

IanCal:
This could be a quick fix, however it looks like the kind of thing that could come up a lot - lots of the lookup code uses polygon I think (finding it a little hard to track around the place).

The config block here:

[getWeb3ConfigByNetwork({ web3: config.get('web3') }, chainId)](https://github.com/cardstack/cardstack/blob/0a20ff65eeab62b9d09d10d8e03c2d6dfca6037b/packages/hub/services/ethers-provider.ts#LL6C18-L6C79)

I think is https://github.com/cardstack/cardstack/blob/0a20ff65eeab62b9d09d10d8e03c2d6dfca6037b/packages/cardpay-sdk/sdk/constants.ts#L225

Has the chain ID as well, so we could pass in the RPC and chainID explicitly here https://github.com/cardstack/cardstack/blob/0a20ff65eeab62b9d09d10d8e03c2d6dfca6037b/packages/hub/services/ethers-provider.ts#L7

Would that just work with passing polygon then?

Do we want to make a larger change around the SDK so things match up otherwise? There might be other places we make this mistake.

jurgenwerk:
@IanCal since we did a successful end to end test using polygon probably other places are wired up correctly. However I share your concern this setup is brittle given network names can change. To avoid these situations completely I think we should base our config lookups on chain ids rather than on network names like we do now. Probably that's a larger task so I'd merge this fix first.

IanCal:
@jurgenwerk Agree on avoiding it entirely

I think the other part is working because it gets the chainID explicitly too:

https://github.com/cardstack/cardstack/blob/0a20ff65eeab62b9d09d10d8e03c2d6dfca6037b/packages/hub/services/scheduled-payments/executor.ts#L28

I don't think the problem is network names changing, it's a mismatch between our naming and ethers. Currently it looks like the hub works with names or chain ids *unless* it's polygon where it'll only work with a chain id. I think if we accept either for our provider, but then only use chain ids when calling off to ethers we can avoid that (as well as future changes to names).

IanCal:
We're proposing extremely similar things, I think we can just move the chainid lookup into our ethers provider class.

jurgenwerk:
@IanCal I reduced network type scope (down to a number) for getting the provider instance as a first line of defense against errors like this - I'll merge this now and we can revisit later. Thanks!

IanCal:
Oh it's only used in scheduled payments, limiting to chainid sounds like a good solution then, I thought it was used across the hub.

PR #3622, opened 2023-02-09
Re-authenticate when account changed

FadhlanR:
Ticket: CS-5223

- [x] Ask for authentication if the user changes the account.
- [x]  Store multiple auth tokens from different accounts. So we don't need to ask for authentication if the account's token is still valid.

https://user-images.githubusercontent.com/12637010/217798472-7f86be85-468b-47b4-ad92-386fc5d3f75d.mov


packages/safe-tools-client/app/services/hub-authentication.ts
```diff
@@ -49,26 +48,36 @@ export default class HubAuthenticationService extends Service {
...
+    if (!this.wallet.address || !authTokens) return null;
+    const authTokensObj = JSON.parse(authTokens);
+    return authTokensObj[this.wallet.address];
   }
 
   set authToken(val: string | null) {
     if (val) {
-      this.storage.setItem('authToken', val);
+      const authTokens = this.storage.getItem('authTokens');
+      const authTokensObj = authTokens ? JSON.parse(authTokens) : {};
```
jurgenwerk, line 60: I think it's safer if we avoid stringifying and parsing an object and store individual entries instead

The keys could be `auth-token-${walletAddress}-${network}` for more clarity and easier debugging/inspection

packages/safe-tools-client/app/services/hub-authentication.ts
```diff
@@ -49,26 +48,36 @@ export default class HubAuthenticationService extends Service {
...
-    );
+    try {
+      const authAddress = this.authToken
+        ? await hubAuth.getAddress(this.authToken)
+        : null;
+      return Boolean(
+        this.wallet.isConnected && authAddress === this.wallet.address
+      );
+    } catch (e) {
+      return false;
```
jurgenwerk, line 79: Is this catch necessary? 

Will this nullify the token in case there are connection issues to the hub?

packages/safe-tools-client/app/services/hub-authentication.ts
FadhlanR: updated

packages/safe-tools-client/app/services/hub-authentication.ts
FadhlanR: Updated by catching a specific error message to return false.

packages/safe-tools-client/app/services/hub-authentication.ts
```diff
@@ -48,27 +49,38 @@ export default class HubAuthenticationService extends Service {
...
-    );
+    try {
+      const authAddress = this.authToken
+        ? await hubAuth.getAddress(this.authToken)
+        : null;
+      return Boolean(
+        this.wallet.isConnected && authAddress === this.wallet.address
+      );
+    } catch (e) {
+      if (e === 'Invalid auth token') {
```
jurgenwerk, line 79: Isn't `e` an `Error` type? Did you mean to use `e.message`?

packages/safe-tools-client/app/services/hub-authentication.ts
```diff
@@ -48,27 +49,38 @@ export default class HubAuthenticationService extends Service {
...
     const hubAuth = await this.getHubAuth();
-
-    return Boolean(
-      this.wallet.isConnected &&
-        this.authToken &&
-        (await hubAuth.checkValidAuth(this.authToken))
-    );
+    try {
+      const authAddress = this.authToken
+        ? await hubAuth.getAddress(this.authToken)
```
jurgenwerk, line 73: I'm confused why fetching the address from the hub is needed. Why not read the address from the wallet service, construct the auth storage key, use it to get the auth token from the storage and perform `checkValidAuth(authToken)`?

FadhlanR: How do we validate the auth token belongs to which address?

packages/safe-tools-client/app/services/hub-authentication.ts
FadhlanR: fixed

packages/safe-tools-client/app/services/hub-authentication.ts
jurgenwerk: Doesn't `await hubAuth.checkValidAuth(this.authToken)` suffice to see if the token is valid?

FadhlanR: Only checking whether the auth token is valid or not. Not checking is that token is used by the right address or not.

jurgenwerk: I see, you're addressing a case where someone could steal your token and plant it in their own localstorage, and prevent communicating to the hub in case the wallet address doesn't match the address encrypted in the token. 

I guess it doesn't hurt to have that extra layer of security.  However that will only work for until the token expires (1 day). And the attacker can easily use the leaked token in curl or use some other http client until the token expires

FadhlanR: Removed the getAddress function.

jurgenwerk (approved):

PR #3624, opened 2023-02-09
chore: Add missing env var for reward-submit-lambda

pcjun97:


jurgenwerk:
@pcjun97 can we merge this?

PR #3626, opened 2023-02-09
Handle time window where the payment does not show up in any lists

lukemelia:


jurgenwerk:
I think there's another problem:

https://github.com/cardstack/cardstack/blob/1a96cacfd5aadc9efee680f7caa50a830b57f4bd/packages/safe-tools-client/app/services/scheduled-payments.ts#L216

After a payment has been scheduled, the list will call the above which might not return the newly scheduled payment since its `payAt` could be lower than the current timestamp. This can happen in the edge case described in the ticket (submitting the form after `payAt` has already passed). Maybe the simplest solution for this would be to add validation that `payAt` must be in the future when scheduling.

lukemelia:
> After a payment has been scheduled, the list will call the above which might not return the newly scheduled payment since its payAt could be lower than the current timestamp

Isn't this query cutting off old scheduled payments (i.e. those that have exceeded validForDays), not cutting off newish ones (whether somewhat in the past or the future)? (I've been getting these data comparisons backwards a lot, so I could be wrong!)

jurgenwerk:
> Isn't this query cutting off old scheduled payments (i.e. those that have exceeded validForDays), not cutting off newish ones (whether somewhat in the past or the future)?

You mean the query for the "Future payments"? Actually that query is just filtering payments by `payAt` and `chainId`.

jurgenwerk:
To cover that edge case I suggest invalidating the form in case the `paymentDate` is larger than the current timestamp. I can work on that but I probably won't be able to work on that until Friday.

jurgenwerk (approved):

PR #3627, opened 2023-02-10
[CS-5239]: Block submission if usdcConversion doesn't exist

danibonilha:
This PR validates the `usdcToGasTokenRate`, at first I went to the most complicated way to handle it, but just this flag seems to suffice, let me know if i'm missing something.


jurgenwerk (approved):
This solution looks good to me

PR #3628, opened 2023-02-10
[CS-5233]: Do not show native token if it's zero

danibonilha:
Showing the empty native token gave an impression that the user should have this value, to avoid confusion this PR hides if there is no balance.


FadhlanR (approved):

lukemelia (approved):
Should we have a test confirm this?

danibonilha:
> Should we have a test confirm this?

I'll add a ticket for that

PR #3630, opened 2023-02-10
safe tools: Fix displaying payment execution error messages

jurgenwerk:
This PR addresses the bug where the payment execution failure was not displayed:

image

It turns out it didn't even get written to the scheduled payment record in the hub, due to an error in the code.

### DIFF
--- a/packages/cardpay-sdk/sdk/scheduled-payment-module.ts
+++ b/packages/cardpay-sdk/sdk/scheduled-payment-module.ts
@@ -1278,7 +1278,7 @@ export default class ScheduledPaymentModule {
       // ExceedMaxGasPrice: gasPrice must be lower than or equal maxGasPrice
       // PaymentExecutionFailed: safe balance is not enough to make payments and pay fees
       // OutOfGas: executionGas to low to execute scheduled payment
-      throw extractSendTransactionError(e, new Interface(ScheduledPaymentABI));
+      throw new Error(extractSendTransactionError(e, new Interface(ScheduledPaymentABI)));
     }
   }
 
--- a/packages/cardpay-sdk/sdk/scheduled-payment-module.ts
+++ b/packages/cardpay-sdk/sdk/scheduled-payment-module.ts
@@ -14,7 +14,7 @@ import {
 } from './utils/module-utils';
 import {
   extractBytesLikeFromError,
-  extractSendTransactionError,
+  extractSendTransactionErrorMessage,
   generateSaltNonce,
   hubRequest,
   isTransactionHash,
@@ -1278,7 +1278,7 @@ export default class ScheduledPaymentModule {
       // ExceedMaxGasPrice: gasPrice must be lower than or equal maxGasPrice
       // PaymentExecutionFailed: safe balance is not enough to make payments and pay fees
       // OutOfGas: executionGas to low to execute scheduled payment
-      throw new Error(extractSendTransactionError(e, new Interface(ScheduledPaymentABI)));
+      throw new Error(extractSendTransactionErrorMessage(e, new Interface(ScheduledPaymentABI)));
     }
   }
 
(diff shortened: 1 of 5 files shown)


packages/cardpay-sdk/sdk/scheduled-payment-module.ts
```diff
@@ -1278,7 +1278,7 @@ export default class ScheduledPaymentModule {
       // ExceedMaxGasPrice: gasPrice must be lower than or equal maxGasPrice
       // PaymentExecutionFailed: safe balance is not enough to make payments and pay fees
       // OutOfGas: executionGas to low to execute scheduled payment
-      throw extractSendTransactionError(e, new Interface(ScheduledPaymentABI));
+      throw new Error(extractSendTransactionError(e, new Interface(ScheduledPaymentABI)));
```
jurgenwerk, line 1281: Because we were throwing a string here, this piece of code in the executor was trying to call `message` on a string, which resulted in a blank value:

https://github.com/cardstack/cardstack/blob/b19a1f838a2d95de25cdf43324a5de431c6a2d65/packages/hub/services/scheduled-payments/executor.ts#L160

packages/safe-tools-client/app/helpers/payment-error-message.ts
```diff
@@ -0,0 +1,28 @@
...
+type PositionalArgs = [string];
+
+interface Signature {
+  Args: {
+    Positional: PositionalArgs;
+  };
+  Return: string;
+}
+
+export function paymentErrorMessage([failureReason]: PositionalArgs) {
```
jurgenwerk, line 12: Known revert reasons can be found here:

https://github.com/cardstack/cardstack/blob/b19a1f838a2d95de25cdf43324a5de431c6a2d65/packages/cardpay-sdk/sdk/scheduled-payment-module.ts#L1273-L1280

packages/safe-tools-client/app/helpers/payment-error-message.ts
```diff
@@ -0,0 +1,28 @@
...
+  // Try to match the error message to a known error explained in a way that makes sense to the user
+  if (failureReason) {
+    if (failureReason.includes('InvalidPeriod'))
+      return 'payment attempted at the wrong time'; // Could be the crank's fault. Not much the user can do about it except report it to support
```
jurgenwerk, line 16: Not sure what to tell the user here, looking for ideas. It's actually quite awkward to say that the payment was attempted at the wrong time since this is the core feature of this tool. Maybe this should go into the "unknown error" category instead and we urge the user to contact support, especially since we don't expect this error to happen often, only if there is a scheduling misconfiguration at our side.

packages/safe-tools-client/app/helpers/payment-error-message.ts
```diff
@@ -0,0 +1,28 @@
...
+    if (failureReason.includes('ExceedMaxGasPrice'))
+      return 'current gas price exceeds the maximum specified';
+    if (failureReason.includes('PaymentExecutionFailed'))
+      return 'insufficient funds to execute the payment'; // This error covers 3 different failure scenarios and it is currently not possible to know which we are dealing with - insufficient funds for either of the: transfer amount, gas fee, service fee
+    if (failureReason.includes('OutOfGas'))
+      return 'execution gas is too low to execute the payment'; // Probably our fault where we estimate execution gas. Not much the user can do about it except report it to support
```
jurgenwerk, line 22: This is also a bit weird to say to the user especially since they can't do anything about it to fix it. I believe this means we underestimated amount of gas units needed for the execution and it's unlikely this payment will succeed even with retries. @FadhlanR can you confirm?

danibonilha (approved):

packages/safe-tools-client/app/helpers/payment-error-message.ts
FadhlanR: Right, in this kind of error, the user can only cancel the scheduled payment and make another one.

FadhlanR (approved):

PR #3632, opened 2023-02-13
[CS-5244]: Add option to cancel failed payment

danibonilha:
This PR extracts the cancel option in its own component `PaymentOptionsDropdown` and adds it to the failed row. It also combine similar types (SchedulePayment) to keep consistency across data structure. 

image

PS: Even though it adds a canceled state, the hub doesn't return it yet, so it won't be reflected until the hub ticket is complete.


FadhlanR (approved):
LGTM

jurgenwerk (approved):
Looks good

lukemelia (approved):

PR #3633, opened 2023-02-13
Helper functions for dealing with the ClaimSettlement module & EIP712 signing

IanCal:
This is just moving the utils off from the module code to the SDK. The tests will then import this.


tintinthong (approved):
Nice

PR #3634, opened 2023-02-15
Fix reconnection issue

FadhlanR:
I fixed this issue by adding a provider to local storage in the `chain-changed` event, not in the `connect` event. Because in the `connect` event the chain id might be still using the default value, not the chain id that the users use in their wallet.

### DIFF
--- a/packages/safe-tools-client/app/utils/chain-connection-manager.ts
+++ b/packages/safe-tools-client/app/utils/chain-connection-manager.ts
@@ -231,7 +231,6 @@ export class ChainConnectionManager {
 
   @action onConnect(accounts: string[]) {
     if (!this.strategy) return;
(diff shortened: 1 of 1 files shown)


jurgenwerk:
> Because in the connect event the chain id might be still using the default value, not the chain id that the users use in their wallet

Hm, what circumstances lead to this? I find it odd that when the wallet connects (on `connect` event), the chain id is different from the one in the wallet.

FadhlanR:
On the `connect` event we haven't read the network from the user's wallet but we still use the default network on `network.ts`

lukemelia (approved):
Is it possible to have test coverage for this?

FadhlanR:
@lukemelia It's a bit complex to add the test coverage for this case. If it is needed, we can create the ticket later.

PR #3637, opened 2023-02-17
Add scheduled payment logging for debugging purposes

jurgenwerk:
Ticket: CS-5293

We are hunting a bug where the payment execution reverts with `UnknownHash` error. That probably means the scheduled payment module does not have the payment params registered with which the crank wants to execute the payment (encoded in `spHash`).

There is a possibility params sent to `createSpHash` when scheduling are different when compared to the params sent to `executeScheduledPayment`. This PR adds logging to help with debugging - we should see the params in Sentry. 

We also have another bug where the contract reverts with `InvalidPeriod` - this logging might be useful for that case too.

### DIFF
--- a/packages/safe-tools-client/app/components/schedule-payment-form-action-card/index.gts
+++ b/packages/safe-tools-client/app/components/schedule-payment-form-action-card/index.gts
@@ -378,22 +378,41 @@ export default class SchedulePaymentFormActionCard extends Component<Signature> 
 
     this.scheduleErrorMessage = undefined;
 
-    try {
-      const { currentSafe } = this.safes;
+    const { currentSafe } = this.safes;
 
+    let payAt = this.selectedPaymentType === 'one-time' ? Math.round(this.paymentDate!.getTime() / 1000) : null;
+    let recurringDayOfMonth = this.selectedPaymentType === 'monthly' ? this.paymentDayOfMonth! : null;
+    let recurringUntil = this.selectedPaymentType === 'monthly' ? Math.round(this.monthlyUntil!.getTime() / 1000) : null
+
+    let params = {
+      safeAddress: currentSafe.address,
+      spModuleAddress: currentSafe.spModuleAddress,
+      tokenAddress: this.paymentTokenQuantity.address,
+      amount: this.paymentTokenQuantity.count,
+      payeeAddress: this.payeeAddress,
+      executionGas: Number(this.gasEstimation.gas),
+      maxGasPrice: maxGasPriceString,
+      gasTokenAddress: this.selectedGasToken.address,
+      salt,
+      payAt,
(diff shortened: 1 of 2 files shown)


packages/safe-tools-client/app/components/schedule-payment-form-action-card/index.gts
```diff
@@ -378,22 +378,41 @@ export default class SchedulePaymentFormActionCard extends Component<Signature>
...
     this.scheduleErrorMessage = undefined;
 
-    try {
-      const { currentSafe } = this.safes;
+    const { currentSafe } = this.safes;
+
+    let payAt = this.selectedPaymentType === 'one-time' ? Math.round(this.paymentDate!.getTime() / 1000) : null;
+    let recurringDayOfMonth = this.selectedPaymentType === 'monthly' ? this.paymentDayOfMonth! : null;
+    let recurringUntil = this.selectedPaymentType === 'monthly' ? Math.round(this.monthlyUntil!.getTime() / 1000) : null
+
```
danibonilha, line 386: I'm wondering why on the executePayment on hub we don't round the value, but here in the client we do, could this cause an inconsistency  ?  Maybe we could add an util on the sdk that sort of formats that so it's the same on multiple place idk

packages/safe-tools-client/app/components/schedule-payment-form-action-card/index.gts
```diff
@@ -378,22 +378,41 @@ export default class SchedulePaymentFormActionCard extends Component<Signature>
...
+        params.tokenAddress,
+        params.amount,
+        params.payeeAddress,
+        params.executionGas,
+        params.maxGasPrice,
+        params.gasTokenAddress,
+        params.salt,
+        params.payAt,
+        params.recurringDayOfMonth,
+        params.recurringUntil,
```
danibonilha, line 415: supernit:  if you want to avoid mapping the params by hand and they are in the correct order you could do something like: 
`perform(...Object.values(params))`

but I also understand that this might be too magical, so I'm ok if this suggestion is ignored

danibonilha (approved):

packages/safe-tools-client/app/components/schedule-payment-form-action-card/index.gts
jurgenwerk: Nice potential catch, it could be the issue. We'll see!

lukemelia (approved):

packages/safe-tools-client/app/components/schedule-payment-form-action-card/index.gts
jurgenwerk: Cool - that's good to know. I'm gonna merge and keep in mind when we revisit it.

PR #3642, opened 2023-02-20
Do not include canceled payments in the "payments that should have been attempted" check

jurgenwerk:
I saw this in Checkly:

image

But this payment had a "canceled" status. We don't want to consider such payments unattempted.


FadhlanR (approved):
Looks good

PR #3644, opened 2023-02-20
Handle blank view in view details column

FadhlanR:
Ticket: CS-5294

https://user-images.githubusercontent.com/12637010/220130223-34d32fbb-e074-44bb-adf4-c734798bf543.mov

### DIFF
--- a/packages/safe-tools-client/app/components/block-explorer-button/index.css
+++ b/packages/safe-tools-client/app/components/block-explorer-button/index.css
@@ -1,0 +1,33 @@
+.block-explorer-button-disabled {
+  position: relative;
+  pointer-events: unset !important;
+  border: none;
+}
+
+.block-explorer-button-disabled::after {
+  content: attr(data-hover);
+
+  opacity: 0;
+  background-color: var(--boxel-light);
+  box-shadow: 0 0 0 1px var(--boxel-light-500);
+  color: var(--boxel-dark);
+  font: var(--boxel-button-font, var(--boxel-font-sm));
+  text-align: center;
+  border-radius: var(--boxel-sp-sm);
+  padding: var(--boxel-sp-sm);
+  transition: opacity 1s ease-in-out;
+
+  position: absolute;
+  z-index: 1;
+  left: 0;
+  top: 110%;
+}
+
+.block-explorer-button-disabled:hover::after {
+  opacity: 1;
+  visibility: visible;
+}
+
+.block-explorer-button-disabled:hover {
+  cursor: default;
+}
--- a/packages/safe-tools-client/app/components/block-explorer-button/index.gts
+++ b/packages/safe-tools-client/app/components/block-explorer-button/index.gts
@@ -1,6 +1,8 @@
 import Component from '@glimmer/component';
 import BoxelButton from '@cardstack/boxel/components/boxel/button';
 import { getConstantByNetwork, TransactionHash } from '@cardstack/cardpay-sdk';
+import cn from '@cardstack/boxel/helpers/cn';
+import './index.css';
 
 interface Signature {
   Element: HTMLAnchorElement | HTMLButtonElement;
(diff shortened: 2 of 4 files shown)


jurgenwerk:
I think the design of the disabled button is too strong. I'd prefer a button with reduced opacity. Something like:

image

Can you show this together with payments with different statuses - confirmed, failed, canceled?

FadhlanR:
@jurgenwerk 
Screenshot 2023-02-22 at 14 38 22

packages/safe-tools-client/app/components/block-explorer-button/index.css
```diff
@@ -0,0 +1,36 @@
+.block-explorer-button-disabled {
```
jurgenwerk, line 1: This looks like a modifier to the `block-explorer-button`. Use the modifier naming: https://getbem.com/naming/

packages/safe-tools-client/app/components/block-explorer-button/index.gts
```diff
@@ -38,9 +41,12 @@ export default class BlockExplorerButton extends Component<Signature> {
       @as="anchor"
       @size="extra-small"
       @kind={{@kind}}
-      href={{this.blockExplorerUrl}} 
+      @href={{this.blockExplorerUrl}}
+      @disabled={{not @transactionHash}}
       target="_blank"
       rel="noopener"
+      class={{cn block-explorer-button-disabled=(not @transactionHash)}}
```
jurgenwerk, line 48: The class on this button when disabled should be:

`block-explorer-button block-explorer-button--disabled`

In BEM, `block-explorer-button` is a "block" and we want to modify it (make a new variant) for some special case. That's when the "modifier" class is added (`block-explorer-button--disabled`).

packages/safe-tools-client/app/components/block-explorer-button/index.css
```diff
@@ -0,0 +1,36 @@
+.block-explorer-button--disabled {
+  position: relative;
+  pointer-events: unset !important;
+
+  --boxel-button-color: transparent !important;
+  --boxel-button-border: 1px solid var(--boxel-purple-200) !important;
+  --boxel-button-text-color: var(--boxel-purple-300) !important;
+}
```
lukemelia, line 8: Is there an opportunity to improve BoxelButton rather than having to do custom styling for this?

FadhlanR: I updated this to use cssVar to do the styling. https://github.com/cardstack/cardstack/pull/3644/commits/fa4912045bb816562332767ada99d6a1ab550a82#diff-b08a86a460d3c830173b2f4d7e0708975a275c21930d7cbe0ff072593ea9a996R50

packages/safe-tools-client/app/components/block-explorer-button/index.gts
```diff
@@ -38,9 +42,17 @@ export default class BlockExplorerButton extends Component<Signature> {
...
       @kind={{@kind}}
-      href={{this.blockExplorerUrl}} 
+      @href={{this.blockExplorerUrl}}
+      @disabled={{not @transactionHash}}
       target="_blank"
       rel="noopener"
+      class={{cn "block-explorer-button" block-explorer-button--disabled=(not @transactionHash)}}
+      style={{cssVar
+        boxel-button-color="transparent"
+        boxel-button-border=(if @transactionHash "1px solid var(--boxel-purple-300)" "1px solid var(--boxel-purple-200)")
```
jurgenwerk, line 52: I'm a bit confused why we have styling in two places - in the CSS file and also using the `cssVar` helper

packages/safe-tools-client/app/components/block-explorer-button/index.css
FadhlanR: I've improved the BoxelButton by adding a tooltip argument.

FadhlanR:
Based on @lukemelia 's comment, I decided to put the tooltip on BoxelButton component.

https://user-images.githubusercontent.com/12637010/221217528-0eda3226-19ac-4dd9-926b-f71bcf6be3a5.mov

jurgenwerk (approved):

PR #3647, opened 2023-02-22
Add unsupported network modal

FadhlanR:
Ticket: CS-5234

https://user-images.githubusercontent.com/12637010/220547553-91c2f48e-b58e-4c85-bb64-224fa91179c6.mov


packages/safe-tools-client/app/components/unsupported-network-modal/index.gts
```diff
@@ -0,0 +1,61 @@
...
+      @onClose={{@onClose}}
+      data-test-unsupported-network-modal
+    >
+    <BoxelActionContainer
+          class='unsupported-network-modal__card'
+          tabindex='-1'
+          as |Section|
+        >
+          <button
+            class='unsupported-network-modal__close-button'
```
jurgenwerk, line 37: Have you considered making this modal the same way as "Authenticate with the Cardstack Hub" is made? You can add the close button in the `ActionChin`. That way you can get rid of most if not all of the custom CSS you have added.

This is a very basic modal so let's try to reuse the generic solutions that we have

packages/safe-tools-client/app/components/unsupported-network-modal/index.gts
```diff
@@ -0,0 +1,61 @@
...
+            type='button'
+            aria-label='Close'
+            {{on 'click' @onClose}}
+          >
+            {{svgJar 'close' width='100%' height='100%' aria-hidden='true'}}
+          </button>
+
+          <Section @title="Unsupported network!">
+            <span>Choose a supported one on your wallet and reconnect.</span>
```
jurgenwerk, line 46: Use `<p></p>` for paeagraphs

jurgenwerk, line 46: Your <MetaMask|WalletConnect> wallet is connected to the <network> network which we don't support.

packages/safe-tools-client/app/components/unsupported-network-modal/index.gts
```diff
@@ -0,0 +1,61 @@
...
+            <span>Supported networks: {{arrayJoin this.network.supportedNetworksName ','}}</span>
```
jurgenwerk, line 47: Copy suggestion:

Your <MetaMask|WalletConnect> wallet is connected to the $NETWORK network, which we don't support.

In your wallet, please first connect to a supported network, and then reload the page.

Supported networks: ...

packages/safe-tools-client/app/services/wallet.ts
```diff
@@ -30,6 +30,7 @@ export default class Wallet extends Service {
   @tracked providerId: WalletProviderId | undefined;
   @tracked address: string | undefined;
   @tracked isConnecting = false;
+  @tracked isConnectUnsupportedNetwork = false;
```
jurgenwerk, line 33:
```suggestion
  @tracked isConnectedToAnUnsupportedNetwork = false;
```

packages/safe-tools-client/app/templates/application.hbs
```diff
@@ -46,7 +46,12 @@
...
+  @isOpen={{and (and this.isShowingConnectModal (not this.wallet.isConnected)) (not this.wallet.isConnectUnsupportedNetwork)}}
   @onClose={{set this 'isShowingConnectModal' false}}
   @isConnected={{this.wallet.isConnected}}
-/>
\ No newline at end of file
+/>
+
+<UnsupportedNetworkModal
+  @isOpen={{this.wallet.isConnectUnsupportedNetwork}}
+  @onClose={{set this.wallet 'isConnectUnsupportedNetwork' false}}
```
jurgenwerk, line 56: Is it conceptually correct to set `isConnectUnsupportedNetwork` to false even when the user is in fact perhaps still connected to an unsupported network?

jurgenwerk (changes requested):

FadhlanR:
@jurgenwerk updated

Screenshot 2023-02-24 at 14 48 37

packages/safe-tools-client/app/templates/application.hbs
```diff
@@ -46,7 +46,12 @@
...
+  @isOpen={{and (and this.isShowingConnectModal (not this.wallet.isConnected)) (eq this.wallet.unsupportedConnectCache undefined)}}
   @onClose={{set this 'isShowingConnectModal' false}}
   @isConnected={{this.wallet.isConnected}}
-/>
\ No newline at end of file
+/>
+
+<UnsupportedNetworkModal
+  @isOpen={{not (eq this.wallet.unsupportedConnectCache undefined)}}
+  @onClose={{set this.wallet 'unsupportedConnectCache' undefined}}
```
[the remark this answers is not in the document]
FadhlanR: Does it mean we keep the modal opened until the user changes the network?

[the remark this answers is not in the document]
jurgenwerk: No, until user closes the modal. Isn't that the desired solution?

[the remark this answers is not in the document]
FadhlanR: Ah I see, sorry I misunderstood your comments.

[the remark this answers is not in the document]
FadhlanR: When do we change the value of `this.didCloseUnsupportedNetworkModal` back to false?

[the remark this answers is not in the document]
FadhlanR: It is better for us to keep this code since adding `didCloseUnsupportedNetworkModal` makes the code too complicated in these scenarios.

1. User closes the `UnsupportedNetworkModal` then the value of `didCloseUnsupportedNetworkModal` is `true`. After that, the user reconnects again but still uses an unsupported network, the `UnsupportedNetworkModal` won't show because the `didCloseUnsupportedNetworkModal` value is still `true`.

2. If we set back the value of `didCloseUnsupportedNetworkModal` to false using a render modifier or when clicking connect button, then the `UnsupportedNetworkModal` and the `ConnectWalletModal` will be overlapping because this condition is still satisfied `this.wallet.isConnectedToUnsupportedNetwork && !this.didCloseUnsupportedNetworkModal`.

jurgenwerk (approved):
I see that the suggested solution with `didCloseUnsupportedNetworkModal` is too complicated so I'm fine with your original solution. I'd like to see some minor changes I suggested and improved code comment on why the cache is needed. Otherwise looking good, feel free to merge after considering the new feedback. 👍

PR #3654, opened 2023-02-27
Fix schedule another payment issue

FadhlanR:
Ticket: CS-5322

**Problem**
There is a 401 error when scheduling another payment. It is caused by an invalid nonce error when authenticating to the hub. 

**Solution** 
Instead of asking for re-authentication every time the user schedules a payment, we can pass the existing hub auth token to the schedule payment function in SDK.


jurgenwerk (approved):
👍 I think we do the same for creating safes.

PR #3655, opened 2023-02-28
Refactor encode data for Claim

tintinthong:
- added new class `SolidityStructExtraData`
- actions which need to specify extra data will inherit this class
- add test to compare with previous data encoding

Note: I also considered making the extra data as a SolidityStruct and merging the two `ERC20TransferCaller` to `MinTokens`. But I don't think its meant it should be elevated to be a `SolidityStruct` since our contracts do not care in the EIP712 format of `MinTokens`


tintinthong:
Will get this review first, then will update accordingly for SDK functions that use this encoding

tintinthong:
Closed in favor of https://github.com/cardstack/cardstack/pull/3658

PR #3656, opened 2023-02-28
Refactor schedule payment for safe module

tintinthong:
`claim-settlement-module` inherited generic functions from `safe-module` (a big chunk of code copied from schedule-payment-module).This is the refactor to get `scheudled-payment` to use the same abstract class `safe-module`


tintinthong:
Closing this first. There are some changes to scheduler sdk.

PR #3659, opened 2023-02-28
hub: persist retry strategy so that it can be read in the client

jurgenwerk:
This PR:

1. Changes how exponential backoff for failed payment is calculated
- Previously, the amount of time needed to pass between payment retries was calculated in the SQL query in the fetcher
- Now we have a new DB field named `next_retry_attempt_at` which gets updated on every payment attempt. This new field was introduced so that we can serve it to the client where the users will be able to see when their payment gets retried (in addition to other new information, listed below). This way is also more flexible for testing, debugging, manual interventions.

2. Adds new fields to the scheduled payment response which we are going to display in the safe tools client:
- `next-retry-attempt-at`
- `scheduled-payment-attempts-in-last-payment-cycle-count` (recurring payments have multiple monthly payment period cycles - we are only interested in the retry info from the last cycle)
- `last-scheduled-payment-attempt-id` (so that we can locate the last attempt in the UI and display retry info there)
- `retries-left`

TODO: 
- [ ] ~~Migration for updating the new fields of historic payments~~  There is no need for a backfill migration. We can delete the payments in staging and in prod since there are only test payments at the moment.


lukemelia (approved):
Good test coverage 👍

FadhlanR (approved):
Looks great

jurgenwerk:
🙏

PR #3663, opened 2023-03-01
Adjust "Details" and "Actions" columns in Scheduled Payments transactions list

lukemelia:
image

### DIFF
--- a/packages/safe-tools-client/app/components/block-explorer-button/index.css
+++ b/packages/safe-tools-client/app/components/block-explorer-button/index.css
@@ -1,0 +1,3 @@
+.BlockExplorerButton svg {
+  margin-left: 2px;
+}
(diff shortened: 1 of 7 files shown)


FadhlanR (approved):

jurgenwerk (approved):
Code looks good but for UI consistency I'd prefer to keep the dropdown trigger always displayed and disable the option inside it in case it doesn't apply.

lukemelia:
>Code looks good but for UI consistency I'd prefer to keep the dropdown trigger always displayed and disable the option inside it in case it doesn't apply.

Sounds good. I'll make that change.

lukemelia:
@jurgenwerk updated. have another look

jurgenwerk (approved):
👍

PR #3666, opened 2023-03-03
Shows only the latest attempts of canceled scheduled payment

FadhlanR:
Ticket: CS-5304

**Before**
Screenshot 2023-03-03 at 21 31 38

**After**
Screenshot 2023-03-03 at 21 28 31


jurgenwerk:
For historic transparency maybe it be better to still display failed payments? And to only show the last one as canceled?

FadhlanR:
@jurgenwerk I have implemented your suggestion. My idea is to add a new field `isCanceled` field in the scheduled payment attempt object in the safe tools.

Screenshot 2023-03-06 at 21 53 57

packages/safe-tools-client/app/services/scheduled-payments.ts
```diff
@@ -178,7 +179,26 @@ export default class ScheduledPaymentsService extends Service {
       'GET'
     );
 
-    return this.deserializeScheduledPaymentAttemptResponse(response);
+    const scheduledPaymentAttempts =
+      this.deserializeScheduledPaymentAttemptResponse(response);
+    const isAlreadyFoundTheLattestAttempt: Record<string, boolean> = {};
```
lukemelia, line 184: Maybe a better name would be `previouslySeenPayments`?

lukemelia (changes requested):
Naming needs improvement

jurgenwerk:
Maybe all this can be reduced to:

```hbs
{{if (eq paymentAttempt.status 'failed')}}
  {{#if (and scheduledPaymentAttempt.scheduledPayment.isCanceled  (eq scheduledPaymentAttempt.id scheduledPaymentAttempt.scheduledPayment.lastScheduledPaymentAttemptId)}}
    Canceled
  {{else}}
    Failed
  {{/if}}
{{/if}}
```

FadhlanR:
@jurgenwerk I didn't realize we had that field, thanks for the suggestion. I've updated the code.

PR #3669, opened 2023-03-06
Remove cardie

pcjun97:


lukemelia:
Cardie v1, we hardly knew ye. I'm sorry that ye had to pass so young in years. Ye lived a lot in those short years, though, ye did. The lessons ye taught us will live on and your !deploy will be present in all of our future /deploys. Godspeed, Cardie v1. 🌩️ 👼

PR #3675, opened 2023-03-07
Fix signature for trezor

tintinthong:


habdelra (approved):
oh interesting, is `0x` getting stripped off? this seems good to me

PR #3676, opened 2023-03-07
Change minimum paymentDate to now instead of the top of the next hour

lukemelia:
- Default payment date to 10 minutes from now
- Update tests accordingly


jurgenwerk (approved):
I've been thinking about proposing this change as well. This will greatly reduce the time needed to get a result when testing

PR #3677, opened 2023-03-08
hub: Fix the scope when writing scheduledPaymentAttemptsInLastCycleCount

jurgenwerk:
Ticket: CS-5345

I noticed `attemptsInLastPaymentCycleCount` cache field on the scheduled payment counts all scheduled payment attempts when it should only count the ones belonging to the scheduled payment that is currently executed. 

The effect of this bug is that this count grows for every scheduled payment when any scheduled payment is attempted 😬 😭And since this count is now bigger than the number of allowed attempts in our exponential backoff configuration (9), no payment gets retried if it fails. 

After this is merged and deployed I intend to run the following script in our deployed environments to fix the count: 

```
let scheduledPayments = await prisma.scheduledPayment.findMany({});
for await (let scheduledPayment of scheduledPayments) {
  let attemptsInLastPaymentCycle = await prisma.scheduledPaymentAttempt.findMany({
    where: {
      startedAt: {
        gte: scheduledPayment.payAt!,
      },
      scheduledPaymentId: scheduledPayment.id,
    }
  });

  await prisma.scheduledPayment.update({
    where: { id: scheduledPayment.id },
    data: {
      scheduledPaymentAttemptsInLastPaymentCycleCount: attemptsInLastPaymentCycle.length,
    },
  });
}
```

I also wrote a regression test that makes sure the count is scoped correctly now.

### DIFF
--- a/packages/hub/node-tests/services/scheduled-payments/executor-test.ts
+++ b/packages/hub/node-tests/services/scheduled-payments/executor-test.ts
@@ -123,6 +123,33 @@ describe('executing scheduled payments', function () {
   it("sets the scheduled payment's status to 'failed' if the transaction fails", async function () {
     sdkError = new Error('UnknownHash');
 
+    // We create otherScheduledPayment and make a failed payment attempt so that we can test that scheduledPaymentAttemptsInLastPaymentCycleCount on the scheduledPayment is updated correctly and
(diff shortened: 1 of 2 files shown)


jurgenwerk:
Deployed to staging and production. Ran the script on both too. This is fixed now.

PR #3678, opened 2023-03-08
Update to glint 1.0.0-beta.3

lukemelia:


packages/safe-tools-client/app/templates/schedule.hbs
```diff
@@ -84,7 +84,7 @@
       {{/if}}
     </div>
 
-    <div class="safe-tools__version">
+    <div class="safe-tools__version hide-in-percy">
```
backspace, line 87: Is this (and the accompanying CSS) mixed in from another branch?

backspace (approved):

PR #3683, opened 2023-03-09
safe tools: Set tomorrow noon as the default scheduled date

jurgenwerk:
I am proposing a change in how default payment date is generated when doing a one-time payment.

The current behaviour feels a little bit odd to me. I think 10 minutes from now doesn't feel too realistic to schedule a payment for an average safe tools user. Also I think it would be better to round the minutes so they are consistent with the minutes in the time picker.

So I propose we change from "10 minutes from now":
image

To "tomorrow noon": 
image

However that comes with a risk of spiking gas fees around that time? But maybe it doesn't matter since it will be retried later anyway. 

What do you think?


lukemelia (approved):
I agree this is an improvement.

PR #3685, opened 2023-03-13
Add private memo field to scheduled payments

lukemelia:


packages/hub/db/migrations/20230313142122856_add-memo-to-scheduled-payments.ts
```diff
@@ -0,0 +1,12 @@
+import { MigrationBuilder } from 'node-pg-migrate';
+
+const TABLE = 'scheduled_payments';
+const PRIVATE_MEMO = 'private_memo';
+
+export async function up(pgm: MigrationBuilder): Promise<void> {
+  pgm.addColumn(TABLE, { [PRIVATE_MEMO]: { type: 'string', notNull: false, default: undefined } });
```
jurgenwerk, line 7: Perhaps `notNull: false, default: undefined` might not be needed? It sounds like it's default.

packages/safe-tools-client/app/components/payment-transactions-list/index.gts
```diff
@@ -195,6 +195,9 @@ class PaymentTransactionsList extends Component {
                 </td>
                 <td class="table__cell blockchain-address transactions-table-item-payee" data-test-scheduled-payment-attempts-item-payee>
                   <TruncatedBlockchainAddress @address={{paymentAttempt.scheduledPayment.payeeAddress}} @isCopyable={{true}} />
+                  <div class="payment-transactions-list__memo" title={{paymentAttempt.scheduledPayment.privateMemo}}>
```
jurgenwerk, line 198: When the memo is blank, maybe it's better to not render the element at all?

jurgenwerk (approved):

PR #3690, opened 2023-03-14
Defining gas price to avoid transaction underpriced error

FadhlanR:
Ticket: CS-5358

We often got a `transaction underpriced` error when executing a scheduled payment in Polygon, It was because we didn't define the gas price fields (gas price, max fee, or max priority fee) of the transaction, and in that case, ethers will define max priority fee equal to `1500000000` which is too low if the network is busy. But if we define the gas price, the max fee and max priority fee will equal to gas price (https://github.com/ethers-io/ethers.js/blob/278f84174409b470fa7992e1f8b5693e6e5d2dac/src.ts/providers/abstract-signer.ts#L136).


jurgenwerk (approved):
Looks good - did you test if the execution succeeds now?

lukemelia:
Merging now so that we can test. I agree with @jurgenwerk's code comment request.

PR #3699, opened 2023-03-17
feat(hub): Add 10 seconds timeout to hub status check

pcjun97:
The status check on hub, specifically on staging, is bumping into timeout issue as sokol rpc is no longer responding properly. This PR attempts to restructure the status check flow by making all checks synchronous, and adding a configurable timeout value. Currently the value is set at 10 seconds, we might want to make that configurable via env var or a different static value.

Screenshot 2023-03-17 at 4 08 15 PM

### DIFF
--- a/packages/hub/routes/status.ts
+++ b/packages/hub/routes/status.ts
@@ -5,6 +5,7 @@ import * as Sentry from '@sentry/node';
 import { JSONAPIDocument } from '../utils/jsonapi-document';
 
 export const DEGRADED_THRESHOLD = 10;
+const STATUS_TIMEOUT = 10 * 1000;
 
 export default class StatusRoute {
   clock = inject('clock');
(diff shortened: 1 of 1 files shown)


jurgenwerk:
By adding this aren't we masking the underlying problem? For example if an rpc node takes a long time to respond, there is likely something wrong with it and we should know about it. Moreover, traditionally a timeout function would throw an error when request expires instead of let it pass.

Since sokol network is 💀 🪦 ([3 months since last validated transaction](https://blockscout.com/poa/sokol/)) maybe a better solution for now would be to ignore that check that tries to talk to the rpc node when the network is sokol. I imagine we should get rid of (or repurpose) the supporting mechanisms for cardpay protocol (such as this endpoint) but AFAIK properly sunsetting isn't a priority right now.

Another simple solution to silence these warnings is to remove Checkly checks that perform checks against this route in staging. I think this route gives status solely about cardpay related things which we don't use in staging anymore

pcjun97:
I don't think this is masking the underlying problem, in fact I think it is able to highlight additional problems. The healthcheck that was fixed is checking for `exchange rates` and not the health of the `rpc`, which has always been "operational". It's the rpc node issue that was causing the timeout.

In the original implementation, the healthcheck returns the exact same response data, but it takes almost a full minute to do so. Checkly's check failed because there is a hard cap of 30 seconds of timeout, so despite the original healthcheck is able to return a useful data, it wasn't able to do so within the Checkly's tolerable timespan.

By defining timeout ourselves, we can now check for more than just the health of each services, as we are also able to check whether the service is responding with an acceptable latency.

Do let me know if my understanding of this issue is wrong.

jurgenwerk:
Ah ok, I get it now! Thanks for the great explanation.

jurgenwerk (approved):

backspace (approved):
🤩

PR #3705, opened 2023-03-20
Fix internal json RPC error and other unknown errors

FadhlanR:
We had some execution attempts that we marked as failed even though it was mined. It is because we use unknown errors such as `Internal JSON RPC error` and `NoNetwork` to determine the attempt status. I think we just need to mark the attempt as failed if we found the known errors only, and create a new waiting task if we found unknown errors.

### DIFF
--- a/packages/hub/tasks/scheduled-payment-on-chain-execution-waiter.ts
+++ b/packages/hub/tasks/scheduled-payment-on-chain-execution-waiter.ts
@@ -51,26 +51,38 @@ export default class ScheduledPaymentOnChainExecutionWaiter {
         }
       }
     } catch (error: any) {
-      // waitUntilTransactionMined will return "Transaction took too long to complete" in case it wasn't mined in 60 minutes.
-      // In this case we want to restart the task and wait some more. We could throw an error here so that the worker would restart the task, but
-      // due to the exponential-backoff it could take some time before the waiting can start again. That's why we just spawn a new task directly.
-      if (error.message.includes('took too long')) {
-        if (isBefore(paymentAttempt.startedAt!, subDays(nowUtc(), 1))) {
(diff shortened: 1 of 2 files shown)


packages/hub/tasks/scheduled-payment-on-chain-execution-waiter.ts
```diff
@@ -51,64 +51,49 @@ export default class ScheduledPaymentOnChainExecutionWaiter {
...
+      let knownErrors = [
+        'was reverted',
+        'UnknownHash',
+        'InvalidPeriod',
+        'ExceedMaxGasPrice',
+        'PaymentExecutionFailed',
+        'OutOfGas',
+      ];
+      let isKnownError = knownErrors.find((knownError) => error.message.includes(knownError));
+      let isWaitTooLongError =
```
jurgenwerk, line 70: I think it should be `let isWaitTooLongError = error.message.includes('took too long')` (without the `startedAt` check) to not mix unrelated things. And then you can have another check for `startedAt` later

FadhlanR: I update the naming variable to `isWaitTooLong` instead. So that variable tells us whether we wait too long for that payment.

jurgenwerk: I think it's still a misleading that way. We have 2 different validations:

1. The allowed time that the waiter process to wait for a verified transaction in the waiter has expired
2. The time that we allow the transaction to be pending or to appear on chain has expired

I suggest the following naming to make things clearer:

`let waiterTimedOut = error.message.includes('took too long');`
`let waitingExpired = isBefore(paymentAttempt.startedAt!, subDays(nowUtc(), 1));`

PR #3707, opened 2023-03-20
Add Typedoc to monorepo (1st PR)

tintinthong:
This is the main change that integrate [typedoc](https://typedoc.org/) into our monorepo. 

The main reason for this
- api changes in the docs (particularly sdk), code generated 
- additionally we use tsdoc so documentation goes in the code 

The docs are available at a Github page https://cardstack.github.io/cardstack/. 

The build process for this 
- `api-docs` github workflow is triggered by any opening PR (when integrate versioning we will only publish the docs by any publish of the sdk)
-  the docs are then commited to `gh-pages` branch
- the docs immediately gets deployed to github page

A follow on [PR](https://github.com/cardstack/cardstack/pull/3693) integrates versioning


packages/cardpay-sdk/README.md
```diff
@@ -3,1142 +3,3 @@ This is a package that provides an SDK to use the Cardpay protocol.
 
 ### Special Considerations <!-- omit in toc -->
  One item to note that all token amounts that are provided to the API must strings and be in native units of the token (usually `wei`) unless otherwise noted. All token amounts returned by the API will also be in native units (again, this usually means `wei`). You can use `Web3.utils.toWei()` and `Web3.utils.fromWei()` to convert to and from units of `wei`. Because ethereum numbers can be so large, it is unsafe to represent these natively in Javascript, and in fact it is very common for a smart contract to return numbers that are simply too large to be represented natively in Javascript. For this reason, within Javascript the only safe way to natively handle numbers coming from Ethereum is as a `string`. If you need to perform math on a number coming from Ethereum use the `BigNumber` feature of the ethers library.
-
-- [Function Parameters](#function-parameters)
```
tintinthong, line 7: The SDK index is now avaialble [here](https://cardstack.github.io/cardstack/modules/_cardstack_cardpay_sdk.html)

packages/cardpay-sdk/contracts/abi/account-registration-nft.ts
```diff
@@ -1,3 +1,6 @@
+/**
```
tintinthong, line 1: The docs get very polluted with too many variables / classes so I had to add the @category tag to organise them

packages/cardpay-sdk/index.ts
```diff
@@ -74,5 +74,51 @@ export {
...
+  Address,
+  TransferERC20ToCaller,
+  NFTOwner,
+  TransferNFTToCaller,
+  Claim,
+  getTypedData,
+} from './sdk/claim-settlement/utils';
 export { default as ClaimSettlementModule } from './sdk/claim-settlement-module';
+
+// add these classes for typedoc to display documentation
```
tintinthong, line 87: We only get docs from things which are exported. I tried not to do this too much but the tradeoff here is if we want better links/references to our doc, we should export them. Thats why typedoc tends to work well with packages we use as "libraries" not "services"

packages/cardpay-sdk/sdk/safe-module.ts
```diff
@@ -34,48 +34,81 @@ import BN from 'bn.js';
 import JsonRpcProvider from '../providers/json-rpc-provider';
 import { camelCase } from 'lodash';
 
+/**
```
tintinthong, line 37: This is a file that is a copy of a lot of scheduler work. I place it as @hidden because it is supposed to be common code shared between the scheduler and claimSettlement module but has not been refactored into the scheduler

typedoc.json
```diff
@@ -0,0 +1,36 @@
...
+    "./packages/compiled",
+    "./packages/ssr-web",
+    "./packages/safe-tools-subgraph",
+    "./packages/test-support",
+    "./packages/safe-tools-client",
+    "./packages/core",
+    "./packages/cardhost",
+    "./packages/base-cards",
+    "./packages/cardpay-subgraph",
+    "./scripts"
```
tintinthong, line 33: All of this has been excluded for now. If we want to push stuff in feel free to uncomment.

habdelra (approved):
:+1:

PR #3710, opened 2023-03-21
Add estimateGas step to capture specific execution error

FadhlanR:
Ticket: CS-5394

We decided to define the gas limit to prevent an out-of-gas error. But it caused the send transaction won't execute the estimate gas function, so we can't capture the specific execution error before the transaction is published. So I added a line to call the estimate gas function.


jurgenwerk (approved):
Nice catch

PR #3715, opened 2023-03-24
Define refund receiver address to avoid 'GS026' error

FadhlanR:
Ticket: CS-5403

I've tested in the polygon https://polygonscan.com/tx/0xb7634aa2825005b193657a0ff9e4cb576ed212cebc21fb102efa036e2f32436b. Now the refund receiver address is defined.

Screenshot 2023-03-24 at 13 46 08


FadhlanR:
Will create another PR with correct ticket number.

PR #3716, opened 2023-03-24
Define refund receiver address to avoid 'GS026' error

FadhlanR:
Ticket: [CS-5403](https://linear.app/cardstack/issue/CS-5403)

I've tested in the polygon https://polygonscan.com/tx/0xb7634aa2825005b193657a0ff9e4cb576ed212cebc21fb102efa036e2f32436b. Now the refund receiver address is defined.

Screenshot 2023-03-24 at 13 46 08

### DIFF
--- a/packages/cardpay-sdk/sdk/utils/signing-utils.ts
+++ b/packages/cardpay-sdk/sdk/utils/signing-utils.ts
@@ -72,7 +72,8 @@ export async function signSafeTx(
   estimate: Estimate,
   nonce: BN,
   from: string,
-  signer?: Signer
+  signer?: Signer,
+  refundReceiver?: string
 ): Promise<Signature[]> {
   let signatures = await signSafeTxAsRSV(
     web3OrEthersProvider,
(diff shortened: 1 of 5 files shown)


jurgenwerk (approved):
Amazing catch!! 🙌

habdelra:
Oh it’s really interesting that the relay server is not filling this out for us in the estimate. Feels like the relay server has a vulnerability.

packages/cardpay-sdk/sdk/utils/signing-utils.ts
```diff
@@ -72,7 +72,8 @@ export async function signSafeTx(
   estimate: Estimate,
   nonce: BN,
   from: string,
-  signer?: Signer
+  signer?: Signer,
+  refundReceiver?: string
```
habdelra, line 76: Doesn’t the `estimate` object already include the `refundReceiver`? Is there a reason you make all the callers pass this is separately? Could we not just pull out the `refundReceiver` from the `estimate`?

packages/cardpay-sdk/sdk/utils/signing-utils.ts
```diff
@@ -84,7 +85,7 @@ export async function signSafeTx(
     estimate.dataGas,
     estimate.gasPrice,
     estimate.gasToken,
-    ZERO_ADDRESS,
+    refundReceiver ?? ZERO_ADDRESS,
```
habdelra, line 88: I think `estimate.refundReceiver` would be better here.

PR #3718, opened 2023-03-28
Fix scheduled recurring payment that had not been executed

FadhlanR:
Ticket: CS-5414

**Problem**
We had scheduled recurring payments that had not been executed. The following is an example:

```
|               id                  |       pay_at        | recurring_day_of_month |   recurring_until   |
| 8a84ce01-5bd5-4e05-9b5d-5060fc2095d0 | 2023-03-22 22:47:53 |                     22 | 2023-03-22 03:00:00 |
```

As can be seen, the payment date falls after the recurring end date, meaning that this payment will never be executed, since the recurring until date has already passed.

**Solution**
- [ ] The recurring until date in the safe tools should be set to the end of the day for the selected day.
- [ ] The scheduled payment fetcher should retrieve scheduled recurring payments that have already passed the recurring end date but are still within the valid period.
- [ ] When scheduling a recurring payment, the calculated payment date must be earlier than the recurring end date.

### DIFF
--- a/packages/hub/routes/scheduled-payments.ts
+++ b/packages/hub/routes/scheduled-payments.ts
@@ -127,8 +127,8 @@ export default class ScheduledPaymentsRoute {
       privateMemo: attrs['private-memo'],
     } as unknown as ScheduledPayment;
 
-    if (params.recurringDayOfMonth != null) {
-      params.payAt = calculateNextPayAt(new Date(), params.recurringDayOfMonth);
+    if (params.recurringDayOfMonth != null && params.recurringUntil != null) {
+      params.payAt = calculateNextPayAt(new Date(), params.recurringDayOfMonth, params.recurringUntil);
     }
 
     let errors = await this.scheduledPaymentValidator.validate(params);
--- a/packages/hub/services/validators/scheduled-payment.ts
+++ b/packages/hub/services/validators/scheduled-payment.ts
@@ -95,6 +95,10 @@ export default class ScheduledPaymentValidator {
       if (isAfter(scheduledPayment.payAt, aYearFromNow)) {
         errors['payAt'].push(`payment date cannot be further than 1 year in the future`);
       }
+
+      if (scheduledPayment.recurringUntil && isAfter(scheduledPayment.payAt, scheduledPayment.recurringUntil)) {
+        errors['payAt'].push(`payment date cannot be more than recurring payment end date`);
+      }
     }
 
     if (scheduledPayment.recurringUntil) {
--- a/packages/hub/utils/scheduled-payments.ts
+++ b/packages/hub/utils/scheduled-payments.ts
@@ -3,7 +3,8 @@ import { convertDateToUTC } from './dates';
 // This function will return the next payment date based on the frequency.
 // fromDate will be either the current date when a recurring scheduled is created,
 // or the date of the last payment.
-export function calculateNextPayAt(fromDate: Date, recurringDay: number) {
+// When a recurring scheduled is created the recurringUntil parameter must be not undefined.
+export function calculateNextPayAt(fromDate: Date, recurringDay: number, recurringUntil?: Date) {
   if (recurringDay < 1 || recurringDay > 31) {
     throw new Error('Recurring day must be in the range of 1-31');
   }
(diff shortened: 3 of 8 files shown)


packages/hub/routes/scheduled-payments.ts
```diff
@@ -127,8 +127,8 @@ export default class ScheduledPaymentsRoute {
       privateMemo: attrs['private-memo'],
     } as unknown as ScheduledPayment;
 
-    if (params.recurringDayOfMonth != null) {
-      params.payAt = calculateNextPayAt(new Date(), params.recurringDayOfMonth);
+    if (params.recurringDayOfMonth != null && params.recurringUntil != null) {
```
jurgenwerk, line 130: Maybe we can add a default value to `recurringUntil` in case it is not defined? I.e one year from now like in the interface

packages/hub/services/validators/scheduled-payment.ts
```diff
@@ -95,6 +95,10 @@ export default class ScheduledPaymentValidator {
       if (isAfter(scheduledPayment.payAt, aYearFromNow)) {
         errors['payAt'].push(`payment date cannot be further than 1 year in the future`);
       }
+
+      if (scheduledPayment.recurringUntil && isAfter(scheduledPayment.payAt, scheduledPayment.recurringUntil)) {
+        errors['payAt'].push(`payment date cannot be more than recurring payment end date`);
```
jurgenwerk, line 100:
```suggestion
        errors['payAt'].push(`payment date cannot be later than recurring payment end date`);
```

packages/hub/utils/scheduled-payments.ts
```diff
@@ -3,7 +3,8 @@ import { convertDateToUTC } from './dates';
 // This function will return the next payment date based on the frequency.
 // fromDate will be either the current date when a recurring scheduled is created,
 // or the date of the last payment.
-export function calculateNextPayAt(fromDate: Date, recurringDay: number) {
+// When a recurring scheduled is created the recurringUntil parameter must be not undefined.
```
jurgenwerk, line 6: If you add the default value for `recurringUntil` then this comment won't be needed

packages/hub/utils/scheduled-payments.ts
```diff
@@ -3,7 +3,8 @@ import { convertDateToUTC } from './dates';
...
+export function calculateNextPayAt(fromDate: Date, recurringDay: number, recurringUntil?: Date) {
```
jurgenwerk, line 7: Conceptually speaking Is there a case where the `recurringUntil` can be undefined? I suppose no recurring payment should be repeating indefinitely?

packages/hub/utils/scheduled-payments.ts
```diff
@@ -14,6 +15,13 @@ export function calculateNextPayAt(fromDate: Date, recurringDay: number) {
   let adjustedRecurringDay = Math.min(recurringDay, daysInMonth(fromDate));
   if (fromDate.getUTCDate() < adjustedRecurringDay) {
     nextPayAtUtc.setUTCDate(adjustedRecurringDay); // next pay this month
+  } else if (recurringUntil && fromDate > recurringUntil && fromDate.getUTCDate() === adjustedRecurringDay) {
```
jurgenwerk, line 18: Instead, would it be simpler to adjust this function like so?

Instead of:

```
return nextPayAtUtc;
```

We can do this:

```
if (nextPayAtUtc <= recurringUntilUtc) {
  return nextPayAtUtc;
} else {
  return null; // Payment expired, there is no next pay date
}

Comment (FadhlanR) [reply packages/hub/utils/scheduled-payments.ts:18]: Hmm, are we allowing a user to schedule recurring payments with recurring until before `dateNow()`?

File: packages/hub/utils/scheduled-payments.ts (hunk 4 above)
Comment (jurgenwerk) [reply packages/hub/utils/scheduled-payments.ts:7]: So `recurringUntil?` can be changed to `recurringUntil` in arg types

File: packages/hub/utils/scheduled-payments.ts (hunk 5 above)
Comment (jurgenwerk) [reply packages/hub/utils/scheduled-payments.ts:18]: We should not allow that.

File: packages/hub/routes/scheduled-payments.ts (hunk 1 above)
Comment (FadhlanR) [reply packages/hub/routes/scheduled-payments.ts:130]: I think `recurringUntil` is mandatory for recurring payments.

Comment (FadhlanR) [reply packages/hub/routes/scheduled-payments.ts]: Added the validation https://github.com/cardstack/cardstack/pull/3718/commits/77c3846608277f7400875dc1f944e47a86de5991

File: packages/hub/routes/scheduled-payments.ts [hunk 6]
```diff
@@ -139,6 +135,11 @@ export default class ScheduledPaymentsRoute {
         errors: serializeErrors(errors),
       };
     } else {
+      if (params.recurringDayOfMonth != null && params.recurringUntil != null) {
+        params.payAt = calculateNextPayAt(new Date(), params.recurringDayOfMonth, params.recurringUntil);
+        if (!params.payAt) throw new Error(`cannot calculate next payment date`);
```
jurgenwerk, line 140: 1. The error message is too vague (would be good to have a reason why it cannot be calculated)
2. Won't the request will fail with a 500 Internal Server error if you throw an error here? In that case it is a better idea to put this check in the validator so that the client knows what went wrong.

packages/hub/services/validators/scheduled-payment.ts
```diff
@@ -77,6 +76,12 @@ export default class ScheduledPaymentValidator {
       }
     }
 
+    if (!scheduledPayment.payAt && (!scheduledPayment.recurringDayOfMonth || !scheduledPayment.recurringUntil)) {
```
jurgenwerk, line 79: In case `payAt` and `recurringUntil` are blank, and `recurringDayOfMonth` is present, then I think `recurringDayOfMonth` will have an error that says it's missing which is incorrect

jurgenwerk (changes requested):

lukemelia (approved):

PR #3719, opened 2023-03-28
hub, sdk: Don't allow executionGasPrice to be 0 to make sure gas fee is reimbursed to the crank

jurgenwerk:
Description in progress


FadhlanR (approved):
Looks good, thanks!

PR #3722, opened 2023-03-30
safe client: Fix gas token gas price display

jurgenwerk:
This PR fixes 2 bugs:

1. Gas tokens in the transaction lists are Unknown tokens, which has negative effects because an Unknown token has 18 decimals while USDC token for example has 6. This way the price calculations in helpers will be wrong. This is easily demonstrable by adding `{{paymentAttempt.scheduledPayment.gasToken.name}}` to the list (for inspection):

image

Since gas tokens are loaded dynamically, the fix includes adding a loading control to the tokens resource and making sure the tokens are loaded before the scheduled payment starts the deserializing process. 

After the fix, the token is recognized correctly:

image

(this text is not part of the design, it's only added to demonstrate which gas token is used in the payment attempt)

2. Very small numbers are rounded to 0, for example 0.000001. In my opinion we should make an extra case for handling this one so that we don't cut off the decimals before we reach the first significant decimal. Here is an illustration of this bug ⬇️

The tooltip says:
image

But actually, the execution gas price is 1 (in smallest amounts of the USDC token), and the tooltip fails to reflect that:
image

If we convert the units from the smallest amount to decimals (human readable amount of tokens), which is 1/10**6 = 0.000001, our current rounding will round this to 0.000, which is misleading, and is shown in the first tooltip screenshot in this section. I think it would be better to not round at the rounding point in case the first significant decimal is further than that. In the 2nd commit there's a fix that makes the tooltip render with this data:

image

### DIFF
--- a/packages/safe-tools-client/app/helpers/native-units-to-decimal.ts
+++ b/packages/safe-tools-client/app/helpers/native-units-to-decimal.ts
@@ -10,16 +10,35 @@ interface Signature {
   Return: string;
 }
 
+export function roundDecimals(number: number, decimals: number) {
+  // Below is a special case for numbers betweeon 0 and 1 whose first significant decimal digit comes after the rounding point.
+  // For example we want to round to to 3 decimals, but the number is 0.0000001. In this case if we use toFixed(3),
+  // we will get 0, but in this case we want to keep 0.0000001 to not throw away the significant part.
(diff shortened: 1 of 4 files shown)


lukemelia (approved):
👍

FadhlanR (approved):

PR #3724, opened 2023-03-31
Fix scheduled payment fetcher test

FadhlanR:


.github/workflows/pr-hub.yml
```diff
@@ -25,7 +25,7 @@ jobs:
       CORDE_BOT_TOKEN: ${{ secrets.CORDE_BOT_TOKEN }}
     services:
       postgres:
-        image: postgres:13
+        image: postgres:14
```
jurgenwerk, line 28: Is this part of the fix?

FadhlanR: Yes, I've tried using version 13 and the CLI failed.

jurgenwerk: I can't see how increasing pg version would solve this problem

packages/hub/node-tests/services/scheduled-payments/fetcher-test.ts
```diff
@@ -193,18 +193,7 @@ describe('fetching scheduled payments that are due', function () {
         payAt: addDays(now, 1),
       });
 
-      //passes recurring until but still in the valid period
-      let sp3 = await createScheduledPayment({
```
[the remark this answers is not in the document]
jurgenwerk: I would avoid deleting things from other tests to make this one work

jurgenwerk:
@FadhlanR here's an alternative fix: https://github.com/cardstack/cardstack/pull/3727

PR #3727, opened 2023-03-31
hub: Avoid unpredictable order of scheduled payment records in tests

jurgenwerk:


lukemelia (approved):

jurgenwerk:
This worked for me locally but it didn't work in the CI. There has to be another problem here. Checking

packages/hub/services/scheduled-payments/fetcher.ts
```diff
@@ -99,7 +99,12 @@ export default class ScheduledPaymentsFetcherService {
         ORDER BY last_failed_payment_attempt.started_at ASC NULLS FIRST
         LIMIT ${limit};`;
 
-    return prisma.scheduledPayment.findMany({ where: { id: { in: results.map((result) => result.id) } } });
+    let ids = results.map((result) => result.id);
+
+    // Make sure the collection is sorted in the same order as the ids array
+    return (await prisma.scheduledPayment.findMany({ where: { id: { in: ids } } })).sort(
```
FadhlanR, line 105: Don't we want to sort scheduled payments by the last failed payment attempt, not id?

jurgenwerk: We do, the `results` are sorted by that.

Then we just make sure the prisma models have the same order as results

FadhlanR (approved):

PR #3729, opened 2023-04-03
Add root hash to db model

tintinthong:
adding root hash to hub model. 

Link to [discord](https://discord.com/channels/584043165066199050/806895222830661672/1090901763797819402) 

Basically, 

- there is some logic that merges rootHash ids that will decide if a claim is claimable in the wallet (hence if the button appears) https://github.com/cardstack/cardwallet/blob/7c920915890203438f185cd569df2fbbab879039/cardstack/src/services/rewards-center/rewards-center-service.ts#L115-L126
 - https://github.com/cardstack/cardstack/blob/main/packages/hub/db/migrations/20221215134519797_reward-proofs.ts when I moved the schema to the hub I ommited rootHash thinking it doesn't affect anything. 
 
 In order to implement change, I need to wipe the db data in reward_proofs and add the new column of rootHash


tintinthong:
Test should be fixed

packages/hub/config/structure.sql
```diff
@@ -2006,69 +1990,72 @@ COPY graphile_worker.migrations (id, ts) FROM stdin;
...
+48	20221215102320386_reward-root-index	2023-03-28 19:42:27.335563
+49	20221215134519797_reward-proofs	2023-03-28 19:42:27.338128
+50	20230104034916208_add-block-number-to-reward-root-index	2023-03-28 19:42:27.34101
+51	20230131060836816_alter-gas-estimation-results-fields	2023-03-28 19:42:27.342054
+52	20230227121110219_add-retry-info-to-scheduled-payments	2023-03-28 19:42:27.344302
+53	20230228132252559_add-execution-gas-price-to-scheduled-payment-attempts	2023-03-28 19:42:27.345444
+54	20230308143158157_add-token-and-gas-token-address-to-gas-estimation-results	2023-03-28 19:42:27.346577
+55	20230313142122856_add-memo-to-scheduled-payments	2023-03-28 19:42:27.349959
+56	20230328025428111_claims	2023-03-28 19:42:27.351166
+57	20230328112435791_signed-claims	2023-03-28 19:42:27.353809
```
tintinthong, line 2049: interesting how these made their way in. Ill fix tomorrow

tintinthong: Fixed

packages/hub/config/structure.sql
```diff
@@ -2,8 +2,8 @@
 -- PostgreSQL database dump
```
tintinthong, line 2: @IanCal this file is used to initialise db only. I don't think there is any consequence

tintinthong: yarn db:test-init

IanCal (approved):
LGTM, thanks for checking the sql changes

tintinthong:
Sorry for the delay. I had to confirm everything worked on wallet. And there were so many things going on. 

If wanting to point the built wallet to the hub server the steps should be 
- copy the dist folder from yarn prepack into node_modules at root. Yes copy the entire folder and don't assume that tweaking mainjs.bundle will work
- `yarn setup`
- `yarn install-bundle && yarn install-pods-justin && yarn build:ios && yarn start`,
- yarn ios

For future reference of anyone tweaking the sdk and running the wallet!!!!! I lost so many hours just figuring out the wallet build (should have just trusted my stashes)

I am deploying now. Have also saved dumps of tables and hub on my local (just in case) tho snapshots are already in place for the hub

PR #3730, opened 2023-04-03
safe tools client: indicate pending or failed creation transactions in future payments list

jurgenwerk:
This PR adds a feature to the Future payments list which indicates whether the payment's on-chain creation transaction is pending or has failed. Currently, such payments are shown without indication which is misleading, because it looks like the payment is scheduled while it could be pending or failed at creation. 

To fix this, icons and tooltips were added to such payments. The message is hidden in the tooltip and it opens when hover over the icon. It looks like this:

### Pending icon
image

### Tooltip content
image

### Disabled cancel option
image

### Icon
image

### Tooltip content
image

### Option will say "Delete Payment" instead of "Cancel Payment"
image

### "Delete Payment" modal design for scheduled payment with a failed creation transaction

image

TODO:

- [x] Figure out how to write a test for the tooltip. The current attempt does not work since the tooltip is a CSS-only solution and it's tricky to manipulate this with JavaScript 
- [x] Add an option to delete a payment (without canceling i.e. removing the sp hash from blockchain, only removing from the crank)


FadhlanR (approved):
Looks good!

jurgenwerk:
@lukemelia can you also please have a look?

lukemelia (approved):
Apologies for the delayed review -- this looks great.

PR #3731, opened 2023-04-03
Update loading message

FadhlanR:
Ticket: CS-5392

Before:
Screenshot 2023-04-03 at 21 31 49

After:
Screenshot 2023-04-03 at 21 27 52


jurgenwerk:
Is "awaiting consensus" the phrase we should be start using instead of "confirming transaction" throughout the whole app? There are several other places where we way "confirming transaction" (or even "mining"). Wondering if we should search & replace.

FadhlanR:
@jurgenwerk I think I have replaced all "confirming transaction..." with "awaiting consensus...".

jurgenwerk (approved):

PR #3733, opened 2023-04-06
Web dids

IanCal:
Adds basic support for resolving web dids. Tests should mock out the remote request but don't currently.

Web dids have a structure of:

did:web:domain

which will try and find a document at https://domain/.well-known/did.json

Or 

did:web:domain:subfolder:subfolder2

which will try and find one at https://domain/subfolder/subfolder2/did.json

It's a bit inconsistent but that's how it goes.

I've put sample ones at

https://superb-conkies-141fd5.netlify.app/.well-known/did.json

and 

https://superb-conkies-141fd5.netlify.app/subfolder/did.json

Which also show the two ways that should now be supported for pointing at resources - one uses alsoKnownAs and the other a service endpoint:

```
  "service": [
    {
      "id": "did:web:example.com#resource1",
      "type": "ResourceService",
      "serviceEndpoint": "https://superb-conkies-141fd5.netlify.app/testme"
    }
  ]```


tintinthong (approved):
lgtm

PR #3735, opened 2023-04-10
Changes to support test dapp

tintinthong:
These are changes that are supposed to be implemented anyway

### DIFF
--- a/packages/cardpay-sdk/sdk/claim-settlement-module.ts
+++ b/packages/cardpay-sdk/sdk/claim-settlement-module.ts
@@ -233,7 +233,8 @@ export default class ClaimSettlementModule extends SafeModule {
     let signer = this.signer ? this.signer : this.ethersProvider.getSigner();
     let callerAddress = await signer.getAddress();
     let tokenAddress = await getAddress('cardToken', this.ethersProvider);
-    return this.stakingClaim(moduleAddress, payeeAddress ?? callerAddress, tokenAddress, '1');
+    let nftAddress = await getAddress('accountRegistrationNft', this.ethersProvider); // by default using our NFT
+    return this.stakingClaim(moduleAddress, payeeAddress ?? callerAddress, tokenAddress, '1', nftAddress);
   }
 
   async stakingClaim(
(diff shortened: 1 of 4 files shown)


packages/cardpay-sdk/sdk/claim-settlement-module.ts
```diff
@@ -155,6 +155,80 @@ export default class ClaimSettlementModule extends SafeModule {
     return await waitUntilTransactionMined(this.ethersProvider, txnHash);
   }
 
+  async removeValidator(txnHash: string): Promise<SuccessfulTransactionReceipt>;
```
tintinthong, line 158: Method to remove validator

packages/cardpay-sdk/sdk/claim-settlement/utils.ts
```diff
@@ -47,7 +47,7 @@ class SolidityStruct {
   asMapping() {
     const data: { [key: string]: any } = {}; // eslint-disable-line @typescript-eslint/no-explicit-any
     this.properties.forEach((o, i) => {
-      data[o.name] = this.values[i];
+      data[o.name] = this.values[i] instanceof BigNumber ? this.values[i].toString() : this.values[i];
```
tintinthong, line 50: asMapping() would print out BigNumber objects.

packages/cardpay-sdk/sdk/safe-module.ts
```diff
@@ -222,26 +219,28 @@ export default abstract class SafeModule {
...
-    if (balance.lt(gasCost)) {
-      throw new Error(
-        `Safe does not have enough balance to enable ${
-          this.name
-        } module. The gas token ${gasTokenAddress} balance of the safe ${safeAddress} is ${utils.formatUnits(
-          balance,
-          decimals
-        )}, the the gas cost is ${utils.formatUnits(gasCost, decimals)} ${symbol}`
-      );
+    if (gasTokenAddress) {
```
tintinthong, line 225: Don't know why creating a safe was limited to gasToken only. We should be able to run it with eth

packages/cardpay-sdk/sdk/safe-module.ts
```diff
@@ -577,4 +576,9 @@ export default abstract class SafeModule {
       expectedModuleAddress: enableModuleTxs.expectedModuleAddress,
     };
   }
+
+  async isModuleEnabled(safeAddress: string, moduleAddress: string): Promise<boolean> {
```
tintinthong, line 580: This checks if a particular moduleAddress is enabled on a safe. I found this function not particularly useful because a safe doesn't share modules with other safes. But its a good way to confirm that you have installed a module given your expectedModuleAddress

PR #3738, opened 2023-04-11
Fix fetch gas estimation

FadhlanR:
Ticket: CS-5483


packages/safe-tools-client/app/services/scheduled-payment-sdk.ts
```diff
@@ -100,12 +99,12 @@ export default class SchedulePaymentSDKService extends Service {
     );
   }
 
-  @action
-  async getScheduledPaymentGasEstimation(
+  @task({ enqueue: true })
```
jurgenwerk, line 102: Previous estimation calls should be canceled, right? Would `restartable` be more accurate here?

jurgenwerk (commented):
Is there a guard in place to prevent the user from submitting the form when the new estimation is still loading?

FadhlanR:
@jurgenwerk I just added it

lukemelia (approved):

PR #3744, opened 2023-04-24
Add did sdk and cli 

tintinthong:


packages/cardpay-cli/claim-settlement/set-did.ts
```diff
@@ -0,0 +1,42 @@
...
+  describe: 'Set did configuration on module',
+  builder(yargs: Argv) {
+    return yargs
+      .positional('moduleAddress', {
+        type: 'string',
+        description: 'Module address enabled on safe',
+      })
+      .positional('safeAddress', {
+        type: 'string',
+        description: 'The address receiving assets from claim',
```
IanCal, line 16: This description needs changing, I guess this is the safe that the module is attached to? Can we get that from the module itself rather than passing it in? Not vital

tintinthong: Uhmmm these descriptions are user-specified and will inform cli even before contacting the module itself. I don't think we can.But this is mostlikey the type of mistkaes ill make with the cli  changes. Ill put the change

IanCal: I mixed two things here, one was just the description, the other was asking if we needed the param at all - if we know the module address (as it's passed in) I think we could get the safe from that. Not a required change though, thanks for fixing the desc.

IanCal (approved):
Not executed, code changes look good

PR #3769, opened 2023-08-25
Add data integrity check for cron task

tintinthong:
This adds a data integrity check that checks when a graphile worker task was last fired. If the task runs every 5 minutes, we give it an allowances of 5*3 = 15 minutes. Right now every task uses 3x multiplier but this tolerance can be adjusted for a specific task if needed

I didn't use prisma. Well because graphile_worker tables exist in the graphile_worker namespace and typically prisma schema file only allow to specify only one namespace

### DIFF
--- a/packages/hub/node-tests/routes/data-integrity-checks-test.ts
+++ b/packages/hub/node-tests/routes/data-integrity-checks-test.ts
@@ -145,7 +145,7 @@ describe('GET /api/data-integrity-checks/cron-tasks', async function () {
         );
         expect(
           o.message?.includes(
-            `"remove-old-sent-notifications" has not run within 4320 minutes tolerance (suposed to be every 1440 minutes).`
+            `"remove-old-sent-notifications" has not run within 4320 minutes tolerance (supposed to be every 1440 minutes).`
           )
         );
         expect(o.message).to.not.include('"execute-scheduled-payments"');


packages/hub/node-tests/routes/data-integrity-checks-test.ts
```diff
@@ -82,28 +84,88 @@ describe('GET /api/data-integrity-checks/scheduled-payments', async function ()
...
+  };
+  let { request, getContainer } = setupHub(this);
+
+  let setupKnownCrontabs = async (db: DBClient, task: CronState) => {
+    const tasks = Object.entries(task).map(async ([identifier, config]) => {
+      const query = `
+        INSERT INTO graphile_worker.known_crontabs (identifier, known_since, last_execution) 
+        VALUES ($1, current_timestamp, current_timestamp - interval '${config.minutesAgo} minutes');
+      `;
+      await db.query(query, [identifier]);
```
jurgenwerk, line 102: This could be a single insert query. Not sure about the performance gain when running all the tests but something to consider.

packages/hub/node-tests/routes/data-integrity-checks-test.ts
```diff
@@ -82,28 +84,88 @@ describe('GET /api/data-integrity-checks/scheduled-payments', async function ()
...
+    });
+
+    await Promise.all(tasks);
+  };
+
+  this.beforeEach(async function () {
+    let dbManager = await getContainer().lookup('database-manager');
+    db = await dbManager.getClient();
+    await getContainer().lookup('data-integrity-checks-cron-tasks');
+    await db.query('DELETE FROM graphile_worker.known_crontabs;');
```
jurgenwerk, line 112: I think conceptually this would be better suited to be in `afterEach` - i.e. cleaning up after each test

packages/hub/node-tests/routes/data-integrity-checks-test.ts
```diff
@@ -82,28 +84,88 @@ describe('GET /api/data-integrity-checks/scheduled-payments', async function ()
...
+      await db.query(query, [identifier]);
+    });
+
+    await Promise.all(tasks);
+  };
+
+  this.beforeEach(async function () {
+    let dbManager = await getContainer().lookup('database-manager');
+    db = await dbManager.getClient();
+    await getContainer().lookup('data-integrity-checks-cron-tasks');
```
jurgenwerk, line 111: Is this needed?

packages/hub/services/data-integrity-checks/cron-tasks.ts
```diff
@@ -0,0 +1,102 @@
...
+          ...acc,
+          `"${check.identifier}" has not run within ${check.checkMinutes} minutes tolerance (supposed to be every ${check.cronMinutes} minutes). Last execution is ${check.lastExecution}`,
+        ];
+      }
+      return acc;
+    }, []);
+
+    return {
+      name: 'cron-tasks',
+      status: checks.filter((o: any) => o.stopped === true).length > 0 ? 'degraded' : 'operational',
```
jurgenwerk, line 92: Can you use `IntegrityCheckResult` instead of `any`?

packages/hub/services/data-integrity-checks/cron-tasks.ts
```diff
@@ -0,0 +1,102 @@
...
+}
+
+interface CronTask {
+  'execute-scheduled-payments': TaskConfig;
+  'check-reward-roots': TaskConfig;
+  'print-queued-jobs': TaskConfig;
+  'remove-old-sent-notifications': TaskConfig;
+}
+
+// graphile worker configuration (copied from worker-client)
```
jurgenwerk, line 24: Is it possible to dynamically import this instead of copying?

packages/hub/services/data-integrity-checks/cron-tasks.ts
```diff
@@ -0,0 +1,102 @@
...
+const TASK: CronTask = {
+  'execute-scheduled-payments': { cronMinutes: 5 },
+  'check-reward-roots': { cronMinutes: 10 },
+  'print-queued-jobs': { cronMinutes: 5 },
+  'remove-old-sent-notifications': { cronMinutes: 600 },
+};
+
+let DEFAULT_MULTIPLIER = 3;
```
jurgenwerk, line 32: In the PR description it says the multiplier is 2. Which one is more suitable?

jurgenwerk (commented):
Overall looks good but would be really good to explore the possibilities of getting the cron job frequency dynamically so that we reduce the chance of getting out of sync with checks in case someone adjusts the tasks frequency.

Also is there a better word to label a task as a task that didn't run but it was supposed to? You used "stopped" but that can also indicate someone stopped it intentionally. I'd prefer maybe "lagging"

packages/hub/node-tests/routes/data-integrity-checks-test.ts
jurgenwerk: Or maybe you can even put the deletion in the beginning of the task setup function

packages/hub/services/data-integrity-checks/cron-tasks.ts
tintinthong: I think this is a good point. It did occur to me too but I thought to be conservative as to try not to touch the original code.

Let me investigate this.

packages/hub/services/data-integrity-checks/cron-tasks.ts
tintinthong: Lets use a multiplier of 3

tintinthong:
@jurgenwerk this was perhaps more involved than I wanted to. I had to parse the way the graphile worker encoded `ParsedCronItem`. Which is their object encoding of the original crontab configuration. But in any case I created classes that would correspond to a conservative understanding of what each ParsedCronItem meant

```
export const CRON_TAB_STRING: string = [
  '0 5 * * * remove-old-sent-notifications ?max=5',
  '*/5 * * * * print-queued-jobs',
  '*/5 * * * * execute-scheduled-payments',
  ...(config.get('rewardsIndexer.enabled') ? ['*/10 * * * * check-reward-roots'] : []),
].join('\n');

-> 

ParsedCronItem[]
```

packages/hub/services/data-integrity-checks/cron-tab-utils.ts
```diff
@@ -0,0 +1,155 @@
...
+    let minuteThreshold = cronMinutes * multiplier;
+    let { rows: data } = await db.query(query, [parsedCronItem.identifier, minuteThreshold]);
+    let r = { ...data[0], cronMinutes };
+    if (r.lagging) {
+      r.errorMessage = this.errorMessage(r);
+    }
+    return r;
+  }
+  isType(item: ParsedCronItem): item is MinuteIntervalLessThanHour {
+    if (item.hours.length !== 24 || item.dows.length !== 7 || item.months.length !== 12 || item.dates.length !== 31) {
```
jurgenwerk, line 70: This is a bit hard to follow for me. I think it went a bit too far in complexity 

Have you considered parsing the cron string, get the interval for each entry, and simply compare whether it has been executed in the last interval + some padding? Or am I missing something?

Looks like ChatGPT can help to spit out the function for that:

```
function calculateCronInterval(expression) {
    const fields = expression.split(' ');
    const [minute, hour, dayOfMonth, month, dayOfWeek] = fields;

    function processField(field, min, max) {
        if (field === '*') {
            return max - min + 1;
        }
        const values = field.split(',').map(value => parseInt(value));
        return Math.min(...values.map(value => max - min + 1));
    }

    const minutesInterval = processField(minute, 0, 59);
    const hoursInterval = processField(hour, 0, 23);
    const daysInterval = processField(dayOfMonth, 1, 31);
    const monthsInterval = processField(month, 1, 12);
    const weekdaysInterval = processField(dayOfWeek, 0, 6);

    // Calculate the overall interval in minutes
    const overallInterval = minutesInterval * hoursInterval * daysInterval * monthsInterval * weekdaysInterval;

    return overallInterval;
}

const cronExpression = "*/15 * * * *"; // Example cron expression
const intervalInMinutes = calculateCronInterval(cronExpression);
console.log(`Interval in minutes: ${intervalInMinutes}`);
```

packages/hub/node-tests/services/data-integrity-checks-test.ts
```diff
@@ -389,4 +393,70 @@ describe('data integrity checks', function () {
...
+      );
+      expect(calculateMinuteInterval('*/5 * * * * print-queued-jobs').minuteInterval).to.equal(5);
+      expect(() => calculateMinuteInterval('0 5 2 * * remove-old-sent-notifications ?max=5')).to.throw(
+        'Cannot parse the provided cron expression: 0 5 2 * * remove-old-sent-notifications ?max=5'
+      );
+      expect(() => calculateMinuteInterval('*/5 */3 * * * print-queued-jobs')).to.throw(
+        'Cannot parse the provided cron expression: */5 */3 * * * print-queued-jobs'
+      );
+    });
+  });
```
tintinthong, line 461: I wrote test here

packages/hub/services/data-integrity-checks/cron-tab-utils.ts
tintinthong: I have simplified everything. This function above is wrong. 

I think we prob don't want to model ENTIRE graphile workers crontab string.

The main case we consider is just the difference `*/15 * * * *` and `0 5 * * *`. Former is every 15 minutes. Latter is every day at 5am UTC. For our purposes, we just perceive these two cases as every 15 minutes and every 24*60 minutes (1 day) from last execution. 

That means that I write the code CONSERVATIVELY in that if you write something `*/15 15 * * *`. Which means every 15 minutes within the 15th hour of the day. I expect to just not parse this and error out. This keeps our code exact but incomplete.

jurgenwerk (approved):

PR #3776, opened 2023-11-06
Handle null master contract addresses for safes

IanCal:
Someone set the master contract address to 0x0 which means all proxied calls revert. This broke code that got the owners.

I found one other place we needed to add an extra check, but searching the code didn't find more to be concerned about.

This is the branch in production for the subgraph.


packages/cardpay-subgraph/README.md
```diff
@@ -68,6 +68,11 @@ $ graph auth https://graph-admin-blue.cardstack.com <Cardstack access token>
 
 After you have performed this authorization you will not need to perform this again for subsequent deployments.
 
+If this does not work, you must take the deployment CLI call from package.json and add the token with `--access-token` 
+
+    --access-token <Cardstack access token>
```
jurgenwerk, line 73: Tbh I never got it working without providing `--access-token` so great that you added this

packages/cardpay-subgraph/src/utils.ts
```diff
@@ -61,6 +61,14 @@ export function makeEOATransaction(event: ethereum.Event, address: string, safe:
 
 export function makeEOATransactionForSafe(event: ethereum.Event, safe: string): void {
   let safeContract = GnosisSafe.bind(Address.fromString(safe));
+  // block 30438533 with transaction 0xf52593bad6b798e1ef40e2ac1497d67926aa5d7597f0a755321a67e6001a54d1
+  // sets the master copy address to 0x0 which is weird but technically valid
+  // Ensure we check that we can actually get the owners before trying to get them
```
jurgenwerk, line 66: Nice catch

jurgenwerk (approved):
Are we fully synced now?

IanCal:
> Are we fully synced now?

On blue yes, which is now live and keeping up to date, which has finally resolved the pagerduty alarm. - https://graph.cardstack.com/subgraphs/name/habdelra/cardpay-xdai/graphql?query=%7B%0A%09_meta+%7B%0A++++block+%7B%0A++++++number%0A++++++timestamp%0A++++%7D%0A++%7D%0A%7D

I've just set green going which will take a few days.

habdelra (approved):

PR #3788, opened 2024-09-05
npx ember-resources-codemod

NullVoxPopuli:
Upgrades ember-resources to v7.

It also upgrades the dependencies (if outdated) to their new minimum version. 
These upgrades should probably happen as separate PRs though.

Implementation of the codemod here: https://github.com/NullVoxPopuli/ember-resources/pull/1147

Here are the breaking changes in v6:
- https://github.com/NullVoxPopuli/ember-resources/releases/tag/ember-resources%406.0.0
  - most of the changes here do not affect cardstack because most of the changes are from utilities that cardstack doesn't import/use (more on this in v7)
  
Here are the breaking changes in v7:
- https://github.com/NullVoxPopuli/ember-resources/releases/tag/v7.0.0-ember-resources
  - All utilities moved to `reactiveweb`, documentation here: https://reactive.nullvoxpopuli.com/
  - This mostly doesn't affect cardstack, because cardstack does not use any of the utilities
  - Additionally, the `Resource` class has moved to a separate package, https://github.com/NullVoxPopuli/ember-modify-based-class-resource/
    - This is to align with the ember ecosystem, and the desire to move away from `modify()` hooks -- as they are a sort of backdoor _effect_, which we don't want to encourage -- instead folks should rely on more data derivation patterns 
    

-----------------------------

Possible remaining steps to do:
- find peer-consumer references and update them


backspace:
I’ve approved this but this repository might be considered obsolete, I wouldn’t be surprised to see an avalanche of failures 😳

